PR #37, opened 2018-10-25
Placing all libs and exes in FHS-like dirs during build

bcaimano:


rtimmons:
@bcaimano would you like to keep this open / pending changes to dsi and others to actually use the new/simpler paths or merge it? Looks like the simpler paths will only be used if the appropriate `-Dwhatever` flags are passed in so it should be backward-compatible, ya?

bcaimano:
@rtimmons, I was just thinking about this one. The install path change isn't necessary anymore and the other stuff I made a build option. Take another look?

bcaimano:
Closing out because the repo changover got everything all weird.

PR #109, opened 2019-01-09
TIG-1325 Introduced RateLimiter concept

bcaimano:
So this is the TIG-1325 precursor I've been promising.

### DIFF
--- a/src/gennylib/include/gennylib/RateLimiter.hpp
+++ b/src/gennylib/include/gennylib/RateLimiter.hpp
@@ -1,0 +1,97 @@
+#ifndef HEADER_C68990DE_CA18_4104_9D1F_070CDC3077B0
+#define HEADER_C68990DE_CA18_4104_9D1F_070CDC3077B0
+
+#include <limits>
+#include <memory>
+#include <sstream>
+#include <type_traits>
+
+#include <gennylib/InvalidConfigurationException.hpp>
+#include <gennylib/config/RateLimiterOptions.hpp>
+
+namespace genny {
+
+/**
+ * RateLimiter is a utility class that enforces time scheduling on a function.
+ *
+ * As a general set of features, for every call to `run()`, a RateLimiter may:
+ * 1) Make sure the function passed to `run()` is invoked at least a minimum duration after the last
+ *    invocation of `run()`.
+ * 2) Enforce a sleep of a specified duration before the function passed to `run()` is invoked.
+ * 3) Enforce a sleep of a specified duration after the function passed to `run()` is invoked.
+ *
+ * As an intentional choice, RateLimiter has a strong API that obscures the actual implementation.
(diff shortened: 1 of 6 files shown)


src/gennylib/include/gennylib/RateLimiter.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * 1) Make sure the function passed to `run()` is invoked at least a minimum duration after the last
+ *    invocation of `run()`.
+ * 2) Enforce a sleep of a specified duration before the function passed to `run()` is invoked.
+ * 3) Enforce a sleep of a specified duration after the function passed to `run()` is invoked.
+ *
+ * As an intentional choice, RateLimiter has a strong API that obscures the actual implementation.
+ * As of now, it uses simple sleeps. The preferred mechanism would be to have a scheduler that
+ * notifies each RateLimiter at specific deadlines.
+ */
+class RateLimiter {
```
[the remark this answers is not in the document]
bcaimano: I've moved it into a `v1` namespaced folder.

bcaimano:
Okay, I've added a basic RateLimiter test for minimum periods but I don't have the sleep tests in yet. I also took out the public config options so I need to add equivalent options to `RunCommandOperationOptions`...I might just inline the old options struct for now.

src/gennylib/include/gennylib/RateLimiter.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * RateLimiter is a utility class that enforces time scheduling on a function.
+ *
+ * As a general set of features, for every call to `run()`, a RateLimiter may:
+ * 1) Make sure the function passed to `run()` is invoked at least a minimum duration after the last
+ *    invocation of `run()`.
+ * 2) Enforce a sleep of a specified duration before the function passed to `run()` is invoked.
+ * 3) Enforce a sleep of a specified duration after the function passed to `run()` is invoked.
+ *
+ * As an intentional choice, RateLimiter has a strong API that obscures the actual implementation.
+ * As of now, it uses simple sleeps. The preferred mechanism would be to have a scheduler that
```
guoyr: This last sentence is intriguing. I found some OS C++ rate limiters on GitHub and they all use sleep in the rate limiter like you do here. Is there a reason to prefer the scheduler approach? (A short sentence will do if the full answer is too long, I'm mostly just curious if you were expecting a different behavior with the scheduler approach)

bcaimano: Well, there are a few different advantages. Notably, scheduled sleeps are usually able to be interrupted and usually involve far fewer calls to the `now()` method of a given clock.

bcaimano:
Closing in favor of https://github.com/10gen/genny/pull/119

PR #110, opened 2019-01-10
TIG-1252: Verify that runCommand respects write concern for inserts

thejonathanma:
I added a couple of tests to validate that runCommand will respect specified write concern for inserts.


thejonathanma:
@bcaimano ready for another look!

PR #113, opened 2019-01-11
TIG-1154 Implement Integer, Time, and Rate Syntax Support

guoyr:
I left out support double support for timespec since I don't think we'll need it. If people think it's still necessary, I'm happy to add it back it, it's just a couple LOC but makes the code a bit messier.

### DIFF
--- a/src/gennylib/include/gennylib/conventions.hpp
+++ b/src/gennylib/include/gennylib/conventions.hpp
@@ -56,7 +56,7 @@ struct TimeSpec {
     TimeSpec() = default;
     ~TimeSpec() = default;
 
-    TimeSpec(std::chrono::nanoseconds v) : value{v} {} // NOLINT(google-explicit-constructor)
+    TimeSpec(std::chrono::nanoseconds v) : value{v} {}  // NOLINT(google-explicit-constructor)
 
     // Allow construction with integers for testing.
     explicit TimeSpec(int64_t v) : value{v} {}


src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -17,26 +21,252 @@ void decodeNodeInto(T& out, const YAML::Node& node, const S& fallback) {
     out = node.as<T>(fallback);
 }
 
+/**
+ * Intermediate state for converting YAML syntax into a native integer type of your choice.
+ *
+ * uint64 is used by default; smaller types can be explicitly converted to as needed.
+ */
+struct UIntSpec {
```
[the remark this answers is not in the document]
bcaimano: Just a heads up, but negative numbers usually make error cases clearer and head off bad loop end conditions. I'd rather we have all signed than all unsigned.

src/gennylib/test/conventions_test.cpp
```diff
@@ -1,37 +1,112 @@
...
+TEST_CASE("genny::TimeSpec conversions") {
+    SECTION("Can convert to genny::TimeSpec") {
+        REQUIRE(YAML::Load("D: 3 seconds")["D"].as<TimeSpec>().count() == 3 * std::pow(10, 9));
+        REQUIRE(YAML::Load("0 second").as<TimeSpec>().count() == 0);
+        REQUIRE(YAML::Load("20 millisecond").as<TimeSpec>().count() == 20 * std::pow(10, 6));
+        REQUIRE(YAML::Load("33 microsecond").as<TimeSpec>().count() == 33 * std::pow(10, 3));
+        REQUIRE(YAML::Load("2e3 microseconds").as<TimeSpec>().count() == 2000 * std::pow(10, 3));
+        REQUIRE(YAML::Load("10.3e2 nanoseconds").as<TimeSpec>().count() == 1030);
+        REQUIRE(YAML::Load("3 hour").as<TimeSpec>().count() == 3 * 3600 * std::pow(10, 9));
+        REQUIRE(YAML::Load("2 minutes").as<TimeSpec>().count() == 2 * 60 * std::pow(10, 9));
```
bcaimano: Just as a heads up, [pow](https://en.cppreference.com/w/cpp/numeric/math/pow) is inately floating point.

src/gennylib/test/conventions_test.cpp
```diff
@@ -1,37 +1,112 @@
...
+        REQUIRE_THROWS(YAML::Load("foo").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("[1,2,3]").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("[]").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("{}").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("what nanoseconds").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("29 picoseconds").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("1e3 centuries").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("mongodb").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("1").as<TimeSpec>());
+        REQUIRE_THROWS(YAML::Load("333").as<TimeSpec>());
```
bcaimano: Waiiiiiiit, does this mean that just a number is no longer pure millisecond? This is the sort of config schema break that caused trouble around NoOp. Maybe the mailing list should be given a poke about this.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -17,26 +21,253 @@ void decodeNodeInto(T& out, const YAML::Node& node, const S& fallback) {
...
+ *
+ * size_t is used by default; smaller types can be explicitly converted to as needed.
+ */
+struct UIntSpec {
+    UIntSpec() = default;
+    ~UIntSpec() = default;
+
+    explicit UIntSpec(size_t v) : value{v} {}
+    // size_t is used by default, you can explicitly cast to another type if needed.
+    size_t value;
```
bcaimano: There may be a missed opportunity here to bake optionality into the config types. There is no default ctor and no bool operator, so this could have tech value.

src/gennylib/test/conventions_test.cpp
guoyr: Ah good point. I will send an email out today.

PR #117, opened 2019-01-14
Placing all libs and exes in FHS-like dirs during build

bcaimano:


cmake/Options.cmake
```diff
@@ -19,3 +19,8 @@ option(GENNY_INSTALL_SSL
        "Install the SSL libraries local to the genny install."
        OFF
 )
+
+option(GENNY_BUILD_FLAT
+       "Use FHS-like build directories."
```
rtimmons: somewhere let's link to FHS (I had to look it up) http://www.pathname.com/fhs/pub/fhs-2.3.html

bcaimano:
@rtimmons, this one is a simple switch to **build** paths. It doesn't even change how things are installed or built. This change is dev personal preference. In terms of DSI and CI tests, I've been lobbing from the get go not to use build paths at all. CMake intentionally does very complex tracking of build targets and then we use a static path.

PR #118, opened 2019-01-14
Fix BigUpdate

guoyr:


src/driver/test/BigUpdate.yml
```diff
@@ -97,7 +97,7 @@ Actors:
       - Operation:
           OperationName: Nop
       - Duration: &phase2_duration 10 minutes
-        MinDelay: 31  # This should be replaced with Rate with TIG-1155.
```
guoyr: Removed this comment since we'll be getting rid of this option entirely with TIG-1155

PR #119, opened 2019-01-14
TIG-1325 Introduced RateLimiter concept

bcaimano:
Re pulling since the github PR trouble.


guoyr:
looking now!

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -67,43 +67,51 @@ inline bool operator==(const UIntSpec& lhs, const UIntSpec& rhs) {
  * choice.
  */
 struct TimeSpec {
```
guoyr: Re your comment on `UIntSpec`:

> There may be a missed opportunity here to bake optionality into the config types.

I was thinking about standardizing on one type throughout to make the YAML config simpler.

> Just a heads up, but negative numbers usually make error cases clearer and head off bad loop end conditions. I'd rather we have all signed than all unsigned.

That sounds reasonable. I'll use signed going forward.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -67,43 +67,51 @@ inline bool operator==(const UIntSpec& lhs, const UIntSpec& rhs) {
...
-    TimeSpec() = default;
+public: // Typedefs/Structs
+    // Use the highest precision internally.
+    using ValueT = std::chrono::nanoseconds;
+
+public: // Ctors/Dtors/Generators
     ~TimeSpec() = default;
 
-    TimeSpec(std::chrono::nanoseconds v) : value{v} {}  // NOLINT(google-explicit-constructor)
+    explicit constexpr TimeSpec(ValueT v) : value{v} {}
```
guoyr: nice

PR #121, opened 2019-01-15
Provide simpler Ubuntu 17.10 Bootstrapping Script

rtimmons:
I installed Ubuntu 17.10 on a virtual machine and got Genny compiling. I then tar'ed up the resulting install directory to make it faster (pre-compiled) for anyone else. This script downloads the tarball and runs the required setup. End-to-end it takes 25 minutes on my virtual-machine, 10 minutes of which are downloading the tarball from S3.

This is a bit of a stop-gap measure until we can provide a simpler cross-platform solution via vcpkg or similar in the "full" work of [TIG-1359](https://jira.mongodb.org/browse/TIG-1359).

(cc @henrikingo as FYI. Didn't add you as full reviewer but please feel free to add yourself if you wish.)

### DIFF
--- a/scripts/bootstrap-ubuntu-1710.sh
+++ b/scripts/bootstrap-ubuntu-1710.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+# This works for Ubuntu 17.10 only.
+
+# system packages
+
+sudo apt-get update
+sudo apt-get upgrade
+
+sudo apt-get install -y gcc make perl
+sudo apt-get install -y \
+    build-essential \
+    git \
+    cmake \
+    software-properties-common \
+    libboost-all-dev \
+    autoconf \
+    automake \
+    libtool \
+    curl \
+    make \
+    g++ \
+    gcc-7 \
+    unzip \
+    pkg-config
+
+# pre-built goodies updated as of 2019-01-14
+if [ ! -d "genny-setup" ]; then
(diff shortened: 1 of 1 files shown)


scripts/bootstrap-ubuntu-1710.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# This works for Ubuntu 17.10 only.
+
+# system packages
+
+sudo apt-get update
+sudo apt-get upgrade
+
+sudo apt-get install -y gcc make perl
+sudo apt-get install -y \
```
guoyr: I'm a little bit nervous about using the system packages because WT has a custom OS image that I know nothing about. Could we keep this list to a minimum and use the mongodb toolchain for some dependencies?

scripts/bootstrap-ubuntu-1710.sh
```diff
@@ -0,0 +1,118 @@
...
+    build-essential \
+    git \
+    cmake \
```
guoyr: this can be removed since we're building cmake ourselves?

guoyr:
I've got the vcpkg version mostly working using this repo: https://github.com/10gen/vcpkg

Will send out a a tarball now of the toolchain.

The C++ driver is wonky as Ben pointed out in the ticket, so I'm using a temporary workaround

rtimmons:
Thanks for feedback but going to close this without merging.
This work will be superceded by the work to use VCPKG (TIG-1359).

PR #122, opened 2019-01-15
SERVER-36672 Added Primary Required Reads workload

bcaimano:
Also put more CMake elbow grease around the workload installs

### DIFF
--- a/etc/CMakeLists.txt
+++ b/etc/CMakeLists.txt
@@ -1,7 +1,0 @@
-install(DIRECTORY workloads/
-        DESTINATION etc/genny/workloads
-        COMPONENT Workloads
-        FILES_MATCHING
-            PATTERN "*.yml"
-            PATTERN "*.yaml"
-)
--- a/etc/CMakeLists.txt
+++ b/etc/CMakeLists.txt
@@ -1,0 +1,7 @@
+install(DIRECTORY workloads/
+        DESTINATION etc/genny/workloads
+        COMPONENT Workloads
+        FILES_MATCHING
+            PATTERN "*.yml"
+            PATTERN "*.yaml"
+)
--- a/etc/workloads/service-architecture/PrimaryRequiredReads.yml
+++ b/etc/workloads/service-architecture/PrimaryRequiredReads.yml
@@ -1,0 +1,67 @@
+SchemaVersion: 2018-07-01
+Pool:
+    QueryOptions:
+        maxPoolSize: 1000
+        socketTimeoutMS: 5000
+        connectTimeoutMS: 5000
+Actors:
+  - Name: StepdownCommand
+    Type: AdminCommand
+    Threads: 1
+    # So this *will* fail the first time because that's what replSetStepdown does.
+    # We don't know if the failure message is the right one
+    # or if the replset ends up in the correct state afterward
+    # See https://jira.mongodb.org/browse/TIG-1322 for what should eventually be here
+    ExecutionStrategy:
+        Retries: 0
+        ThrowOnFailure: false
+    Phases:
+      - # Phase: 0 # Load
+        Operation:
+            OperationName: Nop
+      - Phase: 1 # Run
+        Repeat: 1
+        Operation:
+            OperationMetricsName: "Stepdown"
+            OperationName: AdminCommand
+            OperationPreSleep: 5 seconds
+            OperationPostSleep: 5 seconds
+            OperationCommand:
+                replSetStepDown: 60
+  - Name: PrimaryRequiredRead
+    Type: RunCommand
+    Threads: 100
+    ExecutionStrategy:
+        Retries: 3
+    Database: test
+    Phases:
+      - # Phase: 0 # Load
+        Operation:
+            OperationName: Nop
+      - Phase: 1 # Run
+        # Since there is no way to say "repeat until all other phases finish",
+        # We actually specify here how long this should go
+        Duration: 10 seconds
+        Operation: &find_one
+            MetricsName: "Find"
+            OperationName: RunCommand
+            OperationMinPeriod: 100 microseconds
+            OperationCommand:
+                find: Collection0
+                readConcern: { level: local }
+  - Name: Loader
+    Type: Loader
+    Threads: 1
+    Database: test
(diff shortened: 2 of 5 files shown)


etc/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+install(DIRECTORY workloads/
+        DESTINATION etc/genny/workloads
+        COMPONENT Workloads
+        FILES_MATCHING
+            PATTERN "*.yml"
+            PATTERN "*.yaml"
+)
```
rtimmons: so this makes there two directories where workloads live....we have a ticket (somewhere...) to move the workloads directory to a more sensible place - let's kill this change for now and keep things together.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
```diff
@@ -0,0 +1,67 @@
...
+        Retries: 3
+    Database: test
+    Phases:
+      - # Phase: 0 # Load
+        Operation:
+            OperationName: Nop
+      - Phase: 1 # Run
+        # Since there is no way to say "repeat until all other phases finish",
+        # We actually specify here how long this should go
+        Duration: 10 seconds
```
rtimmons: if you omit `Duration` and `Repeat`, it will run until all the other Actors finish the Phase (i.e. it makes this Actor+Phase non-blocking). But there does need to be one Actor in the Phase that keeps it open for a period of time using `Duration` or `Repeat` or else the Phase will end basically immediately.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
```diff
@@ -0,0 +1,67 @@
+SchemaVersion: 2018-07-01
+Pool:
+    QueryOptions:
+        maxPoolSize: 1000
```
rtimmons: 1. why 1k here? seems like you should only need 100ish. Nothing "wrong" with more connections than needed just wondering if there was a reason for the order-of-magnitude difference. If there is a reason, perhaps a comment would help.
2. Do you also want a min pool size to do all the connection acquisition up front?

etc/workloads/service-architecture/PrimaryRequiredReads.yml
```diff
@@ -0,0 +1,67 @@
...
+    Database: test
+    Phases:
+      - # Phase: 0 # Load
+        Operation:
+            OperationName: Nop
+      - Phase: 1 # Run
+        # Since there is no way to say "repeat until all other phases finish",
+        # We actually specify here how long this should go
+        Duration: 10 seconds
+        Operation: &find_one
```
rtimmons: 💅 I don't think you need the anchor

etc/workloads/service-architecture/PrimaryRequiredReads.yml
```diff
@@ -0,0 +1,67 @@
...
+            MetricsName: "Find"
+            OperationName: RunCommand
+            OperationMinPeriod: 100 microseconds
+            OperationCommand:
+                find: Collection0
+                readConcern: { level: local }
```
rtimmons: is `local` what you want, not `primary`?

etc/workloads/service-architecture/PrimaryRequiredReads.yml
```diff
@@ -0,0 +1,67 @@
...
+Pool:
+    QueryOptions:
+        maxPoolSize: 1000
+        socketTimeoutMS: 5000
+        connectTimeoutMS: 5000
+Actors:
+  - Name: StepdownCommand
+    Type: AdminCommand
+    Threads: 1
+    # So this *will* fail the first time because that's what replSetStepdown does.
```
rtimmons: 1. what do you mean "fail the first time"?
2. I think add a comment that this `ExecutionStrategy` block basically means that the Actor won't throw on exceptions encountered when running the command.

etc/CMakeLists.txt
bcaimano: I think the install change is still worth it to me because it allows separating the workloads by suite. The extra `etc` directory not so much.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
bcaimano: 1. It's mostly about that order of magnitude difference. It needs to be set higher because the default is 100. So an order of magnitude is reasonable.
2. I'm not sure that actually forces connection. It's just the amount of pool entries that can be floating about IIRC.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
bcaimano: Yeah, I can document exactly what the ExecutionStrategy means here. I mean, I have a key below that says "ThrowOnFailure". I'd be pretty disappointed if people got stuck on my wording.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
bcaimano: True, but I don't hurt anything by having it. I suspect you'll find people will do a lot of yamlisms for these workloads as is.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
bcaimano: Damn, I read through that code multiple times and still confused myself. I *really* find the PhaseLoop opaque.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
bcaimano: So, it is, because this is `readConcern`. That said, I'd meant to put in explicit `readPreference: primary` once I had faith it was making its way to the C driver. Which I now do.

etc/workloads/service-architecture/PrimaryRequiredReads.yml
rtimmons: > I really find the PhaseLoop opaque.

Open to suggestions on how to simplify - the complexity imho comes from the extra "support" classes necessary to make range-based `for` work.

PR #123, opened 2019-01-15
TIG-1166 Added `--dry-run` to validate workloads get past init step

bcaimano:
So we've tended to break our existing example workloads. This PR gets us to verifying them in CI as part of compiling the driver. I think @rtimmons put out the [superior driver patch](https://github.com/rtimmons/genny/tree/dryrun), but it should be easy to for him to expand upon the changes here directly.


rtimmons:
The PR builder was borked for a minute due to the repo move - can you push another dummy commit to this PR to trigger the PR builder? Sorry for the inconvenience I'll do the actual review in a bit.

src/driver/test/BigUpdate.yml
```diff
@@ -8,6 +8,9 @@
 
 SchemaVersion: 2018-07-01
 
+Pool:
+  QueryOptions:
+    maxPoolSize: 1000
```
dalyd: Oh, thank you. Could you open a ticket or PR against DSI to remove this setting from test_control.big_update.yml? 
This line: https://github.com/10gen/dsi/blob/f8514c63556b3071b68d5f3b21a4b686c6916a74/configurations/test_control/test_control.big_update.yml#L7

For my education, if set in both places, which one wins? They're the same right now so its moot, but what if they weren't?

src/cast_core/src/actors/Loader.cpp
```diff
@@ -131,7 +131,7 @@ Loader::Loader(genny::ActorContext& context, uint thread)
       _totalBulkLoadTimer{context.timer("totalBulkInsertTime", Loader::id())},
       _individualBulkLoadTimer{context.timer("individualBulkInsertTime", Loader::id())},
       _indexBuildTimer{context.timer("indexBuildTime", Loader::id())},
-      _client{context.client()},
+      _client{std::move(context.client())},
```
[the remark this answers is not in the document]
bcaimano: Well, it's mildly complicated. The pool entry is a wrapper around a unique pointer, so if we were given a reference or the like we would have to move. The way the class is returned from context, it has no choice but to move. So in short, it's not necessary but it is a good thing to make clear that we're using move construction and not copy construction.

src/driver/test/BigUpdate.yml
bcaimano: I believe the config wins right now. I should really put up a PR to make that explicitly clear. Of course, I'll PR this into DSI!

evergreen.yml
```diff
@@ -341,6 +342,21 @@ functions:
...
+  # Requires f_compile to have been run first.
+  ##
+  f_dry_run_workloads:
+  - command: subprocess.exec
+    params:
+      working_dir: .
+      binary: /bin/bash
+      args:
+        - "${workdir}/src/scripts/dry-run-workloads.sh"
+        - "${workdir}/dist/etc/genny/workloads"
```
rtimmons: why require an `install` first and not just do this for `src/driver/test`? (We'll change that in a separate ticket but it is what it is for now)

scripts/dry-run-workloads.sh
```diff
@@ -0,0 +1,24 @@
...
+# Find all of the workload yamls at the specified directory
+WORKLOAD_DIR="${1}"
+WORKLOAD_LIST="${BUILD_DIR}/workload-list.txt"
+find "${WORKLOAD_DIR}" \( -iname "*.yml" -or -iname "*.yaml" \) -print >"${WORKLOAD_LIST}"
+
+# Dry run each file
+while read -r FILE; do
+    1>&2 echo "-- Testing workload at '${FILE}' via --dry-run..."
+    "${CMD[@]}" "${FILE}"
+done <"${WORKLOAD_LIST}"
```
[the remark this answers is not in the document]
bcaimano: Heh, I think this is well below mine. I usually go by 100 lines or so. I agree that a language with introspection becomes desirable once state enters into the equation

bcaimano:
@dalyd, sure, splitting out the cleanup seems like a decent idea.

scripts/dry-run-workloads.sh
```diff
@@ -0,0 +1,24 @@
...
+
+CMD=(
+    "${GENNY:-genny}"
+    --dry-run
+)
+
+# Find all of the workload yamls at the specified directory
+WORKLOAD_DIR="${1}"
+WORKLOAD_LIST="${BUILD_DIR}/workload-list.txt"
+find "${WORKLOAD_DIR}" \( -iname "*.yml" -or -iname "*.yaml" \) -print >"${WORKLOAD_LIST}"
```
rtimmons: 1. Similar to above, the script can know in advance where workload files live I don't know if there's a use-case for wanting to run it against a different directory?
2. we only use the `.yml` suffix so no need to `-or`
3. why cat to a separate file? should be able to just do this "in-memory"?

src/cast_core/src/actors/Loader.cpp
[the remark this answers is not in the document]
rtimmons: hm ya I agree it's better to be explicit but I'm afraid of putting something like this into the skeleton Actor created by `create-new-actor.sh` because rvalues and `std::move` are relatively "scary" parts of modern C++; I fear they'll cause a newbie to throw up their hands. Feedback from Dan was to reduce the C++ complexity as much as possible. TLDR is I'm fine with this for now but I'm hopeful we can eliminate the need for some things like this by maybe pushing some of the complexity into framework code.

src/driver/test/Workload.yml
```diff
@@ -6,7 +6,7 @@ Actors:
     Threads: 2
     Phases:
     - Message: Hello Phase 0 🐳
-      Duration: 50
+      Duration: 50 milliseconds
```
rtimmons: noice

evergreen.yml
bcaimano: I'm mildly eyeing the cases where a) we have certain workloads that we for some reason don't want to test and b) we're separating our workloads somehow from our testing code. Essentially, I view dist as a mutable artifact and source as a fact of the current evergreen task. It's not very important right now.

scripts/dry-run-workloads.sh
bcaimano: 1. Well, if someone decides to maintain their own folder of workloads, this script will still work for them.
2. I mean, we don't enforce the extension as far as I can tell. I'm not sure it's worth enforcing either, yaml is the official extension -- https://yaml.org/faq.html.
3. I tried that. And amazon linux 2 shit the bed. I dunno.

evergreen.yml
rtimmons: Okay seems reasonable

PR #124, opened 2019-01-16
[no-merge] TIG-1359 update README for vcpkg

guoyr:
Updated the readme and use vcpkg for a uniform genny development experience across all platforms.

You can view the rendered version from my fork: https://github.com/guoyr/genny/tree/tig1359

Not ready for prime time yet but should get us over the line for Ubuntu 17.10. Note that the toolchain *ONLY* works on Ubuntu 17.10 and not 16.04 because of system library version changes. I have not tested 18.04.

If it doesn't work on 18.04. You can just recompile the c++ driver and its dependencies:
1. `cd vcpkg` // the tarball from s3 extracts into this folder
2. `./vcpkg --triplet x64-linux-shared remove openssl --recurse` // quicker than deleting each manually individually
3. `./vcpkg --tirplet x64-linux-shared install libmongocxx`

@henrikingo Feel free to ping me anytime if you run into any issues!

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,126 +6,46 @@ C++17.
 
 ## Build and install
 
-### macOS
+1. Install the development tools for your OS.
 
-1. [Download XCode 10](https://developer.apple.com/download/) (around 10GB) and install.
-2. Drag `Xcode.app` into `Applications`. For some reason the installer may put it in `~/Downloads`.
-3. Run the below shell (requires [`brew`](https://brew.sh/))
+    - Ubuntu 16.04+: `sudo apt install build-essential`
+    - Red Hat/CentOS 7+: `sudo yum groupinstall "Development Tools"`
+    - Arch: Grab a beer. Everything should already be set up.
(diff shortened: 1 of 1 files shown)


guoyr:
Oh there's a problem with CATCH2's regex parsing that causes one test to fail with `regex_error`. I'm still looking into the root cause but it shouldn't affect anything within the `genny` binary.

guoyr:
The regex issue is addressed in #128

README.md
```diff
@@ -6,126 +6,46 @@ C++17.
...
-    build-essential \
-    cmake \
-    software-properties-common \
-    clang-6.0 \ # optional
-    libboost-all-dev
+1. Install the development tools for your OS.
+
+    - Ubuntu 16.04+: `sudo apt install build-essential`
+    - Red Hat/CentOS 7+: `sudo yum groupinstall "Development Tools"`
+    - Arch: Grab a beer. Everything should already be set up.
```
rtimmons: 💥

PR #129, opened 2019-01-16
TIG-1366 Make genny tarballs publicly readable

bcaimano:
It seems that s3Copy wasn't setting permissions correctly for what it uploaded. I've switched to fetch/put and it seems much easier to reason about.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -126,7 +126,7 @@ tasks:
         install_dir: dist
         short_name: grpc
   - func: f_compile
-  - func: f_push_revision
+  - func: f_push_patch
     vars:
         display_name: Genny
         short_name: genny
@@ -240,7 +240,7 @@ functions:
   #  short_name: The name for the folder and prefix for the tarball to push
   #  display_name: The name to show in evergreen
   ###
-  f_push_revision:
+  f_push_patch:
   - command: archive.targz_pack
     params:
       source_dir: ${install_dir|dist}
@@ -259,11 +259,11 @@ functions:
       aws_key: ${aws_key}
       aws_secret: ${aws_secret}
       local_file: ${short_name}.tgz
-      remote_file: ${project}/${short_name}/${short_name}-${build_variant}-patch_${build_id}.tgz
+      remote_file: ${project}/patches/${build_id}/${short_name}.tgz
       bucket: mciuploads
       permissions: public-read
       content_type: ${content_type|application/gzip}
-      display_name: "${display_name} - Patch ${build_id}"
+      display_name: "${display_name} Patch"
 
   ##
   # Copy the library tarball for a given revision to the official path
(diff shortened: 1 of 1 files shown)


evergreen.yml
```diff
@@ -259,11 +259,11 @@ functions:
       aws_key: ${aws_key}
       aws_secret: ${aws_secret}
       local_file: ${short_name}.tgz
-      remote_file: ${project}/${short_name}/${short_name}-${build_variant}-patch_${build_id}.tgz
+      remote_file: ${project}/patches/${build_id}/${short_name}.tgz
       bucket: mciuploads
       permissions: public-read
       content_type: ${content_type|application/gzip}
-      display_name: "${display_name} - Patch ${build_id}"
+      display_name: "${display_name} Patch"
```
bcaimano: Somehow I missed how verbose the build id is. I've switched to a subfolder and a simple name.

visemet:
> It seems that s3Copy wasn't setting permissions correctly for what it uploaded.

This doesn't sound correct me. Evergreen uses [`BucketCannedACLPublicRead="public-read"`](https://godoc.org/github.com/hashicorp/aws-sdk-go/gen/s3#pkg-constants) when [copying the file to the destination bucket](https://github.com/evergreen-ci/evergreen/blob/bce89df6602e84543edd7c5250e66d3e16ee54d0/vendor/github.com/aws/aws-sdk-go/service/s3/api.go#L20971).

```golang
destOpts := pail.S3Options{
    Credentials: pail.CreateAWSCredentials(s3CopyReq.AwsKey, s3CopyReq.AwsSecret, ""),
    Region:      region,
    Name:        s3CopyReq.S3DestinationBucket,
    Permission:  s3.BucketCannedACLPublicRead,
}
```

> I've switched to fetch/put and it seems much easier to reason about.

This is potentially dangerous. The `s3Copy.copy` command exists because Evergreen takes steps to ensure that an earlier revision cannot overwrite the contents of the S3 key written by a later revision.

The task logs show that Evergreen isn't performing the second `s3Copy.copy` command [for that reason](https://github.com/evergreen-ci/evergreen/blob/bce89df6602e84543edd7c5250e66d3e16ee54d0/service/api_plugin_s3copy.go#L61-L76) though it seems like https://s3.amazonaws.com/mciuploads/genny/genny/genny-amazon2-aa77667d5f09be64fff1fe542a5111013acb618a.tgz returning "Access Denied" is really saying the first upload never happened.

```
[2019/01/16 16:52:06.727] Running command 's3Copy.copy' in "f_push_official" (step 1.1 of 1)
[2019/01/16 16:52:13.317] Running command 's3Copy.copy' in "f_push_official" (step 1.2 of 1)
[2019/01/16 16:52:13.329] s3Copy response: noop, file mciuploads/genny/genny/genny-amazon2.tgz exists
```

bcaimano:
@visemet, I believe in your wisdom in this domain. Can we take this offline tomorrow to talk about more permanent fixes? I'm worried that this could happen so silently with s3copy.

visemet:
> @visemet, I believe in your wisdom in this domain. Can we take this offline tomorrow to talk about more permanent fixes? I'm worried that this could happen so silently with s3copy.

Absolutely.

PR #131, opened 2019-01-16
TIG-1322: RunCommand does collStats after replSetStepDown commands

rtimmons:
Arguably not the most elegant solution but I think this is the "expected' behavior of anyone doing a `{replSetStepDown:X}` command.

I'm still doing some light testing to this locally and will run in a patch-build with the `RunCommand.yml` file as the workload.

### DIFF
--- a/src/cast_core/include/cast_core/actors/RunCommand.hpp
+++ b/src/cast_core/include/cast_core/actors/RunCommand.hpp
@@ -59,10 +59,6 @@ namespace genny::actor {
 
 class RunCommand : public Actor {
 public:
-    class Operation;
-    struct PhaseState;
-
-public:
     explicit RunCommand(ActorContext& context);
     ~RunCommand() = default;
 
@@ -73,6 +69,9 @@ public:
     void run() override;
 
 private:
+    class Operation;
+    struct PhaseState;
+
     genny::DefaultRandom _rng;
     mongocxx::pool::entry _client;
 
--- a/src/cast_core/include/cast_core/actors/RunCommand.hpp
+++ b/src/cast_core/include/cast_core/actors/RunCommand.hpp
@@ -69,13 +69,11 @@ public:
     void run() override;
 
 private:
-    class Operation;
-    struct PhaseState;
-
     genny::DefaultRandom _rng;
     mongocxx::pool::entry _client;
 
-    PhaseLoop<PhaseState> _loop;
+    struct PhaseConfig;
+    PhaseLoop<PhaseConfig> _loop;
 };
 
 
--- a/src/cast_core/src/actors/RunCommand.cpp
+++ b/src/cast_core/src/actors/RunCommand.cpp
@@ -35,20 +35,56 @@
 
 namespace {
 
-void runStepdown(mongocxx::database& database, bsoncxx::document::view& view) {
+/**
+ * @private
+ * @param command passed to run_command()
+ * @return if the command requires waiting for a stepdown/election to complete before continuing.s
+ */
+bool requiresWaitForStepdown(bsoncxx::document::view& command) {
+    return command.find("replSetStepDown") != command.end();
+}
+
+/**
+ * @private
+ * @param exception exception received from mongocxx::database::run_command()
+ * @return if the exception is a result of a connection error or timeout
+ */
+bool isTimeout(mongocxx::operation_exception& exception) {
+    auto what = std::string{exception.what()};
+    return what.find("socket error or timeout") != std::string::npos;
+}
+
+/**
+ * @private
+ * BSON document that when given to run_command requires any rs elections
+ * to be complet.
+ */
+static auto commandRequiringElectionCompleted = []() -> bsoncxx::document::value {
+    auto builder = bsoncxx::builder::stream::document{};
+    // `{ "dbStats": 1 }`
+    auto request = builder << "dbStats" << 1 << bsoncxx::builder::stream::finalize;
+    return request;
+}();
+
+/**
+ * Runs the given command, expects a timeout, and then awaits all replset
+ * elections to complete.
+ *
+ * @private
+ * @param database database on which to run `run_command()`.
+ * @param command the command to run
+ */
+void runThenAwaitElections(mongocxx::database& database, bsoncxx::document::view& command) {
     try {
-        bsoncxx::document::value val = database.run_command(view);
-    } catch (mongocxx::operation_exception& x) {
-        auto what = std::string{x.what()};
-        if (what.find("socket error or timeout") == std::string::npos) {
+        database.run_command(command);
+        throw std::logic_error("Stepdown should throw operation exception");
+    } catch (mongocxx::operation_exception& exception) {
+        if (!isTimeout(exception)) {
             throw;
         }
-        auto builder = bsoncxx::builder::stream::document{};
-        auto request = builder << "dbStats" << 1 << bsoncxx::builder::stream::finalize;
         try {
-            auto stats = database.run_command(request.view());
-        } catch (mongocxx::operation_exception& x2) {
-            BOOST_LOG_TRIVIAL(error) << "Couldn't run stats " << x2.what();
+            auto stats = database.run_command(commandRequiringElectionCompleted.view());
+        } catch (mongocxx::operation_exception& statsException) {
             throw;
         }
     }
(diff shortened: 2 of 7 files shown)


src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <cast_core/actors/RunCommand.hpp>
```
bcaimano: So usually you include the declaration header first before system headers because it makes the include order more predictable. (This is suggested in the google style guide for instance.)

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
+}
+
+/**
+ * @private
+ * @param exception exception received from mongocxx::database::run_command()
+ * @return if the exception is a result of a connection error or timeout
+ */
+bool isTimeout(mongocxx::operation_exception& exception) {
+    auto what = std::string{exception.what()};
+    return what.find("socket error or timeout") != std::string::npos;
```
bcaimano: Well, due to the ambiguity of the c-driver, this could be other things besides a timeout.

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
+ * elections to complete.
+ *
+ * @private
+ * @param database database on which to run `run_command()`.
+ * @param command the command to run
+ */
+void runThenAwaitElections(mongocxx::database& database, bsoncxx::document::view& command) {
+    try {
+        database.run_command(command);
+        throw std::logic_error("Stepdown should throw operation exception");
```
bcaimano: Ugh, um, can this finally be the moment we have a `genny::RuntimeError`?

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
 #include <boost/log/trivial.hpp>
 
+#include <cast_core/actors/RunCommand.hpp>
+
 #include <gennylib/ExecutionStrategy.hpp>
 #include <gennylib/MongoException.hpp>
 #include <gennylib/v1/RateLimiter.hpp>
 #include <gennylib/value_generators.hpp>
 
+namespace {
```
bcaimano: If you put this functionality into a widget in the `v1` namespace, you can run it in `test_gennylib_with_server` without having to construct an actor.

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
+
+/**
+ * @private
+ * @param command passed to run_command()
+ * @return if the command requires waiting for a stepdown/election to complete before continuing.s
+ */
+bool requiresWaitForStepdown(bsoncxx::document::view& command) {
+    return command.find("replSetStepDown") != command.end();
```
bcaimano: I'm pretty nervous about arbitrarily changing behavior based on the underlying command. That's a good way to make users never trust your config language again. If this *has* to be bundled into the existing RunCommand actor instead of a specialization of it, would you please make it activated via an additional flag in the operation?

(I can kinda see the point of inline to RunCommand, but it brings me back to my angst about how it makes more sense to me for each phase to potentially be a different actor that may share state with actors on previous phases.)

bcaimano: Putting a bit more thought at this, it's perfectly fine to generate a warning that `replSetStepDown` without that flag will behave in a weird way. We can even do that in the config stage.

visemet: This was at my request. Ryan originally proposed exposing a boolean flag on the RunCommand actor to control this behavior, but it seemed to me like something we'd eventually remove in favor of having this behavior managed by a new (but currently nonexistent) Nemesis actor.

src/cast_core/src/actors/RunCommand.cpp
rtimmons: ya...are you suggesting to change anything here though?

bcaimano: Well, ideally I'd like `isTimeout` to be renamed `isNetworkFailure`.

src/cast_core/src/actors/RunCommand.cpp
rtimmons: ya..considered it..but I decided to just leave as-is the logic is simple enough to test with the Actor itself

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
+
+/**
+ * @private
+ * BSON document that when given to run_command requires any rs elections
+ * to be complet.
+ */
+bsoncxx::document::value commandRequiringElectionCompleted = []() {
+    auto builder = bsoncxx::builder::stream::document{};
+    // `{ "dbStats": 1 }`
+    auto request = builder << "dbStats" << 1 << bsoncxx::builder::stream::finalize;
```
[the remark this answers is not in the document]
rtimmons: 👍 Changed to call `collStats`

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -28,11 +26,76 @@
...
+    auto what = std::string{exception.what()};
+    return what.find("socket error or timeout") != std::string::npos;
+}
+
+/**
+ * @private
+ * BSON document that when given to run_command requires any rs elections
+ * to be complet.
+ */
+bsoncxx::document::value commandRequiringElectionCompleted = []() {
```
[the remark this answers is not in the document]
rtimmons: Changed everything to `*stepdown*`

rtimmons:
Thanks for review @bcaimano and @visemet. I've changed this to call `collStats` and I've made the await-stepdown configurable via `OperationAwaitStepdown: true/false`.

I also did some slight refactoring to the phase/operation structs internal to `RunCommand.cpp` since I got lost more than once when adding the new config param (although I'm admittedly pretty dense at tracing plumbing things through). Refactored version is a bit more verbose in a couple places but I think a bit more obvious. Subjective for sure.

src/cast_core/include/cast_core/actors/RunCommand.hpp
```diff
@@ -76,7 +72,8 @@ class RunCommand : public Actor {
     genny::DefaultRandom _rng;
     mongocxx::pool::entry _client;
 
-    PhaseLoop<PhaseState> _loop;
+    struct PhaseConfig;
```
bcaimano: I don't think I'm ever going to agree with calling these things configs. If it has a run method, it's not a config. Still, not gonna die on a molehill.

src/driver/test/RunCommand.yml
```diff
@@ -34,3 +34,15 @@ Actors:
       OperationName: RunCommand
       OperationCommand:
         currentOp: 1
+- Name: Stepdown
```
bcaimano: Just as a heads up, I'm not positive what the intention is for these example workloads, but the stepdown command will probably prevent this one from being run on a standalone with success.

rtimmons:
> I can't help but notice that the "refactoring" was mostly resetting a bunch of stuff back to the massive param lists I tried to move this class away from. I'm sad to see this happen but ultimately it doesn't affect how this class runs.

So there were 5 places where configs were stored/parsed:

- PhaseConfig (PhaseState)
- Operation
- RunCommandConfig
- RunCommandConfig::Operation
- Fixture

This got me super confused when I was trying to add the new configuration param. (And having the `RunCommandConfig` in a separate header file made it even harder to track it down - I'm not sure I'd still recommend that for new Actors given that it separates two things that are coupled by design but willing to see how it plays out a bit longer.) I threw my hands up and decided to go homocidal on one of these 5.

The longer param lists are unfortunate but I think a side-effect of there being so many config classes. I think ideally there would just be one for the Phase and one for the Operation and they would just construct themselves from the `genny::PhaseContext`, but I didn't have the fortitude to do a massive change like that.

Happy to keep the debate open though. I suspect this Actor will see one or two more refactors before we find the right pattern. It's important to get this Actor right because it will be a copy-pasta source for other Actors.

src/cast_core/include/cast_core/actors/RunCommand.hpp
rtimmons: ya I see what you're saying - but `State` also didn't feel right because it's not the state of the current phase or anything, it's a pre-configured per-phase mini-actor-lite thing...maybe `PhaseAction` or something? idk. open to other names but we should be consistent in what our Actors call it to make it less confusing for new Actor-authors.

PR #132, opened 2019-01-16
TIG-1187 Crud Actor

jasonjhchan:
Extremely WIP, putting it up to get another set of eyes on it because there definitely will be room for improvements. The skeleton is there for the 'aggregation' and 'bulk_write' commands which seem to be the trickiest part of the actor. 

Still TBD:
How to bind YAML syntax for operation Options.

The current syntax is reflected in the YAML configuration defined in the first test case of CrudActor_test.cpp.

### DIFF
--- a/src/cast_core/include/cast_core/actors/CrudActor.hpp
+++ b/src/cast_core/include/cast_core/actors/CrudActor.hpp
@@ -1,0 +1,71 @@
+#ifndef HEADER_338F3734_B052_443C_A358_60215BA9D5A0_INCLUDED
+#define HEADER_338F3734_B052_443C_A358_60215BA9D5A0_INCLUDED
+
+#include <string_view>
+
+#include <mongocxx/database.hpp>
+#include <mongocxx/pool.hpp>
+
+#include <gennylib/Actor.hpp>
+#include <gennylib/DefaultRandom.hpp>
+#include <gennylib/ExecutionStrategy.hpp>
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+#include <gennylib/value_generators.hpp>
+
+#include <cast_core/config/RunCommandConfig.hpp>
+
+namespace genny::actor {
+
+/**
+ * TODO: document me
+ */
+class CrudActor : public Actor {
+public:
+    class Operation {
+
+    public:
+        using OpCallback = std::function<void(CrudActor::Operation&)>;
+        using Options = config::RunCommandConfig::Operation;
+        struct Fixture {
+            PhaseContext& phaseContext;
+            mongocxx::database database;
+            OpCallback& op;
+        };
+
+        OpCallback& callback;
+        const YAML::Node& opCommand;
+
+    public:
+        Operation(Fixture fixture, const YAML::Node& opCommand, Options opts);
+        void run(const mongocxx::client_session& session);
+
+    private:
+        mongocxx::database _database;
+        std::unique_ptr<value_generators::DocumentGenerator> _doc;
+        Options _options;
+        std::optional<metrics::Timer> _timer;
+    };
+
+public:
+    explicit CrudActor(ActorContext& context);
+    ~CrudActor() = default;
+
+    static std::string_view defaultName() {
+        return "CrudActor";
+    }
+
+    void run() override;
+
+private:
+    ExecutionStrategy _strategy;
+    mongocxx::pool::entry _client;
+    genny::DefaultRandom _rng;
+
+    struct PhaseConfig;
+    PhaseLoop<PhaseConfig> _loop;
+};
+
+}  // namespace genny::actor
+
+#endif  // HEADER_338F3734_B052_443C_A358_60215BA9D5A0_INCLUDED
--- a/src/cast_core/include/cast_core/actors/CrudActor.hpp
+++ b/src/cast_core/include/cast_core/actors/CrudActor.hpp
@@ -21,31 +21,9 @@ namespace genny::actor {
  * TODO: document me
  */
 class CrudActor : public Actor {
+
 public:
-    class Operation {
-
-    public:
-        using OpCallback = std::function<void(CrudActor::Operation&)>;
-        using Options = config::RunCommandConfig::Operation;
-        struct Fixture {
-            PhaseContext& phaseContext;
-            mongocxx::database database;
-            OpCallback& op;
-        };
-
-        OpCallback& callback;
-        const YAML::Node& opCommand;
-
-    public:
-        Operation(Fixture fixture, const YAML::Node& opCommand, Options opts);
-        void run(const mongocxx::client_session& session);
-
-    private:
-        mongocxx::database _database;
-        std::unique_ptr<value_generators::DocumentGenerator> _doc;
-        Options _options;
-        std::optional<metrics::Timer> _timer;
-    };
+    class Operation;
 
 public:
     explicit CrudActor(ActorContext& context);
--- a/src/cast_core/include/cast_core/actors/CrudActor.hpp
+++ b/src/cast_core/include/cast_core/actors/CrudActor.hpp
@@ -18,7 +18,54 @@
 namespace genny::actor {
 
 /**
- * TODO: document me
+ * CrudActor is an actor that performs CRUD operations on a collection and has the ability to start
+ * and commit transactions. This actor aims to support the operations in the mongocxx-driver
+ * collections API. The actor supports the usage of a list of operations for a single phase.
+ *
+ * Example:
+ *
+ * ```yaml
+ * Actors:
+ * - Name: BulkWriteInTransaction
+ *   Type: CrudActor
+ *   Database:
+ *   Type: CrudActor
+ *   Database: testdb
+ *   ExecutionStrategy:
+ *     ThrowOnFailure: true
+ *   Phases:
+ *   - Repeat: 1
+ *     Collection: test
+ *     Operations:
+ *     - OperationName: startTransaction
+ *       OperationCommand:
+ *         Options:
+ *           WriteConcern:
+ *             Level: majority
+ *             Journal: true
+ *           ReadConcern:
+ *             Level: snapshot
+ *           ReadPreference:
+ *             ReadMode: primaryPreferred
+ *             MaxStalenessSeconds: 1000
+ *     - OperationName: bulkWrite
+ *       OperationCommand:
+ *         WriteOperations:
+ *         - WriteCommand: insertOne
+ *           Document: { a: 1 }
+ *         - WriteCommand: updateOne
+ *           Filter: { a: 1 }
+ *           Update: { $set: { a: 5 } }
+ *         Options:
+ *           Ordered: true
+ *           WriteConcern:
+ *             Level: majority
+ *         OnSession: true
+ *     - OperationName: commitTransaction
+ *   - Repeat: 1
+ *     Collection: test
+ *     Operation:
+ *       OperationName: drop
  */
 class CrudActor : public Actor {
 
@@ -36,7 +83,6 @@ public:
     void run() override;
 
 private:
-    ExecutionStrategy _strategy;
     mongocxx::pool::entry _client;
     genny::DefaultRandom _rng;
 
--- a/src/cast_core/include/cast_core/actors/CrudActor.hpp
+++ b/src/cast_core/include/cast_core/actors/CrudActor.hpp
@@ -7,11 +7,11 @@
 #include <mongocxx/pool.hpp>
 
 #include <gennylib/Actor.hpp>
-#include <gennylib/DefaultRandom.hpp>
 #include <gennylib/ExecutionStrategy.hpp>
 #include <gennylib/PhaseLoop.hpp>
 #include <gennylib/context.hpp>
-#include <gennylib/value_generators.hpp>
+#include <value_generators/DefaultRandom.hpp>
+#include <value_generators/value_generators.hpp>
 
 #include <cast_core/config/RunCommandConfig.hpp>
 
--- a/src/cast_core/src/actors/CrudActor.cpp
+++ b/src/cast_core/src/actors/CrudActor.cpp
@@ -1162,6 +1162,7 @@ std::unordered_map<std::string, OpCallback&> opConstructors = {
     {"insertOne", createInsertOneOp},
     {"deleteOne", createDeleteOneOp},
     {"deleteMany", createDeleteManyOp},
+    {"updateOne", createUpdateOneOp},
     {"updateMany", createUpdateManyOp},
     {"replaceOne", createReplaceOneOp}};
 };  // namespace
--- a/src/cast_core/src/actors/CrudActor.cpp
+++ b/src/cast_core/src/actors/CrudActor.cpp
@@ -1,4 +1,5 @@
 #include <cast_core/actors/CrudActor.hpp>
+
 #include <chrono>
 #include <memory>
 
@@ -8,14 +9,12 @@
 #include <mongocxx/collection.hpp>
 
 #include <boost/log/trivial.hpp>
-
-#include <boost/log/trivial.hpp>
 #include <bsoncxx/json.hpp>
-#include <gennylib/Cast.hpp>
-#include <gennylib/context.hpp>
 
+#include <gennylib/Cast.hpp>
 #include <gennylib/ExecutionStrategy.hpp>
 #include <gennylib/MongoException.hpp>
+#include <gennylib/context.hpp>
 #include <gennylib/conventions.hpp>
 
 using BsonView = bsoncxx::document::view;
@@ -44,7 +43,7 @@ struct convert<mongocxx::read_preference> {
         }
         auto maxStaleness = rhs.max_staleness();
         if (maxStaleness) {
-            node["MaxStalenessSeconds"] = genny::TimeSpec(*maxStaleness).count();
+            node["MaxStalenessSeconds"] = genny::TimeSpec(*maxStaleness);
         }
         return node;
     }
@@ -58,7 +57,6 @@ struct convert<mongocxx::read_preference> {
             return false;
         }
         auto readMode = node["ReadMode"].as<std::string>();
-        rhs = mongocxx::read_preference{};
         if (readMode == "primary") {
             rhs.mode(ReadMode::k_primary);
         } else if (readMode == "primaryPreferred") {
@@ -73,8 +71,8 @@ struct convert<mongocxx::read_preference> {
             return false;
         }
         if (node["MaxStalenessSeconds"]) {
-            auto maxStaleness = node["MaxStalenessSeconds"].as<int>();
-            rhs.max_staleness(std::chrono::seconds(maxStaleness));
+            auto maxStaleness = node["MaxStalenessSeconds"].as<genny::TimeSpec>();
+            rhs.max_staleness(std::chrono::seconds{maxStaleness});
         }
         return true;
     }
@@ -121,9 +119,7 @@ struct convert<mongocxx::write_concern> {
     using WriteConcern = mongocxx::write_concern;
     static Node encode(const WriteConcern& rhs) {
         Node node;
-        auto timeout = rhs.timeout();
-        auto timeoutCount = genny::TimeSpec{timeout}.count();
-        node["TimeoutMillis"] = timeoutCount;
+        node["TimeoutMillis"] = genny::TimeSpec{rhs.timeout()};
 
         auto journal = rhs.journal();
         node["Journal"] = journal;
@@ -149,19 +145,21 @@ struct convert<mongocxx::write_concern> {
             return false;
         }
         auto level = node["Level"].as<std::string>();
-        bool isNum = level.find_first_not_of("0123456789") == std::string::npos;
-        if (isNum) {
-            auto levelNum = node["Level"].as<int>();
-            rhs.nodes(levelNum);
-        } else if (level == "majority") {
-            rhs.majority(std::chrono::milliseconds{0});
-        } else {
-            // writeConcern level must be of valid integer or 'majority'.
-            return false;
+        try {
+            auto level = node["Level"].as<int>();
+            rhs.nodes(level);
+        } catch (const BadConversion& e) {
+            auto level = node["Level"].as<std::string>();
+            if (level == "majority") {
+                rhs.majority(std::chrono::milliseconds{0});
+            } else {
+                // writeConcern level must be of valid integer or 'majority'.
+                return false;
+            }
         }
         if (node["TimeoutMillis"]) {
-            auto timeout = node["TimeoutMillis"].as<int>();
-            rhs.timeout(std::chrono::milliseconds(timeout));
+            auto timeout = node["TimeoutMillis"].as<genny::TimeSpec>();
+            rhs.timeout(std::chrono::milliseconds{timeout});
         }
         if (node["Journal"]) {
             auto journal = node["Journal"].as<bool>();
@@ -275,8 +273,7 @@ struct convert<mongocxx::options::count> {
         }
         auto maxTime = rhs.max_time();
         if (maxTime) {
-            auto maxTimeSpec = genny::TimeSpec(*maxTime);
-            node["MaxTimeMillis"] = maxTimeSpec.count();
+            node["MaxTimeMillis"] = genny::TimeSpec(*maxTime);
         }
         return node;
     }
@@ -295,8 +292,8 @@ struct convert<mongocxx::options::count> {
             rhs.limit(limit);
         }
         if (node["MaxTimeMillis"]) {
-            auto maxTime = node["MaxTimeMillis"].as<int>();
-            rhs.max_time(std::chrono::milliseconds(maxTime));
+            auto maxTime = node["MaxTimeMillis"].as<genny::TimeSpec>();
+            rhs.max_time(std::chrono::milliseconds{maxTime});
         }
         if (node["ReadPreference"]) {
             auto readPref = node["ReadPreference"].as<mongocxx::read_preference>();
@@ -339,15 +336,6 @@ namespace genny::actor {
 
 namespace {
 
-enum class StageType {
-    Bucket,
-    Count,
-    // TODO: fill in rest of aggregate stages.
-};
-
-std::unordered_map<std::string, StageType> stringToStageType = {{"bucket", StageType::Bucket},
-                                                                {"count", StageType::Count}};
-
 struct BaseOperation {
     virtual void run(mongocxx::client_session& session) = 0;
     virtual ~BaseOperation() = default;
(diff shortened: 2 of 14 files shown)


jasonjhchan:
@visemet In case you wanted to take a preliminary look - I managed to get the write concern tests working. For some reason, I don't seem to have write access to the repository and I can't add anyone as a reviewer.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+            auto levelNum = node["Level"].as<int>();
+            rhs.nodes(levelNum);
+        } else if (level == "majority") {
+            rhs.majority(std::chrono::milliseconds{0});
+        } else {
+            // writeConcern level must be of valid integer or 'majority'.
+            return false;
+        }
+        if (node["TimeoutMillis"]) {
+            auto timeout = node["TimeoutMillis"].as<int>();
```
visemet: Per genny's conventions of having time-based units be a first-class concept, you'll want to have this be a `TimeSpec` rather than an `int`. This can be achieved by delegating to `TimeSpec`'s decoder.

```cpp
if (node["Timeout"]) {
    auto timeout = node["Timeout"].as<genny::TimeSpec>();
    rhs.timeout(std::chrono::milliseconds{timeout});
}
```

Similar for the `encode()` function  you'll want to do `node["Timeout"] = genny::TimeSpec{rhs.timeout()}`.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+            rhs.mode(ReadMode::k_secondary);
+        } else if (readMode == "secondaryPreferred") {
+            rhs.mode(ReadMode::k_secondary_preferred);
+        } else if (readMode == "nearest") {
+            rhs.mode(ReadMode::k_nearest);
+        } else {
+            return false;
+        }
+        if (node["MaxStalenessSeconds"]) {
+            auto maxStaleness = node["MaxStalenessSeconds"].as<int>();
```
visemet: Per genny's conventions of having time-based units be a first-class concept, you'll want to have this be a `TimeSpec` rather than an `int`. See my later comment for handling `wtimeout`.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+            auto h = node["Hint"].as<std::string>();
+            auto hint = mongocxx::hint(h);
+            rhs.hint(hint);
+        }
+        if (node["Limit"]) {
+            auto limit = node["Limit"].as<int>();
+            rhs.limit(limit);
+        }
+        if (node["MaxTimeMillis"]) {
+            auto maxTime = node["MaxTimeMillis"].as<int>();
```
visemet: Per genny's conventions of having time-based units be a first-class concept, you'll want to have this be a `TimeSpec` rather than an `int`. See my earlier comment for handling `wtimeout`.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+    static bool decode(const Node& node, WriteConcern& rhs) {
+        if (!node.IsMap()) {
+            return false;
+        }
+        if (!node["Level"]) {
+            // writeConcern must specify the write concern level.
+            return false;
+        }
+        auto level = node["Level"].as<std::string>();
+        bool isNum = level.find_first_not_of("0123456789") == std::string::npos;
```
visemet: [question] is there a reason you chose to check if the value is a number yourself instead of letting yaml-cpp handle it for you?

```cpp
try {
    auto level = node["Level"].as<int>();
    ...
} catch (const BadConversion& e) {
    auto level = node["Level"].as<std::string>();
    ...
}
```

https://stackoverflow.com/questions/19994312/obtain-type-of-value-stored-in-yamlnode-for-yaml-cpp

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+
+struct BaseOperation {
+    virtual void run(const mongocxx::client_session& session) = 0;
+    virtual ~BaseOperation() = default;
+};
+
+using OpCallback = std::function<std::unique_ptr<BaseOperation>(
+    YAML::Node, bool, mongocxx::collection, genny::DefaultRandom&)>;
+
+struct AggregateOperation : public BaseOperation {
```
visemet: I'm dubious of using [`mongocxx::pipeline`](http://mongocxx.org/api/current/classmongocxx_1_1pipeline.html) to run an aggregation pipeline because it doesn't have an escape hatch to run an aggregation stage that isn't known to the C++ driver. This seems like something that would limit the ability for the Query team to use genny to write performance tests for new aggregation stages. My preference would be to do away with the `AggregateOperation` class for the time being and revisit it in a separate pull request. CC @rtimmons

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+    }
+
+    void run(const mongocxx::client_session& session) override {
+        auto bulk = (_onSession) ? _collection.create_bulk_write(session, _options)
+                                 : _collection.create_bulk_write(_options);
+        for (auto&& writeOp : _writeOps) {
+            bulk.append(writeOp);
+        }
+        auto result = bulk.execute();
+        if (!result) {
```
visemet: It looks like [the `stdx::optional<result::bulk_write>` is only disengaged (i.e. `== std::nullopt`) when using unacknowledged writes](https://github.com/mongodb/mongo-cxx-driver/blob/r3.4.0/src/mongocxx/bulk_write.cpp#L175-L182). I think this if-statement is safe to remove.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,615 @@
...
+    PhaseConfig(PhaseContext& phaseContext, genny::DefaultRandom& rng, const mongocxx::database& db)
+        : collection{db[phaseContext.get<std::string>("Collection")]},
+          options{ExecutionStrategy::getOptionsFrom(phaseContext, "ExecutionsStrategy")} {
+
+        auto addOperation = [&](YAML::Node node) {
+            auto yamlCommand = node["OperationCommand"];
+            auto opName = node["OperationName"].as<std::string>();
+            auto onSession = yamlCommand["Session"] && yamlCommand["Session"].as<bool>();
+
+            // Grab the appropriate
```
visemet: Finish the thought?

src/driver/test/CrudActor.yml
```diff
@@ -0,0 +1,24 @@
...
+        WriteOperations:
+        - WriteCommand: insertOne
+          Document: { a: 1 }
+        - WriteCommand: updateOne
+          Filter: { a: 1 }
+          Update: { $set: { a: 5 } }
+          Options:
+            WriteConcern:
+              Level: majority
+              TimeoutMillis: 5000
```
visemet: [lint] Add a newline to the end of this file and address the other whitespace-related errors in CrudActor_test.cpp.

```
$ git diff upstream/master... --check
src/driver/test/CrudActor.yml:2: trailing whitespace.
+  
src/gennylib/test/CrudActor_test.cpp:653: trailing whitespace.
+                  Options: 
src/gennylib/test/CrudActor_test.cpp:691: trailing whitespace.
+                  Options: 
```

src/gennylib/src/context.cpp
```diff
@@ -58,9 +59,17 @@ WorkloadContext::WorkloadContext(YAML::Node node,
 
     _clientPool = poolFactory.makePool();
 
+    // TODO: Remove hasApmOpts when CDRIVER-2931 is resolved.
+    auto apmOpts = clientOpts.apm_opts();
+    bool hasApmOpts = false;
```
visemet: You could condense lines 64-67 to just `auto hasApmOpts = bool{clientOpts.apm_opts()}`.

src/gennylib/include/gennylib/PoolFactory.hpp
```diff
@@ -45,7 +45,7 @@ class PoolFactory {
     };
 
 public:
-    PoolFactory(std::string_view uri);
+    PoolFactory(std::string_view uri, mongocxx::options::client& clientOpts);
```
visemet: Since `PoolFactory` owns the `mongocxx::options::client` options you should just take it by value and `std::move()` it into the constructor for places it is being used. I'd recommend having it default to `mongocxx::options::client{}` and discarding your changes to PoolFactory_test.cpp.

src/gennylib/test/ActorHelper.hpp
```diff
@@ -84,6 +85,7 @@ class ActorHelper {
     std::unique_ptr<WorkloadContext> _wlc;
 
     std::stringstream _metricsOutput;
+    mongocxx::options::client _clientOpts;
```
visemet: Why does the `ActorHelper` class need to store a copy of the `mongocxx::options::client` options?

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+}  // namespace YAML
+
+namespace genny::actor {
+
+namespace {
+
+enum class StageType {
+    Bucket,
+    Count,
+    // TODO: fill in rest of aggregate stages.
```
rtimmons: remove for now since we're only doing these

src/cast_core/src/actors/CrudActor.cpp
rtimmons: ya I would vote for that too

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+    void createOps(const YAML::Node& writeOp, genny::DefaultRandom& rng) {
+        auto writeCommand = writeOp["WriteCommand"].as<std::string>();
+        if (writeCommand == "insertOne") {
+            auto insertDoc = writeOp["Document"];
+            if (!insertDoc) {
+                throw InvalidConfigurationException("'insertOne' expects a 'Document' field.");
+            }
+            auto doc = value_generators::makeDoc(insertDoc, rng);
+            bsoncxx::builder::stream::document document{};
+            doc->view(document);
```
rtimmons: this applies the value generator at ctor time not at runtime

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+        } else if (writeCommand == "replaceOne") {
+            auto filter = writeOp["Filter"];
+            auto replacement = writeOp["Replacement"];
+            if (!filter || !replacement) {
+                throw InvalidConfigurationException(
+                    "'replaceOne' expects 'Filter' and 'Replacement' fields.");
+            }
+            auto filterTemplate = value_generators::makeDoc(filter, rng);
+            bsoncxx::builder::stream::document filterDocument{};
+            filterTemplate->view(filterDocument);
```
rtimmons: similarly here - need to apply the generator at runtime

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+            auto updateTemplate = value_generators::makeDoc(update, rng);
+            bsoncxx::builder::stream::document updateDocument{};
+            updateTemplate->view(updateDocument);
+            _writeOps.push_back(
+                mongocxx::model::update_one(filterDocument.extract(), updateDocument.extract()));
+        } else if (writeCommand == "updateMany") {
```
rtimmons: hrm the number of `else if`s here is kinda worrisome - does it make sense to use the same `map<string,callback>` pattern here as you do for the operation name?

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+                               bool onSession,
+                               mongocxx::collection collection,
+                               genny::DefaultRandom& rng,
+                               metrics::Operation operation) {}
+
+    void run(mongocxx::client_session& session) override {
+        try {
+            session.commit_transaction();
+        } catch (mongocxx::operation_exception& exception) {
+            throw;
```
rtimmons: what benefit do you get by catch and then immediately throw?

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,926 @@
...
+                throw InvalidConfigurationException("'Operations' must be of sequence type.");
+            }
+            for (auto&& op : *operationList) {
+                addOperation(op);
+            }
+        } else if (operationUnit) {
+            addOperation(*operationUnit);
+        } else if (!operationUnit && !operationList) {
+            throw InvalidConfigurationException("No operations found in RunCommand Actor.");
+        }
```
rtimmons: you can use the new `context.getPlural()` method here

src/driver/test/CrudActor.yml
rtimmons: plz extend this example to include the opeartions that are supported and show a variety of value-generator usages as well.

src/gennylib/include/gennylib/context.hpp
```diff
@@ -226,10 +227,11 @@ class PhaseContext;
  */
 class ActorContext final : public V1::ConfigNode {
 public:
-    ActorContext(YAML::Node node, WorkloadContext& workloadContext)
+    ActorContext(YAML::Node node, WorkloadContext& workloadContext, bool hasApmOpts)
```
rtimmons: please document this and/or choose a better name for the param `apm` means nothing to me 😄

src/gennylib/src/context.cpp
```diff
@@ -42,7 +43,7 @@ WorkloadContext::WorkloadContext(YAML::Node node,
     mongocxx::instance::current();
 
     // TODO: make this optional and default to mongodb://localhost:27017
-    auto poolFactory = PoolFactory(mongoUri);
+    auto poolFactory = PoolFactory(mongoUri, clientOpts);
```
rtimmons: I think this muddies how the connection should be configured - previously it was only from the uri and/or yaml configs and PoolFactory worked out how to munge those together. Now there is an additional `clientOpts` in the works. Is there a way we can push this concern into the poolfactory so there's not another layer of indirection?

src/gennylib/src/context.cpp
```diff
@@ -130,6 +139,10 @@ std::unordered_map<PhaseNumber, std::unique_ptr<PhaseContext>> ActorContext::con
 }
 
 mongocxx::pool::entry ActorContext::client() {
+    if (_hasApmOpts) {
+        // TODO: Remove this conditional when CDRIVER-2931 is resolved.
+        return _workload->_clientPool->acquire();
+    }
```
rtimmons: I think you can move this method to `WorkloadContext` and just delegate to it. This would prevent you from having to pass down the boolean.


```c++
mongocxx::pool::entry ActorContext::client() {
  return _workload->client();
}
```

src/gennylib/src/context.cpp
```diff
@@ -58,9 +59,17 @@ WorkloadContext::WorkloadContext(YAML::Node node,
 
     _clientPool = poolFactory.makePool();
 
+    // TODO: Remove hasApmOpts when CDRIVER-2931 is resolved.
+    auto apmOpts = clientOpts.apm_opts();
```
rtimmons: I'm confused - when do we actually set `clientOpts` and when does it get passed in? The driver is the only thing that creates the `WorkloadContext` and that's not being changed here from what I see so it seems like the `apm_opts()` will always just be whatever it is from the default ctor, ya?

src/gennylib/src/PoolFactory.cpp
```diff
@@ -177,7 +177,8 @@ struct PoolFactory::Config {
     };
 };
 
-PoolFactory::PoolFactory(std::string_view rawUri) : _config(std::make_unique<Config>(rawUri)) {}
+PoolFactory::PoolFactory(std::string_view rawUri, mongocxx::options::client& clientOpts)
+    : _config(std::make_unique<Config>(rawUri)), _clientOpts{clientOpts} {}
```
rtimmons: if the only thing you actually care about and use from the client options is the apm flag maybe just pass that boolean in? that way it's obvious that only that flag is being respected.

src/gennylib/src/context.cpp
jasonjhchan: The clientOpts are being set with the apm options here: https://github.com/mongodb/genny/blob/733e2991c30f8215c07d60a952cc502f02222e67/src/gennylib/test/CrudActor_test.cpp#L27-L39. A `SessionTest` object is initialized in the necessary test cases and the client options are then passed in the constructor of ActorHelper.

src/cast_core/include/cast_core/actors/CrudActor.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <gennylib/context.hpp>
+#include <value_generators/DefaultRandom.hpp>
+#include <value_generators/value_generators.hpp>
+
+#include <cast_core/config/RunCommandConfig.hpp>
+
+namespace genny::actor {
+
+/**
+ * TODO: document me
```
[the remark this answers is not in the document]
jasonjhchan: 👍

src/cast_core/src/actors/CrudActor.cpp
jasonjhchan: This slipped my mind :(

src/gennylib/src/context.cpp
rtimmons: do we need to do the same thing in the driver code then? not sure why tests would be different? or since we're only intending the user to pass in a callback, maybe the API should take the callback as part of the ctor and let the PoolFactory construct everything? just throwing out ideas

src/gennylib/src/PoolFactory.cpp
jasonjhchan: I don't think a boolean is sufficient here. The client options are being passed in with the apm options already initialized from the tests. We need a way to pass those options down to the `pool` constructor.

jasonjhchan:
@visemet @rtimmons I made some changes to how bulkWrite is structured in order to call `view()` at runtime rather than ctor time. In doing so, some WriteOperations like `insertOne`, `deleteMany`, etc. ended up being implemented as well. I added some preliminary tests for these operations. The encoding and decoding of the `options` for these operations still need to be implemented and tested.

Based on Ryan's comments, I also made it so the APM callback was passed into PoolFactory rather than the client options. 

I plan to update the workload yml to better reflect the current functionality of the CRUD actor with my remaining time (and of course fixing up CR comments).

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+        return true;
+    }
+};
+}  // namespace YAML
+
+namespace genny::actor {
+
+namespace {
+
+struct BaseOperation {
```
rtimmons: nit but I think all these helper things can be just inside an anonymous namespace not inside `genny::actor` - I think just move the opening `namespace genny::actor {` line below all the helper things.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+        auto filter = opNode["Filter"];
+        auto update = opNode["Update"];
+        if (!filter || !update) {
+            throw InvalidConfigurationException(
+                "'updateOne' expects 'Filter' and 'Update' fields.");
+        }
+        _filterTemplate = value_generators::makeDoc(filter, rng);
+        _updateTemplate = value_generators::makeDoc(update, rng);
+
+        // TODO: parse update options.
```
rtimmons: not sure what this and the other TODO means - can you elaborate? Seems like it's already parsed?

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+        (_onSession) ? _collection.update_one(session, filterView, updateView, _options)
+                     : _collection.update_one(filterView, updateView, _options);
+        ctx.success();
+    }
+
+private:
+    bool _onSession;
+    mongocxx::collection _collection;
+    std::unique_ptr<DocGenerator> _filterTemplate;
+    std::unique_ptr<DocGenerator> _updateTemplate;
```
rtimmons: imho template is a much better name than value-generator. nice.

consider adding `using UDocGenerator = std::unique_ptr<DocGenerator>;`

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+};
+
+WriteOpCallback createUpdateMany =
+    [](YAML::Node opNode,
+       bool onSession,
+       mongocxx::collection collection,
+       genny::DefaultRandom& rng,
+       metrics::Operation operation) -> std::unique_ptr<WriteOperation> {
+    return std::make_unique<UpdateManyOperation>(opNode, onSession, collection, rng, operation);
+};
```
rtimmons: I *think* you can template this guy to remove all the callback boilerplate:

```c++
template<class P, class C, class O>
C baseCallback =
        [](YAML::Node opNode,
           bool onSession,
           mongocxx::collection collection,
           genny::DefaultRandom& rng,
           metrics::Operation operation) -> std::unique_ptr<P> {
            return std::make_unique<O>(opNode, onSession, collection, rng, operation);
        };
```
and you can move these inline into the ctors map

```c++
bulkWriteConstructors = {
   ...
  {"updateOne", baseCallback<WriteOperation, WriteOpCallback, UpdateOneOperation>},
  ...
};
```

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+            _options = opNode["Options"].as<mongocxx::options::bulk_write>();
+        }
+    }
+
+    void createOps(const YAML::Node& writeOp, genny::DefaultRandom& rng) {
+        auto writeCommand = writeOp["WriteCommand"].as<std::string>();
+        auto writeOpConstructor = bulkWriteConstructors.find(writeCommand);
+        if (writeOpConstructor == bulkWriteConstructors.end()) {
+            throw InvalidConfigurationException("WriteCommand '" + writeCommand +
+                                                "' not supported in bulkWrite perations.");
```
rtimmons: typo: operations

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+            _options = opNode["Options"].as<mongocxx::options::transaction>();
+        }
+    }
+
+    void run(mongocxx::client_session& session) override {
+        try {
+            session.start_transaction(_options);
+        } catch (mongocxx::operation_exception& exception) {
+            throw;
+        }
```
rtimmons: no need to catch&rethrow

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+    {"commitTransaction", createCommitTransaction},
+    {"setReadConcern", createSetReadConcern},
+    {"drop", createDrop},
+    {"insertOne", createInsertOneOp},
+    {"deleteOne", createDeleteOneOp},
+    {"deleteMany", createDeleteManyOp},
+    {"updateOne", createUpdateOneOp},
+    {"updateMany", createUpdateManyOp},
+    {"replaceOne", createReplaceOneOp}};
+};  // namespace
```
rtimmons: I think the above template thing would work here - probably same template too.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+            }
+
+            // Grab the appropriate Operation struct defined by 'OperationName'.
+            auto op = opConstructors.find(opName);
+            if (op == opConstructors.end()) {
+                throw InvalidConfigurationException("Operation '" + opName +
+                                                    "' not supported in Crud Actor.");
+            }
+            auto createOperation = op->second;
+            std::vector<std::pair<std::string, std::unique_ptr<DocGenerator>>> docs;
```
rtimmons: looks like `docs` isn't used anymore

src/driver/test/CrudActor.yml
```diff
@@ -0,0 +1,38 @@
...
+        - { a : {^RandomString: {length: {^RandomInt: {min: 3, max: 5} }}} }
+        - { b : {^RandomInt: {min: 5, max: 15} } }
+  - Repeat: 1
+    Collection: test
+    Operation:
+      OperationName: drop
+      OperationCommand:
+        Options:
+          WriteConcern:
+            Level: majority
```
rtimmons: I like the example you had in the class docs - can you put that here (or in a separate yaml file) and then just comment in the class docs to see example in the file?

src/gennylib/include/gennylib/PoolFactory.hpp
```diff
@@ -45,7 +45,9 @@ class PoolFactory {
     };
 
 public:
-    PoolFactory(std::string_view uri);
+    using ApmCallback = std::function<void(const mongocxx::events::command_started_event&)>;
```
rtimmons: add note that this is temporary. Let's create a TIG ticket to track the CDRIVER ticket and add that ticket number to the comment

src/gennylib/src/context.cpp
```diff
@@ -96,6 +101,18 @@ ActorVector WorkloadContext::_constructActors(const Cast& cast,
     return actors;
 }
 
+mongocxx::pool::entry WorkloadContext::client() {
+    if (_hasApmOpts) {
+        // TODO: Remove this conditional when CDRIVER-2931 is resolved.
```
rtimmons: in this and other places related to the APM callback let's add a `// TODO(TIG-XYZ) ...`

src/cast_core/src/actors/CrudActor.cpp
guoyr: Even better, create a Jira ticket for all the TODOs in this PR

src/cast_core/src/actors/CrudActor.cpp
guoyr: In addition, could you add a TODO or add to the TODO ticket to wrap all the run methods with a try/catch and rethrow as genny::MongoException. Not particularly useful here but should come in handy for CRUD operations

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1241 @@
...
+struct DropOperation : public BaseOperation {
+
+    DropOperation(YAML::Node opNode,
+                  bool onSession,
+                  mongocxx::collection collection,
+                  genny::DefaultRandom& rng,
+                  metrics::Operation operation)
+        : _onSession{onSession}, _collection{collection}, _operation{operation} {
+        if (!opNode)
+            return;
```
guoyr: Should this throw an error?

src/cast_core/src/actors/CrudActor.cpp
guoyr: Ditto for `OpCallback`s below

src/cast_core/src/actors/CrudActor.cpp
jasonjhchan: No, I decided to make it that for a `collection.drop()` command, the user doesn't have to specify the `OperationCommand` block unless specified with options (like write concern) as the command takes no arguments .

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1077 @@
...
+/**
+ * Example usage:
+ *    Operations:
+ *    - OperationName: setReadConcern
+ *      OperationCommand:
+ *        ReadConcern:
+ *          Level: majority
+ */
+
+struct SetReadConcernOperation : public BaseOperation {
```
visemet: [question] Why did this end up being a separate operation instead of additional options to the commands? I was expected there'd be a way to make use of `ConfigNode` so the read and write concerns wouldn't need to be repeated for each operation in the list.

src/cast_core/src/actors/CrudActor.cpp
```diff
@@ -0,0 +1,1077 @@
...
+struct CommitTransactionOperation : public BaseOperation {
+
+    CommitTransactionOperation(YAML::Node opNode,
+                               bool onSession,
+                               mongocxx::collection collection,
+                               genny::DefaultRandom& rng,
+                               metrics::Operation operation) {}
+
+    void run(mongocxx::client_session& session) override {
+        session.commit_transaction();
```
visemet: @rtimmons, @guoyr, do either of you have thoughts on making whether to automatically retry the entire transaction be part of `ExecutionStrategy`? If some operation performed inside of the transaction fails due to a write conflict, then we're probably going to appear as doing very little database work very quickly. Restructuring the code into a `WithTransactionOperation` object might also make the handling the retries easier.

src/cast_core/src/actors/CrudActor.cpp
jasonjhchan: My rationale for adding a separate operation was mostly to reflect the mongocxx::collections api command as it seemed like something that might be useful. I will add a note in the TODO ticket to consider adding the ability to set ReadConcern at the Phase level.

PR #133, opened 2019-01-17
Add context.getPlural

rtimmons:
We have a few occasions where we want to support both a singular and a plural syntax. Notably `Operation` and `Operations` but probably others too. This adds `context.getPlural` to simplify the calling code.

### DIFF
--- a/src/cast_core/src/actors/RunCommand.cpp
+++ b/src/cast_core/src/actors/RunCommand.cpp
@@ -122,31 +122,15 @@ struct actor::RunCommand::PhaseState {
             (*client)[database],
         };
 
-        auto addOperation = [&](YAML::Node node) {
+        auto createOperation = [&](YAML::Node node) {
             auto yamlCommand = node["OperationCommand"];
             auto doc = value_generators::makeDoc(yamlCommand, rng);
 
             auto options = node.as<Operation::Options>(Operation::Options{});
-            operations.push_back(std::make_unique<Operation>(fixture, std::move(doc), options));
+            return std::make_unique<Operation>(fixture, std::move(doc), options);
         };
 
-        auto operationList = context.get<YAML::Node, false>("Operations");
-        auto operationUnit = context.get<YAML::Node, false>("Operation");
-        if (operationList && operationUnit) {
-            throw InvalidConfigurationException(
-                "Can't have both 'Operations' and 'Operation' in YAML config.");
-        } else if (operationList) {
-            if (!operationList->IsSequence()) {
-                throw InvalidConfigurationException("'Operations' must be of sequence type.");
-            }
-            for (auto&& op : *operationList) {
-                addOperation(op);
-            }
-        } else if (operationUnit) {
-            addOperation(*operationUnit);
-        } else if (!operationUnit && !operationList) {
-            throw InvalidConfigurationException("No operations found in RunCommand Actor.");
-        }
+        operations = context.getPlural<std::unique_ptr<Operation>>("Operation", "Operations", createOperation);
     }
 
     ExecutionStrategy strategy;
--- a/src/gennylib/include/gennylib/v1/ConfigNode.hpp
+++ b/src/gennylib/include/gennylib/v1/ConfigNode.hpp
@@ -171,6 +171,50 @@ public:
         return this->_delegateNode->template get<Out, Required>(std::forward<Args>(args)...);
     };
 
+    /**
+     * Extract a vector of items by supporting both singular and plural keys.
+     *
+     * Example YAML that this supports:
+     *
+     * ```yaml
+     * # Calling getPlural<int>("Number","Numbers") returns [7]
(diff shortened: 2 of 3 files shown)


src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -122,31 +122,15 @@ struct actor::RunCommand::PhaseState {
...
-            }
-            for (auto&& op : *operationList) {
-                addOperation(op);
-            }
-        } else if (operationUnit) {
-            addOperation(*operationUnit);
-        } else if (!operationUnit && !operationList) {
-            throw InvalidConfigurationException("No operations found in RunCommand Actor.");
-        }
+        operations = context.getPlural<std::unique_ptr<Operation>>("Operation", "Operations", createOperation);
```
guoyr: In the interest of using this actor as the canonical example for `getPlural()`, I think simplifying this line by using a standard `convert<Operation>` and dropping the `unique_ptr` might make this code marginally more streamlined for future readers?

src/gennylib/include/gennylib/v1/ConfigNode.hpp
```diff
@@ -170,6 +171,85 @@ class ConfigNode {
...
+     * @param f callback function mapping from the found node to a T instance. If not specified,
+     * will use the built-in YAML::convert<T> function.
+     * @return
+     */
+    template <typename T, typename F = std::function<T(YAML::Node)>>
+    std::vector<T> getPlural(
+        const std::string& singular, const std::string& plural, F&& f = [](YAML::Node n) {
+            return n.as<T>();
+        }) {
+        static_assert(std::is_same_v<T, decltype(f(YAML::Node{}))>,
```
bcaimano: I think maybe it's best to leave this out. You could return something that is implicitly convertible to T. The error message is probably pretty parsable as long as you structure your expressions in a nice way.

src/cast_core/src/actors/RunCommand.cpp
rtimmons: - Ya I thought about that (it's what lead me to allow `convert<>` if the function isn't passed in), but the `Operation` needs other things like the `PhaseContext` etc which aren't available in the `convert<>` struct methods. So the function/callback style is necessary here.
- The unique_ptr is there to allow virtual dispatch on the `Operation`

src/gennylib/include/gennylib/v1/ConfigNode.hpp
```diff
@@ -170,6 +171,85 @@ class ConfigNode {
...
+            str << "Can't have both '" << singular << "' and '" << plural << "'.";
+            throw InvalidConfigurationException(str.str());
+        } else if (pluralValue) {
+            if (!pluralValue->IsSequence()) {
+                std::stringstream str;
+                str << "'" << plural << "' must be a sequence type.";
+                throw InvalidConfigurationException(str.str());
+            }
+            for (auto&& val : *pluralValue) {
+                out.push_back(std::invoke(f, val));
```
bcaimano: I'd suggest
```
T val = std::invoke(f, val);
out.emplace_back(std::move(val));
```
It makes the typing more permissive and probably will have a more clear error message in weird compile cases.

src/gennylib/test/context_test.cpp
```diff
@@ -462,6 +462,102 @@ TEST_CASE("Actors Share WorkloadContext State") {
...
+    onContext(createYaml("Foo: 71"), [](ActorContext& c) {
+        REQUIRE(c.getPlural<int>("Foo", "Foos") == std::vector<int>{71});
+    });
+
+    onContext(createYaml("{ Foo: 9, Foos: 1 }"), [](ActorContext& c) {
+        REQUIRE_THROWS_WITH([&]() { c.getPlural<int>("Foo", "Foos"); }(),
+                            Matches("Can't have both 'Foo' and 'Foos'."));
+    });
+
+    onContext(createYaml("Number: 7"), [](ActorContext& c) {
```
bcaimano: Maybe you want to have a `Numbers` here?

src/gennylib/include/gennylib/v1/ConfigNode.hpp
rtimmons: Ya good call about convertible. Removing the assert.

src/gennylib/include/gennylib/v1/ConfigNode.hpp
rtimmons: Good idea - changing.

src/gennylib/test/context_test.cpp
rtimmons: The "parent" context already has `Numbers` so this test asserts that the child inherits from the parent and the error still happens.

rtimmons:
Thanks for review @guoyr and @bcaimano - I merged in haste so will cut a quick PR for the changes suggested.

PR #134, opened 2019-01-18
Update some docs and make doxygen more strict

rtimmons:
Finally got doxygen config tweaked so that missing docstrings result in warnings. There are currently about 500 warnings related to missing docstrings.

Uploaded generated docs here: http://genny-docs.s3-website-us-east-1.amazonaws.com/html/index.html


.doxygen
```diff
@@ -160,8 +160,9 @@ FILE_PATTERNS          = *.c \
                          *.ucf \
                          *.qsf
 RECURSIVE              = YES
-EXCLUDE                = mwg build src/python src/third_party \
-                         src/gennylib/test src/driver/test src/cast_core/test
+EXCLUDE                = build src/python src/third_party \
```
bcaimano: I noticed this before, but I'm not certain `EXCLUDE` works for directories. Perhaps [EXCLUDE_PATTERNS](https://stackoverflow.com/questions/5507055/excluding-directories-for-doxygen)

src/gennylib/include/gennylib/Cast.hpp
```diff
@@ -27,17 +27,20 @@ namespace genny {
 /**
  * A cast is a map of strings to shared ActorProducer instances.
  *
+ * (Actors belong to a cast. So it's a cast of actors. Get it??)
```
bcaimano: :)

.doxygen
rtimmons: I tried a few times with `EXCLUDE_PATTERNS` and it always ends up grabbing `third_party` etc.  FWIW `EXCLUDE` does seem to work as-is here

src/HitchhikersGuide.md
```diff
@@ -1,7 +1,7 @@
 ## 🚀 Hitchhikers' Guide to the Genny Repo 🚀 {#mainpage}
 
 #### 🔰 Intro 🔰
-This guide is meant to serve as a readme for anyone working in the Genny Repo. This is not meant as a guide to use Genny as a tool - that is in the [README.md](README.md) file in this same directory. The Genny Repo has a lot of different layers 🍰 to it, and before getting started it is useful to know what part of it to work on. This guide should give you an overview of where to get started!
+This guide is meant to serve as a readme for anyone working in the Genny Repo. This is not meant as a guide to use Genny as a tool - that is in the README.md file in the parent directory. The Genny Repo has a lot of different layers 🍰 to it, and before getting started it is useful to know what part of it to work on. This guide should give you an overview of where to get started!
```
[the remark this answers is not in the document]
rtimmons: I was trying to get the links working in both the doxygen-generated output and the github-rendered version but alas. I just removed the markdown links. The [doxygen generated version](http://genny-docs.s3-website-us-east-1.amazonaws.com/html/) does an okay version of detecting the links on its own.

PR #136, opened 2019-01-18
Switch ExecutionStrategy to use metrics::Operation

rtimmons:
Existing metrics types are on their way out and this is a first step to using the new `metrics::Operation` more pervasively. Not strictly backward-compatible and I'm still pondering on how metrics want to be recorded for retries etc but this is a good first-start.


src/gennylib/include/gennylib/ExecutionStrategy.hpp
```diff
@@ -71,19 +72,18 @@ class ExecutionStrategy {
...
-                timer.report();
+                fun(ctx);
 
+                ctx.success();
                 result.wasSuccessful = true;
                 shouldContinue = false;
             } catch (const boost::exception& e) {
-                _recordError(e);
+                BOOST_LOG_TRIVIAL(debug) << "Caught error: " << boost::diagnostic_information(e);
+                ctx.fail();
```
rtimmons: we start the `ctx` at every retry/iteration so no need to scope-guard close it I think

src/gennylib/include/gennylib/ExecutionStrategy.hpp
```diff
@@ -99,34 +99,15 @@ class ExecutionStrategy {
         }
     }
 
-    // This function takes the existing counter metrics recorded in this class, and adds them into
-    // the metrics namespace timeseries. This may be desired in specific edge cases or at the end of
-    // each phase. This will always be called when the ExecutionStrategy is dtor'd.
-    void recordMetrics();
```
rtimmons: I'm not sure ExecutionStrategy should have raii semantics - it should have the same lifecycle as the Actor itself; if there's still anything pending being recorded during actor dtor then I think there's something seriously wrong.

src/gennylib/include/gennylib/ExecutionStrategy.hpp
```diff
@@ -32,8 +31,9 @@ class ActorContext;
  * A small wrapper for running Mongo commands and recording metrics.
  *
  * This class is intended to make it painless and safe to run mongo commands that may throw
- * operation_exceptions. It maintains a timer for successful operations, an always-updated genny
- * gauge for failed operations, and an amortized guage for all operation. The ExecutionStrategy also
+ * boost exceptions.
+ *
+ * The ExecutionStrategy also
```
bcaimano: [nit] Mind fixing the line ragging? My mom designed brochures for a living before she retired. :)

src/gennylib/include/gennylib/ExecutionStrategy.hpp
bcaimano: So I agree, but just for posterity, people tend to like scope guards because they're relatively resilient to refactors or unexpected cases.

src/gennylib/include/gennylib/ExecutionStrategy.hpp
bcaimano: Ehhh, so I agree, but removing cheap safeguards because they're not necessary now is usually the start of warning stories. Again, really doesn't matter here but I felt the need to put a soapbox under my feet again.

src/gennylib/include/gennylib/ExecutionStrategy.hpp
```diff
@@ -99,34 +99,15 @@ class ExecutionStrategy {
         }
     }
 
-    // This function takes the existing counter metrics recorded in this class, and adds them into
-    // the metrics namespace timeseries. This may be desired in specific edge cases or at the end of
-    // each phase. This will always be called when the ExecutionStrategy is dtor'd.
-    void recordMetrics();
-
-    size_t errors() const {
```
bcaimano: I'm mildly surprised I didn't have a test case that used that.

src/gennylib/include/gennylib/ExecutionStrategy.hpp
rtimmons: > removing cheap safeguards because they're not necessary now is usually the start of warning stories

It took me a while to reason about why the scope-guard was necessary here and how it could be re-worked with the changed API fwiw. So counter-argument is that treating something with RAII semantics that doesn't really "need it" for correctness could cause confusion and incorrect usage in the future.

PR #137, opened 2019-01-18
TIG-1350 Add documentation on methods of third party includes

bcaimano:
I'm certain there are additional details you two would like, but this is a start.

### DIFF
--- a/src/third_party/README.md
+++ b/src/third_party/README.md
@@ -1,0 +1,49 @@
+# Third Party Libraries
+
+Genny depends on a set of third party libraries to compile and run.
+These libraries are included in one of three general ways:
+1. Toolchain libraries via find_package
+2. Vendored libraries via add_subdirectory
+3. User libraries via find_package
+
+## Toolchain Libaries
+
+Toolchain libraries are generally monolithic pieces of code with a large collection of headers.
+If they can be linked via a static library, they should be. Generally speaking, these libraries are
+expected to have a version that is acceptable for genny for the foreseeable future. Often, these
+libraries are so essential that the CMake find_package module searches for their core headers at
+known locations. For alternative locations, usually the module will allow the user to specify a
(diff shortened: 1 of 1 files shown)


src/third_party/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Genny currently uses:
+* yaml-cpp
+* Loki
+* Catch2
+* Jasper
+
+## User Libraries
+
+User libraries are modular pieces with a known ABI and some dependencies. Usually, linking to
```
guoyr: So I understand from this comment how user libraries are functionally different to toolchain libraries; does that mean we dynamically link them due to convenience? Asking because I'm about to change the genny toolchain to be all vcpkg and I'd like to include the user libraries in there as well.

bcaimano: That's a good question. So there are two concerns here, one that we care about quite a bit, and the other one that we surprisingly don't care about very much:
* User libraries usually have extra deps to other libraries. (We've deliberately limited that with our distributable builds of these libraries.) We shared link because we may or may not have those dependency libraries linked into our executable by default and we'd rather not care about these deps even if the cmake module pulls them in. For mongocxx, that's mongoc. For grpc, that's protobuf and c-ares.
* People get touchy around static linking and licenses. (See [here](https://softwareengineering.stackexchange.com/questions/167773/how-does-the-gpl-static-vs-dynamic-linking-rule-apply-to-interpreted-languages) Another hallmark of libraries that fit this bill is that they may have some restrictive licensing conditions. Somehow, ours are both Apache. It's weird.

bcaimano: I think I should put in a section on linking in general here.

rtimmons: This is good and well-reasoned info - would love a touch of color on linking since I don't have a good intuition about what to link shared vs static and I suspect it's not actually that complicated in the general case.

PR #139, opened 2019-01-19
TIG-1329: Refactor cmake into genny_subdir function

rtimmons:
So TIG-1329 explicitly says its goal isn't to create a cmake function but in doing #138 it was pretty tedious to make sure the right things were copy/pasted around. And it turns out writing the cmake function was pretty easy. CMake is a pretty opaque build-system but doing programmatic things like this actually aren't too bad.

<strike>Resmoke still mystifies me so I haven't tried this change under resmoke but I'll do that before merging</strike> Update: ran get-mongo-source and then run-resmoke and resmoke worked and it passed 💥


src/cast_core/CMakeLists.txt
```diff
@@ -14,46 +14,10 @@
 
 project(cast_core VERSION 0.0.1 LANGUAGES CXX)
 
-# Pacify the CLion
```
rtimmons: clion upgraded the version of cmake or something so now this is no longer needed 🎉

src/cast_core/CMakeLists.txt
```diff
@@ -14,46 +14,10 @@
...
-add_library(cast_core SHARED
-    # ActorsStart
-    src/actors/HelloWorld.cpp
-    src/actors/Insert.cpp
-    src/actors/InsertRemove.cpp
-    src/actors/Loader.cpp
-    src/actors/MultiCollectionQuery.cpp
-    src/actors/MultiCollectionUpdate.cpp
-    src/actors/RunCommand.cpp
-    # ActorsEnd
```
rtimmons: we glob all cpps now so this isn't needed and removed the grossness in `create-new-actor.sh` to update this file 🎉

src/cast_core/test/actors_test.cpp
```diff
@@ -12,16 +12,15 @@
...
 #include <bsoncxx/builder/stream/document.hpp>
 #include <bsoncxx/json.hpp>
 #include <bsoncxx/types.hpp>
 
 #include <mongocxx/stdx.hpp>
 #include <mongocxx/uri.hpp>
 
-#include "MongoTestFixture.hpp"
+#include <testlib/MongoTestFixture.hpp>
+#include <testlib/helpers.hpp>
```
rtimmons: do we want to be stricter about the namespaces lining up with the include dir? making `testlib` files define only things in the `testlib::` namespace? Seems "right" but also like it could get tedious.

src/gennylib/CMakeLists.txt
```diff
@@ -16,143 +16,16 @@
...
-    yaml-cpp
-    loki
-    metrics
-    value_generators
-    Boost::boost Boost::log
-    MongoCxx::mongocxx
-)
-
-# Concepts to do with Actors
-add_library(gennylib_cast STATIC
```
rtimmons: I *think* there were a few sub-targets within `gennylib` in order to maybe improve compile-time and maybe to keep some things separate that probably didn't need to be together, is that right @bcaimano?

The new `genny_subdir` thing isn't quite so flexible and would force us to create separate cmake projects for each of these sub-targets - do you think that's worth it? My instinct is no but open to ideas here since I'm not sure of all the reasons why this split was added (I was out while it went in).

src/gennylib/CMakeLists.txt
```diff
@@ -16,143 +16,16 @@
...
-add_executable(benchmark_gennylib
-    test/PhaseLoop_benchmark.cpp
-    test/RateLimiter_benchmark.cpp
-    test/orchestrator_benchmark.cpp
-)
-
-target_link_libraries(benchmark_gennylib gennylib_testutil)
-
-
-add_executable(test_gennylib_with_server
```
rtimmons: I moved Actor tests to `cast_core/test` and I didn't create a distinction for "`_with_server`" tests - I don't think there's much value in separating out tests by target when we have catch2 tags that could facilitate isolating which tests to run.

src/resmokeconfig/genny_sharded.yml
```diff
@@ -12,7 +12,7 @@ executor:
     - "12345"
     - "--success"
     - "--out"
-    - "build/test_gennylib_with_server.junit.xml"
+    - "build/cast_core_test.junit.xml"
```
rtimmons: so now resmoke only runs `cast_core_test` so any tests within `gennylib/test` that require a mongo db won't be run by resmoke. I don't see a good way around this but open to ideas - @guoyr is it possible to specify multiple binaries or something?

I think the main value of running via resmoke is for Actor-authors who probably don't care much about running genny's own internal tests so I'm not super concerned but if it's easy to fix I'd like to do so.

src/third_party/catch2/CMakeLists.txt
```diff
@@ -1,20 +1,17 @@
...
-add_library(catch2 OBJECT ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
+add_library(catch2 SHARED ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
 target_include_directories(catch2
-    PRIVATE
+    PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
         $<INSTALL_INTERFACE:include>
 )
 
+install(TARGETS  catch2
```
rtimmons: this..may be "wrong" because it means we export catch2 but this made things a lot easier

src/third_party/catch2/CMakeLists.txt
```diff
@@ -1,20 +1,17 @@
...
-target_include_directories(catch2_headers
-    INTERFACE
-        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-        $<INSTALL_INTERFACE:include>
-)
 
-# For more info on object libraries,
-# see https://cmake.org/cmake/help/v3.10/command/add_library.html#object-libraries
-add_library(catch2 OBJECT ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
+add_library(catch2 SHARED ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
```
rtimmons: treating catch2 just like any other library got around some awkwardness with object librareis not being able to be added to target includes or somesuch and it also gets rid of a cmake_policy warning which is nice. This said I'm not sure of the reasons why catch2 wanted to be an object library (I read the link) so depending on those reasons portions of the linking with catch2 in the `genny_subdir` function may need to be changed.

cmake/genny_subdir.cmake
```diff
@@ -0,0 +1,223 @@
...
+#     is given.
+# -   Assumes regular tests are `test/*.cpp` and get parsed by
+#     `ParseAndAddCatchTests`. The test binary is named
+#     `project_name_test`.
+# -   Assumes benchmark tests are `benchmark/*.cpp` and get parsed by
+#     `ParseAndAddCatchTests`. The benchmark test binary is named
+#     `project_name_benchmark`.
+# -   Installs the library to the `GennyLibraryConfig` export
+#
+function(GENNY_SUBDIR)
```
rtimmons: maybe this file and function should be called `GennySubdir` to be consistent with our other `.cmake` files?

Either way I don't really like "subdir" being part of the name. Cmake calls each dir a project but `GennyProject` seemed wrong and `GennySubProject` feels incorrect and contrived...open to ideas about other names though.

cmake/genny_subdir.cmake
```diff
@@ -0,0 +1,223 @@
...
+            set(_gs_list_target "_gs_test_depends")
+            continue()
+        else()
+            # if no _gs_want* set and no keyword types
+            # like NAME/TYPE etc then we must be in a
+            # list i.e. DEPENDS or TEST_DEPENDS so
+            # add the arg to teh appropriate list
+            list(APPEND "${_gs_list_target}" "${arg}")
+        endif()
+    endforeach()
```
rtimmons: I decided against being super pedantic about parsing errors e.g. handling missing name/type value or specifying e.g. `genny_subdir(TYPE NAME foo)` or whatever...I don't think we'll add new subdirs that often and the errors that come from improper usage were (to me) pretty obvious when I was playing around with it.

src/third_party/catch2/CMakeLists.txt
guoyr: Along the same line, Does catch2 need a `main.cpp` that just includes the header file? If not, we can make catch2 an INTERFACE library and simplify a lot of things.

src/resmokeconfig/genny_sharded.yml
guoyr: Not at the moment; it's a known limitation. I'll have a CR up to add support. Should be easy fix.

guoyr: Filed https://jira.mongodb.org/browse/SERVER-39125

src/cast_core/test/actors_test.cpp
guoyr: If we only have to fix up files in testlib it seems worth it. (2c)

cmake/genny_subdir.cmake
guoyr: Noted. Have you looked into `https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html`? It seems to provide most of the functionality you need but I'm not sure how well it does error handling?

src/CMakeLists.txt
```diff
@@ -17,11 +17,12 @@ add_subdirectory(third_party)
 add_subdirectory(metrics)
 add_subdirectory(value_generators)
 add_subdirectory(gennylib)
+add_subdirectory(testlib)
 add_subdirectory(cast_core)
 add_subdirectory(driver)
 
 # This makes the project importable from the build directory
-export(TARGETS value_generators metrics gennylib gennylib_cast gennylib_core loki cast_core
+export(TARGETS value_generators metrics gennylib loki cast_core
```
guoyr: You should be able to replace this line with `export(EXPORT GennyLibraryConfig` now that all the targets are being installed to `GennyLibraryConfig`

src/third_party/catch2/CMakeLists.txt
bcaimano: Uhhhhh, so the catch2 library does main functions in a weird way that I haven't looked into too deeply. I had it as an object library so that each test would have its main function included in its binary.

That said. I honestly don't believe this file should be a thing. `catch2` should be an interface library and each test executable should have its own `test_foo.cpp` that looks like this main and whatever else we care about. (I suspect there is some sort of global teardown/setup in catch2 that might differ per unit.)

src/third_party/catch2/CMakeLists.txt
bcaimano: So maybe now's the time to bring up that I don't believe this `GennyLibraryConfig` is bringing very much value at the moment. It would be reasonable if people were developing actors outside of genny, but that's not the case and I doubt it will be for some time to come.

cmake/genny_subdir.cmake
bcaimano: Yeah, just use [cmake_parse_arguments](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html#command:cmake_parse_arguments).

cmake/genny_subdir.cmake
bcaimano: I honestly believe that you should have a `add_genny_library` and `add_genny_test` function. This is a standard that tends to come up pretty regularly. They can be almost entirely passthru besides the initial target variable:

```
function(add_genny_library TARGET)
    ...
    add_library(${ARGV})
    ...
endfunction(add_genny_library)

function(add_genny_test TARGET <SOURCES>)
    ...
    add_executable(test_${TARGET} ${SOURCES})
    target_link_library(test_${TARGET} ${TARGET})
    ...
endfunction(add_genny_test)
```

```
add_genny_library(mylib SHARED) # Defines mylib as normal, includes the local includes folder, links in the core deps
add_genny_test(mylib ...) # Defines test_mylib, links it to mylib, adds that silly ParseAndCombulate script
```

src/third_party/catch2/CMakeLists.txt
rtimmons: Ya good call, I killed `GennyLibraryConfig`

src/third_party/catch2/CMakeLists.txt
rtimmons: It turned out to be straightforward and easy to treat catch2 just like a regular genny subdir so moved things around a bit and just used `genny_subdir` (now renamed to `CreateGennyTargets`).

Catch2 can't be *truly8 header-only because there needs to be exactly one cpp file in the test-targets that have the main defined (comes in by defining `CATCH_CONFIG_MAIN` before call to include catch2.hpp). Each genny test subdir could have its own main.cpp that does this but that seems like needless boilerplate; the main object-code may as well live with the catcch2 lib.

cmake/genny_subdir.cmake
rtimmons: Inspired by your suggestion - renamed to `CreateGennyTargets`

cmake/genny_subdir.cmake
rtimmons: Done

cmake/genny_subdir.cmake
rtimmons: > I honestly believe that you should have a add_genny_library and add_genny_test function...

Not quite sure I see the benefit - What does this get you apart from the ability to specify multiple distinct test targets? why would we want multiple of them beyond just the "regular" ones and "benchmark" ones?

rtimmons:
Thanks for review, @bcaimano and @guoyr. I changed to use `cmake_parse_arguments` and renamed the function to `CreateGennyTargets`. Ready for another quick look when you can.

> I'm uneasy about the CMake changes. Every time you try to mock up and chain down CMake, it gets more brittle and you limit your options. Specialized add_library or add_executable(test) functions are a nice limited way to keep your maintenance pain down.

I hear ya. This may go a bit overboard in terms of "everything must follow the conventions," but I think that's kind of a feature for now.

I wouldn't want anybody to have to question "should I make this a new library or use an existing one" or "should this be a separate test binary or part of an existing one." Specifying there are up to 2 test binaries (regular and benchmark) and one object-library per genny subdir feels like the right balance for now. However if we end up wanting more customization we'd be open to a subdir having a bit of additional cmake (there is always the escape-hatch of putting your own cmake after the `CreateGennyTargets` invocation; this is what `driver` still does for the workload yamls).

Does that allay enough of your concerns? If not, I can take another pass through and try splitting up `CreateGennyTargets` into `CreateGennyLibrary` and `CreateGennyTest` kinda like I think you suggest but I'm not sure what benefit that gives and it does add a bit more cmake syntax to each of the subdir CMakeLists.txt files.

cmake/genny_subdir.cmake
bcaimano: Well, I can think of a few different short term goals that this would thwart:
1. You no longer have the direct ability to bundle code dependent upon features into a static library that gets added if needed. (Ala what was proposed with jasperlib)
2. You no longer get the ability separate long running tests from short running tests or test that need specific features from test that are vanilla unit tests.
3. You now have a weight of cmake function to support for every bit of your linking and you'll slowly have to modify it to fit your needs.

bcaimano:
@rtimmons, I'm not going to put you to undue effort for my beliefs in CMake. That said I do believe you're generating more work for yourself in the build system for the sake of problems that can be easily addressed in PR discussions. Also, `create_genny_library` :p. Nothing like reversed conventions, eh?

src/CMakeLists.txt
guoyr: ⬆️ 

Alternatively, could we rename the `GennyLibraryConfig` export-name to something else if we don't want it to be the content of `GennyLibraryConfig.cmake`?

rtimmons: > Alternatively, could we rename the GennyLibraryConfig export-name to something else if we don't want it to be the content of GennyLibraryConfig.cmake?

I tried my best to just kill `GennyLibraryConfig` but it was failing to compile on linux variants without it (but was working on my laptop). So I just threw my hands up. I like your idea but I can't bear to spend another test iteration on this so just gonna leave it 😔

PR #140, opened 2019-01-22
TIG-1155 add global rate limiter

guoyr:
Did a drive-by to rename _hwCounter to _helloCounter and UIntSpec to IntegerSpec.

### DIFF
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -20,6 +20,7 @@
 #include <iterator>
 #include <optional>
 #include <sstream>
+#include <thread>
 #include <type_traits>
 #include <unordered_map>
 #include <utility>
@@ -27,6 +28,7 @@
 #include <gennylib/InvalidConfigurationException.hpp>
 #include <gennylib/Orchestrator.hpp>
 #include <gennylib/context.hpp>
+#include <gennylib/v1/GlobalRateLimiter.hpp>
 
 /**
  * @file
@@ -42,7 +44,7 @@ namespace genny {
 namespace V1 {
 
 /**
- * Determine if we're done iterating for a given Phase.
+ * Determine the conditions for continuing to iterate a given Phase.
  *
  * One of these is constructed for each `ActorPhase<T>` (below)
  * using a PhaseContext's `Repeat` and `Duration` keys. It is
@@ -54,17 +56,16 @@ namespace V1 {
  * iteration start time) are passed into the IterationCompletionCheck
  * to determine if the loop should continue iterating.
  */
-class IterationCompletionCheck final {
+class IterationChecker final {
 
 public:
-    IterationCompletionCheck(std::optional<TimeSpec> minDuration,
-                             std::optional<UIntSpec> minIterations,
-                             bool isNop)
+    IterationChecker(std::optional<TimeSpec> minDuration,
+                     std::optional<IntegerSpec> minIterations,
+                     bool isNop)
         : _minDuration{minDuration},
           // If it is a nop then should iterate 0 times.
-          _minIterations{isNop ? UIntSpec(0l) : minIterations},
+          _minIterations{isNop ? IntegerSpec(0l) : minIterations},
           _doesBlock{_minIterations || _minDuration} {
-
         if (minDuration && minDuration->count() < 0) {
             std::stringstream str;
             str << "Need non-negative duration. Gave " << minDuration->count() << " milliseconds";
@@ -77,10 +78,35 @@ public:
         }
     }
 
-    explicit IterationCompletionCheck(PhaseContext& phaseContext)
-        : IterationCompletionCheck(phaseContext.get<TimeSpec, false>("Duration"),
-                                   phaseContext.get<UIntSpec, false>("Repeat"),
-                                   phaseContext.isNop()) {}
+    explicit IterationChecker(PhaseContext& phaseContext)
+        : IterationChecker(phaseContext.get<TimeSpec, false>("Duration"),
+                           phaseContext.get<IntegerSpec, false>("Repeat"),
+                           phaseContext.isNop()) {
+        auto rateSpec = phaseContext.get<RateSpec, false>("Rate");
+        const auto rateLimiterName =
+            phaseContext.get<std::string, false>("RateLimiterName").value_or("defaultRateLimiter");
+
+        if (rateSpec) {
+            _rateLimiter =
+                phaseContext.workload().getRateLimiter(rateLimiterName, rateSpec.value());
+        }
+    }
+
+    bool shouldLimitRate(int64_t currentIteration) const {
+        // Only rate limit if the current iteration is a muliple of the burst size.
+        return _rateLimiter && (currentIteration % _rateLimiter->getBurstSize() != 0);
+    }
+
+    bool limitRate() const {
+        return _rateLimiter->consume();
+    }
+
+    void wait() const {
+        // Sleep for a little bit. This is not efficient if there's a large number of
+        // threads. If there's a bottleneck, we can use exponential back off or wait
+        // based on the number of actors using this rate limiter.
+        std::this_thread::sleep_for(std::chrono::nanoseconds(_rateLimiter->getRate()));
+    }
 
     std::chrono::steady_clock::time_point computeReferenceStartingPoint() const {
         // avoid doing now() if no minDuration configured
@@ -96,11 +122,11 @@ public:
                                           std::chrono::steady_clock::now() - startedAt));
     }
 
-    bool operator==(const IterationCompletionCheck& other) const {
+    bool operator==(const IterationChecker& other) const {
         return _minDuration == other._minDuration && _minIterations == other._minIterations;
     }
 
-    bool doesBlock() const {
+    bool doesBlockCompletion() const {
         return _doesBlock;
     }
 
@@ -110,7 +136,8 @@ private:
     // .end() iterator needs an instance of this, so it's weird
 
     const std::optional<TimeSpec> _minDuration;
-    const std::optional<UIntSpec> _minIterations;
+    const std::optional<IntegerSpec> _minIterations;
+    v1::GlobalRateLimiter* _rateLimiter = nullptr;
     const bool _doesBlock;  // Computed/cached value. Computed at ctor time.
 };
 
@@ -126,12 +153,12 @@ private:
 class ActorPhaseIterator final {
 
 public:
-    // Normally we'd use const IterationCompletionCheck& (ref rather than *)
+    // Normally we'd use const IterationChecker& (ref rather than *)
     // but we actually *want* nullptr for the end iterator. This is a bit of
     // an over-optimization, but it adds very little complexity at the benefit
     // of not having to construct a useless object.
     ActorPhaseIterator(Orchestrator& orchestrator,
-                       const IterationCompletionCheck* iterationCheck,
+                       const IterationChecker* iterationCheck,
                        PhaseNumber inPhase,
                        bool isEndIterator)
         : _orchestrator{std::addressof(orchestrator)},
@@ -156,6 +183,18 @@ public:
 
     ActorPhaseIterator& operator++() {
         ++_currentIteration;
+        // Limit rate after incrementing currentIteration so the first few iterations
+        // are never blocked.
+        if (_iterationCheck && _iterationCheck->shouldLimitRate(_currentIteration)) {
+            while (true) {
+                auto success = _iterationCheck->limitRate();
+                if (!success && _orchestrator->currentPhase() == _inPhase) {
+                    _iterationCheck->wait();
+                    continue;
+                }
+                break;
+            }
+        }
         return *this;
     }
 
@@ -169,7 +208,7 @@ public:
                      // ...or...
                      // if we block, then check to see if we're done in current phase
                      // else check to see if current phase has expired
-                     (_iterationCheck->doesBlock() ? _iterationCheck->isDone(_referenceStartingPoint, _currentIteration)
+                     (_iterationCheck->doesBlockCompletion() ? _iterationCheck->isDone(_referenceStartingPoint, _currentIteration)
                                                    : _orchestrator->currentPhase() != _inPhase)))
 
                 // Below checks are mostly for pure correctness;
@@ -203,11 +242,11 @@ public:
 
 private:
     Orchestrator* _orchestrator;
-    const IterationCompletionCheck* _iterationCheck;
+    const IterationChecker* _iterationCheck;
     const std::chrono::steady_clock::time_point _referenceStartingPoint;
     const PhaseNumber _inPhase;
     const bool _isEndIterator;
-    unsigned int _currentIteration;
+    int64_t _currentIteration;
 
 public:
     // <iterator-concept>
@@ -239,7 +278,7 @@ public:
      */
     template <class... Args>
     ActorPhase(Orchestrator& orchestrator,
-               std::unique_ptr<const IterationCompletionCheck> iterationCheck,
+               std::unique_ptr<const IterationChecker> iterationCheck,
                PhaseNumber currentPhase,
                Args&&... args)
         : _orchestrator{orchestrator},
@@ -262,7 +301,7 @@ public:
           _value{!phaseContext.isNop()
                      ? std::make_optional<>(std::make_unique<T>(std::forward<Args>(args)...))
                      : std::nullopt},
-          _iterationCheck{std::make_unique<IterationCompletionCheck>(phaseContext)} {
+          _iterationCheck{std::make_unique<IterationChecker>(phaseContext)} {
         static_assert(std::is_constructible_v<T, Args...>);
     }
 
@@ -274,9 +313,9 @@ public:
         return ActorPhaseIterator{_orchestrator, nullptr, _currentPhase, true};
     };
 
-    // Used by PhaseLoopIterator::doesBlock()
+    // Used by PhaseLoopIterator::doesBlockCompletion()
     bool doesBlock() const {
-        return _iterationCheck->doesBlock();
+        return _iterationCheck->doesBlockCompletion();
     }
 
     // Checks if the actor is performing a nullOp. Used only for testing.
@@ -319,7 +358,7 @@ private:
     Orchestrator& _orchestrator;
     const PhaseNumber _currentPhase;
     const std::optional<std::unique_ptr<T>> _value;  // Is nullopt iff operation is Nop
-    const std::unique_ptr<const IterationCompletionCheck> _iterationCheck;
+    const std::unique_ptr<const IterationChecker> _iterationCheck;
 
 };  // class ActorPhase
 
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -98,12 +98,13 @@ public:
     }
 
     bool limitRate() const {
+        // Return if we're within the rate limit.
         return _rateLimiter->consume();
     }
 
     void wait() const {
         // Sleep for a little bit. This is not efficient if there's a large number of
-        // threads. If there's a bottleneck, we can use exponential back off or wait
+        // threads. If this is a problem, we can use exponential back off or wait
         // based on the number of actors using this rate limiter.
         std::this_thread::sleep_for(std::chrono::nanoseconds(_rateLimiter->getRate()));
     }
@@ -137,6 +138,8 @@ private:
 
     const std::optional<TimeSpec> _minDuration;
     const std::optional<IntegerSpec> _minIterations;
+
+    // The rate limiter is owned by the workload context.
     v1::GlobalRateLimiter* _rateLimiter = nullptr;
     const bool _doesBlock;  // Computed/cached value. Computed at ctor time.
 };
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -117,8 +117,8 @@ public:
                 if (!success && (o.currentPhase() == inPhase)) {
                     // Add some jitter to avoid threads waking up at once.
                     std::this_thread::sleep_for(
-                        std::chrono::nanoseconds(
-                            (_rateLimiter->getRate() + std::rand() % 1000) * _rateLimiter->getNumUsers()));
+                        std::chrono::nanoseconds((_rateLimiter->getRate() + std::rand() % 1000) *
+                                                 _rateLimiter->getNumUsers()));
                     continue;
                 }
                 break;
@@ -132,12 +132,11 @@ public:
     }
 
     bool isDone(SteadyClock::time_point startedAt, int64_t currentIteration) {
-        return
-            (!_minIterations || currentIteration >= (*_minIterations).value) &&
+        return (!_minIterations || currentIteration >= (*_minIterations).value) &&
             (!_minDuration ||
-            // check is last to avoid doing now() call unnecessarily
-            (*_minDuration).value <= std::chrono::duration_cast<std::chrono::nanoseconds>(
-                SteadyClock::now() - startedAt));
+             // check is last to avoid doing now() call unnecessarily
+             (*_minDuration).value <= std::chrono::duration_cast<std::chrono::nanoseconds>(
+                                          SteadyClock::now() - startedAt));
     }
 
     bool operator==(const IterationChecker& other) const {
@@ -183,7 +182,8 @@ public:
                        bool isEndIterator)
         : _orchestrator{std::addressof(orchestrator)},
           _iterationCheck{iterationCheck},
-          _referenceStartingPoint{isEndIterator ? SteadyClock::time_point::min() : _iterationCheck->computeReferenceStartingPoint()},
+          _referenceStartingPoint{isEndIterator ? SteadyClock::time_point::min()
+                                                : _iterationCheck->computeReferenceStartingPoint()},
           _inPhase{inPhase},
           _isEndIterator{isEndIterator},
           _currentIteration{0} {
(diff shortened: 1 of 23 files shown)


guoyr:
Tests have been added.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    constexpr int64_t getRate() const {
+        return _rateNS;
+    }
+
+private:
+    // Manually align _emptiedTimeNS here to vastly improve performance.
+    // Lazily initialized by the first call to consume().
+    alignas(BaseGlobalRateLimiter::CacheLineSize) std::atomic_int64_t _emptiedTimeNS = 0;
```
rtimmons: maybe `_lastEmptiedTimeNS` to indicate that it's continually updated?

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -0,0 +1,146 @@
...
+        int64_t newEmptiedTime;
+
+        do {
+            newEmptiedTime = curEmptiedTime + _rateNS;
+
+            // If the new emptied time is in the future, the bucket is empty. Return early.
+            if (startTime < newEmptiedTime) {
+                return false;
+            }
+        } while (!_emptiedTimeNS.compare_exchange_weak(curEmptiedTime, newEmptiedTime));
```
rtimmons: plz add comment about why the `_weak` variant is used.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -155,7 +185,22 @@ class ActorPhaseIterator final {
...
     ActorPhaseIterator& operator++() {
+        // This function is called after each iteration, so we never rate limit the
+        // first iteration. This means the number of completed operations is always
+        // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
+        // _burstSize. `m` here is the number of threads using the rate limiter.
+        if (_iterationCheck && _iterationCheck->shouldLimitRate(_currentIteration)) {
+            while (true) {
+                auto success = _iterationCheck->limitRate();
+                if (!success && (_orchestrator->currentPhase() == _inPhase)) {
+                    _iterationCheck->wait();
```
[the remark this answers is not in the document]
guoyr: If there's no repeat, there's no need for a rate limiter.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,36 @@ class IterationCompletionCheck final {
...
+    bool limitRate() const {
+        // Return if we're within the rate limit.
+        return _rateLimiter->consume();
+    }
+
+    void wait() const {
+        // Sleep for a little bit. This is not efficient if there's a large number of
+        // threads. If this is a problem, we can use exponential back off or wait
+        // based on the number of actors using this rate limiter.
+        std::this_thread::sleep_for(std::chrono::nanoseconds(_rateLimiter->getRate()));
```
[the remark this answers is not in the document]
guoyr: The idea of using sleep here is that the token bucket algorithm is lock free; using a condition variable would defeat the lock free-ness. Also, folly's own blocking version of consume() uses sleep: https://github.com/facebook/folly/blob/74ea53886c3333eda4eaf457d538a678ceaa5add/folly/TokenBucket.h#L266

I believe folly's design goal is to make the rate limit as light-weight as possible and leave the heavy-lifting (e.g. figuring how long to sleep) to callers. Using condition variables is probably going to make genny faster because we usually have many calling threads, but it would make the rate limiter itself slower.

Since using condition variables is going to make the rate limiter worse, I'm not sure it's something we should do. If we want to achieve the optimal performance for genny's immediate use cases, I think we should implement a separate "scheduler" and not use a rate limiter at all.

The distinction between the two is that the scheduler knowns all the information (rate spec and number of threads) in advance; all a thread has to do at run time is to sleep for a pre-computed duration (minus the length of the operation) This also guarantees that threads will go in some fixed order and never be starved.

A rate limiter is more versatile and allows different Y and X values for each call, but it also adds more runtime overhead.

cc @bcaimano in case you have thoughts

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -0,0 +1,147 @@
...
+    // Lazily initialized by the first call to consume().
+    alignas(BaseGlobalRateLimiter::CacheLineSize) std::atomic_int64_t _lastEmptiedTimeNS = 0;
+
+    // Flag to determine when consume() is first called.
+    std::atomic_bool _emptiedTimeNSHasBeenInitialized = false;
+
+    // Note that the rate limiter as-is doesn't use the burst size, but it is cleaner to
+    // store the burst size and the rate together, since they're specified together in
+    // the YAML as RateSpec.
+    const int64_t _burstSize;
```
rtimmons: ¯\\\_(ツ)\_\/¯ no strong opinion but I'd just as well leave off the burstSize; at least don't expose the `getBurstSize()` getter since I wouldn't want anybody to build on it.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,44 @@ class IterationCompletionCheck final {
...
+        // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
+        // _burstSize. `m` here is the number of threads using the rate limiter.
+        if (shouldLimitRate(currentIteration)) {
+            while (true) {
+                auto success = _rateLimiter->consume();;
+                if (!success && (o.currentPhase() == inPhase)) {
+                    // Sleep for a little bit. This is not efficient if there's a large number of
+                    // threads. If this is a problem, we can use exponential back off or wait
+                    // based on the number of actors using this rate limiter.
+                    std::this_thread::sleep_for(std::chrono::nanoseconds(_rateLimiter->getRate()));
```
rtimmons: thinking a bit more about this....so this works well for bursty workloads, but will perf be good enough here to support a fast but constant-rate workload? E.g. imagine wanting to do `"1 per 100 microseconds"`. If we have 10k threads all competing for the same token will we actually be able to achieve that rate given the sleeps? May warrant a bit of playing around with it and/or a benchmark test.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,44 @@ class IterationCompletionCheck final {
...
 
-    explicit IterationCompletionCheck(PhaseContext& phaseContext)
-        : IterationCompletionCheck(phaseContext.get<TimeSpec, false>("Duration"),
-                                   phaseContext.get<UIntSpec, false>("Repeat"),
-                                   phaseContext.isNop()) {}
+    explicit IterationChecker(PhaseContext& phaseContext)
+        : IterationChecker(phaseContext.get<TimeSpec, false>("Duration"),
+                           phaseContext.get<IntegerSpec, false>("Repeat"),
+                           phaseContext.isNop()) {
+        auto rateSpec = phaseContext.get<RateSpec, false>("Rate");
```
rtimmons: it would be a mis-configuration to have `Rate` but not `Repeat`/`Duration`. Can we check that and bomb out if specified `Rate` but not `Repeat` or `Duration`?

src/gennylib/include/gennylib/PhaseLoop.hpp
guoyr: I believe yes. With the current approach, we only need one thread, any thread, to get waken up per 100us for the workload to proceed. The scheduler should be smart enough to do this. Additionally, since sleep() and the rate limiter are both lock free, the number of threads is irrelevant, 100k threads should behave similarly to 10k threads.

With the condition variable approach, you'd have 10k threads all trying to acquire some mutex. Acquiring a mutex takes a non-trivial amount of time (50ns according to this guy: https://stackoverflow.com/a/49712993), so in the pathological worst case if the threads are all scheduled sequentially when one thread is in the critical section, you'd need 50ns * 10k = 500us for the thread doing the workload to proceed.

I'll write up a benchmark to test out that the rate limiter is able to proceed with >10k threads and a 1 per 100 microseconds rate.

For the BigUpdate workload, I saw no performance difference between this approach and the previous ad-hoc approach: https://evergreen.mongodb.com/version/5c4a3b1a3066153a37fb754a (see the first execution). But funnily, I believe the BigUpdate workload currently isn't using the rate limiter at all, because the number of actual operations is lower than what the rate limiter requires.

I've restarted the task with a lower rate limit to verify that the rate limiter does anything at all.

rtimmons: > I'll write up a benchmark to test out that the rate limiter is able to proceed with >10k threads and a 1 per 100 microseconds rate.

Thanks - if the benchmark bares it out no need to speculate further. 

Thinking aloud: A benchmark test could have 10k threads that increment an atomic counter rate-limited to "1 per 100 microseconds" and runs for 5000000 microseconds (5 seconds). Counter should be exactly 50000 at the end.

guoyr:
I added a benchmark to test the rate limiter. macOS has a hard limit of about 1000 procs per 8GB of ram, so I ended up using 1500 threads instead of 10k on my macbook pro with 16GB of ram.

Since we want to know the overhead of sleep(), I created additional calls by increasing the rate instead of the number of threads and should yield similar results.

On my laptop, with 1500 threads, I got down to `1 per 100 nanoseconds` with my 5 second benchmark, which corresponds to 10M operations per second. For the 10k threads with a 1 per 1us rate, there only needs 100k sleeps per second, which should be more than enough for the rate limiter to handle.

As a sanity check, I ran some other combinations of rate and thread count:
control group: 1500 threads, 1 per 100 nanoseconds: 60M-70M ops per 5sec.
experiment group:
- 1500 threads, 1 per 10 nanoseconds: ~70M ops per 5sec
- 1500 threads, 1 per 1 microsecond: ~6M ops per 5 sec
- 500 threads 1 per 100 nanoseconds: 55M - 60M ops per 5 sec
- 1 thread, 1 per 100 nanoseconds: 49.999M ops per 5sec

src/gennylib/test/GlobalRateLimiter_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+        auto getCurState = []() {
+            return WorkloadContext::getActorSharedState<IncActor, IncActor::IncCounter>().load();
+        };
+
+        ah.run();
+
+        // Result is greater than the expected output of 5sec * 10k per second.
+
+        // Commented out for CI testing. I got about 12M ops/s on my 2.8GHz Haswell CPU.
+        // REQUIRE(getCurState() > 5 * 1000 * 1000 * 100);
```
rtimmons: we leave benchmarks uncommented with the caveat that they're allowed to fail on developer laptops or outside of CI.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,51 @@ class IterationCompletionCheck final {
...
+        const auto rateLimiterName =
+            phaseContext.get<std::string, false>("RateLimiterName").value_or("defaultRateLimiter");
+
+        if (rateSpec) {
+            _rng = std::make_optional(phaseContext.workload().createRNG());
```
rtimmons: not sure you need the relatively heavyweight genny RNG to generate the jitter. I think you could safely use `rand() % 10000`

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,51 @@ class IterationCompletionCheck final {
...
+                   const Orchestrator& o,
+                   const PhaseNumber inPhase) {
+        // This function is called after each iteration, so we never rate limit the
+        // first iteration. This means the number of completed operations is always
+        // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
+        // _burstSize. `m` here is the number of threads using the rate limiter.
+        if (shouldLimitRate(currentIteration)) {
+            while (true) {
+                auto success = _rateLimiter->consume();
+                ;
```
rtimmons: 💅

src/gennylib/include/gennylib/PhaseLoop.hpp
guoyr: Ah good point! I misunderstood your previous question about rate limited ops potentially running in the wrong phase.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
guoyr: Right now the phase loop uses `getBurstSize()` to figure out which iterations need rate limiting applied. Would you prefer storing the burstSize elsewhere?

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,54 @@ class IterationCompletionCheck final {
...
 
-    explicit IterationCompletionCheck(PhaseContext& phaseContext)
-        : IterationCompletionCheck(phaseContext.get<TimeSpec, false>("Duration"),
-                                   phaseContext.get<UIntSpec, false>("Repeat"),
-                                   phaseContext.isNop()) {}
+    explicit IterationChecker(PhaseContext& phaseContext)
+        : IterationChecker(phaseContext.get<TimeSpec, false>("Duration"),
+                           phaseContext.get<IntegerSpec, false>("Repeat"),
+                           phaseContext.isNop()) {
+        auto rateSpec = phaseContext.get<RateSpec, false>("Rate");
```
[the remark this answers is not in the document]
guoyr: Hm, I think this is fine. The iteration checker is constructed during phaseloop ctor time already. Here's an abbreviated stacktrace:

{noformat}
    frame #3: 0x0000000100216f34 test_gennylib`genny::v1::IterationChecker::IterationChecker(this=0x0000000103b00320, minDuration=optional<genny::TimeSpec> @ 0x00007ffeefbf7988, minIterations=optional<genny::IntegerSpec> @ 0x00007ffeefbf7978, isNop=false) at PhaseLoop.hpp:70
    frame #4: 0x000000010020d8f1 test_gennylib`genny::v1::IterationChecker::IterationChecker(this=0x0000000103b00320, minDuration=optional<genny::TimeSpec> @ 0x00007ffeefbf7c90, minIterations=optional<genny::IntegerSpec> @ 0x00007ffeefbf7c80, isNop=false) at PhaseLoop.hpp:69
    frame #5: 0x000000010020d31d test_gennylib`genny::v1::IterationChecker::IterationChecker(this=0x0000000103b00320, phaseContext=0x0000000103b00270) at PhaseLoop.hpp:84
    frame #18: 0x00000001002b3079 test_gennylib`std::__1::unordered_map<unsigned int, genny::v1::ActorPhase<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor::PhaseConfig>, std::__1::hash<unsigned int>, std::__1::equal_to<unsigned int>, std::__1::allocator<std::__1::pair<unsigned int const, genny::v1::ActorPhase<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor::PhaseConfig> > > > genny::PhaseLoop<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7(actorContext=0x0000000103b002a0)::IncActor::PhaseConfig>::constructPhaseMap<>(genny::ActorContext&) at PhaseLoop.hpp:603
    frame #19: 0x00000001002b2578 test_gennylib`genny::PhaseLoop<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor::PhaseConfig>::PhaseLoop<>(this=0x0000000103b00368, context=0x0000000103b002a0) at PhaseLoop.hpp:568
    frame #20: 0x00000001002b24f2 test_gennylib`genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7(this=0x0000000103b00350, ac=0x0000000103b002a0)::IncActor::IncActor(genny::ActorContext&) at GlobalRateLimiter_test.cpp:109
    frame #21: 0x00000001002b246d test_gennylib`genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7(this=0x0000000103b00350, ac=0x0000000103b002a0)::IncActor::IncActor(genny::ActorContext&) at GlobalRateLimiter_test.cpp:110
    frame #22: 0x00000001002b189a test_gennylib`genny::DefaultActorProducer<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor>::produceInto(std::__1::vector<std::__1::unique_ptr<genny::Actor, std::__1::default_delete<genny::Actor> >, std::__1::allocator<std::__1::unique_ptr<genny::Actor, std::__1::default_delete<genny::Actor> > > >&, genny::ActorContext&) [inlined] std::__1::__unique_if<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor>::__unique_single std::__1::make_unique<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7(__args=0x0000000103b002a0)::IncActor, genny::ActorContext&>(genny::ActorContext&&&) at memory:3078
    frame #23: 0x00000001002b1845 test_gennylib`genny::DefaultActorProducer<genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7()::IncActor>::produceInto(this=0x0000000103c01b08, out=size=0, context=0x0000000103b002a0) at ActorProducer.hpp:86
    frame #24: 0x00000001003353e1 test_gennylib`genny::ParallelizedActorProducer::produce(this=0x0000000103c01b08, context=0x0000000103b002a0) at ActorProducer.cpp:26
    frame #25: 0x00000001007e2c4d libgennylib.dylib`genny::WorkloadContext::_constructActors(cast=0x0000000103c03350, actorContext=0x0000000103b000b0) at context.cpp:96
    frame #26: 0x00000001007dd6ad libgennylib.dylib`genny::WorkloadContext::WorkloadContext(this=0x0000000105006400, node=<unavailable>, registry=0x0000000103c02dd0, orchestrator=0x0000000103c061c0, mongoUri="mongodb://localhost:27017", cast=0x0000000103c03350) at context.cpp:74
    frame #27: 0x00000001007e3345 libgennylib.dylib`genny::WorkloadContext::WorkloadContext(this=0x0000000105006400, node=<unavailable>, registry=0x0000000103c02dd0, orchestrator=0x0000000103c061c0, mongoUri="mongodb://localhost:27017", cast=0x0000000103c03350) at context.cpp:37
    frame #28: 0x000000010032d842 test_gennylib`genny::ActorHelper::ActorHelper(YAML::Node const&, int, std::initializer_list<std::__1::pair<std::__1::basic_string_view<char, std::__1::char_traits<char> > const, std::__1::shared_ptr<genny::ActorProducer> > >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) [inlined] std::__1::__unique_if<genny::WorkloadContext>::__unique_single std::__1::make_unique<genny::WorkloadContext, YAML::Node const&, genny::metrics::Registry&, genny::Orchestrator&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, genny::Cast&>(__args=0x00007ffeefbfd5d8, __args=0x0000000103c02dd0, __args=0x0000000103c061c0, __args="mongodb://localhost:27017", __args=0x0000000103c03350) at memory:3078
    frame #29: 0x000000010032d786 test_gennylib`genny::ActorHelper::ActorHelper(this=0x00007ffeefbfd488, config=0x00007ffeefbfd5d8, tokenCount=1000, castInitializer=genny::Cast::List @ 0x00007ffeefbfc348, uri="mongodb://localhost:27017") at ActorHelper.cpp:41
    frame #30: 0x000000010032e0eb test_gennylib`genny::ActorHelper::ActorHelper(this=0x00007ffeefbfd488, config=0x00007ffeefbfd5d8, tokenCount=1000, castInitializer=genny::Cast::List @ 0x00007ffeefbfcf80, uri="mongodb://localhost:27017") at ActorHelper.cpp:30
    frame #31: 0x00000001002ae66a test_gennylib`genny::testing::(anonymous namespace)::____C_A_T_C_H____T_E_S_T____7() at GlobalRateLimiter_test.cpp:205
{noformat}

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,10 +78,54 @@ class IterationCompletionCheck final {
...
+                    "no guarantee the rate limited operation will run in the correct phase");
+            }
+            _rateLimiter =
+                phaseContext.workload().getRateLimiter(rateLimiterName, rateSpec.value());
+        }
+    }
+
+    bool shouldLimitRate(int64_t currentIteration) const {
+        // Only rate limit if the current iteration is a muliple of the burst size.
+        return _rateLimiter && (currentIteration % _rateLimiter->getBurstSize() == 0);
```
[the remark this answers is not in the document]
guoyr: I'm going to push back again on this a bit (Sorry). I started changing `getBurstSize` to `mightExceedTokens` but didn't like the new interface. The problem is that it's now it mandates the caller to store `currentIteration`, which is not a thing that the caller needs and isn't coherent with the rest the token bucket class.

I image it should be perfectly fine to do something like:

```c++

if (rl.consume()) {
  for (i : rl.getBurstSize()) {
    // do something. 
  }
}
```

...Or the canonical token bucket example of slicing a burstSize() chunk of an array to be consumed. Neither of which require a notion of `currentIteration`.

Feel free to correct me if I'm wrong, but my understanding of your hesitation to exposing `consume()` and `burstSize()` is that these are implementation details of the rate limiter itself that should be exposed to the caller. A simpler interface would be exposing just `rateLimiter.doYourThing() # Blocking`. I would prefer this interface, but it makes the rate limiter's implementation a bit worse, because you'd need an atomic thing in the rate limiter to track the iteration, which creates a bottleneck.

If you prefer the simpler interface, I'd be happy to switch to the single call version of the rate limiter.

guoyr:
Made some changes:
1. Cache now() to save one call per iteration per thread. Improves perf by 10%
2. Sleep duration increases with the number of threads. Improves perf by 25%
3. rename consume to consumeIfWithinRate

src/gennylib/include/gennylib/PhaseLoop.hpp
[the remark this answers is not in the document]
rtimmons: ah so I'm fine with this as-is - why I said "consider" 😺 thanks for considering though seems like it's fine to leave I agree with your understanding.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -109,9 +153,20 @@ class IterationCompletionCheck final {
     // referenceStartingPoint time (versus having those in the ActorPhaseIterator). BUT: even the
     // .end() iterator needs an instance of this, so it's weird
 
+    SteadyClock::time_point& _now() {
+        _cachedNow = SteadyClock::now();
+        return _cachedNow;
+    }
```
rtimmons: based on what we were looking at earlier today I think the `now()` calls weren't really an important factor here? If so, can we remove this to keep the logic slightly simpler?

guoyr: Done!

I also added a comment in Jira for my latest hypothesis: https://jira.mongodb.org/browse/TIG-1155?focusedCommentId=2131381&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2131381

The actual rate limit is perfectly fine. I'll send an email once this goes in to clarify the expected behavior of the rate limiter.

guoyr: Oh I also checked the naive version of the rate limiter. It is definitely worse than the global rate limiter.

In the IncActor, I was getting 60% of the expected throughput with 1000 threads and 100us sleeps . I believe this is because `sleep_for` only guarantees a lower bound but seems to sleep for a bit longer than the specified duration.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -0,0 +1,146 @@
...
+        return _rateNS;
+    }
+
+    constexpr int64_t getNumUsers() const {
+        return _numUsers;
+    }
+
+    void addUser() {
+        _numUsers++;
+    }
```
rtimmons: since the rate-limiter itself doesn't ever use this value, can it instead be stored in the thing that knows the value (the workload context)? workload context could instead have `map<string,pair<limiter,count>>` and the `getRateLimiter` could increment that count?

Either that or please add comment about how this value is intended to be used.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -77,30 +80,69 @@ class IterationCompletionCheck final {
...
+        // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
+        // _burstSize. `m` here is the number of threads using the rate limiter.
+        if (shouldLimitRate(currentIteration)) {
+            while (true) {
+                auto success = _rateLimiter->consumeIfWithinRate(SteadyClock::now());
+                if (!success && (o.currentPhase() == inPhase)) {
+                    // Add some jitter to avoid threads waking up at once.
+                    std::this_thread::sleep_for(
+                        std::chrono::nanoseconds((_rateLimiter->getRate() + std::rand() % 1000) *
+                                                 _rateLimiter->getNumUsers()));
```
rtimmons: plz add comment justifying the multiplication by `getNumUsers`

guoyr: Done, added a comment in getNumUsers() along with explaining what the number is for.

guoyr:
> As part of this ticket are we also going to kill the existing rate-limiter and incorporate this code into the existing actors?

Nah: https://jira.mongodb.org/browse/TIG-1400

PR #141, opened 2019-01-22
SERVER-36673: Add Secondary Allowed workload

thejonathanma:
Currently the workload only runs a stepdown command to trigger an election.

### DIFF
--- a/src/driver/test/SecondaryAllowed.yml
+++ b/src/driver/test/SecondaryAllowed.yml
@@ -1,57 +1,54 @@
 SchemaVersion: 2018-07-01
-
 Actors:
-- Name: Loader
-  Type: Loader
-  Threads: 1
-  Database: test
-  CollectionCount: 1
-  DocumentCount: 100
-  BatchSize: 100
-  Document:
-      x: {$randomint: {min: 0, max: 20}}
-  Indexes:
-  - keys: {x: 1}
-  Phases:
-  - Repeat: 1
-  - Operation:
-      OperationName: Nop
-- Name: SecondaryPreferredRead
-  Type: RunCommand
-  Threads: 1
-  ExecutionStrategy:
-      Retries: 3
-  Database: test
-  Phases:
-  - Operation:
-      OperationName: Nop
-  - Phase: 1
-    Repeat: 1
-    Operation: &find_one
-      MetricsName: "Find"
-      OperationName: RunCommand
-      OperationMinPeriod: 100 microseconds
-      OperationCommand:
-            find: Collection0
-            # readConcern: { level: local }
-            # readPreference: primary
-# - Name: SecondaryAllowedInserts
-#   Type: RunCommand
-#   Threads: 1
-#   Phases:
-#   - Repeat: 1
-#     Database: test
-#     Operations:
-#     - OperationName: RunCommand
-#       OperationCommand:
-#           insert: testCollection 
-#           documents: [{rating: 9}, {rating: 10}, {rating: 8}, {rating: 4}, {rating: 3}]
-#     - OperationMetricsName: Find
-#       OperationCommand: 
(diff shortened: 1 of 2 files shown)


src/driver/test/SecondaryAllowed.yml
```diff
@@ -0,0 +1,51 @@
...
+      Threads: 1
+      ExecutionStrategy:
+          Retries: 0
+          ThrowOnFailure: false
+      Phases:
+        - NoOp: true
+        - Phase: 1
+          Repeat: 1
+          Operation:
+              OperationMetricsName: "Stepdown"
```
bcaimano: Can I convince you to test out OperationAwaitStepdown from [TIG-1322](https://github.com/mongodb/genny/commit/e24bca65dcc8a052c00d54fa0346cd898d75a560)?

src/driver/test/SecondaryAllowed.yml
```diff
@@ -0,0 +1,51 @@
...
+              OperationName: AdminCommand
+              OperationSleepBefore: 30 seconds
+              OperationCommand:
+                  replSetStepDown: 60
+    - Name: SecondaryPreferredRead
+      Type: RunCommand
+      Threads: 10
```
bcaimano: Maybe increase this thread limit to at least 100+?

src/driver/test/SecondaryAllowed.yml
```diff
@@ -0,0 +1,51 @@
+SchemaVersion: 2018-07-01
```
bcaimano: I suspect you'll want some [QueryOptions](https://github.com/mongodb/genny/blob/master/src/driver/test/service-architecture/PrimaryRequiredReads.yml#L3)

thejonathanma:
Had to force push because I accidentally rebased master on this branch :( 

@bcaimano ready for another look!

src/driver/test/service-architecture/SecondaryAllowed.yml
```diff
@@ -0,0 +1,57 @@
...
+                  $readPreference: { mode: secondary }
+    - Name: Loader
+      Type: Loader
+      Threads: 1
+      Database: test
+      CollectionCount: 1
+      DocumentCount: 100
+      BatchSize: 100
+      Document:
+          x: {$randomint: {min: 0, max: 20}}
```
bcaimano: Just stumbled across this myself, I think this now needs to be `^RandomInt`. (See [here](https://github.com/mongodb/genny/commit/fb7bc6a2d2b9406328ea67c6a461b4d713e751cf))

src/driver/test/service-architecture/SecondaryAllowed.yml
```diff
@@ -0,0 +1,57 @@
...
+          ThrowOnFailure: false
+      Phases:
+        - NoOp: true
+        - Phase: 1
+          Repeat: 1
+          Operation:
+              OperationMetricsName: "Stepdown"
+              OperationName: AdminCommand
+              OperationSleepBefore: 30 seconds
+              OperationCommand:
```
bcaimano: It's probably also worth attaching an `OperationSleepAfter` because the option awaits the stepdown but doesn't necessarily go till the stepup.

rtimmons: > but doesn't necessarily go till the stepup.

should it? is that a shortcoming of the RunCommand actor?

bcaimano: So kinda. I think that waiting for a proper StepUp is a much more programatically complex task that requires looping until we can get a proper primary again. The StepDown itself means a handoff of primary status that will thrash readPreference like we expect. What I'm trying to say is that I think this is good enough for now even if there is some fudge in the box.

PR #142, opened 2019-01-22
Move integration tests to amazon2, use Evergreen binary archive

bcaimano:
Hopefully there is nothing too crazy here.

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -66,7 +66,7 @@ buildvariants:
   expansions:
     cmake: "/opt/cmake/bin/cmake"
     toolchain: "/opt/mongodbtoolchain/v3/bin"
-    mongodb_archive_url: "https://s3.amazonaws.com/mciuploads/mongodb-mongo-v4.0/enterprise-amazon2/3739429dd92b92d1b0ab120911a23d50bf03c412/binaries/mongo-mongodb_mongo_v4.0_enterprise_amazon2_3739429dd92b92d1b0ab120911a23d50bf03c412_18_12_13_18_53_46.tgz"
+    mongodb_archive_url: "https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-amazon2-4.1.7.tgz"
   tasks:
   - name: t_test_with_server
   - name: t_push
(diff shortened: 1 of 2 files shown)


scripts/run-resmoke.sh
```diff
@@ -42,7 +42,8 @@ cat << EOF >> "${SENTINEL_XML}"
 EOF
 
 # Add the mongo dir to the end of the path so we can find mongo executables
-export PATH="${PATH}:${MONGO_DIR}"
+export PATH+=":${MONGO_DIR}"
```
guoyr: remove this line?

evergreen.yml
```diff
@@ -66,8 +66,9 @@ buildvariants:
   expansions:
     cmake: "/opt/cmake/bin/cmake"
     toolchain: "/opt/mongodbtoolchain/v3/bin"
+    mongodb_archive_url: "https://s3.amazonaws.com/mciuploads/mongodb-mongo-v4.0/enterprise-amazon2/3739429dd92b92d1b0ab120911a23d50bf03c412/binaries/mongo-mongodb_mongo_v4.0_enterprise_amazon2_3739429dd92b92d1b0ab120911a23d50bf03c412_18_12_13_18_53_46.tgz"
```
guoyr: Since you're modifying the binary download anyway, I think we might as well just use the nightly build of master to get us closer to what DSI does. 

https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-amazon2-4.1.7.tgz

scripts/run-resmoke.sh
bcaimano: It allows people to run with local builds. The `bin` form works with our binary archives.

visemet: Making it easy for engineers to be able to run with local builds sounds useful to me. I think this is a quirk of how we're running resmoke.py from a location that isn't the top-level of the mongodb/mongo repository - it is going to check for `mongod`, etc. in the top-level of the mongodb/genny repository.

evergreen.yml
```diff
@@ -182,7 +183,8 @@ tasks:
 ##                ⚡️ Task Groups ⚡️
 
 task_groups:
-- name: t_test
+- &test_group
```
visemet: Calling this `test_group` is probably going to confuse someone. How about `compile_and_run_tests`?

bcaimano: Yeah, that's a good idea. Thanks!

PR #146, opened 2019-01-25
Fix evergreen yaml indent

rtimmons:
💅


rtimmons:
Thanks for reviews. Good call on the 100-char length. I'll do that in the future. This file isn't linted in CI now, I was just re-baselining it to be consistent. Once it's linted in CI I'm fine to change the indentation scheme.

PR #147, opened 2019-01-28
TIG-1245: Add Expression class hierarchy for generating random values

visemet:
Its API is modeled off [the `Expression` class](https://github.com/mongodb/mongo/blob/r4.1.7/src/mongo/db/pipeline/expression.h) in the aggregation framework for representing aggregation expressions.

----

I originally attempted to support value generation within array values by making a much smaller change to the existing code, but found the `parseMap()`, `parseSequence()`, and `applyOverrideLevel()` functions difficult to follow. The `Expression` class avoids implementing a second round of top-down traversal logic for BSON in addition to YAML by building the generated `Value` from the bottom-up.

I've only done the integration into the RunCommand actor (see 3c9e6765108172409fbed4a1e4523fcd71d633ea) in order to first build consensus on the overall direction. Depending on the feedback, I'll move forward with replacing the other usages of `genny::value_generators::makeDoc()` in a separate pull request.


src/value_generators/src/value_generators.cpp
```diff
@@ -33,5 +39,444 @@ std::unique_ptr<DocumentGenerator> makeDoc(const YAML::Node node, genny::Default
...
+    using Ts::operator()...;
+};
+
+template <class... Ts>
+overloaded(Ts...)->overloaded<Ts...>;
+
+int64_t Value::getInt64(std::string_view name) {
+    int64_t ret;
+
+    std::visit(
```
visemet: [note] Using `std::visit()` requires macOS 10.14 (Mojave) which may be problematic if people haven't upgraded already. (I hadn't.)

https://stackoverflow.com/questions/52310835/xcode-10-call-to-unavailable-function-stdvisit

src/value_generators/include/value_generators/value_generators.hpp
```diff
@@ -69,6 +77,262 @@ class DocumentGenerator {
...
+ * instance.
+ *
+ * The bsoncxx::types::b_document and bsoncxx::types::b_array cases that are part of
+ * bsoncxx::types::value do not allow for BSON documents or arrays to own their underlying buffer
+ * and can only be constructed from their respective view types.
+ *
+ * Note that we explicitly subclass from std::variant rather than alias it in order to define a
+ * custom std::ostream::operator<< for it.
+ */
+class Value : public std::variant<bool,
```
visemet: Subclassing from `std::variant` appears to cause compilation to fail with gcc...

https://stackoverflow.com/questions/51309467/using-stdvisit-on-a-class-inheriting-from-stdvariant-libstdc-vs-libc

src/cast_core/test/RunCommandActor_test.cpp
```diff
@@ -490,7 +492,7 @@ TEST_CASE_METHOD(MongoTestFixture,
         auto verifyFn = [&db, adminDb, view](const WorkloadContext& context) {
             REQUIRE_FALSE(adminDb.has_collection("testCollection"));
             REQUIRE(db.has_collection("testCollection"));
-            REQUIRE(db.collection("testCollection").count(view) == 1);
+            REQUIRE(db.collection("testCollection").count(view) == 2);
```
rtimmons: ♨️ ♨️ ♨️

src/cast_core/src/actors/RunCommand.cpp
```diff
@@ -139,8 +141,9 @@ class Operation {
 
 private:
     void _run() {
-        bsoncxx::builder::stream::document document{};
-        auto view = _doc->view(document);
+        auto command = std::get<bsoncxx::document::view_or_value>(_commandExpr->evaluate(_rng));
```
rtimmons: can there be a simpler way here? E.g. `_commandExpr->value(_rng)` that does L144-145 under the covers and returns a `bsoncxx::document::value`?

src/value_generators/include/value_generators/value_generators.hpp
```diff
@@ -69,6 +77,262 @@ class DocumentGenerator {
...
+                                  int32_t,
+                                  int64_t,
+                                  double,
+                                  std::string,
+                                  bsoncxx::types::b_null,
+                                  bsoncxx::document::view_or_value,
+                                  bsoncxx::array::view_or_value> {
```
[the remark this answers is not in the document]
visemet: I expanded on the comment for the `Expression` class in 7520cd3e8c6f9216283053011849768ca3cb7971. Let me know if that helps to clarify the usage of `view_or_value` here.

src/value_generators/include/value_generators/value_generators.hpp
```diff
@@ -69,6 +77,262 @@ class DocumentGenerator {
...
+public:
+    int64_t getInt64(std::string_view name);
```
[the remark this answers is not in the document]
visemet: Only `int64_t` has special semantics about treating `int32_t`s as `int64_t`s. We could theoretically do the same for certain `double` values but I didn't want to take that on right now.

This method has been changed to `std::optional<int64_t> tryAsInt64()` with the context about the parameter we're processing done as a separate function in value_generators.cpp.

src/cast_core/src/actors/RunCommand.cpp
visemet: Changing to encapsulate the `std::variant` made it so you can write `_commandExpr->value(_rng).getDocument()`. Is that simple enough? The only reason I did `auto view = command.view()` on the immediately following line was so I didn't have change the places where the `view` variable was being used already.

Regarding `bsoncxx::document::value`, I'd like to emphasize that we'd only ever return a `bsoncxx::document::view` from `Expression::evaluate()` if the underlying buffer was owned by the `UniqueExpression`. Using `.view()` ensures we aren't making unnecessary copies of the BSON document.

src/value_generators/include/value_generators/value_generators.hpp
```diff
@@ -69,6 +77,262 @@ class DocumentGenerator {
...
+    /**
+     * Parse `node` into one of the following:
+     *   - one of the Expression types registered in the `parserMap`,
+     *   - a DocumentExpression instance containing further nested Expression types,
+     *   - an ArrayExpression instance containing a list Expression types,
+     *   - a ConstantExpression instance representing a scalar value.
+     *
+     * @throws InvalidConfigurationException if `node` is malformed.
+     */
+    static UniqueExpression parseOperand(YAML::Node node);
```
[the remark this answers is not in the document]
visemet: I would say that `Expression::parseOperand()` is the most permissive in terms of what it parses. The choice of which function should be used depends on the context in the YAML configuration. For example, using `ArrayExpression::parse()` would throw an exception at `parse()`-time if it wasn't going to generate an array value rather than throwing an exception at `evaluate()`-time when calling `.getArray()` on the result.

I'm open to suggestions on how to make the choice more clear to future actor authors.

src/value_generators/include/value_generators/value_generators.hpp
```diff
@@ -69,6 +77,262 @@ class DocumentGenerator {
...
+    const std::vector<ElementType> _elements;
+};
+
+
+/**
+ * Base class for generating random int64_t values.
+ */
+class RandomIntExpression : public Expression {
+public:
+    static UniqueExpression parse(YAML::Node node);
```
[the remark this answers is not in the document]
visemet: > (Or, rather, can these be moved into `genny::v1` and a simple "use this function to extract value-generators" provided?)

Maybe. I had considered making `Expression::parseExpression()` and `Expression::parseObject()` `protected` functions so they could only be used when defining `Expression` subclasses, but then it seemed somewhat arbitrary those functions wouldn't be part of the public API and yet `ArrayExpression::parse()` would.

We could make all but `Expression::parseOperand()` `protected` functions if you don't feel strongly about having callers use a different one to require a certain type. (We could also add some other scheme to the `Expression` class in the future to learn the type without `evaluate()`ing a `Value` for it.) I didn't attempt to hide much of the implementation because all `Expression` subclasses need to be added to `parserMap` until we add a more dynamic registration scheme akin to `Cast::Registration`.

visemet:
> This does seem like an appropriate place for `std::variant` but the `visit` syntax is kinda icky. But: As long as it's properly hidden inside the `value_generators` impl though I don't think it matters that much.

Composition seems to be the only option with how gcc fails to compile `std::variant` subclasses. It's mostly the same but just with more boilerplate.

PR #149, opened 2019-01-28
Fix RPATH on OSX

rtimmons:
I'm working on a quick python script to make it more idiomatic to run genny in dsi. In testing it locally I realized that `make install`ed things don't run on osx.

Took a bit of toying but ultimately this worked on my laptop and [on a patch-build](https://evergreen.mongodb.com/version/5c4f76521e2d17321fe6c257).

I only like 50% of the way really understand RPATH and the existing cmake logic for it so this was some trial-and-error.


cmake/ConfigureInstall.cmake
```diff
@@ -14,14 +14,7 @@ list(APPEND CMAKE_PREFIX_PATH "${GENNY_INSTALL_DIR}")
 
 include(GNUInstallDirs)
 
-# This line allows us to run genny executables from a distribution folder where libraries live in
-# bin/../lib64. This is a nice thing for driver-like applications, because you can pass around
-# archives trivially.
-set(CMAKE_INSTALL_RPATH                 "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"   CACHE PATH "")
```
bcaimano: This will break DSI. At the least, keep the ORIGIN if not also doing `@executable_path` (I've only done that via the macos post link tools. I think it should work.)

Some helpful links:
https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath
https://stackoverflow.com/questions/40146437/how-to-set-multiple-rpath-directories-using-cmake-on-macos

cmake/ConfigureInstall.cmake
```diff
@@ -14,14 +14,7 @@ list(APPEND CMAKE_PREFIX_PATH "${GENNY_INSTALL_DIR}")
...
-
-# Since we're shooting for redistribution, we change up the RPATH configuration
-set(CMAKE_SKIP_BUILD_RPATH              false   CACHE BOOL "")
-set(CMAKE_SKIP_INSTALL_RPATH            false   CACHE BOOL "")
-set(CMAKE_BUILD_WITH_INSTALL_RPATH      false   CACHE BOOL "")
-set(CMAKE_INSTALL_RPATH_USE_LINK_PATH   true    CACHE BOOL "")
-set(CMAKE_MACOSX_RPATH                  ON      CACHE BOOL "Ensure that RPATH is used on OSX")
+set(CMAKE_MACOSX_RPATH                  true                          CACHE BOOL "Ensure that RPATH is used on OSX")
```
bcaimano: [nit] I'm not sure how I feel about lowercase true. It's definitely mentioned as acceptable but it's completely outside the convention of CMake. (See https://cmake.org/cmake/help/latest/command/if.html)

scripts/bootstrap.sh
```diff
@@ -6,8 +6,8 @@ set -o nounset
 
 set -x
 
-SCRIPTS_DIR="$(dirname "${BASH_SOURCE[0]}")"
-ROOT_DIR="$(cd "${SCRIPTS_DIR}/.." && pwd)"
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
```
bcaimano: I'm not sure why one would want to do an extra invocation here. Like I guess potentially the scripts dir could be a symlink path and root dir wouldn't be, but you shouldn't be doing file modification on scripts anyway.

cmake/ConfigureInstall.cmake
rtimmons: 1. so it compiles on arch and alinux2 without this line, but I haven't tested running it from a genny patch-build yet.
2. BUT with this line in there the `dist/bin/genny` version doesn't work on osx. The `build/src/driver/genny` version works but not when it's moved to `dist`.

```sh
# dist version with the CMAKE_INSTALL_RPATH line
$> ../dist/bin/genny
dyld: Library not loaded: @rpath/libcast_core.dylib
  Referenced from: /Users/rtimmons/CLionProjects/squeeze/build/../dist/bin/genny
  Reason: image not found
zsh: abort      ../dist/bin/genny
```

bcaimano: Yeah, so it will work in install before it's redistributed but not in a different directory. This is because `$ORIGIN` is the only way linux knows about relative library paths. `@executable_path`/`@rpath` is the macOS version of that. Not that your `rpath` there isn't relative ala `@rpath/../lib`. I suspect that it's due to `CMAKE_MACOSX_RPATH`.

rtimmons:
This goes away with #158 so closing in favor of that

PR #150, opened 2019-01-30
TIG-1245: Replace usages of value_generators::makeDoc()

visemet:
`value_generators::Expression::parseOperand()` is now used in its place.

Also fixes the `create-new-actor.sh` script to account for the changes from b5eecf2fdd626e02dbd422b95d65916a9693fa6a.


visemet:
> I think also need to update the create-new-actor script

@rtimmons, I did though? See 7b1f016c1c7b70c2df21e20efc35ba2e50d9a6ef.

rtimmons:
Ha nice - guess I missed it 😎

PR #152, opened 2019-01-30
TIG-1205 Error if No Actors Produced for an Actor YAML Block

vrachev:
Added a test to check that producing an actor with a mistyped 'Type' will result in an InvalidConfigurationException.

### DIFF
--- a/src/gennylib/src/context.cpp
+++ b/src/gennylib/src/context.cpp
@@ -91,8 +91,7 @@ ActorVector WorkloadContext::_constructActors(const Cast& cast,
     } catch (const std::out_of_range&) {
         std::ostringstream stream;
         stream << "Unable to construct actors: No producer for '" << name << "'." << std::endl;
-        cast.streamProducersTo(stream);
-        throw std::out_of_range(stream.str());
+        throw InvalidConfigurationException(stream.str());
     }
 
     for (auto&& actor : producer->produce(*actorContext)) {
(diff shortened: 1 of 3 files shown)


src/gennylib/src/context.cpp
```diff
@@ -91,8 +91,7 @@ ActorVector WorkloadContext::_constructActors(const Cast& cast,
     } catch (const std::out_of_range&) {
         std::ostringstream stream;
         stream << "Unable to construct actors: No producer for '" << name << "'." << std::endl;
-        cast.streamProducersTo(stream);
```
vrachev: Is this method necessary here? It was adding garbage characters to the stream, and it seems like the error message here is sufficient enough. I couldn't understand why it was printing out garbage, and adding null checks to see that the values existed didn't eliminate the issue.

visemet: I wouldn't expect it to lead to printing out garbage. I'm happy to take a look over your shoulder.

This may be unrelated but - I think `Cast::streamProducersTo()` should be using `_producers` [rather than `globalCast().getProducers()`](https://github.com/mongodb/genny/blob/b0992e276d8fb8f1c1f72bc62001a1d434617171/src/gennylib/src/Cast.cpp#L34) given that those two necessarily aren't the same for testing purposes.

vrachev: Strangely it only printed out garbage on my machine but not on Max's. In any case I changed it to _producers and that fixed the issue.

PR #154, opened 2019-01-31
TIG-1359 use genny toolchain for compilation

guoyr:
Waiting for the patch build to finish first before requesting a review

PR #158, opened 2019-02-07
TIG-1359 Use script to compile genny

guoyr:
This replaces #154. I had macOS put the toolchain in /data/mci so we can still get CI integration for the genny toolchain

sys-perf-atlas patch build: https://evergreen.mongodb.com/version/5c5c2b17850e615f60330c2d
sys-perf patch build: https://evergreen.mongodb.com/version/5c5bc7787742ae4c1ac92d86
genny patch build: https://evergreen.mongodb.com/version/5c5c2c7be3c33145206b6539

The macOS build is failing because of the rpath issue. Happy to remove it for now or just do compile if we don't want a red box.


guoyr:
macOS rpath should be fixed but the Evergreen task still hangs due to https://jira.mongodb.org/browse/CDRIVER-2825. When we dry run the AuthNInsert workload, the certs don't exist, which hangs on macOS.

The fix is due to go in the next mongoc driver release. I'll see if there's an easy workaround

README.md
```diff
@@ -4,128 +4,28 @@ Genny 🧞‍
...
-- https://mongodb.github.io/mongo-cxx-driver/mongocxx-v3/installation/
-- Note that you need to specifically tell it to install into `/usr/local`!
-
-Note that on Ubuntu 18.04 you need a custom linker option:
-
-```sh
-cmake -E chdir "build" cmake -DCMAKE_EXE_LINKER_FLAGS=-Wl,--no-as-needed ..
-make -j8 -C "build" genny
-```
+1. `python3 lamp.py [--linux-distro ubuntu1804/rhel7/amazon2/arch]`
```
rtimmons: I ran this locally and while it works the warnings it produces may scare a new user to think that things aren't right. Mention that 

1. `#  define MONGOCXX_DEPRECATED __attribute__ ((__deprecated__))` errors are known (and link to ticket to update to latest driver to kill the warning0
2. `boost::system::error_category::std_category::equivalent` errors are known (I think..we didn't see those prior to the toolchain so is there something still not quite right there?)

README.md
```diff
@@ -4,128 +4,28 @@ Genny 🧞‍
...
-- icu
-- mongo-cxx-driver
-- protobuf (3.6.1)
-
-To build genny use the following commands:
+Ubuntu 16.04+: sudo apt install build-essential
+Red Hat/CentOS 7+: sudo yum groupinstall "Development Tools"
+Arch: Grab a beer. Everything should already be set up.
+macOS: xcode-select --install
+Windows: https://visualstudio.microsoft.com/
```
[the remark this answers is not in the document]
rtimmons: Also add whatever is needed to build on Amazon Linux 2, ya? Is this `sudo yum groupinstall "Development Tools"` (per https://github.com/10gen/dsi/pull/285/files)?

README.md
```diff
@@ -4,128 +4,28 @@ Genny 🧞‍
...
-- grpc (>=1.16)
-- icu
-- mongo-cxx-driver
-- protobuf (3.6.1)
-
-To build genny use the following commands:
+Ubuntu 16.04+: sudo apt install build-essential
+Red Hat/CentOS 7+: sudo yum groupinstall "Development Tools"
+Arch: Grab a beer. Everything should already be set up.
+macOS: xcode-select --install
```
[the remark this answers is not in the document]
rtimmons: I just ran `brew rm python` and the script fails so I think it is still necessary to `brew install python`. Similarly python3 will be necessary for the other OSes too

README.md
```diff
@@ -142,7 +42,8 @@ it's not going to make common editing environments go wonky.
 Genny has self-tests using Catch2. You can run them with the following command:
 
 ```sh
-make -j8 -C "build" gennylib_test driver_test test
+# Build genny first: `python3 lamp.py [...]`
+python3 lamp.py cmake-test
```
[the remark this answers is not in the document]
guoyr: Done

lamp.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env python3
+
+from __future__ import print_function
```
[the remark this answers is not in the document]
bcaimano: In general, I'd recommend having some sort of internal python directory hierarchy. It makes it a lot easier to reuse your code pieces, even if it's just being able to refer to `import tools.myFoo`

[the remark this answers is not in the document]
guoyr: 1. lol because lamps give you 🧞 .  Happy to pick a more conventional name though if this one's not intuitive. XD

2. scripts/ sounds good. 
3. Yeah that's the goal. the script is in Python to be testable and I imagine we'd want to migrate more sh scripts in `scripts/` to anything-but-bash. Not going to do it now since testability wasn't a requirement, but happy to file a followup ticket.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+        elif distro == 'rhel7':
+            build_id = 'genny_toolchain_rhel70_b15f51760a9e651fb66bc0593e2eea738e15ffa2_19_02_07_07_04_06'
+        elif distro == 'amazon2':
+            build_id = 'genny_toolchain_amazon2_b15f51760a9e651fb66bc0593e2eea738e15ffa2_19_02_07_07_04_06'
+        elif distro == 'arch':
+            build_id = 'genny_toolchain_archlinux_b15f51760a9e651fb66bc0593e2eea738e15ffa2_19_02_07_07_04_06'
+        else:
+            raise ValueError('Unknown Linux distro {}; please set "--linux-distro"'.format(distro))
+    else:
+        raise ValueError('Unsupported platform: {}'.format(os_family))
```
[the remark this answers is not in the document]
guoyr: So a couple of reasons for doing it this way.

1. I actually want to raise a different error for linux (fixable) and unsupported platforms (not fixable).
2. different os's can share the same toolchain (e.g. ubuntu 1710 can share the ubuntu1804 toolchain). I want to be able to easily add do a one-to-many mapping between distro and toolchain without making the dictionary more complex.

I did use a dictionary for a similar purpose in `triplet_os` and `cmake_bin_relative_dir` :)

2. Yep done.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+        logging.critical(
+            'Please ensure you have write access to the parent directory for gennytoolchain: '
+            '`sudo chown $USER %s`', install_dir)
+        sys.exit(1)
+
+    toolchain_tarball = os.path.join(install_dir, 'gennytoolchain.tgz')
+    if os.path.isfile(toolchain_tarball):
+        logging.info('Skipping downloading %s', toolchain_tarball)
+    else:
+        logging.info('Downloading gennytoolchain, please wait...')
```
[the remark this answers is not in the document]
guoyr: Per your comment below, I also think the URL should be a debug-only log, so not logging it here. (It's logged on line 35 already).

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+        urllib.request.urlretrieve(url, toolchain_tarball)
+        logging.info('Finished Downloading gennytoolchain as %s', toolchain_tarball)
+
+    toolchain_dir = os.path.join(install_dir, 'gennytoolchain')
+    if os.path.exists(toolchain_dir):
+        logging.info('Skipping extracting the toolchain into: %s', toolchain_dir)
+    else:
+        logging.info('Extracting gennytoolchain, please wait...')
```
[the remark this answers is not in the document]
guoyr: Done.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+        tarball = tarfile.open(toolchain_tarball)
+        tarball.extractall(toolchain_dir)
+        logging.info('Finished extracting gennytoolchain into %s', toolchain_dir)
```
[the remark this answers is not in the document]
guoyr: Yeah I understand your sentiment, which is why I didn't include the URL and the output dir. The otherwise useless log here is just to tell the user the script is working. Without it, it looks like the script hangs with no output at all.

The logging.info below are to prepare users for the cmake/ninja/ctest output. I want to log the command we used so the cmake output can make sense if people.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def compile(env):
+    compile_cmd = ['ninja', '-C', 'build']
+    logging.info('Running ninja: %s', compile_cmd)
+    subprocess.run(compile_cmd, env=env)
+
+
+def install(env):
+    install_cmd = ['ninja', '-C', 'build', 'install']
```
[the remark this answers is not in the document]
guoyr: The readme doesn't mention either ninja or make. Everything's done through the wrapper.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+    cmake_cmd += [
+        '-DCMAKE_PREFIX_PATH={}'.format(cmake_prefix_path),
+        '-DCMAKE_TOOLCHAIN_FILE={}'.format(cmake_toolchain_file),
+        '-DVCPKG_TARGET_TRIPLET=x64-{}-static'.format(triplet_os),
+    ]
+
+    cmake_cmd += cmdline_cmake_args
+
+    logging.info('Running cmake: %s', cmake_cmd)
+    subprocess.run(cmake_cmd, env=env)
```
bcaimano: I think you'll really want to specify an output file, subprocess can have issues with output buffer flood.

lamp.py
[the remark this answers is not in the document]
bcaimano: Yeah, I do really like ninja but not everybody has it. Can we give people options here? (`--use-ninja`, `--use-make`)

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+    subprocess.run(compile_cmd, env=env)
+
+
+def install(env):
+    install_cmd = ['ninja', '-C', 'build', 'install']
+    logging.info('Running ninja install: %s', install_cmd)
+    subprocess.run(install_cmd, env=env)
+
+
+def cmake_test(env):
```
bcaimano: Not saying anything actionable here, but since catch2 does provide cohesive test suite options, it would be really nice to ditch ctest. I've rarely seen it provide anything of value.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+    ctest_cmd = [
+        'ctest',
+        '--label-exclude',
+        '(standalone|sharded|single_node_replset|three_node_replset)'
+    ]
+    subprocess.run(ctest_cmd, cwd=workdir, env=env)
+
+
+def parse_args(args, os_family):
+    parser = argparse.ArgumentParser(
```
bcaimano: It's worth moving the parser into the module namespace, it's a nice step toward composable argparse parsers.

lamp.py
```diff
@@ -0,0 +1,209 @@
...
+    subparsers = parser.add_subparsers(
+        dest='subcommand',
+        description='subcommands perform specific actions; make sure you run this script without any '
+                    'subcommand first to initialize the environment')
+    subparsers.add_parser(
+        'cmake-test', help='run cmake unit tests that don\'t connect to a MongoDB cluster')
+    subparsers.add_parser('compile', help='just run the compile step for genny')
+    subparsers.add_parser('install', help='just run the install step for genny')
+
+    known_args, unknown_args = parser.parse_known_args(args)
```
bcaimano: It is worth adding a `--cmake-arg` as an append argument instead. (It would be used like `--cmake-arg "-DFOO=BAR"`.) It will make it much much clearer what is consumed and what is passed along.

lamp.py
bcaimano: Well, it's never pretty if you want it to happen entirely through CMake. You can introduce a `check` target and have everything hang off. Personally, I'd rather just enumerate via text, it's not like people will be adding new tests every day.

lamp.py
bcaimano: Hmmmm, that is true if you don't quote the cmake args, but we really should anyway.

Just making sure I wasn't crazy:
```
In [25]: parser = argparse.ArgumentParser()                                                                                                                  

In [26]: parser.add_argument("--cmake-arg", nargs=1, action='append')                                                                                        
Out[26]: _AppendAction(option_strings=['--cmake-arg'], dest='cmake_arg', nargs=1, const=None, default=None, type=None, choices=None, help=None, metavar=None)

In [27]: parser.parse_known_args('--cmake-arg "-DFOO=BAR" --cmake-arg "-DBAR"'.split())                                                                      
Out[27]: (Namespace(cmake_arg=[['"-DFOO=BAR"'], ['"-DBAR"']]), [])
```

evergreen.yml
```diff
@@ -335,47 +340,28 @@ functions:
...
+        # for python3
+        export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
+
+        export DEVELOPER_DIR="/Applications/Xcode10.1.app"
+
+        ./scripts/lamp \
+          -d=${distro|not-linux} \
+          -DENABLE_JASPER:BOOL=ON \
+          -DGENNY_INSTALL_DIR:PATH=${workdir}/dist \
+          -DGENNY_INSTALL_SSL:BOOL=ON
```
rtimmons: I think you need to quote each of these args to prevent spaces from wreaking havoc

```sh
./scripts/lamp \
  "-d=${distro|not-linux}" \
  ...
  "-DGENNY_INSTALL_DIR=${workdir/dist" \
```

don't technically need to quote the lines without var expansions but I would just for consistency

scripts/lamp
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python3
+
+"""
+Because genies come from lamps.
```
rtimmons: 😸

guoyr:
The build is breaking again because we reverted giving the mci user access to `/opt`. I'm going to put the toolchain in `/data/mci` for now until we find a more permanent solution.

https://github.com/10gen/buildhost-configuration/commit/2469615afdc40c096f451e0cb690dd03414db45a#diff-93052275de1596ce5518360c53ba149a

README.md
rtimmons: > If it can't be fixed easily (because it's boost), would you be OK with a banner + followup ticket for now?

Ya fine by me but compiling warning-free is definitely something I value highly so I'd want to prioritize as a priority bug-fix. Just my $0.02.

guoyr:
Fixed the visibility issue by compiling genny with `-fvisibility=hidden` as recommended for compilers and libraries that support visibility=hidden.

As expected, this fixes one thing but creates a new problem in that catch2 doesn't understand visibility when we link it as a shared library. Changing it to a static library triggers the actor producer not found issue.

I'm going to investigate the actor producer not found issue next

bcaimano:
Hmmmm, @guoyr, take a look at `--whole-archive`.

guoyr:
@bcaimano Beautiful!! That did the trick after statically linking `cast_core`.

```
SET(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wl,-all_load") # For clang.
```

I might  take another stab at fixing the visibility issue in boost, it's more prevalent than I thought.

guoyr:
Ready for another round. I confirmed in the latest commit that there are no new compile warnings or anything else unexpected on all the build variants.

Also made a couple tweaks to make the toolchain versioned and the macOS binaries relocatable.

CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ find_package(
 )
 # </Boost>
 
+# <yaml-cpp>
+find_package(yaml-cpp CONFIG REQUIRED)
+# <yaml-cpp>
```
rtimmons: maybe put the `find_package(catch2)` in here as well so `src/third_party` is only used for vendored libs

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
 find_package(
         Boost
-        1.58 # Minimum version, vetted via Ubuntu 16.04
+        1.68
```
bcaimano: So this is actually a minimum version number, not an explicit version. FindBoost is wonky. Did you need the upgrade?

cmake/ConfigureInstall.cmake
```diff
@@ -17,7 +17,7 @@ include(GNUInstallDirs)
 # This line allows us to run genny executables from a distribution folder where libraries live in
 # bin/../lib64. This is a nice thing for driver-like applications, because you can pass around
 # archives trivially.
-set(CMAKE_INSTALL_RPATH                 "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"   CACHE PATH "")
+set(CMAKE_INSTALL_RPATH "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}" "@loader_path/../${CMAKE_INSTALL_LIBDIR}" CACHE PATH "")
```
bcaimano: Yep, that looks about like what I'd expect for loader path.

CMakeLists.txt
guoyr: Nah, didn't run into any problems. Thought I'd change this number so people can use 1.68 features.

CMakeLists.txt
rtimmons: Nevermind here - I see the `find_package` is still there for the "real" catch2 impl and we just have the "main" file now. Fine to leave as-is!

README.md
```diff
@@ -4,131 +4,28 @@ Genny 🧞‍
...
-
-```sh
-cmake -E chdir "build" cmake -DCMAKE_EXE_LINKER_FLAGS=-Wl,--no-as-needed ..
-make -j8 -C "build" genny
-```
+1. `./scripts/lamp [--linux-distro ubuntu1804/rhel7/amazon2/arch]`
+ 
+    This command downloads genny's toolchain, compiles genny and
+    installs genny to `dist/`. You can rerun this command at any time
+    to rebuild genny. If your OS isn't the supported, please let us
```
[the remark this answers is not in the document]
guoyr: macOS has the same command. The linux-distro part is in brackets as they are optional. I added a sentence at the end to clarify that the linux distro part is not needed for macOS

guoyr:
Re notes:

> Still getting `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: .../genny/dist/lib/libvalue_generators.a(DefaultRandom.cpp.o) has no symbols`. The file just has an explicit template instantiation but not sure why it wouldn't have symbols. Not worth fixing imho but maybe put a "known warning" or something there.

I'll file a ticket to look into it. Should be a quick fix.

* After running `lamp`, CLion complains with `Cannot generate into ... It was created with incompatible generator 'Ninja'`. I can `cd build && rm -rf * && cmake ..` myself but being able to specify that I want `Unix Makefiles` instead of `Ninja` would be nice.

Added a section to the readme to remove the `-G Ninja` part when using an IDE. Alternatively you can install ninja.

* It took a while to download the toolchain. I'm on a 30mbps connection here and wasn't sure what to expect. Maybe somehow print out the file size (even if it's hard-coded e.g. `Downloading gennytoolchain (approx XYZmb) please wait..`)

Done

PR #159, opened 2019-02-07
TIG-1408 update README with simplified genny patch build steps

guoyr:
README change for #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,10 +258,9 @@ You will see an output like this:
       Finalized : No
 ```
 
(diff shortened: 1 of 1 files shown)


guoyr:
macOS rpath should be fixed but the Evergreen task still hangs due to https://jira.mongodb.org/browse/CDRIVER-2825. When we dry run the AuthNInsert workload, the certs don't exist, which hangs on macOS.

The fix is due to go in the next mongoc driver release. I'll see if there's an easy workaround

PR #160, opened 2019-02-08
TIG-1434 use count_documents instead of count

guoyr:
Based on #158


guoyr:
notes from standup. Add two new crud actors for the two counts

PR #161, opened 2019-02-08
TIG-1140: Support Named Connection Pools

rtimmons:
This is a followup for #157. This replaces the `Pool` yaml syntax keyword with `Clients` that is now a map to allow for multiple distinct connection pools to be used with multiple instances of each of those pools as well.


src/gennylib/src/PoolManager.cpp
```diff
@@ -79,3 +79,11 @@ mongocxx::pool::entry genny::PoolManager::client(const std::string& name,
     }
     return std::move(*entry);
 }
+
+std::unordered_map<std::string, size_t> genny::v1::PoolManager::instanceCount() {
+    auto out = std::unordered_map<std::string, size_t>();
```
visemet: Add `std::lock_guard<std::mutex> getLock{this->_poolsGet}` above this line? I realize this is meant to only be used for testing purposes but it seems easy enough to make it thread-safe.

src/gennylib/test/PoolFactory_test.cpp
```diff
@@ -231,15 +231,15 @@ TEST_CASE("PoolFactory behavior") {
...
-        auto w = ah.workload();
-        w->client("Foo", 0);
-        w->client("Foo", 10);
-        w->client("Bar", 1);
+        genny::v1::PoolManager manager{"mongodb:://localhost:27017", {}};
+        genny::v1::ConfigNode config{YAML::Load(R"()")};
+
+        auto foo0 = manager.client("Foo", 0, config);
+        auto foo0again = manager.client("Foo", 0, config);
+        auto foo100 = manager.client("Foo", 10, config);
```
visemet: Change foo100 -> foo10.

src/gennylib/test/PoolFactory_test.cpp
```diff
@@ -231,15 +231,15 @@ TEST_CASE("PoolFactory behavior") {
...
+        auto bar0 = manager.client("Bar", 0, config);
+
+        REQUIRE((manager.instanceCount() ==
+                 std::unordered_map<std::string, size_t>({{"Foo", 2}, {"Bar", 1}})));
```
visemet: [note] This assertion doesn't actually verify that we aren't calling `createPool()` again when running `manager.client("Foo", 0, config)` a second time. A different style of trying to write this test is to register a callback which gets called by `createPool()` and use that to "spy on" the `name` and `instance` for which `createPool()` gets called. I'm fine leaving this as-is but just wanted to mention it as another example of where something like [TIG-1191](https://jira.mongodb.org/browse/TIG-1191) would probably be helpful.

src/gennylib/src/PoolManager.cpp
rtimmons: Good idea, done

src/gennylib/test/PoolFactory_test.cpp
rtimmons: Ya I considered going to that extreme but didn't feel worth it. I added a comment to this test (borrowing most of your words) for future maintainers. If this is ever seriously refactored that's a consideration that should be taken into mind.

rtimmons:
Thanks for review @visemet. Ready for another look when you can. No rush.

src/gennylib/src/PoolManager.cpp
visemet: You can use `std::lock_guard` instead of `std::unique_lock` because you aren't doing anything fancy with unlocking the mutex partway through. (This came up before in https://github.com/mongodb/genny/pull/24#discussion_r218973538.)

PR #162, opened 2019-02-08
Remove cruft

rtimmons:
Had an itch to scratch. Removed a few `WorkloadContext` methods that aren't used and move some impls to `.cpp` files to be consistent.

### DIFF
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -118,7 +118,7 @@ public:
     /**
      * @return all the actors produced. This should only be called by workload drivers.
      */
-    const ActorVector& actors() const {
+    constexpr const ActorVector& actors() const {
         return _actors;
     }
 
@@ -270,14 +270,14 @@ public:
     /**
      * @return top-level workload configuration
      */
-    WorkloadContext& workload() const {
+    constexpr WorkloadContext& workload() const {
         return *this->_workload;
     }
 
     /**
      * @return the workload-wide Orchestrator
      */
-    Orchestrator& orchestrator() const {
+    constexpr Orchestrator& orchestrator() const {
         return *this->workload()._orchestrator;
     }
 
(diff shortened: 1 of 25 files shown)


src/gennylib/include/gennylib/context.hpp
```diff
@@ -384,51 +373,6 @@ class ActorContext final : public v1::ConfigNode {
         return this->_workload->_registry->operation(name);
     }
 
-    /**
```
guoyr: yeah!

src/gennylib/include/gennylib/context.hpp
```diff
@@ -112,35 +112,25 @@ class WorkloadContext : public v1::ConfigNode {
...
+    WorkloadContext(WorkloadContext&&) = delete;
     void operator=(WorkloadContext&&) = delete;
 
     /**
      * @return all the actors produced. This should only be called by workload drivers.
      */
-    constexpr const ActorVector& actors() const {
-        return _actors;
-    }
+    const ActorVector& actors() const;
```
guoyr: Hm, what's the benefit of putting this in the CPP file? I think in general people should be encouraged to `inline` and `constexpr` things on contexts since they could be performance critical (e.g. interactions with ShareableState). This particular function is probably unlikely to matter, but I'm less sure about some one liners below on the actor and phase contexts.

FWIW The Google style guide says it's OK to inline small functions. https://google.github.io/styleguide/cppguide.html#Inline_Functions

rtimmons: Okay good call - reverted most of these moves to the `.cpp`

rtimmons:
Thanks for review, @guoyr. Ready for another look whenever you can. No rush obv.

src/gennylib/include/gennylib/context.hpp
```diff
@@ -112,27 +112,21 @@ class WorkloadContext : public v1::ConfigNode {
...
     void operator=(WorkloadContext&) = delete;
-    WorkloadContext(WorkloadContext&&) = default;
+    WorkloadContext(WorkloadContext&&) = delete;
     void operator=(WorkloadContext&&) = delete;
 
     /**
      * @return all the actors produced. This should only be called by workload drivers.
      */
-    constexpr const ActorVector& actors() const {
+    const ActorVector& actors() const {
```
guoyr: Keep the `constexpr`?

PR #163, opened 2019-02-11
TIG-1328 Migrate all Actors from previous metrics types to metrics::Operation

vrachev:
Please check that I correctly deprecated the previous metrics types. 

Some existing genny classes are still using metrics::registry and metrics::guage (such as orchestrator.hpp). Not sure if we'll want to do that at some point or not. But all of the actors in cast_core are now using metrics::operation.

### DIFF
--- a/src/cast_core/src/MultiCollectionUpdate.cpp
+++ b/src/cast_core/src/MultiCollectionUpdate.cpp
@@ -83,7 +83,7 @@ void MultiCollectionUpdate::run() {
 MultiCollectionUpdate::MultiCollectionUpdate(genny::ActorContext& context)
     : Actor(context),
       _rng{context.workload().createRNG()},
-      _updateOp{context.operation("update", MultiCollectionUpdate::id())},
+      _updateOp{context.operation("Update", MultiCollectionUpdate::id())},
       _client{std::move(context.client())},
       _loop{context, _client} {}
 
--- a/src/cast_core/src/MultiCollectionUpdate.cpp
+++ b/src/cast_core/src/MultiCollectionUpdate.cpp
@@ -71,9 +71,10 @@ void MultiCollectionUpdate::run() {
             // BOOST_LOG_TRIVIAL(info) << "Collection Name is " << collectionName;
             {
                 // Only time the actual update, not the setup of arguments
-                auto op = _updateTimer.raii();
+                auto opCtx = _updateOp.start();
                 auto result = collection.update_many(std::move(filter), std::move(update));
-                _updateCount.incr(result->modified_count());
+                opCtx.addOps(result->modified_count());
+                opCtx.success();
             }
         }
     }
@@ -82,8 +83,7 @@ void MultiCollectionUpdate::run() {
 MultiCollectionUpdate::MultiCollectionUpdate(genny::ActorContext& context)
     : Actor(context),
       _rng{context.workload().createRNG()},
-      _updateTimer{context.timer("updateTime", MultiCollectionUpdate::id())},
-      _updateCount{context.counter("updatedDocuments", MultiCollectionUpdate::id())},
+      _updateOp{context.operation("updateOperation", MultiCollectionUpdate::id())},
       _client{std::move(context.client())},
       _loop{context, _client} {}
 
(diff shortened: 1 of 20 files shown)


src/gennylib/include/gennylib/Orchestrator.hpp
```diff
@@ -107,7 +107,7 @@ class Orchestrator {
     PhaseNumber _current = 0;
 
     // _phaseNumberGauge should keep up to date with _current.
-    metrics::Gauge _phaseNumberGauge;
+    metrics::v1::Gauge _phaseNumberGauge;
```
rtimmons: Let's just kill `_phaseNumberGauge`. We'll probably want something similar in the future but I think it really wants to be a custom gauge within metrics::Operation (requires TIG-1420).

src/metrics/include/metrics/metrics.hpp
```diff
@@ -555,14 +554,14 @@ class Registry {
...
     }
-    Timer timer(const std::string& name) {
-        return Timer{this->_timers[name]};
+    v1::Timer timer(const std::string& name) {
+        return v1::Timer{this->_timers[name]};
     }
-    Gauge gauge(const std::string& name) {
-        return Gauge{this->_gauges[name]};
+    v1::Gauge gauge(const std::string& name) {
+        return v1::Gauge{this->_gauges[name]};
```
rtimmons: Let's mark these and the associated methods in `ActorContext` as deprecated - see [this SO answer](https://stackoverflow.com/questions/295120/c-mark-as-deprecated).

src/cast_core/src/MultiCollectionUpdate.cpp
```diff
@@ -82,8 +83,7 @@ void MultiCollectionUpdate::run() {
 MultiCollectionUpdate::MultiCollectionUpdate(genny::ActorContext& context)
     : Actor(context),
       _rng{context.workload().createRNG()},
-      _updateTimer{context.timer("updateTime", MultiCollectionUpdate::id())},
-      _updateCount{context.counter("updatedDocuments", MultiCollectionUpdate::id())},
+      _updateOp{context.operation("updateOperation", MultiCollectionUpdate::id())},
```
rtimmons: no need for the "operation" suffix and let's use PascalCase. So just call it `context.operation("Update")`

src/gennylib/include/gennylib/Orchestrator.hpp
rtimmons: Actually `timer()`, `gauge()` and `counter()` methods on `ActorContext` can probably just be removed now, ya?

src/cast_core/src/MultiCollectionUpdate.cpp
vrachev: Currently all of the actors use camelCase in these situations. Do we prefer if they were all PascalCase?

rtimmons: > Do we prefer if they were all PascalCase?

Yes.

vrachev: 👍

src/metrics/include/metrics/metrics.hpp
vrachev: After talking about this on slack with Ryan we decided not to officially deprecate them, as they are still widely used internally and will spam the compile logs.

vrachev:
Ready for another look.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -86,8 +86,9 @@ using gauged_at_time = std::pair<time_point, gauged_type>;
 
 class Reporter;
 // The v1 namespace is here for two reasons:
-// 1) it's a step towards an ABI. These classes are basically the pimpls of the outer classes
-// 2) it prevents auto-completion of metrics::{X}Impl when you really want metrics::{X}
+// 1) it's a step towards an ABI. These classes are basically the pimpls of the outer classes.
+// 2) it prevents auto-completion of metrics::{X}Impl when you really want metrics::{X}.
+// 3) Counter, Timer and Guage have been deprecated. Use Operation instead when writing actors.
```
rtimmons: nit - this isn't one of the reasons why the v1 namespace is here :) Just put a blank line here and nix the number

```c++
// 2) ...
//
// Note: Counter, Timer, and Gauge...
```

PR #165, opened 2019-02-12
Allow user to build with make rather than ninja

rtimmons:
CLion got me down so added the option to build with make 😸

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ copy and paste the options into Clion. The options
 should look something like this:
 
 ```bash
--G "Unix Makefiles" \
+-G some-build-system \
 -DCMAKE_PREFIX_PATH=/data/mci/gennytoolchain/installed/x64-osx-shared \
 -DCMAKE_TOOLCHAIN_FILE=/data/mci/gennytoolchain/scripts/buildsystems/vcpkg.cmake \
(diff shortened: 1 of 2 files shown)


guoyr:
Thanks for putting up the fix! I'm kind of surprised at the behavior you're seeing.

For me, I'm using the default CLion CMake path of `cmake-build-debug` and the default manual build CMake path of `build` so there isn't a conflict. Even if you specify `-G Ninja`, CLion will override it with `-G "CodeBlocks - Unix Makefiles"` and I don't have to do anything.

Are you doing anything special with your CLion cmake? If not, could you post the logs of its cmake output when you run "Reload Project", I'm curious about what it's doing.

rtimmons:
I have CLion also building into `build` rather than its weird `cmake-build-debug` dir. I tried it its way but I kept getting confused about which binary/library I was looking at. Plus doing it this way (having `lamp` specify `-G makefiles`) you don't have to enter any custom cmake settings in your build config (you just have to run `lamp` first before opening it up in clion)

rtimmons:
@guoyr thanks for review.
Sorry in advance but I just realized this will probably cause conflicts for you on #167 - lemme know if you want a hand with the merge

guoyr:
Haha no worries. I'll test out `-b` again after a manual merge.

PR #166, opened 2019-02-12
TIG-1223: Warn if accessing phaseConfig-> on nop phases

rtimmons:
Decided in pointing to defer TIG-1223 and instead just warn for now.

This is our first real use for `NDEBUG`. I don't know if the server does something fancier that we may want to adopt?

Also changed to use vanilla `unique_ptr<T>` rather than `optional<unique_ptr<T>>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,14 @@ When creating a new actor, `create-new-actor.sh` will generate a new test case
 template to ensure the new actor can run against different MongoDB topologies,
 please update the template as needed so it uses the newly created actor.
 
+### Debugging
+
+IDEs can debug Genny if it is built with the `Debug` build type:
+
+```sh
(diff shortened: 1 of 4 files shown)


guoyr:
(Sorry about the test failures. They should be fixed after you rebase)

rtimmons:
> FYI The mongo server [sets a global bool](https://github.com/mongodb/mongo/blob/d34aa0fd73e9fc5aef3024059fce1e22067b9e54/src/mongo/util/debug_util.h#L40) so you can use if-statement to check a debug build instead of resorting to the preprocessor. The compiler should be smart enough to remove the if statement completely on opt builds.

Good to know I'll create a ticket to standardize on that and actually use debug builds as part of ci testing to catch things or something.

PR #167, opened 2019-02-13
TIG-1404 bubble up test failures

guoyr:
As part of this work
- fixed global rate limiter test
- split up the lamp script
- added unit test for the new test failure code.


scripts/lamplib/tasks/run_tests.py
```diff
@@ -0,0 +1,39 @@
+import os
```
guoyr: This file and `test_run_tests.py` are the only two pieces of new Python code. The rest of the Python files is mostly splitting up the original script.

scripts/lamplib/__main__.py
```diff
@@ -0,0 +1,63 @@
...
+        tasks.compile_all(env)
+
+    if args.subcommand == 'install':
+        tasks.install(env)
+    elif args.subcommand == 'cmake-test':
+        tasks.run_tests.cmake_test(env)
+
+
+if __name__ == '__main__':
+    main()
```
rtimmons: I like moving/splitting the files but can we please put inside `src/python`?

scripts/lamplib/context.py
```diff
@@ -0,0 +1,40 @@
...
+class Context:
+    # Permanent constants.
+    TOOLCHAIN_BUILD_ID = '7297534c3dc1df9f6d315098174171eaf75c5846_19_02_09_02_19_44'
+    TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split('_')[0]
+
+    # User configurable constants.
+    IGNORE_TOOLCHAIN_VERSION = False
+
+    # User configurable dynamic context.
+    get_compile_environment = _create_compile_environment
```
rtimmons: why not define it in the class?

```python
class Context:
  def get_compile_enviornment(self, ...): ...
```

evergreen.yml
```diff
@@ -418,6 +419,25 @@ functions:
...
+      shell: bash
+      script: |
+        set -o errexit
+        set -o pipefail
+        set -o nounset
+
+        # for python3
+        export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
+
+        ./scripts/lamp self-test
```
rtimmons: this new task goes away if you put `lamplib` into `src/python`, ya?

PR #169, opened 2019-02-19
TIG-1101: Add support for faking out steady and system clocks

visemet:
All of the calls to `metrics::clock::now()` have been consolidated into `MetricsClockSource::now()`.

----

I left 9f7bebf61b6e9da620c33aa7cf627c5ea157f920 in the commit history so you could see the path I tried to go down (but ultimately abandoned) for templatizing `metrics::Registry`.


src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -70,18 +70,27 @@ class Reporter {
...
      *            data-points to this ostream.
      * @param metricsFormat the format to use. Either "csv" or "sys-perf".
-     * @param perm passkey permission (must be friends with metrics::Registry)
      */
+    template <class ClockSource = SystemClockSource>
     void report(std::ostream& out,
-                const std::string& metricsFormat,
-                v1::Permission perm = {}) const {
+                const std::string& metricsFormat) const {
+        v1::Permission perm;
```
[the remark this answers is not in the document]
visemet: I might be missing some part of how the passkey idiom works, but I was surprised the code compiled previously. My understanding is that only `genny::metrics::Reporter` is a friend of `genny::metrics::v1::Permission`. How would callers of methods on the `genny::metrics::Reporter` class be able to construct `genny::metrics::v1::Permission` instances? To me, the `genny::metrics::Reporter` class should be taking care of that for the caller. Note that when I introduced the templating to the `.report()` method—and further still when I introduced the templating to the `genny::metrics::v1::ReporterT` class)—compilation broke.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -84,6 +84,18 @@ using duration_at_time = std::pair<time_point, period>;
 using count_at_time = std::pair<time_point, count_type>;
 using gauged_at_time = std::pair<time_point, gauged_type>;
 
+class ClockSource {
```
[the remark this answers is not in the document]
visemet: I reverted back to my original approach of adding a `ClockSource` template parameter to the `genny::metrics::Registry` class. There's no longer a `ClockSource` class with `virtual time_point now()` method.

(The key was to also define additional `using` aliases for `genny::metrics::Operation` and `genny::metrics::OperationContext` to hide away the template parameter as we had discussed in-person.)

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -22,108 +22,106 @@
...
-                v1::Permission perm = {}) const {
+    template <typename ReporterClockSource = SystemClockSource>
+    void report(std::ostream& out, const std::string& metricsFormat) const {
+        v1::Permission perm;
+
         // should these values come from the registry, and should they be recorded at
         // time of registry-creation?
-        auto systemTime = std::chrono::system_clock::now().time_since_epoch().count();
-        auto metricsTime = _registry->now(perm).time_since_epoch().count();
+        auto systemTime = std::chrono::duration_cast<std::chrono::nanoseconds>(
```
visemet: It looks like we were calling `.count()` on the `std::chrono::duration` returned by the `.time_since_epoch()` without knowing what units were being used for the system clock. It turns out to be different between Amazon Linux 2 and Arch Linux (nanoseconds) and RHEL 7.0, Ubuntu 18.04, and macOS (microseconds).

src/metrics/include/metrics/metrics.hpp
```diff
@@ -557,55 +570,84 @@ class Operation {
...
+};
+
+using Registry = v1::RegistryT<MetricsClockSource>;
+
 static_assert(std::is_move_constructible<Registry>::value, "move");
 static_assert(std::is_move_assignable<Registry>::value, "move");
 
+// Registry::clock is actually the ClockSource template parameter, so we check the underlying
+// clock_type in a roundabout way by going through the time_point type it exposes.
+static_assert(Registry::clock::time_point::clock::is_steady, "clock must be steady");
```
rtimmons: consider making this a typedef on `Registry`? idk.
```c++
class Registry { public: 
  using clock_is_steady = ClockSource::time_point::clock::is_steady;
};
```

src/metrics/test/metrics_test.cpp
```diff
@@ -63,6 +106,62 @@ TEST_CASE("example metrics usage") {
...
+        "14,client.query,0\n"
+        "17,client.query,1\n"
+        "19,client.query,0\n"
+        "21,client.query,0\n"
+        "24,client.query,1\n"
+        "\n";
+
+    std::ostringstream out;
+    reporter.report<ReporterClockSourceStub>(out, "csv");
+    REQUIRE(out.str() == expected);
```
rtimmons: this is great!

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -22,108 +22,106 @@
...
-
-        auto micros = std::chrono::duration_cast<std::chrono::microseconds>(total);
-
-        out << R"({"storageEngine":"wiredTiger", "results":[{
- "name":"dummy_inserts", "workload":"genny_dummy_workload",
- "start":1537815283.968272, "end":1537817860.423682,
- "results":{"4":{"ops_per_sec":)"
-            << micros.count() << R"(, "ops_per_sec_values":[)" << micros.count() << R"(]}}}]}
-)";
-    }
```
rtimmons: love to see this gone. I think also need to remove mention of the format in the program-options thing in DefaultDriver.

src/metrics/include/metrics/MetricsReporter.hpp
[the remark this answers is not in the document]
visemet: > In any event, here and elsewhere in this `Reporter` class I think you can remove the extra instantiation of the permission:
> 
> ```c++
>          auto& x = _registry->getGauges(v1::Permission{}); 
> ```

Done.

src/metrics/include/metrics/MetricsReporter.hpp
visemet: The server codebase defines [a `durationCount()` helper function](https://github.com/mongodb/mongo/blob/r4.1.8/src/mongo/util/duration.h#L125-L141) to make it so the number of ticks is logged according to the intended units. I added a `nanosecondsCount()` helper function since we don't need the extra parameterization.

> consider making the unit a typedef here so refactoring couldn't cause the units of metricsTime and systemTime to change independently.

I'd rather we say "nanoseconds" so it is more obvious to future readers what the units in the output are without needing to look at another typedef. (I also plan to add a header for the "Clocks" section that says "nanoseconds" in the "csv" file too.)

src/metrics/include/metrics/metrics.hpp
```diff
@@ -557,55 +570,84 @@ class Operation {
...
-    std::unordered_map<std::string, v1::TimerImpl> _timers;
-    std::unordered_map<std::string, v1::GaugeImpl> _gauges;
+    std::unordered_map<std::string, v1::CounterImpl<ClockSource>> _counters;
+    std::unordered_map<std::string, v1::TimerImpl<ClockSource>> _timers;
+    std::unordered_map<std::string, v1::GaugeImpl<ClockSource>> _gauges;
 };
 
+}  // namespace v1
+
+class MetricsClockSource {
```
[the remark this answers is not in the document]
visemet: Done.

src/metrics/include/metrics/MetricsReporter.hpp
visemet: I had grep'd for "sys-perf" - [it doesn't look like there are any mentions of the valid options in the DefaultDriver help text](https://github.com/mongodb/genny/blob/5a8c92e7e1003af6ed099419a6a2fe88f8ca273e/src/driver/src/DefaultDriver.cpp#L197-L199).

src/metrics/include/metrics/metrics.hpp
visemet: `Registry::clock` is the source of truth for how `::now()` is called so my preference is to check against that still. I'm inclined to leave this as-is.

src/gennylib/benchmark/orchestrator_benchmark.cpp
```diff
@@ -32,7 +30,6 @@ using namespace std::chrono;
 TEST_CASE("Orchestrator Perf", "[benchmark]") {
     const auto dur = milliseconds{200};
 
-    genny::metrics::Registry metrics;
```
visemet: Due to https://bugs.archlinux.org/task/61786, I wasn't able to run `perf` on Arch Linux to profile why after the changes from 5a8c92e7e1003af6ed099419a6a2fe88f8ca273e slowed things down. There's likely some difference in the generated code that may be impacting inlining of some of the function calls. What's especially odd to me is that `genny::Orchestrator` doesn't even make use of `genny::metrics::Registry`.

PR #170, opened 2019-02-19
TIG-1465: Resolve Genny Warnings

rtimmons:
- Git-ignore `__pycache__` (generated when you run `scripts/lamp`)
- Allow `metrics::OperationContext` to be moved so it can be constructed inside a `std::optional` and not go out of scope as an rval
- RunCommand should still call `op.fail()` if it catches an exception
- Kill `DefaultRandom.cpp` because it causes a linker warning on macOS
- Don't include `metrics.hpp` on Orchestrator. It's not needed anymore.
- A few `clang-tidy` recommendations


src/metrics/include/metrics/metrics.hpp
```diff
@@ -473,9 +473,18 @@ class Timer {
...
 
 public:
-    OperationContext(v1::OperationImpl& op)
+    explicit OperationContext(v1::OperationImpl& op)
         : _op{std::addressof(op)}, _started{metrics::clock::now()} {}
 
+    OperationContext(OperationContext&& rhs) noexcept
+        : _op{std::move(rhs._op)},
+          _started{std::move(rhs._started)},
+          _isClosed{std::move(rhs._isClosed)},
```
visemet: [optional] Consider writing `_isClosed{std::exchange(rhs._isClosed, true)}` and removing line 485.

rtimmons: Interesting. Considered but that seemed harder to reason about so left as-is 😄

PR #172, opened 2019-02-21
TIG-1386 add test for create-new-actor.sh

guoyr:


evergreen.yml
```diff
@@ -399,15 +413,32 @@ functions:
...
+      shell: bash
+      script: |
+        set -o errexit
+        set -o pipefail
+        set -o nounset
+
+        # for python3
+        export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
+
+        ./scripts/lamp resmoke-test ${resmoke_test_args}
```
rtimmons: 1. can we make `lamp` smart enough to know about the toolchain rather than copy/pasting the `export PATH` in all the places we call it?
2. also I think `set -eou pipefail` is good enough for the first 3 lines

with both of these this block becomes just 2 lines which would help keep our evergreen yaml burden down.

src/lamplib/parser.py
```diff
@@ -7,7 +7,8 @@
 def parse_args(args, os_family):
     parser = argparse.ArgumentParser(
         description='Script for building genny',
-        epilog='Unknown positional arguments will be forwarded verbatim to the cmake invocation'
+        epilog='Unknown positional arguments will be forwarded verbatim to the cmake'
+               'invocation where relevant'
```
rtimmons: add a space at beginning of second string `' invocation...'`

scripts/create-new-actor.sh
```diff
@@ -322,24 +322,23 @@ create_impl                  "$uuid_tag" "$actor_name"
 create_test                  "$actor_name"
 create_workload_yml          "$actor_name"
 
+actor_name_lowercase=$(echo "$actor_name" | tr '[:upper:]' '[:lower:]');
```
rtimmons: why do we lowercase the actor name? keeping it consistent would get us out of having multiple case conventions to worry about.

guoyr: Hm, catch2's cmake integration does it for you: https://github.com/mongodb/genny/blob/916a2c20a3eb7813dbbc9595418a381099e09c89/cmake/ParseAndAddCatchTests.cmake#L144

No idea why.

I commented out the above line to have the tags' casing be consistent catch2 and ctest.

guoyr:
For posterity, here are two patch builds to check that test failures bubble up correctly:
1. bad create-new-actor.sh generated unit test: https://evergreen.mongodb.com/version/5c6eb859306615475af65d31
2. bad generated actor yaml: https://evergreen.mongodb.com/version/5c6eb432a4cf4741c45c2bd1

PR #173, opened 2019-02-23
TIG-1101: Remove .counter(), .gauge(), and .timer() APIs

visemet:
* Implements the "Genny.Setup" and "Genny.ActiveActors" metrics in terms of `metrics::Operation`s.
* Renames the `period` class to `Period` (uppercase).
* Moves the `Operation*`, `Period`, `Permission`, and `TimeSeries` classes to their own .hpp files.
* Adds `addErrors()` and `addIterations()` methods to `OperationContext`.
* Renames `OperationContext::addOps()` to `addDocuments()`.
* Removes `OperationContext::fail()` in favor of `failure()` and `discard()`. The former still reports the operation, whereas the latter doesn't. ExecutionStrategy and the RunCommand actor continue to not report the operation if it leads to an exception.

----

[Patch build](https://evergreen.mongodb.com/task/sys_perf_linux_1_node_replSet_insert_remove_patch_cbf90926f4bdc5f3fd6d20291a5e24218e5f77ba_5c71a5600305b91f4777677b_19_02_23_19_57_20) to show that the "csv" metrics format output hasn't changed and is still being processed correctly by DSI.


src/HitchhikersGuide.md
```diff
@@ -132,37 +132,90 @@ above.
...
+In your actor class, declare a `metrics::Operation` member and
+initialize it using the `ActorContext::operation()` method in the
+actor's constructor.
 
-You can increment a counter (for example, increment all documents inserted in a bulkinsert):
-`ctx.addOps(1);`
+```cpp
+class MyActor {
+    MyActor(ActorContext& context)
+        : _insertOp{context.operation("Insert", MyActor::id())} {}
```
rtimmons: is the order of the params right here?

src/HitchhikersGuide.md
```diff
@@ -132,37 +132,90 @@ above.
...
 
-You can also count bytes:
-`ctx.addBytes(document.view().length());`
+    // Note that while here the MyActor class here stores the
+    // metrics::Operation type, most other actors wrap it in an
+    // ExecutionStrategy in order to be able to handle retry logic.
```
rtimmons: ¯\\\_(ツ)_/¯ I'm not super keen on `ExecutionStrategy` proliferating as-is...maybe say "some actors"

src/HitchhikersGuide.md
```diff
@@ -132,37 +132,90 @@ above.
...
+```
 
+You can also use `metrics::OperationContext::discard()` to abandon the
+data points that were being collected for the operation.
 
+One of `success()`, `failure()`, or `discard()` must be called before
+the `metrics::OperationContext` instance goes out of scope. A warning
+message will be logged if they aren't as not explicitly handling any
+expected exceptions represents either a programmer error or unexpected
+server behavior.
```
rtimmons: 👍 nicely worded

src/gennylib/include/gennylib/context.hpp
```diff
@@ -354,69 +354,23 @@ class ActorContext final : public v1::ConfigNode {
...
-    auto counter(const std::string& operationName, ActorId id = 0u) const {
-        auto name = this->metricsName(operationName, id);
-        return this->_workload->_registry->counter(name);
-    }
-
     auto operation(const std::string& operationName, ActorId id = 0u) const {
-        auto name = this->metricsName(operationName, id);
-        return this->_workload->_registry->operation(name);
+        return this->_workload->_registry->operation(
+            id, this->get<std::string>("Name"), operationName);
```
rtimmons: no strong opinion but maybe we should remove the `id=0u` default now. If Actors don't specify it in two threads they'll unintentionally create races. I'm trying to think of a use-case when you'd actually want the default and can't think of any.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -595,54 +70,35 @@ class MetricsClockSource {
  */
 template <typename ClockSource>
 class RegistryT {
+private:
+    using EventSeries = typename v1::OperationImpl<ClockSource>::EventSeries;
+    using OperationsMap = std::unordered_map<
+        std::string,
+        std::unordered_map<std::string, std::unordered_map<ActorId, EventSeries>>>;
```
rtimmons: Type seems right but pretty dense. Let's add comments about what each of these sub-fields means:

```c++
using OperationsMap = std::unordered_map<
  std::string, // actor name
  std::unordered_map<std::string, // operation name
                     std::unordered_map<ActorId, EventSeries> // the series for each instance of the Actor
  >>
```

or create simpler `using`s for the sub-fields e.g.
```c++
using SeriesPerActorId = std::unordered_map<ActorId, EventSeries>>;
```
and then use those so you end up with something like
```c++
using OperationsMap = std::ordered_map<std::string, OperationsForActor>
```

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+    OperationContextT(OperationContextT<ClockSource>&& other) noexcept
+        : _op{std::move(other._op)},
+          _started{std::move(other._started)},
+          _event{std::move(other._event)},
+          _isClosed{std::exchange(other._isClosed, true)} {}
+
+    ~OperationContextT() {
+        if (!_isClosed) {
+            BOOST_LOG_TRIVIAL(warning)
```
rtimmons: maybe `error`? it is programmer-error that affects data-quality.

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+    void addBytes(count_type size) {
+        _event.size += size;
+    }
+
+    /**
+     * Increments the counter for the number of write errors, transient transaction errors, etc.
+     */
+    void addErrors(count_type errors) {
+        _event.errors += errors;
+    }
```
rtimmons: maybe default the params to 1 for all the `add*` methods?

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+        _isClosed = true;
+    }
+
+private:
+    void reportOutcome(OutcomeType outcome) {
+        auto finished = ClockSource::now();
+        _event.duration = finished - _started;
+        _event.outcome = outcome;
+
+        if (_event.iters == 0) {
```
rtimmons: maybe have this be `-1` to indicate unset - that way it's possible to indicate that no actual operations happened in this iteration. could be useful in cases where outcome=failure to indicate that maybe something happened during setup that prevented the operation from happening.

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+            // We default the event to represent a single iteration of a loop if addIterations() was
+            // never called.
+            _event.iters = 1;
+        }
+
+        _op->reportAt(finished, std::move(_event));
+        _isClosed = true;
```
rtimmons: since we moved-from the event should we put ourselves into a moved-from state here as well? (basically `swap` this with a new instance)

if the user calls `.failure()` and `.success()` (or either more than once) what will the actual behavior be? (But otoh you've got good comments and `@warning`s on the various methods so not a huge deal if this turns out to be painful to change)

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+public:
+    explicit OperationT(v1::OperationImpl<ClockSource> op) : _op{std::move(op)} {}
+
+    OperationContextT<ClockSource> start() {
+        return OperationContextT<ClockSource>(this->_op);
+    }
+
+private:
+    v1::OperationImpl<ClockSource> _op;
+};
```
rtimmons: I think the `OperationImpl` wants to be owned by the `Registry`, ya? If it is, the same actor could get the same `OperationImpl` by repeated calls, plus it makes the `Operation` a cheap value type. So I think either make the registry the owner or make this class non-copy/move-able.

src/metrics/include/metrics/operation.hpp
```diff
@@ -0,0 +1,260 @@
...
+    v1::OperationImpl<ClockSource>* const _op;
+    const time_point _started;
+
+    OperationEvent<ClockSource> _event;
+    bool _isClosed = false;
+};
+
+
+template <typename ClockSource>
+class OperationT {
```
rtimmons: we haven't been super consistent with it, but I think make all these classes `final` - they are not designed to be subclassed and it seems like good hygiene to document that intention.

src/metrics/include/metrics/passkey.hpp
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * The passkey idiom is way for a class to govern how its private members can be accessed by another
+ * class. It can be thought of as a finer-grained way to express friendship in C++. The passkey
+ * idiom works by defining a class with a private constructor and having that class be friends with
+ * the class seeking access. Since the constructor is private, only the class seeking access is able
+ * to construct instances of it (though could possibly share them). This means that the governing
+ * class can then define a public function that requires it as an argument and still be ensured that
+ * only the class seeking access is able to call that function.
+ *
+ * \sa https://arne-mertz.de/2016/10/passkey-idiom/
```
rtimmons: not documented but we've been using javadoc style tags e.g. `@sa` instead of `\sa`. Plus I think you can also use `@see` which is a bit more reader-friendly.

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -116,60 +81,158 @@ class ReporterT {
...
+            out << "Genny.ActiveActors";
+            out << ",";
+            out << numActors;
+            out << std::endl;
+        };
+
+        // The termination condition of the while-loop is based only on `finishedActors` because
+        // there should always be as many ActorFinished events as there are ActorStarted events and
+        // an actor can only be finished after it has been started.
+        while (finishedEventIt != finishedActors.end()) {
```
rtimmons: I admire your patience in writing this code considering it'll likely get tossed in your next PR 😸

src/HitchhikersGuide.md
visemet: It is - when we were going over the code in-person the `v1::RegistryT::operation()` is what had `ActorId` listed first. I realized it is probably confusing for it to have them in a different order from `ActorContext::operation()` so I changed `v1::RegistryT::operation()` to have the `ActorId` listed last.

src/HitchhikersGuide.md
visemet: Done.

src/gennylib/include/gennylib/context.hpp
visemet: Done.

src/metrics/include/metrics/metrics.hpp
visemet: I added a couple `using` declarations but went with the names `OperationsByThread` and `OperationsByType` to match the local variables I declared in `v1::RegistryT::operation()`.

src/metrics/include/metrics/operation.hpp
visemet: Done.

src/metrics/include/metrics/operation.hpp
visemet: Maybe @tychoish can say whether reporting `n=1` is (1) something Cedar currently supports, and (2) if so, something Cedar would intentionally keep supporting. Either way, I'm inclined to leave this as-is.

> could be useful in cases where outcome=failure to indicate that maybe something happened during setup that prevented the operation from happening.

I think I'd want to know more about the circumstances to understand why if the operation failed to happen that we'd want to report anything rather than log something and try to cause the Evergreen task to fail.

src/metrics/include/metrics/operation.hpp
visemet: I'm not sold on more syntax sugar helping all that much. I think not needing to call `addIterations(1)` altogether is the biggest win.

src/metrics/include/metrics/operation.hpp
visemet: > if the user calls `.failure()` and `.success()` (or either more than once) what will the actual behavior be?

I'm not sure `std::move(_event)` can realistically lead to anything other than the time series inserting a copy of `_event`. I suspect calling `success()` or `failure()` multiple times to report events as though multiple of them were started at the same time. How does Cedar handle effectively concurrent operations for the same (Actor, Operation, Thread) triplet?

> since we moved-from the event should we put ourselves into a moved-from state here as well? (basically `swap` this with a new instance)

What exactly are you trying to achieve? I'm curious if you're actually looking for `OperationContextT` to define a move assignment operator so the start time is able to be updated.

```cpp
OperationContextT& operator=(OperationContextT&& other) {
    _op = std::move(other._op);
    _started = std::move(other._started);
    _event = std::move(other._event);
    _isClosed = std::exchange(other._isClosed, true);
    return *this;
}
```

src/metrics/include/metrics/operation.hpp
visemet: Done. It meant exposing the `v1::OperationImpl<>::EventSeries` publicly because the "metrics::OperationContext interface" test case doesn't have a class name to reference as a friend. We could add one to `v1::Permission` that says it is for testing purposes only. /shrug

src/metrics/include/metrics/operation.hpp
visemet: Done.

src/metrics/include/metrics/passkey.hpp
visemet: Done.

visemet:
> but nothing that would warrant another look unless you'd like one.

@rtimmons could you take a look at the comment threads I left unresolved and respond to any of them you have more opinions on?

src/metrics/include/metrics/operation.hpp
rtimmons: > How does Cedar handle effectively concurrent operations for the same (Actor, Operation, Thread) triplet?

Cedar doesn't really care about thread numbers so it'd basically treat it the same as two threads reporting the same data at the same time.

> What exactly are you trying to achieve?

Designing types is hard especially given that we have an explicit move-from for only a single instance-var in a non-move operator/ctor. I get nervous whenever I see `std::move` on an instance-variable that doesn't result in the whole instance being in a moved-from state. I'm not sure if you need `operator=(&&)` as such (and it would force you to make some fields non-`const`).

I'm not sure what to suggest other than to have a test or two that calls both `failure()` and `success()` one or more times and asserts that we don't blow up. "Undefined reporting behavior" seems okay but segfaulting doesn't. I don't think this will segfault or anything but a refactor could change that.

src/metrics/include/metrics/operation.hpp
rtimmons: 👍 That's reasonable

src/metrics/include/metrics/operation.hpp
tychoish: I believe that cedar would appreciate itter counts

src/metrics/include/metrics/operation.hpp
visemet: > I get nervous whenever I see `std::move` on an instance-variable that doesn't result in the whole instance being in a moved-from state. I'm not sure if you need `operator=(&&)` as such (and it would force you to make some fields non-`const`).

I believe an instance-variable being in a moved-from state but the whole instance not being in one is how the `extract()` methods on the `bsoncxx::builder::basic::array` and `bsoncxx::builder ::basic::document` behave. The idea is that you could `std::move()` into them again (using the move assignment operator) and put them back into a valid state again.

> I'm not sure what to suggest other than to have a test or two that calls both `failure()` and `success()` one or more times and asserts that we don't blow up. "Undefined reporting behavior" seems okay but segfaulting doesn't. I don't think this will segfault or anything but a refactor could change that.

I'd rather genny crash loudly in cases of programmer error like this. While Catch2 doesn't support death tests, we could potentially [lift the functionality from the server codebase](https://github.com/mongodb/mongo/blob/r4.1.8/src/mongo/unittest/death_test.cpp).

PR #174, opened 2019-02-23
TIG-1101: Add support for the "cedar-csv" metrics format

visemet:
GitHub doesn't have an interface to be able to stack code reviews so don't bother looking at this one until #173 goes in first.


src/metrics/test/metrics_test.cpp
```diff
@@ -242,6 +242,33 @@ TEST_CASE("metrics output format") {
         reporter.report<ReporterClockSourceStub>(out, "csv");
         REQUIRE(out.str() == expected);
     }
+
+    SECTION("cedar-csv reporting") {
+        auto expected =
```
guoyr: I'm going to put this string into its own file in #175. Feel free to use it for this test as well. Sharing a file could be useful for picking up format changes on the Python side without having to run genny live.

guoyr: I've tentatively placed the file in `src/python/tests/fixtures/cedar/shared_with_cxx_metrics_test.csv`

See: https://github.com/mongodb/genny/pull/175/commits/f85df473566f438a73d3ac6d921b58b17ab72c88

visemet: I attempted to change metrics_test.cpp to read the `src/python/tests/fixtures/cedar/shared_with_cxx_metrics_test.csv` file in 359743a9c1d8862feae90e6b82e82d254195c6ae but ultimately back those changes out c99d12655d2b4ff68d6662bbd4b3b3ab78bdc7da in when I realized that my preferred way of running `./build/src/metrics/metrics_test` isn't compatible without we're running `ctest` out of the `./build/src/metrics` directory.

guoyr: Ah oh well. Thanks for trying it out!

guoyr: After reading some creative solutions to reading files in C++ on SO, I'm convinced it's far easier to just have python parse the C++ source file and extract this string.

guoyr: Sorry for the vagueness of the above comment; I posted it mostly for my own amusement.

A more elaborate description of a mechanism for C++ and Python tests to share fixtures is here: https://jira.mongodb.org/browse/TIG-1479

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -235,6 +242,75 @@ class ReporterT final {
...
+                if (actorName == "Genny" &&
+                    (opName == "ActorStarted" || opName == "ActorFinished")) {
+                    // The cedar-csv metrics format ignores the Genny.ActorStarted and
+                    // Genny.ActorFinished operations reported by the DefaultDriver because the
+                    // OperationThreadCounts section effectively tracks the number of concurrent
+                    // actors and that number isn't meaningfully changing over time.
+                    continue;
+                }
+
+                opThreadCounts[std::make_pair(actorName, opName)] += opsByThread.size();
```
rtimmons: any reason to use a map versus a [multiset](http://www.cplusplus.com/reference/unordered_set/unordered_multiset/)? I prefer the map version but some may argue multiset is more appropriate.

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -235,6 +242,75 @@ class ReporterT final {
...
+            out << count << std::endl;
+        }
+        out << std::endl;
+
+        out << "Operations" << std::endl;
+        out << "timestamp,actor,thread,operation,duration,outcome,n,ops,errors,size" << std::endl;
+        for (const auto& [actorName, opsByType] : _registry->getOps(perm)) {
+            for (const auto& [opName, opsByThread] : opsByType) {
+                if (actorName == "Genny" &&
+                    (opName == "ActorStarted" || opName == "ActorFinished")) {
```
rtimmons: consider putting this repeated `if` statement into a function `if skipReporting(actorName, opName)`

src/metrics/test/metrics_test.cpp
rtimmons: (we could easily "findup" the repo root by looking for `.git` and agree on the path from there. but fine to have this convo in TIG-1479)

src/metrics/include/metrics/MetricsReporter.hpp
visemet: Done.

src/metrics/include/metrics/MetricsReporter.hpp
visemet: I hadn't looked at the multiset type in C++ before. I think because we want the count of all elements and not just the count for a specific key that the `map<std::pair<std::string, std::string>, size_t>` presents an appropriate interface.

PR #176, opened 2019-02-28
TIG-1399 Support pre and post sleep

guoyr:


src/testlib/include/testlib/clocks.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+#include <metrics/Period.hpp>
+
+namespace genny::testing {
+using namespace std::literals::chrono_literals;
+
+// The template is to allow different DummyClocks classes to be created so calling static methods
+// won't conflict.
+template <typename DummyT>
+struct DummyClock {
```
guoyr: The changes to this file is unrelated to the feature but I think it's worth putting clocks together before we write another clock for testing sleep-related functionality

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -202,14 +226,19 @@ class ActorPhaseIterator final {
     }
 
     ActorPhaseIterator& operator++() {
-        if (_iterationCheck)
+        if (_iterationCheck) {
             _iterationCheck->limitRate(_currentIteration, std::cref(*_orchestrator), _inPhase);
+            _iterationCheck->sleepAfter(std::cref(*_orchestrator), _inPhase);
```
rtimmons: why is the `std::cref` necessary? Can you add a comment about it? I see type-decorators like that and it makes me think there's a hole in the type-system-design.

src/gennylib/include/gennylib/v1/Sleeper.hpp
```diff
@@ -0,0 +1,53 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef HEADER_A2EF8CA3_1185_4B6C_979B_BD40D6893EF8_INCLUDED
+#define HEADER_A2EF8CA3_1185_4B6C_979B_BD40D6893EF8_INCLUDED
+
+#include <chrono>
+#include <gennylib/conventions.hpp>
```
rtimmons: 💅 newline here they're separate "groups" of imports

src/gennylib/include/gennylib/v1/Sleeper.hpp
```diff
@@ -0,0 +1,53 @@
...
+public:
+    Sleeper(Duration before, Duration after) : _before(before), _after(after) {}
+
+    constexpr void before(const Orchestrator& o, const PhaseNumber pn) const {
+        if (_before.count() && o.currentPhase() == pn) {
+            std::this_thread::sleep_for(_before);
+        }
+    }
+
+    constexpr void after(const Orchestrator& o, const PhaseNumber pn) const {
```
rtimmons: please add docstrings for all methods and ctors.

src/gennylib/test/PhaseLoop_test.cpp
```diff
@@ -377,4 +378,29 @@ TEST_CASE("Actual Actor Example") {
...
+        ActorHelper ah(config, 1, {{"Inc", imvProducer}});
+
+        auto start = std::chrono::high_resolution_clock::now();
+        ah.run();
+
+        auto duration = std::chrono::high_resolution_clock::now() - start;
+
+        REQUIRE(duration > 450ms);
+        REQUIRE(duration < 550ms);
+    }
```
rtimmons: also:

1. a case where sleep is negative (should blow up)
2. a case where sleep is zero
3. No sleep

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -100,6 +101,18 @@ class IterationChecker final {
...
+        auto sleepAfter = phaseContext.get<TimeSpec, false>("SleepAfter").value_or(TimeSpec());
+
+        if ((sleepBefore || sleepAfter) && rateSpec) {
+            throw InvalidConfigurationException(
+                "Rate must *not* be specified alongside wither sleepBefore or sleepAfter"
+                "genny cannot enforce the global rate when there are mandatory sleeps in"
+                "each thread");
+        }
+        if (sleepBefore || sleepAfter)
+            _sleeper = std::make_unique<v1::Sleeper>(sleepBefore, sleepAfter);
```
rtimmons: why have it be a ptr instead of a value? We don't need runtime indirection (virtuals). Our styleguide isn't very clear on this but imho it's cleaner and better to use values rather than owning ptrs unless you need the non-copyable/runtime-dispatch that the pointer provides.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -149,6 +162,16 @@ class IterationChecker final {
         return _doesBlock;
     }
 
+    void sleepBefore(const Orchestrator& o, const PhaseNumber pn) const {
+        if (_sleeper)
+            _sleeper->before(o, pn);
+    }
```
rtimmons: no strong opinion on this but consider just always calling `_sleeper->before()` (and `after()`) and just let the sleeper itself decide if it wants to sleep.

I realize this is a change also to not use `unique_ptr` and to potentially use something like `std::optional` in `Sleeper` (or skip the `sleep_for` if the duration is zero).

This way the sleeper owns all the logic about when to sleep and how much etc. E.g. a change could conceivably be made to alter the sleep by coordinating more closely with the Orchestrator or something.

src/gennylib/test/PhaseLoop_test.cpp
```diff
@@ -377,4 +378,29 @@ TEST_CASE("Actual Actor Example") {
...
+                Key: 71
+        )");
+
+        auto imvProducer = std::make_shared<CounterProducer<IncrementsMapValues>>("Inc");
+        ActorHelper ah(config, 1, {{"Inc", imvProducer}});
+
+        auto start = std::chrono::high_resolution_clock::now();
+        ah.run();
+
+        auto duration = std::chrono::high_resolution_clock::now() - start;
```
rtimmons: 1. you've done the work to put the test clock into a new header is it worth using that to do the sleep? Like incorporate a `sleep` method into the clock?
2. depending on ↑, consider templating `Sleeper` on the clock similar to what the metrics types do. I.e. rename this to `SleeperT` and have a `using Sleeper = SleeperT<default-clock-or-whatever-foo>` line.

src/gennylib/test/PhaseLoop_test.cpp
guoyr: 3 would be covered by every other test :p . I replaced it with one that has both sleep and rate, which should error.

PR #177, opened 2019-03-04
Moved around some of the descriptions for testing

vrachev:
I forgot how to run a specific testcase, and somehow missed it when reading the readme (my mind only registered the cmake way of running the test), so I wanted to emphasize that you can run the binary itself with the catch2 parameters. I think it would be more clear if both testing options were in the same section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,18 @@ Genny has self-tests using Catch2. You can run them with the following command:
 ./scripts/lamp cmake-test
 ```
 
+For more fine-tuned testing (eg. running a single test or excluding some) you
+can manually invoke the test binaries:
+
+```sh
+# Build genny first: `./scripts/lamp [...]`
+./build/src/gennylib/gennylib_test "My testcase"
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,6 +70,18 @@ Genny has self-tests using Catch2. You can run them with the following command:
...
+can manually invoke the test binaries:
+
+```sh
+# Build genny first: `./scripts/lamp [...]`
+./build/src/gennylib/gennylib_test "My testcase"
+```
+
+Read more about what parameters you can pass [here][s].
+
+[s]: https://github.com/catchorg/Catch2/blob/master/docs/command-line.md#specifying-which-tests-to-run
```
guoyr: suggest `s/master/v2.5.0` in the URL in case they add parameters. So far there's no difference between the version we're using and master.

visemet: Can we also name the link "catch2" instead of "s"? 🙂

vrachev: Done

PR #178, opened 2019-03-05
TIG-1202 Support Phase Range Syntax

vrachev:
Added `PhaseRangeSpec` to handle the `..` syntax. `PhaseRangeSpec` handles both `IntegerSpec..IntegerSpec` and `IntegerSpec` formats. The `..` syntax is inclusive and resembles Ruby's `..`. 

Note that the `decode` method for `PhaseRangeSpec` was  implemented to throw if a user inputted invalid syntax, but the correct behavior should be to return false instead and let the caller decide whether to throw or have other behavior. I've kept the current behavior to be consistent with how the other specs have been implemented, and opened up [TIG-1507](https://jira.mongodb.org/browse/TIG-1507) to fix them all at once.

### DIFF
--- a/src/gennylib/include/gennylib/conventions.hpp
+++ b/src/gennylib/include/gennylib/conventions.hpp
@@ -15,6 +15,7 @@
 #ifndef HEADER_CC9B7EF0_9FB9_4AD4_B64C_DC7AE48F72A6_INCLUDED
 #define HEADER_CC9B7EF0_9FB9_4AD4_B64C_DC7AE48F72A6_INCLUDED
 
+#include <cassert>
 #include <chrono>
 #include <cmath>
 #include <sstream>
@@ -22,6 +23,7 @@
 #include <yaml-cpp/yaml.h>
 
 #include <gennylib/InvalidConfigurationException.hpp>
+#include <gennylib/Orchestrator.hpp>
 
 namespace genny {
 
@@ -124,11 +126,90 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
     return (lhs.per == rhs.per) && (lhs.operations == rhs.operations);
 }
 
+struct PhaseRangeSpec {
+    PhaseRangeSpec() = default;
+    ~PhaseRangeSpec() = default;
+
+    PhaseRangeSpec(genny::IntegerSpec s, genny::IntegerSpec e) {
+        assert(s.value >= 0 && e.value >= 0 && s.value <= UINT_MAX && e.value <= UINT_MAX && s.value <= e.value);
+        start = static_cast<genny::PhaseNumber>(s.value);
+        end = static_cast<genny::PhaseNumber>(e.value);
+    }
+    PhaseRangeSpec(genny::IntegerSpec s) : PhaseRangeSpec(s, s) {}
+
+    // Allow construction with unsigned ints for testing
+    PhaseRangeSpec(u_int32_t s, u_int32_t e) : start{s}, end{e} {}
+    PhaseRangeSpec(u_int32_t s) : PhaseRangeSpec(s, s) {}
+
+    genny::PhaseNumber start;
+    genny::PhaseNumber end;
+};
+
 }  // namespace genny
 
 namespace YAML {
 
 using genny::decodeNodeInto;
+
+/**
+ * Convert between YAML and genny::PhaseRange
+ *
+ * The YAML syntax accepts "[genny::Integer]..[genny::Integer]".
+ * This is used to stipulate repeating a phase N number of times
+ */
+template <>
+struct convert<genny::PhaseRangeSpec> {
+    static Node encode(const genny::PhaseRangeSpec rhs) {
+        std::stringstream msg;
+        msg << rhs.start << ".." << rhs.end;
+        return Node{msg.str()};
+    }
+
+    static bool decode(const Node& node, genny::PhaseRangeSpec& rhs) {
+        if (node.IsSequence() || node.IsMap()) {
+            return false;
+        }
+        auto strRepr = node.as<std::string>();
+
+        // use '..' as delimiter.
+        const std::string delimiter = "..";
+        auto delimPos = strRepr.find(delimiter);
+
+        if (delimPos == std::string::npos) {
+            // check if user inputted an integer
+            try {
+                auto phaseNumberYaml = node.as<genny::IntegerSpec>();
+                rhs = genny::PhaseRangeSpec(phaseNumberYaml);
+                return true;
+            } catch (const genny::InvalidConfigurationException) {
+                std::stringstream msg;
+                msg << "Invalid value for genny::PhaseRangeSpec: '" << strRepr << "'."
+                    << " The correct syntax is either a single integer or two integers delimited "
+                       "by '..'";
+                throw genny::InvalidConfigurationException(msg.str());
+            }
+        }
+
+        genny::IntegerSpec start;
+        genny::IntegerSpec end;
+
+        auto startYaml = Load(strRepr.substr(0, delimPos));
+        auto endYaml = Load(strRepr.substr(delimPos + delimiter.size()));
+
+        try {
+            start = startYaml.as<genny::IntegerSpec>();
+            end = endYaml.as<genny::IntegerSpec>();
+        } catch (genny::InvalidConfigurationException) {
+            std::stringstream msg;
+            msg << "Invalid value for genny::PhaseRangeSpec: '" << strRepr << "'."
+                << " The correct syntax is two integers delimited by '..'";
+            throw genny::InvalidConfigurationException(msg.str());
+        }
+        rhs = genny::PhaseRangeSpec(start, end);
+
+        return true;
+    }
+};
 
 /**
  * Convert between YAML and genny::Rate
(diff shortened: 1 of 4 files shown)


src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +126,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
 
+struct PhaseRangeSpec {
+    PhaseRangeSpec() = default;
+    ~PhaseRangeSpec() = default;
+
+    PhaseRangeSpec(genny::IntegerSpec s, genny::IntegerSpec e) {
+        assert(s.value >= 0 && e.value >= 0 && s.value <= UINT_MAX && e.value <= UINT_MAX && s.value <= e.value);
+        start = static_cast<genny::PhaseNumber>(s.value);
+        end = static_cast<genny::PhaseNumber>(e.value);
+    }
```
rtimmons: you can make the assignment into initialization:

```c++
PhaseRangeSpec(IntSpec s, IntSpec e)
: start{static_cast<PhaseNumber>(s)}, end{...} {
  assert(...)
}
```

Moreover I think you want to throw an `InvalidConfigurationException` rather than `assert` since `assert`s will get compiled away.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +126,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
+
+    static bool decode(const Node& node, genny::PhaseRangeSpec& rhs) {
+        if (node.IsSequence() || node.IsMap()) {
+            return false;
+        }
+        auto strRepr = node.as<std::string>();
+
+        // use '..' as delimiter.
+        const std::string delimiter = "..";
```
rtimmons: `static` as well so it can be allocated in `.data`.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +126,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
+        auto delimPos = strRepr.find(delimiter);
+
+        if (delimPos == std::string::npos) {
+            // check if user inputted an integer
+            try {
+                auto phaseNumberYaml = node.as<genny::IntegerSpec>();
+                rhs = genny::PhaseRangeSpec(phaseNumberYaml);
+                return true;
+            } catch (const genny::InvalidConfigurationException) {
```
rtimmons: catch by const-ref

```
catch(const InvalidExcept& x)
```

this avoid splicing the exception

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +126,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
+        genny::IntegerSpec start;
+        genny::IntegerSpec end;
+
+        auto startYaml = Load(strRepr.substr(0, delimPos));
+        auto endYaml = Load(strRepr.substr(delimPos + delimiter.size()));
+
+        try {
+            start = startYaml.as<genny::IntegerSpec>();
+            end = endYaml.as<genny::IntegerSpec>();
+        } catch (genny::InvalidConfigurationException) {
```
rtimmons: same here - const ref

src/gennylib/src/context.cpp
```diff
@@ -124,16 +123,18 @@ std::unordered_map<PhaseNumber, std::unique_ptr<PhaseContext>> ActorContext::con
...
+            phase["Phase"].as<PhaseRangeSpec>(PhaseRangeSpec{IntegerSpec{totalIndex}});
+        for (int rangeIndex = configuredRange.start; rangeIndex <= configuredRange.end; rangeIndex++) {
+            auto [it, success] =
+                out.try_emplace(rangeIndex, std::make_unique<PhaseContext>(phase, *actorContext));
+            if (!success) {
+                std::stringstream msg;
+                msg << "Duplicate phase " << rangeIndex;
+                throw InvalidConfigurationException(msg.str());
+            }
+            totalIndex++;
```
rtimmons: suggestion - rename `totalIndex` to `lastPhaseNumber` since that's what it's tracking

src/gennylib/test/conventions_test.cpp
```diff
@@ -124,5 +124,42 @@ TEST_CASE("genny::RateSpec conversions") {
...
+    }
+
+    SECTION("Barfs on invalid values") {
+        REQUIRE_THROWS(YAML::Load("0....20").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("0.1").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("0abc..20").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("20..25abc").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("12abc").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("{foo}").as<PhaseRangeSpec>());
+        REQUIRE_THROWS(YAML::Load("").as<PhaseRangeSpec>());
```
rtimmons: how do negative numbers behave? ideally they'd cause an exception

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +127,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
+                       "by '..'";
+                throw genny::InvalidConfigurationException(msg.str());
+            }
+        }
+
+        genny::IntegerSpec start;
+        genny::IntegerSpec end;
+
+        auto startYaml = Load(strRepr.substr(0, delimPos));
+        auto endYaml = Load(strRepr.substr(delimPos + delimiter.size()));
```
rtimmons: it would be nice to be able to space-separate them e.g.

```yaml
Phase: 1 .. 5
```

I think if you do [`atoi`](http://www.cplusplus.com/reference/cstdlib/atoi/) to both sides of the `..` it'll correctly strip leading/trailing spaces. But it may also mask invalid input since I think `atoi("123abc") == 123`. Your call if you think this is worth it or not. Either way please add a test case for what happens if you specify `Phase: 1 .. 3`.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -124,12 +127,91 @@ inline bool operator==(const RateSpec& lhs, const RateSpec& rhs) {
...
+
+    PhaseRangeSpec(genny::IntegerSpec s, genny::IntegerSpec e) {
+        assert(s.value >= 0 && e.value >= 0 && s.value <= UINT_MAX && e.value <= UINT_MAX && s.value <= e.value);
+        start = static_cast<genny::PhaseNumber>(s.value);
+        end = static_cast<genny::PhaseNumber>(e.value);
+    }
+    PhaseRangeSpec(genny::IntegerSpec s) : PhaseRangeSpec(s, s) {}
+
+    // Allow construction with unsigned ints for testing
+    PhaseRangeSpec(u_int32_t s, u_int32_t e) : start{s}, end{e} {}
```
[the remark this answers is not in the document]
guoyr: Or, even better, you can remove these two for-testing constructors altogether and pass in `int64_t` for testing. C++ can do one standard conversion and one user-defined conversion to turn a `int64_t` into a `IntegerSpec`. This way you get to test the behavior of negative numbers.

You'll also need to remove the `explicit` line from `IntegerSpec`'s constructor around line 47.

src/gennylib/include/gennylib/conventions.hpp
vrachev: Done

src/gennylib/include/gennylib/conventions.hpp
vrachev: I made it a constexpr.

src/gennylib/include/gennylib/conventions.hpp
vrachev: done

src/gennylib/include/gennylib/conventions.hpp
vrachev: done

src/gennylib/test/conventions_test.cpp
vrachev: Added some tests for this

src/gennylib/include/gennylib/conventions.hpp
vrachev: Added a test for this. Parsing the two sides of the `..` into an `IntegerSpec` handles this for free which is nice.

src/gennylib/include/gennylib/conventions.hpp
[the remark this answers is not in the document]
vrachev: Done

vrachev:
Ready for another round.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -15,13 +15,16 @@
 #ifndef HEADER_CC9B7EF0_9FB9_4AD4_B64C_DC7AE48F72A6_INCLUDED
 #define HEADER_CC9B7EF0_9FB9_4AD4_B64C_DC7AE48F72A6_INCLUDED
 
+#include <cassert>
```
guoyr: 🗑

PR #180, opened 2019-03-11
nomerge: TIG-1209: Make ValueGenerators typed internally

rtimmons:
Wasn't so bad to pass type information through the various `parse` statics.

This adds a new `UniqueTypedExpression<T>` type and uses it internally wherever `Expression`s want recursive/sub-expressions.

Assuming okay with this approach I'll 

1. hide all of the current implementation(s) behind a `v1` namespace
2. expose various `UniqueTypedExpression<T>` aliases/wrappers so calling code never has to see `UniqueExpression` and will instead only see things like `IntGenerator`, `DocGenerator`, etc. These will have the various `operator T()` conversions on them.
3. See if we still need `UniqueExpression` or if we can get away with just `UniqueTypedExpression`.
4. add more testing for all of the above

### DIFF
--- a/src/value_generators/test/expression_test.cpp
+++ b/src/value_generators/test/expression_test.cpp
@@ -238,6 +238,13 @@ TEST_CASE("Expression parsing with ConstantExpression::parse") {
     genny::DefaultRandom rng{};
     rng.seed(269849313357703264LL);
 
+    SECTION("recursive application caught at setup time") {
+        auto yaml = YAML::Load(R"(
+{^RandomInt: {min: [7], max: 100}}
+        )");
+        REQUIRE_THROWS(Expression::parseExpression(yaml));
+    }
+
     SECTION("valid syntax") {
         auto yaml = YAML::Load(R"(
 1
--- a/src/value_generators/include/value_generators/value_generators.hpp
+++ b/src/value_generators/include/value_generators/value_generators.hpp
@@ -134,6 +134,15 @@ class Expression;
 using UniqueExpression = std::unique_ptr<Expression>;
 
 
+enum class ValueType {
+    Integer,
+    String,
+    Array,
+    Boolean,
+    Document,
+    Null,
+};
+
 /**
  * Base class for generating values.
  *
@@ -173,13 +182,6 @@ public:
 
     virtual ~Expression() = default;
 
-    enum class ValueType {
-        Integer,
-        String,
-        Array,
-        Document,
-    };
-
     virtual ValueType valueType() const = 0;
 
     /**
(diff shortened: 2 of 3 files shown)


guoyr:
This looks pretty elegant and intuitive. Okay'd!

Separately, I'm making a note here of additional changes we may make to value generators after Max and I chatted this afternoon about composable workloads.

Max's proposal in [WRITING-3681](https://jira.mongodb.org/browse/WRITING-3681) of explicitly specifying which fields can be overridden in parent workloads using the value generator syntax means we may be adding these new features to value generation:
1. Support either a value generator or literal for any field.
2. Support variable substitution with a new `^Variable` key.

Implementation-wise, I think we'd want to blur the line between YAML and Value generation while moving some existing yaml functionality to value generators. Obviously the details are still to be hashed out, but we'll likely do something along the following over time:
1. Merge `^Verbatim` with all non-value-generator YAML values. Maybe removing `YAML.as<>` in favor of `value_generators::Expression::parseOperand`.
2. Generalize the phase/actor/workload context chain to allow unlimited nesting.
3. Allow expressions to be evaluated based on the context chain. One way is to store a context pointer in each expression and use RNGs anything else dangled from the context in `Expression::evaluate()`

No action needed right now and the above implementation may not be something we do in the short term.

src/value_generators/test/expression_test.cpp
```diff
@@ -238,6 +238,13 @@ TEST_CASE("Expression parsing with ConstantExpression::parse") {
     genny::DefaultRandom rng{};
     rng.seed(269849313357703264LL);
 
+    SECTION("recursive application caught at setup time") {
```
visemet: Is this so much of being recursive as just being the wrong type?

rtimmons: This is proving that the type-checking is done recursively at ctor time not at runtime. This didn't used to fail until `evaluate()` was called.

rtimmons:
Thanks for initial thoughtful feedback @guoyr and @visemet. I'm going to close this while I continue to work.

> Merge `^Verbatim` with all non-value-generator YAML values. Maybe removing `YAML.as<>` in favor of `value_generators::Expression::parseOperand`.

Definitely want to blur the line a bit between value-generators and `ConfigNode`. I'd like to completely hide the fact that we're using `YAML::Node` and instead just expose `ConfigNode`s everywhere. I think that gives us a nice way to hook into value-generators.

> Generalize the phase/actor/workload context chain to allow unlimited nesting.

I think so? There's still utility imho in having e.g. `ActorContext` distinct from `PhaseContext` etc. But related to prior point, if we change the default `T` in `ConfigNode::get<T,boolean>()` to be `ConfigNode` (and if we specialize to prohibit `get<YAML::Node>`) then we allow unlimited nesting and make it possible to pass things from the ActorContext etc.

> I was under the impression that templating would also mean we'd get rid of the Value class variant type.

The `Value` type may go away not sure yet. It definitely won't be exposed outside of the `v1` namespace.

> I think the `evaluate()` method of `template <ValueType T> TypedExpression` should therefore return a T.

Ya `evaluate()` here will go away and will return the generated type. I think this will require using a type-trait rather than an `enum ValueType`. The trait will indicate the return type and a conversion from `Value` to the return-type.

> I think getting this type able to be used externally (i.e. in actor code) is going to be most valuable.

Sorry if the intention / vision wasn't super clear here. Ya using it internally is just step one. Next step will be to define the appropriate type-aliases and make the `evaluate()` API a bit simpler. Then I'll replace all `UniqueExpression` instances and uses in the code. *Then* it'll be integrating more closely with `ConfigNode` to make creating them simpler.

PR #183, opened 2019-03-15
TIG-1209: Add Simple YAML to BSON Conversion in testlib

rtimmons:
Tests want to write bson but it's painful with the builders. This cribs Max's work on `Value` to build a simple utility to convert yaml to bson.

There's already `bsoncxx::from_json` so tests could write bson in the form of json and some should probably do just that, but I've got an idea to put a whole suite of tests for overhauled value-generators into a yaml file and I want to be able to write the `expected` case as yaml not a stringified json or something....

### DIFF
--- a/src/testlib/include/testlib/yamlToBson.hpp
+++ b/src/testlib/include/testlib/yamlToBson.hpp
@@ -1,0 +1,40 @@
+// Copyright 2019-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef HEADER_765C17F3_36B6_4EDC_BA8C_61DD618CCA80_INCLUDED
+#define HEADER_765C17F3_36B6_4EDC_BA8C_61DD618CCA80_INCLUDED
+
+#include <exception>
+
+#include <bsoncxx/document/view_or_value.hpp>
+#include <bsoncxx/array/view_or_value.hpp>
+#include <bsoncxx/types.hpp>
+
+#include <yaml-cpp/yaml.h>
+
+namespace genny::testing {
+
+class InvalidYAMLToBsonException : public std::invalid_argument {
+    using std::invalid_argument::invalid_argument;
+};
+
+bsoncxx::document::view_or_value toDocumentBson(const YAML::Node& node);
+
+bsoncxx::array::view_or_value toArrayBson(const YAML::Node& node);
+
+}
+
+
+
+#endif  // HEADER_765C17F3_36B6_4EDC_BA8C_61DD618CCA80_INCLUDED
--- a/src/testlib/include/testlib/yamlToBson.hpp
+++ b/src/testlib/include/testlib/yamlToBson.hpp
@@ -17,8 +17,8 @@
 
 #include <exception>
 
-#include <bsoncxx/document/view_or_value.hpp>
 #include <bsoncxx/array/view_or_value.hpp>
+#include <bsoncxx/document/view_or_value.hpp>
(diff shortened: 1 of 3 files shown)


src/testlib/include/testlib/yamlToBson.hpp
```diff
@@ -0,0 +1,49 @@
...
+class InvalidYAMLToBsonException : public std::invalid_argument {
+    using std::invalid_argument::invalid_argument;
+};
+
+/**
+ * @param node yaml map node
+ * @return bson representation of it.
+ * @throws InvalidYAMLToBsonException if node isn't a map
+ */
+bsoncxx::document::view_or_value toDocumentBson(const YAML::Node& node);
```
visemet: This should just be `bsoncxx::document::value` because `YAML::Node` cannot already hold a `bsoncxx::document::value` where a copy could be avoided.

Ditto with `bsoncxx::array::value` and `toArrayBson()`.
Ditto with the `std::variant<>` type.
Ditto with the `bson()` and `fromYaml()` function in `yamlToBson_test.cpp`.

src/testlib/test/yamlToBson_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include <iostream>
+#include <testlib/helpers.hpp>
+#include <testlib/yamlToBson.hpp>
+
+namespace genny::testing {
+
+using Catch::Matchers::Matches;
+
+namespace {
```
visemet: C++ unit tests should be entirely put into an anonymous namespace because there's nothing defined within them that requires external linkage.

src/testlib/test/yamlToBson_test.cpp
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+
+#include <ios>
+#include <string>
+
+#include <yaml-cpp/yaml.h>
+
+#include <bsoncxx/json.hpp>
+
+#include <iostream>
```
visemet: [style] Move this include immediately after line 15 to keep the built-ins grouped together.

Separately, do you need both it and `#include <ios>`?

src/testlib/src/yamlToBson.cpp
```diff
@@ -0,0 +1,179 @@
...
+                                     bsoncxx::document::view_or_value,
+                                     bsoncxx::array::view_or_value>;
+    VariantType _value;
+};
+
+
+std::string to_string(const YAML::Node& node) {
+    YAML::Emitter e;
+    e << node;
+    return std::string{e.c_str()};
```
visemet: [question] Does YAML not allow embedding null bytes?

visemet: Nevermind, this is just part of yaml-cpp's API.

src/testlib/test/yamlToBson_test.cpp
rtimmons: CLion snuck this in - removed.

src/testlib/test/yamlToBson_test.cpp
rtimmons: The tests themselves aren't in the anonymous namespace only helper functions are.

rtimmons:
Thanks for review - changed to `bson...:value` and killed some cruft. Assuming CI tests come back clear I'll merge in a bit unless objections.

src/testlib/test/yamlToBson_test.cpp
rtimmons: I mis-read your comment - good call - moved the tests themselves to the anon namespace as well.

visemet: Gotta save those microseconds on link times!

src/testlib/src/yamlToBson.cpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+private:
+    using VariantType = std::variant<bool,
+                                     int32_t,
+                                     int64_t,
+                                     double,
+                                     std::string,
+                                     bsoncxx::types::b_null,
+                                     bsoncxx::document::view_or_value,
```
visemet: > Ditto with the `std::variant<>` type.
> https://github.com/mongodb/genny/pull/183#discussion_r266061112

I'd say just have this line `bsoncxx::document::value` and the next line `bsoncxx::array::value`.

rtimmons: ohai the the `bsoncxx*` types didn't actually need to be a part of the `variant` so just removed them entirely 👍

PR #184, opened 2019-03-18
TIG-1209: Add findRepoRoot to assist with loading test fixtures

rtimmons:
I'm writing a test that wants to load up a yaml file but the cwd of the test binary isn't really well-defined. This lets us look for a special ".genny-root" file by checking `.` then `./..` and then `./../..` etc.

I considered looking for `.git` but that seems a bit cheeky and I think could break if we ever add submodules or something.

### DIFF
--- a/src/testlib/src/findRepoRoot.cpp
+++ b/src/testlib/src/findRepoRoot.cpp
@@ -1,0 +1,53 @@
+// Copyright 2019-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


src/testlib/src/findRepoRoot.cpp
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <testlib/findRepoRoot.hpp>
+
+#include <boost/filesystem.hpp>
```
guoyr: Add `Boost::filesystem` to CMakeLists.txt?

PR #185, opened 2019-03-18
TIG-1209: Replace expression_test.cpp with a yaml-based set of tests

rtimmons:
`Expression` is getting the axe so moving its tests to an "api-neutral" (or "public API only") form. This will let the internals of `DocumentGenerator` change without having to change its tests.


src/testlib/test/yamlToBson_test.cpp
```diff
@@ -51,8 +52,29 @@ void test(const std::string& yaml, const std::string& json) {
...
+
+
+namespace BasicBson = bsoncxx::builder::basic;
+namespace BsonTypes = bsoncxx::types;
 
 TEST_CASE("YAML To BSON Simple") {
+    // use the smallest type that will fit
+    test("foo: 0", BasicBson::make_document(BasicBson::kvp("foo", BsonTypes::b_int32{0})));
+
+    // max valeu for int32 ↓
```
guoyr: 💅

PR #186, opened 2019-03-20
PERF-1816: Increase run time for Insert remove test

dalyd:
Need to open a ticket, but here's the change. I changed the log message to debug while here. 
Sys-perf patch here: https://evergreen.mongodb.com/version/5c9241775623433487e7d6bd

I assume it will break history. That's fine.

### DIFF
--- a/src/cast_core/src/InsertRemove.cpp
+++ b/src/cast_core/src/InsertRemove.cpp
@@ -69,7 +69,7 @@ struct InsertRemove::PhaseConfig {
 void InsertRemove::run() {
     for (auto&& config : _loop) {
         for (auto&& _ : config) {
-            BOOST_LOG_TRIVIAL(info) << " Inserting and then removing";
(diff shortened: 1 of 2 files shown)


dalyd:
Opened PERF-1816 for this.

dalyd:
Any idea why the patch build is failing? I've restarted it a few times. Passes locally for me. The first one look suspect to me, but the remaining ones looked like an evergreen problem.

visemet:
> Any idea why the patch build is failing? I've restarted it a few times. Passes locally for me. The first one look suspect to me, but the remaining ones looked like an evergreen problem.

@dalyd, are you restarting the whole build? [Those tasks are all part of a single-host task group](https://github.com/mongodb/genny/blob/50ef8adbfe77a88ca5005a2cefe76e91ea694612/evergreen.yml#L209-L220).

PR #187, opened 2019-03-21
PERF-1748 Add 2 CommitLatency workloads

henrikingo:


henrikingo:
I didn't do the unit tests yet, but this works so would like to get a review now. (Will also add a little bit of logging.)

One remaining issue: Currently genny outputs the statistics for all phases exactly the same. So they are lumped together under the same id. I'd like to get statistics separately for each phase. How can I do that?

The other issue is that both for transactions and sessions I need access to _client, but current layout of genny doesn't provide for that. Plan is to commit this first and deal with that in a follow-up. Note that both the added yml files work and provide value as it is, and this is the motivation to get them committed and into use.

henrikingo:
@rtimmons Could you tag your favorite C++ driver engineer as second reviewer? I know you've worked with them.

rtimmons:
Sorry I ran out of time to take a look here today - I'll do this first thing tomorrow morning eastern-time. At a high-level I'm curious what part of this could/"should" be implemented in CrudActor.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,195 @@
...
+    bool useSession;
+    int64_t repeat;
+    int64_t threads;
+    std::uniform_int_distribution<int64_t> amountDistribution;
+    ExecutionStrategy::RunOptions options;
+
+    PhaseConfig(PhaseContext& phaseContext, const mongocxx::database& db)
+        : collection{db[phaseContext.get<std::string>("Collection")]},
+          rp_string{phaseContext.get<std::string>("ReadPreference")},
+          useSession{phaseContext.get<bool, false>("Session")},
```
rtimmons: This syntax may not be doing what you think and I had to stare at it for a bit myself - the `get<T,false>` form returns a `std::optional`. This is effectively casting the `std::optional` to a bool by using an implicit-conversion. The `std::optional<boolean>` will itself evaluate to `true` _iff_ it's populated (if the user specified a value for the key). So if the user specifies `Session: false` in the YAML the `useSession` variable will be true. I think what you want here is

```c++
useSession{phaseContext.get<bool, false>("Session").value_or(false)}
```

Similar applies to elswhere here where this does `.get<bool,false>`

I see you did the `value_or` thing for a few others so this could be intentional (in which case may be worthy of a comment).

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,195 @@
...
+          repeat{phaseContext.get<IntegerSpec>("Repeat")},
+          threads{phaseContext.get<IntegerSpec>("Threads")},
+          amountDistribution{-100, 100},
+          options{ExecutionStrategy::getOptionsFrom(phaseContext, "ExecutionsStrategy")} {
+              // write_concern
+              if( phaseContext.get<bool, false>("WriteConcernMajority") ) {
```
rtimmons: do you want to allow `WriteConcernMajority` to specify the value in `milliseconds` (via a `TimeSpec`)?

```c++
if(auto maj = phaseContext.get<TimeSpec, false>("WriteConcernMajority") ) {
  wc.majorty(maj);
}
```

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,195 @@
...
+              }
+              optionsUpdate.write_concern(wc);
+
+              // read_concern
+              rc.acknowledge_string(phaseContext.get<std::string>("ReadConcern"));
+              // Ugh... read_concern cannot be set for operations individually, only through client, database, or collection.
+              // And out of those, only client is valid inside transactions. (DRIVERS-619)
+              // It seems mongocxx::database doesn't provide a getter to get the client, nor is the client available in this context.
+              // Ergo, transactions cannot be supported...
+              // TODO: Should set _client. (And options::find etc should support read_concern!)
```
rtimmons: Similar to how you pass in the `mongocxx::database` you could pass in the `mongocxx::client` as well?

src/cast_core/test/CommitLatency_test.cpp
```diff
@@ -0,0 +1,73 @@
...
+          ExecutionStrategy:
+            ThrowOnFailure: true
+          Phases:
+          - Repeat: 100
+            Collection: mycoll
+            Document: {foo: {^RandomInt: {min: 0, max: 100}}}
+    )");
+
+
+    SECTION("Inserts documents into the database.") {
```
rtimmons: I think you can just delete this test file unless you plan to fill it in later (which of course would be a <i>good idea</i>™️ but it's your call)

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,195 @@
...
+                  // TODO: This actually sends primaryPreferred to server. -> File bug in Jira.
+                  // This is benign for this test, but potentially bad for real apps!
+                  // Note that k_secondary works correctly.
+                  rp.mode(mongocxx::read_preference::read_mode::k_primary);
+              }
+              else if ( rp_string == "SECONDARY" ) {
+                  rp.mode(mongocxx::read_preference::read_mode::k_secondary);
+              }
+              else {
+                  throw InvalidConfigurationException("ReadPreference must be PRIMARY or SECONDARY.");
```
rtimmons: it's not well-documented or linted yet, but we've found it's better to write this as

```c++
#include <boost/throw_exception.hpp>
....
BOOST_THROW_EXCEPTION(InvalidConfigurationException("ReadPreference must be PRIMARY or SECONDARY."));
```

since then you get line-numbers in exceptions.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,195 @@
...
+                    // Check for isolation or consistency errors
+                    // Note: If there's skew in the sum total, we never fix it. So after the first
+                    // error, this counter is likely to increment on every loop after that. The
+                    // exact value of the metric is therefore uninteresting, it's only significant
+                    // whether it is zero or non-zero.
+                    for (auto doc : cursor) {
+                        if (doc["total"].get_int64() != 200) {
+                            ctx.addErrors(1);
+                        }
+                    }
```
rtimmons: you may also want to `ctx.addDocs(1)` for each doc and `ctx.addBytes(doc.size())`. These can serve as a sort of canary that the behavior of the test (number of bytes/docs) hasn't changed dramatically from run-over-run.

src/workloads/networking/CommitLatencySingleUpdate.yml
```diff
@@ -0,0 +1,60 @@
+---
+# Single threaded updates to measure commit latency for various write concern.
```
rtimmons: We don't have strong guidelines, but most of the other product-perf workloads are in `workloads/scale` and platforms-team has been in `workloads/networking`. I can make a case for `networking` here though - as long as David's in agreement with where it belongs it's fine.

henrikingo:
> You could make the metrics::Operation prefixed with the phase number

I don't think the nr is very informative. I would like to either provide a PhaseName in the yaml, or generate a descriptive name in PhaseConfig.

If this is not possible now, then a workaround for this seems to be to just have 1 phase per yaml file, and a different yaml file for each permutation of settings. That can easily be supplied from a test_control.yml file. I can then use a different ActorName for each small file.

> You may be able to model this workload by using the CrudActor but I actually like how explicit and direct the logic is here, and I think would be messier in YAML, so it's a matter of personal taste.

CrudActor didn't exist when I started this work, but I can learn from it now to see how I should create a session and transaction.

That said, I suspect CrudActor may be missing a couple things I need: It's a bit trivial, but this is supposed to be a real transaction, so there's actually a variable `amount` used across several operations. Then there's also an if clause at the end.

And then, as you say, this would become unreadable in CrudActor yaml. This is another case where some kind of sub routine (like in Alex' LargeScale workload discussion) would be necessary to keep the yaml reasonably clean. If it had been possible, 

> We haven't been bringing in driver folks for questions about using the driver

My specific reason in this case is indeed to audit the correct usage of the concerns and preferences. They aren't very consistent with each other, documentation was light (but enough to get this done) and no tutorials available on the internet. Also, across the drivers in general there are obscure surprises like [DRIVERS-619](https://jira.mongodb.org/browse/DRIVERS-619), so we definitively want the driver author to verify that this actually tests what it looks like on the outside. (I of course can verify quite a lot from mongod.log side, but would still like driver author to share some responsibility.)

But I'll take another stab at transactions and sessions first, before bringing in outside reviewer.

> Similar to how you pass in the mongocxx::database you could pass in the mongocxx::client as well?

Cool. I'll try that and if it works I add to this PR. I wasn't sure if I can / allowed to do that, so wanted to ask first.

src/cast_core/src/CommitLatency.cpp
henrikingo: I understood the functionality of std::optional such that if my default is false, then the `value_or(false)` is redundant. But I can add them for clarity if that's our style. Clarity is always nice.

src/cast_core/src/CommitLatency.cpp
henrikingo: Not really. I think a boolean is more intuitive and matches other options that are also boolean. But thanks for the example.

src/cast_core/src/CommitLatency.cpp
henrikingo: Cool. Will try that. I wasn't sure I can / allowed to change this interface.

src/cast_core/src/CommitLatency.cpp
henrikingo: Intended use case is to run exactly 500 repeats, so it felt more noisy than useful to keep those.

src/cast_core/test/CommitLatency_test.cpp
henrikingo: If management allows me to spend the time, I'd rather like to write some unit tests. Part of the motivation for this was also to learn genny.

src/workloads/networking/CommitLatencySingleUpdate.yml
henrikingo: InsertRemove being under scale is funny too, but in any case I can't put a single threaded test under scale. My alternatives were to create a new directory called either `transactions` or `consistency`. You can pick.

rtimmons: > I can't put a single threaded test under scale

lol..agreed...

> My alternatives were to create a new directory called either transactions or consistency. You can pick.

🤷‍♂️ no real opinion I'll let @dalyd comment there or leave in `networking` since it's not really wrong. Bikeshedding is fun.

rtimmons:
Thanks Henrik - nothing requiring further input from me - please ping when you're ready for more eyes.

I'm actually away next week (back April 1), so I'm going to remove myself and add back STM in case anyone else wants to take another look. 

> we definitively want the driver author to verify that this actually tests what it looks like on the outside.

I don't have anyone in particular from drivers who I'd bring in: Jesse is no longer on drivers but Kevin may be willing to spare the cycles. We can first reach out on `#cxx-driver` slack- they're usually pretty helpful.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+                    mongocxx::pipeline p{};
+                    bsoncxx::document::value doc_group = bsoncxx::builder::stream::document{}
+                            << "_id" << "foo"
+                            << "total" << bsoncxx::builder::stream::open_document
+                                << "$sum" << "$n"
+                            << bsoncxx::builder::stream::close_document
+                        << bsoncxx::builder::stream::finalize;
+                    p.group(doc_group.view());
+                    auto cursor = config->collection.aggregate(*_session, p, config->optionsAggregate);
+                    // Check for isolation or consistency errors
```
guoyr: [Question] If the following loop is used to check "isolation or consistency errors", did you want to wrap it with a `if (config->useSession)`? If not, suggest updating the comment.

src/workloads/networking/CommitLatency.yml
```diff
@@ -0,0 +1,208 @@
...
+
+Collection: &Collection CommitLatency
+
+Actors:
+- Name: Load phase
+  Type: RunCommand
+  Threads: 1
+  Database: test
+  Phases:
+  - Repeat: 1
```
guoyr: You can do this!
```
 - &ResetCollection
    Repeat: 1
    Operations:
    - OperationName: RunCommand
      OperationCommand:
        drop: *Collection
    - OperationName: RunCommand
      OperationCommand:
        insert: *Collection
        documents: [{_id: 1, n: 100}, {_id: 2, n: 100}]
  - &Noop
    Operation:
       OperationName: Nop
  - *ResetCollection
  - *Noop
  - *ResetCollection
  - *Noop
...
...
```

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+    bool useSession;
+    bool useTransaction;
+    int64_t repeat;
+    int64_t threads;
+    std::uniform_int_distribution<int64_t> amountDistribution;
+    ExecutionStrategy::RunOptions options;
+
+    PhaseConfig(PhaseContext& phaseContext, const mongocxx::database& db)
+        : collection{db[phaseContext.get<std::string>("Collection")]},
+          rp_string{phaseContext.get<std::string>("ReadPreference")},
```
guoyr: You can set the read concern like this after you rebase to latest master:
`rc{context.get<mongocxx::read_concern>("ReadConcern")`

src/workloads/networking/CommitLatency.yml
```diff
@@ -0,0 +1,208 @@
...
+      OperationName: Nop # Drop and reload
+  - WriteConcernMajority: true
+    ReadConcern: majority
+    ReadPreference: SECONDARY
+    Session: True
+  - Operation:
+      OperationName: Nop # Drop and reload
+  - WriteConcernMajority: true
+    ReadConcern: snapshot
+    ReadPreference: PRIMARY
```
guoyr: Would a secondary read preference +  transaction scenario be interesting?

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+          useSession{phaseContext.get<bool, false>("Session").value_or(false)},
+          useTransaction{phaseContext.get<bool, false>("Transaction").value_or(false)},
+          repeat{phaseContext.get<IntegerSpec>("Repeat")},
+          threads{phaseContext.get<IntegerSpec>("Threads")},
+          amountDistribution{-100, 100},
+          options{ExecutionStrategy::getOptionsFrom(phaseContext, "ExecutionsStrategy")} {
+              // write_concern
+              if( phaseContext.get<bool, false>("WriteConcernMajority").value_or(false) ) {
```
guoyr: I'm going to expose the write concern and read preference parsing code in Genny so you don't have to manually parse those guys. Should be ready by EOD today. If not, I'll let you know!

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+                  rp.mode(mongocxx::read_preference::read_mode::k_secondary);
+              }
+              else {
+                  BOOST_THROW_EXCEPTION(InvalidConfigurationException("ReadPreference must be PRIMARY or SECONDARY."));
+              }
+              if ( useTransaction ) {
+                  optionsTransaction.read_preference(rp);
+              }
+              else {
+                  optionsFind.read_preference(rp);
```
guoyr: Since you're only setting read and write preferences on find, update and aggregate, I think you can just set them directly on the collection. This way you don't need to store `optionsFind`, `optionsUpdate`, and `optionsAggregate`

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+                  optionsAggregate.read_preference(rp);
+              }
+
+          }
+};
+
+void CommitLatency::run() {
+    for (auto&& config : _loop) {
```
guoyr: If you add `using namespace bsoncxx::builder::stream;` to this function, all the `bsoncxx::builder::stream::` can be removed 😄

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,227 @@
...
+                        << bsoncxx::builder::stream::finalize;
+                    auto result1 = config->collection.find_one(*_session, doc_filter1.view(), config->optionsFind);
+
+                    // db.hltest.update_one( {'_id': 1}, {'$inc': {'n': -amount}}, session=session )
+                    bsoncxx::document::value doc_update1 = bsoncxx::builder::stream::document{}
+                        << "$inc" << bsoncxx::builder::stream::open_document
+                            << "n" << -amount
+                        << bsoncxx::builder::stream::close_document
+                        << bsoncxx::builder::stream::finalize;
+                    config->collection.update_one(*_session, doc_filter1.view(), doc_update1.view(),
```
guoyr: You might have to explicitly call the non-session version of `update_one` `if (!_session)`

src/workloads/networking/CommitLatencySingleUpdate.yml
```diff
@@ -0,0 +1,60 @@
...
+  - Name: SingleThreadUpdate
+    Type: RunCommand
+    Threads: 1
+    Database: test
+    Phases:
+      - Repeat: 500
+        Operations:
+        - OperationCommand:
+            update: accounts
+            updates: [ { q: { _id: 1 }, u: { $inc: { n: 1 } } } ]
```
guoyr: Should there be an insert somewhere before these updates?

henrikingo: Note: While adding this I also learned that RunCommand exits on error, such as when dropping a non-existing collection. So I have also refactored CommitLatency.yml so that drop and insert are 2 distinct phases.

src/cast_core/src/CommitLatency.cpp
guoyr: Yep. Genny has some magic to convert strings to `mongocxx::read_conern`. Here's the list of supported values FYI. Your existing YAML should just work out of the box. 

https://github.com/mongodb/genny/blob/cc460549bc482ba30c218bd88f62d4d0a5048124/src/gennylib/include/gennylib/conventions.hpp#L181

src/cast_core/src/CommitLatency.cpp
guoyr: OK I pinged you on the PR to expose parsing for write concern and read preference. The syntax should mirror the BSON command structure but with capitalized keys.

src/cast_core/src/CommitLatency.cpp
guoyr: Noted. My preference is still to add a comment here to indicate your concern since you're never setting `optionsTransaction` and on the collection at the same time.

No strong feelings though. Feel free to keep it as is.

src/cast_core/src/CommitLatency.cpp
guoyr: So there's no specific rule for *when* to use `using namespace`..

There are a couple of best practices though.
1. You want to load it at the narrowest block scope to avoid polluting unrelated functions in the same namespace and/or file.
2. If you only need one thing from a namespace frequently, use `using toJson = bsoncxx::to_json;` instead of loading the entire namespace.

src/cast_core/src/CommitLatency.cpp
guoyr: Thanks for the clarification. Maybe rename this comment to "Record isolation or consistency violations in non-transactional operations"? "Errors" sounds really ominous to me XD.

src/workloads/networking/CommitLatency.yml
guoyr: Interesting, even in 4.2! TIL.

src/cast_core/src/CommitLatency.cpp
henrikingo: Comments have been updated and reference jira tickets too.

src/cast_core/src/CommitLatency.cpp
henrikingo: I added a sentence explaining when they are expected.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,199 @@
...
+                            << "_id" << "foo"
+                            << "total" << open_document
+                                << "$sum" << "$n"
+                            << close_document
+                        << finalize;
+                    p.group(doc_group.view());
+                    auto cursor = config->collection.aggregate(*_session, p, config->optionsAggregate);
+                    // Check for isolation or consistency errors.
+                    //
+                    // These are expected when read_preference is secondary and not in a session.
```
guoyr: 👍

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,199 @@
...
+                    // result1 = db.hltest.find_one( { '_id': 1 }, session=session )
+                    bsoncxx::document::value doc_filter1 = document{}
+                        << "_id" << 1
+                        << finalize;
+                    auto result1 = config->collection.find_one(*_session, doc_filter1.view(), config->optionsFind);
+
+                    // db.hltest.update_one( {'_id': 1}, {'$inc': {'n': -amount}}, session=session )
+                    bsoncxx::document::value doc_update1 = document{}
+                        << "$inc" << open_document
+                            << "n" << -amount
```
guoyr: Please confirm that that these streams are indented correctly after clang format. (You shouldn't need to format anything manually)

henrikingo: What is "correctly"? clang-format of course isn't aware of idiomatic mongocxx style and has mangled these to single line. It's not clear to me that's what you expected?

henrikingo:
> For your discussion with Ryan on reporting operations at the phase level, did you have a preference on how you'd like to proceed?

My response to Ryan was

> I don't think the nr is very informative. I would like to either provide a PhaseName in the yaml, or generate a descriptive name in PhaseConfig.

> If this is not possible now, then a workaround for this seems to be to just have 1 phase per yaml file, and a different yaml file for each permutation of settings. That can easily be supplied from a test_control.yml file. I can then use a different ActorName for each small file.

To be clear, I haven't understood there is any option at all to achieve this at the moment. Ryan said he would file a ticket to provide something.

guoyr:
> To be clear, I haven't understood there is any option at all to achieve this at the moment. Ryan said he would file a ticket to provide something.

Yes there is no option to do this at the moment. It's easy to add support `PhaseName` in genny but not so easy to visualize the results at a per-phase level in Evergreen. I wasn't sure if you or Ryan had discussed what that ticket would contain outside of this PR. If not, that's fine. I'll just wait for Ryan to file that ticket.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,189 @@
...
+          useTransaction{phaseContext.get<bool, false>("Transaction").value_or(false)},
+          repeat{phaseContext.get<IntegerSpec>("Repeat")},
+          threads{phaseContext.get<IntegerSpec>("Threads")},
+          amountDistribution{-100, 100},
+          options{ExecutionStrategy::getOptionsFrom(phaseContext, "ExecutionsStrategy")} {
+        if (useTransaction) {
+            optionsTransaction.write_concern(wc);
+            optionsTransaction.read_preference(rp);
+            // Transactions ignore the read_concern of a collection handle, must set transaction
+            // options (DRIVERS-619).
```
kevinAlbs: FYI that's a private Jira ticket. Not sure if it matters, but if this is a public repo any outside readers will get a permissions error.

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,189 @@
...
+          useSession{phaseContext.get<bool, false>("Session").value_or(false)},
+          useTransaction{phaseContext.get<bool, false>("Transaction").value_or(false)},
+          repeat{phaseContext.get<IntegerSpec>("Repeat")},
+          threads{phaseContext.get<IntegerSpec>("Threads")},
+          amountDistribution{-100, 100},
+          options{ExecutionStrategy::getOptionsFrom(phaseContext, "ExecutionsStrategy")} {
+        if (useTransaction) {
+            optionsTransaction.write_concern(wc);
+            optionsTransaction.read_preference(rp);
+            // Transactions ignore the read_concern of a collection handle, must set transaction
```
kevinAlbs: I'm not sure if it's helpful here, but you could also inherit the read_concern/write_concern/read_preference from the session's default_transaction_opts, or failing that, the client's.

See: https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#starttransaction

src/cast_core/src/CommitLatency.cpp
```diff
@@ -0,0 +1,189 @@
...
+            // options (DRIVERS-619).
+            optionsTransaction.read_concern(rc);
+        } else {
+            optionsUpdate.write_concern(wc);
+            optionsFind.read_preference(rp);
+            optionsAggregate.read_preference(rp);
+            // Ugh... read_concern cannot be set for operations individually, only through client,
+            // database, or collection. (CXX-1748)
```
kevinAlbs: Agree, we should add this. Thanks for filing CXX-1748

src/cast_core/src/CommitLatency.cpp
henrikingo: Yes. I learned that the hard way when I first did this in pymongo. Since this test may or may not use transactions and sessions, I've tried to use only the options that are available in all permutations. I guess that wasn't fully possible anyway, but that was where I was coming from.

src/cast_core/src/CommitLatency.cpp
henrikingo: I will remove. Thanks.

rtimmons:
Sorry for my delay in getting this merged. Thanks for adding integration tests they look great.

*But* it seems the integration tests are failing because the commit latency is too high during integration tests - see [evergreen logs](https://evergreen.mongodb.com/task_log_raw/genny_amazon2_t_integration_test_sharded_patch_4de7565615f2b09a58c5a4563c2e55a171ada9e3_5ca6215f32f4177e9f50d058_19_04_04_15_23_13/0?type=T).

I've just restarted it to see if it's transient but I'm wondering if it makes sense to make the `_test.cpp` a bit more resilient to potential high latencies. I'll check back on the integration test I restarted in a bit.

henrikingo:
Well that was new. I think. I removed the [sharded] tag from the unit test. I'm unsure whether transactions work with sharded clusters yet. And actually I don't know the mongod version used in that test! But in any case I was really focused on replica sets so far.

henrikingo:
Seems to pass now :-)

rtimmons:
💥 nice - merging! Thanks for your patience in getting this across the finish-line.

PR #189, opened 2019-03-21
TIG-1578: Fix DocumentGenerators produce different values across systems

rtimmons:


src/value_generators/include/value_generators/DefaultRandom.hpp
```diff
@@ -105,7 +106,12 @@ class Random {
 /**
  * DefaultRandom should be used if you need a random number generator.
  */
-using DefaultRandom = v1::Random<std::mt19937_64>;
+// Note he use boost::random because its distributions are
+// cross-platform. The downside to this is that some <algorithm>
+// functions e.g. std::shuffle() don't work with DefaultRandom
+// because Boost's RNGs don't have constexpr min() and max()
+// functions - https://svn.boost.org/trac10/ticket/12251
```
guoyr: We can take the PR for that ticket and apply it to our toolchain boost 😃 

Just have to download the PR and add `PATCHES constexpr-min-max.patch` to [here](https://github.com/10gen/vcpkg/blob/242b616fd730882aadd4b54ed05121fa266dcb3f/ports/boost-random/portfile.cmake#L11) if the PR applies cleanly.

rtimmons: lol nice - I have [a PR](https://github.com/boostorg/random/pull/24/files) open to the boost.random maintainers. Even if that's merged successfully it'll probably be a while before it's available in the next boost release so we may end up using that option as a stopgap.

guoyr: Nice! I'm glad they responded so quickly. It looks like boost branches for a release 7 weeks prior, so yeah it's likely still worth using a vcpkg patch if we decide to use boost random.

rtimmons:
This is ready for a quick look if you have the chance @guoyr

PR #190, opened 2019-03-21
TIG-1546 Add ReadConcern, Limit, Sort to the MultiCollectionQuery Actor

guoyr:


src/cast_core/src/MultiCollectionQuery.cpp
```diff
@@ -41,8 +42,19 @@ struct MultiCollectionQuery::PhaseConfig {
         : rng{rng},
           database{(*client)[context.get<std::string>("Database")]},
           numCollections{context.get<IntegerSpec, true>("CollectionCount")},
+          readConcern{context.get<mongocxx::read_concern, false>("ReadConcern")},
           filterExpr{DocumentGenerator::create(context.get("Filter"), rng)},
-          uniformDistribution{0, numCollections} {}
+          uniformDistribution{0, numCollections} {
+        const auto limit = context.get<int64_t, false>("Limit");
+        if (limit) {
+            options.limit(limit.value());
```
rtimmons: `.value()` is awkward - I think it wants you to use `*`:

```c++
       if (limit) {
            options.limit(*limit);
```

PR #191, opened 2019-03-22
TIG-1209: Refactor Value-Generators Internals

rtimmons:
I'm not super happy with how this turned out.

- Templating the `ValueGenerator` type turned awkward with the document/array container types. 
- But overall I feel it's 2 steps forward 1 step back so I guess you could say this is a positive change.
- But that's subjective.
- So I'm not opposed to totally scrapping this.

Pros of this change:

- It's more type-safe internally
- It doesn't use `std::variant` which--while appropriate for the problem in many ways--is kinda awkward in practice
- It's fully PImpl'd so changes within the impl don't require recompilation

Cons of this change:

- it's a lot of discrete types that are kinda hard to piece together mentally
- similarly, the free functions e.g. `int64OperandBasedOnDistribution` are kinda annoying and were hard to group together in a succinct way
- The `Verbatim`-as-a-template feels a bit too clever but it also worked quite nicely

I tried for like 4 hours to split this into nice and tidy header files and multiple cpp files, but the linker really hated the idea of `parserMap` so I gave up except for `Appendable.hpp`...

This depends on #185 which is blocked pending some back-and-forth on random-numbers. Assume that PR is already merged for this so consider this PR based off of master.

I'm out next week, back 4/1. I'll address any feedback then. This change should be transparent to Actors so no pressing need for this.


visemet:
> For `verbatimOperand`, I have a crazy thought. What do you think about removing `^Verbatim` altogether? I don't see it being used anywhere at the moment and I don't foresee any workload desperately needing to have a hat-prefixed key.

I tried out this crazy thought once before with the old "$useval" generator. The result was David Daly emailing me to adding it back just in case.

guoyr:
> > For `verbatimOperand`, I have a crazy thought. What do you think about removing `^Verbatim` altogether? I don't see it being used anywhere at the moment and I don't foresee any workload desperately needing to have a hat-prefixed key.
> 
> I tried out this crazy thought once before with the old "$useval" generator. The result was David Daly emailing me to adding it back just in case.

Good to know! I'm going to try my luck and see if David will be more agreeable this time around after your `s/$/^` change.

guoyr:
@dalyd One more syntax debate for ya. Would you be open to removing support for `^` prefixed keys in Genny? There's a bit of magic that has to happen to parse `^Verbatim` and it seems less mission-critical than $-prefixed keys.

rtimmons:
Thanks for review. I'm going to leave `^Verbatim` in here for now since it works and the code would only lose about 100 lines with it gone.

I did a pass through everything with fresh eyes after being away for a week. Didn't change that much but I removed some cruft and changed some ordering to hopefully make things clearer.

> There are two types of "discrete types": generators and operands...

Ya this is a good way to think about it. I did a bit of renaming and shuffling to make this a bit more explicit in the code.

> For operands, maybe fastRandomStringOperand, randomStringOperand, and verbatimOperand can be replaced with a genericOperand that just maps to a templated generator?

I just tried this, and it ended up worse because we don't know if we're in the top-level context `{^RandomString:{length:10}}` or if we're in the inner context `{length:10}`.

However I was able to eliminate some of the mess and confusion here by making the `fastRandomStringOperand` etc functions into lambda-functions which much more clearly express the intent. The only exception is the int-generator versus int-based-on-operand generator.

I also did end up creating `template<T> class Generator` which removed about 50 lines of boilerplate without sacrificing too much clarity imho.

I'm still not 100% convinced this is worth merging but I feel better about it now than I did before (fwiw).

dalyd:
Thanks for preserving the ^Verbatim. That was added for me. Max added that one for me. I can't sweat it will ever be used, but it is comforting to me to know I can represent *any* document. If the ^Verbatim ends up being complicated, that's a good sign of something needing clean up.

PR #193, opened 2019-03-25
TIG-1555 move read_preference and write_concern parsing to public header

guoyr:


src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -164,6 +166,116 @@ namespace YAML {
...
+            node["ReadMode"] = "nearest";
+        }
+        auto maxStaleness = rhs.max_staleness();
+        if (maxStaleness) {
+            node["MaxStaleness"] = genny::TimeSpec(*maxStaleness);
+        }
+        return node;
+    }
+
+    static bool decode(const Node& node, ReadPreference& rhs) {
```
guoyr: @henrikingo Here's the code in genny to natively parse `mongocxx::read_preference` and `mongocxx::write_concern`. For read preference, it accepts the "ReadMode" and "MaxStaleness" top level keys. "ReadMode" accepts a number of strings as defined on lines 203 to 212.

PR #194, opened 2019-03-28
TIG-1543 support composable phase configs

guoyr:


guoyr:
I've been thinking about the phase config syntax and the fact that workloads can reuse phase configs but phase configs can't reuse each other keeps nagging me. I feel like there's a risk where we'll end up pushing the duplication down to the phase configs.

I'm going to move this PR back into draft for now and add support for setting top-level keys to be loaded from phase configs.

guoyr:
OK added support for parsing top-level keys. Re-requesting review.

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+    auto replacement = loadConfig(path.string());
+
+    if (external["Parameters"]) {
+        nodeSize++;
+        auto newParams = external["Parameters"].as<YamlParameters>();
+        params.insert(newParams.begin(), newParams.end());
+    }
+
+    if (external["Key"]) {
+        nodeSize++;
```
rtimmons: consider `keysSeen` or something as var name - `nodeSize` confused me on first-read.

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+        const auto key = external["Key"].as<std::string>();
+        if (!replacement[key]) {
+            auto os = std::ostringstream();
```
rtimmons: I've usually used `std::stringstream` since I think it's more universal but it looks like `std::ostringstream` may be slightly faster... No change required but I wonder if worth indicating the preference for `ostringstream` in the styleguide.

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+
+YAML::Node recursiveParse(const YAML::Node& node,
+                          YamlParameters& params,
+                          const fs::path& phaseConfig) {
+    YAML::Node out;
+    switch (node.Type()) {
+        case YAML::NodeType::Map: {
+            std::string fileName;
+            for (YAML::const_iterator it = node.begin(); it != node.end(); ++it) {
+                if (it->first.as<std::string>() == "^Parameter") {
```
rtimmons: any reason to do old-skool versus range-based-for e.g. `for(auto&& kvp : node)`?

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
     }
 }
 
+YAML::Node parseExternal(const YAML::Node& external,
```
rtimmons: would it make things any easier to test or re-use if this code were in `ConfigNode` i.e. if you could just do `ConfigNode parsed = ConfigNode::load(filePath)`?

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+            os << "Could not find top-level key: " << key
+               << " in phase config YAML file: " << path.string();
+            throw InvalidConfigurationException(os.str());
+        }
+        replacement = replacement[key];
+    }
+
+    if (external.size() != nodeSize) {
+        throw InvalidConfigurationException(
+            "Invalid keys for 'External'. Please set 'Path' and if any, 'Parameters");
```
rtimmons: also print the `external` `Node` - you can use `YAML::Emitter` for this - something like the following:

```c++
auto toString(YAML::Node n) {
  YAML::Emitter e;
  e << n;
  return e.str();
}
```

this would also be useful in other places where this code throws on bad yaml - print the full yaml that's in scope wherever possible.

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+    if (!input["Name"] || !input["Default"]) {
+        throw InvalidConfigurationException(
+            "Invalid keys for '^Parameter', please set 'Name' and 'Default'");
+    }
+
+    auto name = input["Name"].as<std::string>();
+    // The default value is mandatory.
+    auto defaultVal = input["Default"];
+
+    auto paramVal = params.find(name);
```
rtimmons: this to the end may benefit from init-statement if:

```c++
if(auto paramVal = params.find(name); paramVal != params.end()) {
  return paramVal->second;
} else {
  return input["Default"];
}
```

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,103 @@ YAML::Node loadConfig(const std::string& source,
...
+    return out;
+}
+
+YAML::Node recursiveParse(const YAML::Node& node,
+                          YamlParameters& params,
+                          const fs::path& phaseConfig) {
+    YAML::Node out;
+    switch (node.Type()) {
+        case YAML::NodeType::Map: {
+            std::string fileName;
```
rtimmons: `fileName` is unused

src/driver/src/DefaultDriver.cpp
```diff
@@ -89,7 +197,31 @@ genny::driver::DefaultDriver::OutcomeCode doRunLogic(
...
+         *     /phase_configs
+         *         /[name-of-actor]
+         *             /[name-of-phase-config-snippet.yml]
+         *
+         * The path to the phase config snippet is obtained as a relative path from the workload
+         * file.
+         */
+        phaseConfigSource =
+            fs::path(options.workloadSource).parent_path().parent_path().parent_path() /
+            "phase_configs";
```
rtimmons: 1. Not sure what to suggest but relying on such a hard-coded path may come back to haunt us.
2. We've used PascalCase so instead lets have the subdir be called `PhaseConfigs` (or maybe `phases` if multi-word-mixed-casing-conventions is too unpleasant)

src/driver/test/Driver_test.cpp
```diff
@@ -326,4 +328,105 @@ TEST_CASE("Various Actor Behaviors") {
...
+        auto [code, opts] = outcome(R"(
+        SchemaVersion: 2018-07-01
+        Actors:
+          - Type: Fails
+            Threads: 1
+            Phases:
+            - ExternalPhaseConfig:
+                Path: "src/phase_configs/self_tests/Good.yml"
+                Parameters:
+                  Repeat: 2
```
rtimmons: I thought we didn't want `Repeat` and ilk to be external? I was imagining the syntax looking like

```yaml
- Type: Fails
  Threads: 1
  Phases:
  - ExternalPhaseConfig:
      Path: self_tests/Good.yml
    Repeat: 10
```

Rationale being that the external Phase yaml can't really both specify `Repeat` and `Duration` as parameters.

src/phase_configs/self_tests/Good.yml
```diff
@@ -0,0 +1,2 @@
+Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
+Mode: NoException
```
rtimmons: I'd really like to avoid having self-test yamls shipped to production machines. When debugging the first thing you do is start to suspect that test yamls are getting read etc. If we don't hard-code the path does that help?

src/driver/src/DefaultDriver.cpp
guoyr: I think there are legit use cases for all three. I copied this line from the description printing code and didn't look closely at `stringstream`. I'm personally not bothered by genny using both. It's like sometimes people use `\n` and some `endl`. But happy to put a note in the style guide if you'd like.

src/driver/src/DefaultDriver.cpp
guoyr: Done. I used `YAML::Dump()` instead to save a `.str()` call.

src/driver/src/DefaultDriver.cpp
guoyr: Nope. Goofballed.

src/driver/src/DefaultDriver.cpp
guoyr: 1. I wanted to have the path be set in an environment variable but people didn't seem to like this at the 6week checkin meeting. Which would you prefer between env var and hard code?
2. So we've been using snake case for directories like `third_party` and `value_party`. I just want to confirm that you want to change the casing. (I think `PhaseConfigs` is preferable to `phases` because it's more explicit)

src/driver/test/Driver_test.cpp
guoyr: Oh yeah. Thanks for catching this!

I'll add a test tomorrow.

src/phase_configs/self_tests/Good.yml
guoyr: I think so. We can then put the self test YAMLs elsewhere and not install them.

src/driver/test/Driver_test.cpp
guoyr: I read over Henrik's sample workload again and he wanted to reuse Duration. So I'm keeping support for `Duration` and `Repeat` in because it's also easier code-wise. The solution is to allow merging of keys specified in external yamls and workload yamls. Conflicts can be caught by `--evaluate`.

The alternative solution which I didn't do is to hard code in checks for top-level keys the external YAML can't contain.

src/driver/src/DefaultDriver.cpp
```diff
@@ -55,6 +66,107 @@ YAML::Node loadConfig(const std::string& source,
...
+
+    // Nested params are ignored for simplicity.
+    if (auto paramVal = params.find(name); paramVal != params.end()) {
+        return paramVal->second;
+    } else {
+        return input["Default"];
+    }
+}
+
+YAML::Node recursiveParse(const YAML::Node& node,
```
rtimmons: I've run into weird problems trying to pass `const YAML::Node&`. It's not well-documented but `YAML::Node`s are basically ref-counted so they're cheap to pass around. Please change this and others to just `YAML::Node`.

src/driver/src/DefaultDriver.cpp
```diff
@@ -89,7 +201,26 @@ genny::driver::DefaultDriver::OutcomeCode doRunLogic(
...
+         *         /[name-of-workload-theme]
+         *             /[my-workload.yml]
+         * The path to the phase config snippet is obtained as a relative path from the workload
+         * file.
+         */
+        phaseConfigSource = fs::path(options.workloadSource).parent_path();
+    }
+
+    std::cout << YAML::Dump(yaml) << std::endl;
+    return genny::driver::DefaultDriver::OutcomeCode::kSuccess;
```
rtimmons: this code modifies `phaseConfigSource` but then the code that actually does `YAML::Dump` doesn't use that - is there some magic I'm missing?

src/driver/test/Driver_test.cpp
```diff
@@ -155,6 +156,7 @@ std::pair<DefaultDriver::OutcomeCode, DefaultDriver::ProgramOptions> outcome(
 
 
 TEST_CASE("Various Actor Behaviors") {
+    boost::filesystem::current_path(genny::findRepoRoot());
```
rtimmons: do you need this?

src/phases/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Install all the phase config snippets.
+install(DIRECTORY .
+    DESTINATION etc/genny/phase_configs
+    COMPONENT PhaseConfigs
+    FILES_MATCHING
+    PATTERN "*.yml"
+)
```
rtimmons: 💅 newline

src/testlib/phases/Good.yml
```diff
@@ -0,0 +1,2 @@
+Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
```
rtimmons: can't comment on the actual file since it's empty but name `src/testlib/phases/GooNoPhase.yml` doesn't make sense

src/testlib/phases/GoodWithKey.yml
```diff
@@ -0,0 +1,3 @@
+ForSelfTest:
+  Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
+  Mode: NoException
```
rtimmons: I think we should require a `SchemaVersion` on included files and basically assert that the `SchemaVersion` matches the top-level workload file.

guoyr:
> I think we should require a SchemaVersion on included files and basically assert that the SchemaVersion matches the top-level workload file.

I thought about the use case for schema version a bit more and I'm wondering if it's better to do without a new one for now.

The benefit of a new schema version is to allow people to run workloads and the genny binaries from different points in time. But this seems like an unlikely use case because 1. genny and workloads live in the same repo 2. we don't guarantee that running the same workload with different versions of genny will produce the same result.

I'm also thinking of protobuf's versioning strategy, where schema versions are used exceedingly sparingly. Any change that is "append only" does not require a new version. Since the changes in this ticket are all append only (which implies non backwards-breaking), theoretically you could include a YAML snippet that uses an older schema version.

Thoughts @rtimmons ?

src/driver/src/DefaultDriver.cpp
guoyr: I left in some debugging code :(

src/testlib/phases/Good.yml
guoyr: lol. Fixed the typo.

src/testlib/phases/GoodWithKey.yml
guoyr: Done

src/driver/test/Driver_test.cpp
rtimmons: ooh..I missed that this sets cwd....hm....that seems like it could make tests order-dependent. Fine with this for now but I'd prefer tests to never assume cwd and call `findRepoRoot` on their own if they need to access a file at a known location.

guoyr: Yeah I had intended to set the CWD for this `TEST_CASE` only. Other test cases will have to set their own to avoid order dependencies.

PR #195, opened 2019-03-28
TIG-1563 Improve metrics parsing performance and log csv parsing progress

guoyr:
I threw in some changes to improve the perf. We should see roughly double the throughput of the first half of csv parsing. (double `tottime` of `method 'writerow' of '_csv.writer' objects`).

I haven't benchmarked csvsort yet but I'm less confident about making large improvements there. If this PR doesn't visibly improve the overall performance of metrics parsing, we may have to do something else entirely for the longer term.

```
// python -m cProfile -s tottime -m genny.cedar_report
// BEFORE
         17957893 function calls (17954266 primitive calls) in 11.364 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   637040    5.467    0.000    7.051    0.000 csv2.py:85(_parse_into_intermediate_csv)
   637040    2.391    0.000    2.391    0.000 {method 'writerow' of '_csv.writer' objects}
637057/637041    0.978    0.000    0.999    0.000 {built-in method builtins.next}
  6370400    0.575    0.000    0.575    0.000 {method 'isdigit' of 'str' objects}
  6370419    0.525    0.000    0.525    0.000 {method 'strip' of 'str' objects}
        1    0.332    0.332   11.096   11.096 cedar.py:173(split_into_actor_csv_files)
   637040    0.321    0.000    8.372    0.000 csv2.py:82(__next__)
   637040    0.273    0.000    0.273    0.000 csv2.py:109(<listcomp>)
   637041    0.112    0.000    0.112    0.000 csv2.py:273(default_columns)
1280275/1279786    0.100    0.000    0.100    0.000 {built-in method builtins.len}

// AFTER
         7486149 function calls (7482522 primitive calls) in 11.639 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  1226057    4.472    0.000    4.472    0.000 {method 'writerow' of '_csv.writer' objects}
  1226057    3.817    0.000    4.007    0.000 csv2.py:85(_parse_into_intermediate_csv)
1226074/1226058    1.739    0.000    1.774    0.000 {built-in method builtins.next}
  1226057    0.608    0.000    6.388    0.000 csv2.py:82(__next__)
        1    0.582    0.582   11.445   11.445 cedar.py:173(split_into_actor_csv_files)
  2457609    0.190    0.000    0.190    0.000 {method 'rstrip' of 'str' objects}
       17    0.029    0.002    0.030    0.002 {method 'get_filename' of 'zipimport.zipimporter' objects}
     17/1    0.029    0.002    0.075    0.075 {method 'load_module' of 'zipimport.zipimporter' objects}
     9399    0.019    0.000    0.019    0.000 {built-in method _codecs.utf_8_decode}
       25    0.016    0.001    0.017    0.001 {built-in method _imp.create_dynamic}
     9399    0.016    0.000    0.034    0.000 codecs.py:319(decode)
      180    0.016    0.000    0.016    0.000 {built-in method marshal.loads}
```

### DIFF
--- a/src/python/genny/cedar.py
+++ b/src/python/genny/cedar.py
@@ -179,6 +179,9 @@ def split_into_actor_csv_files(data_reader, out_dir):
     cur_actor = (None, None)
     output_files = []
 
+    # Print progress to prevent the CI task from timing out.
+    counter = 0
+
     for line, actor in data_reader:
         if actor != cur_actor:
             cur_actor = actor
@@ -197,6 +200,10 @@ def split_into_actor_csv_files(data_reader, out_dir):
             cur_out_csv.writerow(IntermediateCSVColumns.default_columns())
 
         cur_out_csv.writerow(line)
+
+        counter += 1
+        if not (counter % 1e6):
+            print('Parsed {} metrics'.format(counter))
 
     cur_out_fh.close()
 
--- a/src/python/genny/cedar.py
+++ b/src/python/genny/cedar.py
@@ -202,7 +202,7 @@ def split_into_actor_csv_files(data_reader, out_dir):
         cur_out_csv.writerow(line)
 
         counter += 1
-        if not (counter % 1e6):
+        if counter % 1e6 == 0:
             print('Parsed {} metrics'.format(counter))
 
     cur_out_fh.close()
--- a/src/python/genny/cedar_report.py
+++ b/src/python/genny/cedar_report.py
@@ -286,26 +286,30 @@ def main__cedar_report(argv=sys.argv[1:], env=None, cert_retriever_cls=CertRetri
     parser = build_parser()
     args = parser.parse_args(argv)
 
-    if not env:
-        with open(args.expansions_file, 'r') as f:
-            env = yaml.safe_load(f)
-
-    if args.test_name:
-        env['test_name'] = args.test_name
-    else:
-        env['test_name'] = env['task_name']
+    # if not env:
+    #     with open(args.expansions_file, 'r') as f:
+    #         env = yaml.safe_load(f)
+    #
+    # if args.test_name:
+    #     env['test_name'] = args.test_name
+    # else:
+    #     env['test_name'] = env['task_name']
 
     metrics_file_names, test_run_time = cedar.run(args)
-    config = _Config(env, metrics_file_names, test_run_time)
+    # config = _Config(env, metrics_file_names, test_run_time)
+    #
+    # report_dict = build_report(config)
+    #
+    # with open(args.report_file, 'w') as f:
+    #     json.dump(report_dict, f, cls=RFCDateTimeEncoder)
+    #
+    # jira_user = env['perf_jira_user']
+    # jira_pwd = env['perf_jira_pw']
+    #
+    # cr = cert_retriever_cls(jira_user, jira_pwd)
+    # runner = ShellCuratorRunner(cr, args.report_file)
+    # runner.run(runner.get_command())
 
-    report_dict = build_report(config)
 
-    with open(args.report_file, 'w') as f:
-        json.dump(report_dict, f, cls=RFCDateTimeEncoder)
-
-    jira_user = env['perf_jira_user']
-    jira_pwd = env['perf_jira_pw']
-
-    cr = cert_retriever_cls(jira_user, jira_pwd)
-    runner = ShellCuratorRunner(cr, args.report_file)
-    runner.run(runner.get_command())
+if __name__ == '__main__':
+    main__cedar_report()
--- a/src/python/genny/cedar_report.py
+++ b/src/python/genny/cedar_report.py
@@ -286,29 +286,29 @@ def main__cedar_report(argv=sys.argv[1:], env=None, cert_retriever_cls=CertRetri
     parser = build_parser()
     args = parser.parse_args(argv)
 
-    # if not env:
-    #     with open(args.expansions_file, 'r') as f:
-    #         env = yaml.safe_load(f)
-    #
-    # if args.test_name:
-    #     env['test_name'] = args.test_name
-    # else:
-    #     env['test_name'] = env['task_name']
+    if not env:
+        with open(args.expansions_file, 'r') as f:
+            env = yaml.safe_load(f)
+
+    if args.test_name:
+        env['test_name'] = args.test_name
+    else:
+        env['test_name'] = env['task_name']
 
     metrics_file_names, test_run_time = cedar.run(args)
-    # config = _Config(env, metrics_file_names, test_run_time)
-    #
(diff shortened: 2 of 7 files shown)


src/python/genny/csv2.py
```diff
@@ -82,42 +82,43 @@ def __iter__(self):
...
+        op = csv_line[_OpColumns.OPERATION].rstrip()
+        actor = csv_line[_OpColumns.ACTOR].rstrip()
+        # Convert timestamp from ns to ms and add offset.
+        unix_time = (int(csv_line[_OpColumns.TIMESTAMP]) + self.unix_time_offset) / (1000 * 1000)
 
-            # Convert string digits to ints
-            if line[i].isdigit():
-                line[i] = int(line[i])
+        # Transform output into IntermediateCSV format.
+        # Python list assignment too slow so we construct the output array directly.
```
visemet: 30 nanoseconds per list element assignment doesn't seem that slow to me. I think the problem comes down to how the list is being constructed via a list comprehension.

If the number of columns isn't expected to change at runtime, then we should we using tuples over lists.

```
$ python3 -m timeit 'a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]'
$ python3 -m timeit 'a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]'
5000000 loops, best of 5: 96.8 nsec per loop

$ python3 -m timeit 'a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]; a[0] = 0; a[1] = 1; a[2] = 2; a[3] = 3; a[4] = 4; a[5] = 5; a[6] = 6; a[7] = 7; a[8] = 8; a[9] = 9; a[10] = 10'
1000000 loops, best of 5: 401 nsec per loop

$ python3 -m timeit 'a = [None for _ in range(len(["unix_time", "ts", "thread", "operation", "duration", "outcome", "n", "ops", "errors", "size", "workers"]))]; a[0] = 0; a[1] = 1; a[2] = 2; a[3] = 3; a[4] = 4; a[5] = 5; a[6] = 6; a[7] = 7; a[8] = 8; a[9] = 9; a[10] = 10'
200000 loops, best of 5: 1.41 usec per loop

$ python3 -m timeit 'a = [None for _ in range(len(["unix_time", "ts", "thread", "operation", "duration", "outcome", "n", "ops", "errors", "size", "workers"]))]'
200000 loops, best of 5: 1.05 usec per loop

$ python3 -m timeit 'a = [None] * 11'
2000000 loops, best of 5: 149 nsec per loop

$ python3 -m timeit 'a = [None] * len(["unix_time", "ts", "thread", "operation", "duration", "outcome", "n", "ops", "errors", "size", "workers"])'
1000000 loops, best of 5: 297 nsec per loop

$ python3 -m timeit 'a = [None] * len(("unix_time", "ts", "thread", "operation", "duration", "outcome", "n", "ops", "errors", "size", "workers"))'
1000000 loops, best of 5: 228 nsec per loop
```

guoyr:
> If I'm reading the profiler output correctly, then split_into_actor_csv_files() is the slowest step.
Are you referring to the `cumtime` column? The cumulative time is the total time spent in a function including function calls. `tottime` is  the time spent outside of function calls and is the thing I we care about. Since `split_into_actor_csv_files` calls `writerow` and `_parse_into_intermediate_csv `, its cumulative time is going to be the largest. But its `tottime` should be very low.

> Why wouldn't we want to focus on that the most because it seems like any performance bottlenecks in _parse_into_intermediate_csv() could maybe be ignored by just using multiprocessing to handle each csv file.
multiprocessing could work. But it's not a trivial amount of work. Genny produces one large metrics file and the script splits it up into multiple files. We'd need to parse the one file in parallel to get benefits from multiprocessing

src/python/genny/cedar_report.py
```diff
@@ -286,26 +286,30 @@ def main__cedar_report(argv=sys.argv[1:], env=None, cert_retriever_cls=CertRetri
...
 
-    if not env:
-        with open(args.expansions_file, 'r') as f:
-            env = yaml.safe_load(f)
-
-    if args.test_name:
-        env['test_name'] = args.test_name
-    else:
-        env['test_name'] = env['task_name']
+    # if not env:
```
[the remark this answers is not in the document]
guoyr: Sorry those were for local testing. Reverted.

src/python/genny/csv2.py
guoyr: Thanks for the tip! I removed list comprehension in favor of multiplication

src/python/genny/cedar_report.py
```diff
@@ -309,3 +309,7 @@ def main__cedar_report(argv=sys.argv[1:], env=None, cert_retriever_cls=CertRetri
     cr = cert_retriever_cls(jira_user, jira_pwd)
     runner = ShellCuratorRunner(cr, args.report_file)
     runner.run(runner.get_command())
+
+
+if __name__ == '__main__':
```
guoyr: I left this in here so people can run the code snippet I posted in TIG-1566

src/python/genny/cedar.py
```diff
@@ -198,6 +201,10 @@ def split_into_actor_csv_files(data_reader, out_dir):
 
         cur_out_csv.writerow(line)
 
+        counter += 1
+        if counter % 1e6 == 0:
+            print('Parsed {} metrics'.format(counter))
```
visemet: [note] You can also use f-strings as long as we're willing to set a floor of Python 3.6.

https://docs.python.org/3/reference/lexical_analysis.html#f-strings

src/python/genny/csv2.py
```diff
@@ -84,10 +84,6 @@ def __next__(self):
 
     def _parse_into_intermediate_csv(self, line):
         for i in range(len(line)):
-
-            # Strip spaces from lines
-            line[i] = line[i].strip()
-
             # Convert string digits to ints
             if line[i].isdigit():
```
visemet: FWIW, the `str.isdigit()` approach performs a lot faster than the `try/except ValueError` approach.

```
$ python3 -m timeit '
a = "1"
if a.isdigit():
    int(a)
'
1000000 loops, best of 5: 295 nsec per loop

$ python3 -m timeit '
a = "a"
if a.isdigit():
    int(a)
'
5000000 loops, best of 5: 73.4 nsec per loop

$ python3 -m timeit '
a = "1"
try:
    int(a)
except ValueError:
    pass
'
1000000 loops, best of 5: 241 nsec per loop

$ python3 -m timeit '
a = "a"
try:
    int(a)
except ValueError:
    pass
'
200000 loops, best of 5: 1.33 usec per loop
```

PR #196, opened 2019-03-29
TIG-1542 Support --run-mode

guoyr:
Here's an example output:

```
(venv) ➜  build git:(tig1542) ✗ ./src/driver/genny --run-mode evaluate ../src/workloads/docs/HelloWorld.yml
[2019-03-29 13:29:07.384822] [0x00000001125b65c0] [info]    Confirmed logging through boost
🧞 Genny Version 0.0.1 💝🐹🌇⛔
evaluate
Printing evaluated workload YAML file:
SchemaVersion: 2018-07-01
Owner: "@mongodb/stm"
Actors:
  - Name: HelloWorld
    Type: HelloWorld
    Threads: 2
    Phases:
      - Message: Hello Phase 0 🐳
        Duration: 50 milliseconds
      - Message: Hello Phase 1 👬
        Repeat: 100
(venv) ➜  build git:(tig1542) ✗ ./src/driver/genny --run-mode list-actors ../src/workloads/docs/HelloWorld.yml
[2019-03-29 13:29:13.906353] [0x00000001152115c0] [info]    Confirmed logging through boost
🧞 Genny Version 0.0.1 💝🐹🌇⛔
list-actors
AdminCommand is AdminCommand
CrudActor is CrudActor
HelloWorld is HelloWorld
Insert is Insert
InsertRemove is InsertRemove
Loader is Loader
MultiCollectionQuery is MultiCollectionQuery
MultiCollectionUpdate is MultiCollectionUpdate
RunCommand is RunCommand
(venv) ➜  build git:(tig1542) ✗ ./src/driver/genny --run-mode dry-run ../src/workloads/docs/HelloWorld.yml
[2019-03-29 13:29:26.688154] [0x00000001096ef5c0] [info]    Confirmed logging through boost
🧞 Genny Version 0.0.1 💝🐹🌇⛔
dry-run
Workload context constructed without errors.
```


src/driver/include/driver/DefaultDriver.hpp
```diff
@@ -28,8 +28,24 @@ namespace genny::driver {
  * Basic workload driver that spins up one thread per actor.
  */
 class DefaultDriver {
-
 public:
+    enum class RunMode {
+        kNormal = 0,
```
visemet: [style] Is there any value to specify `= 0` here? Unlike the `OutcomeCode` enum we aren't trying to convert the `RunMode` enum cases to their underlying integer value.

src/driver/include/driver/DefaultDriver.hpp
```diff
@@ -28,8 +28,24 @@ namespace genny::driver {
...
+        kDryRun = 1,
+        kEvaluate = 2,
+        kListActors = 3,
+        kSmokeTest = 4,
```
visemet: Unused? I know Ryan mentioned it in the ticket description but it seems like when we add support for it in code we can the enum case.

src/driver/src/DefaultDriver.cpp
```diff
@@ -251,3 +283,4 @@ genny::driver::DefaultDriver::ProgramOptions::ProgramOptions(int argc, char** ar
         this->workloadSourceType = YamlSource::kString;
     }
 }
+}  // namespace genny::driver
```
visemet: [lint] Add a newline to the end of this file.

src/driver/src/DefaultDriver.cpp
```diff
@@ -188,58 +193,85 @@ std::string normalizeOutputFile(const std::string& str) {
...
+    if (vm.count("list-actors") >= 1 || runModeStr == "list-actors")
+        this->runMode = RunMode::kListActors;
+
+    if (vm.count("dry-run") >= 1 || runModeStr == "dry-run")
+        this->runMode = RunMode::kDryRun;
+
+    if (runModeStr == "evaluate")
+        this->runMode = RunMode::kEvaluate;
+
+    std::cout << runModeStr << "\n";
```
visemet: Remove this debugging line?

src/driver/src/DefaultDriver.cpp
```diff
@@ -37,10 +37,10 @@
 
 #include <driver/DefaultDriver.hpp>
 
+namespace genny::driver {
 namespace {
 
 using namespace genny;
```
visemet: Seems like this line can also be removed because all the usages of types are within the `genny` and `genny::driver` namespaces.

src/driver/src/DefaultDriver.cpp
```diff
@@ -77,27 +77,34 @@ void runActor(Actor&& actor,
...
+        setupCtx.success();
+        return DefaultDriver::OutcomeCode::kSuccess;
+    }
+
     auto yaml = loadConfig(options.workloadSource, options.workloadSourceType);
     auto orchestrator = Orchestrator{};
 
+    if (options.runMode == DefaultDriver::RunMode::kEvaluate) {
+        std::cout << "Printing evaluated workload YAML file:" << std::endl;
+        std::cout << YAML::Dump(yaml) << std::endl;
```
visemet: I find it a little odd that I can run `./build/src/driver/genny --run-mode evaluate` and not get an error from not having specified a file.

src/driver/src/DefaultDriver.cpp
```diff
@@ -188,58 +193,85 @@ std::string normalizeOutputFile(const std::string& str) {
...
     po::variables_map vm;
     po::store(run, vm);
     po::notify(vm);
 
+    const auto runModeStr = vm["run-mode"].as<std::string>();
+
+    if (vm.count("list-actors") >= 1 || runModeStr == "list-actors")
+        this->runMode = RunMode::kListActors;
+
+    if (vm.count("dry-run") >= 1 || runModeStr == "dry-run")
```
visemet: [question] Is it worth making the `--list-actors` and `--run-mode` conflict or do we not care about being able to do inane things like `./build/src/driver/genny --list-actors --run-mode dry-run src/workloads/docs/HelloWorld.yml`?

src/driver/include/driver/DefaultDriver.hpp
guoyr: Nah. Removed.

src/driver/include/driver/DefaultDriver.hpp
guoyr: SG. Removed

src/driver/include/driver/DefaultDriver.hpp
```diff
@@ -28,8 +28,24 @@ namespace genny::driver {
...
+    };
+
+
+    enum class OutcomeCode {
+        kSuccess = 0,
+        kStandardException = 1,
+        kBoostException = 2,
+        kInternalException = 3,
+        kUnknownException = 10,
```
guoyr: FWIW this is also unused.

src/driver/src/DefaultDriver.cpp
guoyr: yep

src/driver/src/DefaultDriver.cpp
guoyr: I added error handling for empty `workloadSource` when not in `list-actors` mode.

src/driver/src/DefaultDriver.cpp
guoyr: Hm, would you be open to deleting the non-run-mode version here and in DSI? I'll of course also send out an email.

visemet: It feels like maybe more trouble than its worth to do in this ticket right now to remove those command line options. We should get the DSI code and any other scripting switched over to using the new command line options in the next iteration though.

src/driver/src/DefaultDriver.cpp
```diff
@@ -77,27 +75,40 @@ void runActor(Actor&& actor,
...
+        setupCtx.failure();
+        return DefaultDriver::OutcomeCode::kUserException;
+    }
+
     auto yaml = loadConfig(options.workloadSource, options.workloadSourceType);
     auto orchestrator = Orchestrator{};
 
+    if (options.runMode == DefaultDriver::RunMode::kEvaluate) {
+        std::cout << "Printing evaluated workload YAML file:" << std::endl;
+        std::cout << YAML::Dump(yaml) << std::endl;
```
rtimmons: As it is this will prevent doing someting like `genny --evaluate | yamllint` since the yaml and the logs are interspersed. 

Let's either:

1. dump the log messages and the file to different ostreams (e.g. logs to stderr and yaml to stdout)
2. not do logging for `--evaluate`

src/driver/src/DefaultDriver.cpp
```diff
@@ -188,58 +197,82 @@ std::string normalizeOutputFile(const std::string& str) {
...
-            "Mongo URI to use for the default connection-pool.")
-    ;
+            ("metrics-output-file,o",
+             po::value<std::string>()->default_value("/dev/stdout"),
+             "Save metrics data to this file. Use `-` or `/dev/stdout` for stdout.")
+            ("workload-file,w",
+             po::value<std::string>(),
+             "Path to workload configuration yaml file. "
+             "Paths are relative to the program's cwd. "
+             "Can also specify as the last positional argument.")
```
rtimmons: similarly this seems like it "wants" to be the first positional param for the `genny run` subcommand:

```sh
genny run ./workloads/foo.yml
```

(I'm saying all this without really knowing how fancy or not `boost::program_options` is...it gets messy even in Python so I wouldn't be surprised if this makes things go off the rails.)

src/driver/src/DefaultDriver.cpp
```diff
@@ -188,58 +197,82 @@ std::string normalizeOutputFile(const std::string& str) {
...
+
+    list-actors
+        List all actors available for use
+    )";
+
+    progDescription.add_options()
+            ("help,h",
+             "Show help message")
+            ("run-mode", po::value<std::string>()->default_value("normal"),
+             runModeHelp)
```
rtimmons: do we want to require the `--run-mode` param or can it just be the first positional param similar to `git` et al. E.g.

```sh
genny list-actors
genny run # aka "normal"
```

This would change the "run mode" to be more of a "command" or "action". It seems clunky to require an option flag to tell it what to do. Imagine if you had to type `git --action checkout` 😸
I don't have a strong opinion here so please push back if you think it's not worth it.

src/driver/src/DefaultDriver.cpp
```diff
@@ -188,58 +197,82 @@ std::string normalizeOutputFile(const std::string& str) {
...
+            ("list-actors",
+                    "DEPRECATED. Please use the \"list-actors\" run mode.")
+            ("dry-run",
+                    "DEPRECATED. Please the \"dry-run\" run mode.")
```
rtimmons: imho it's fine to break old usages at least for now - I'd just remove these for now - the change to DSI is relatively easy to do simultaneously when merging this.

src/driver/src/DefaultDriver.cpp
rtimmons: My vote would be to remove the old options - seems easy to change dsi and genny at the same time and this will do a lot to make the options more sane.

src/driver/src/main.cpp
```diff
@@ -36,6 +36,7 @@ int main(int argc, char** argv) {
 
     auto opts = genny::driver::DefaultDriver::ProgramOptions(argc, argv);
     if (opts.isHelp) {
+        std::cout << u8"🧞 Genny Usage: " << argv[0] << "[options] <workload-file>" << std::endl;
```
rtimmons: 😻

src/driver/src/DefaultDriver.cpp
guoyr: Funny you should ask! Everything before the "[don't use subcommands](https://github.com/mongodb/genny/pull/196/commits/205956bdf52ddc2113ad9790f93341d1ab8dffda)" commit used subcommands as you described.

The problem is the workload yaml is also a positional argument and genny can't distinguish between the two easily. I didn't want to want to have genny guess whether the first argument is a workload file or a subcommand.

What do you think about removing the workload positional argument?

rtimmons: > genny can't distinguish between the two easily

it seems like it should always assume that the first argument is the subcommand and any after that are positional arguments to the subcommand. So running just `genny foo.yaml` shouldn't work because `foo.yaml` isn't a subcommand but `genny run foo.yaml` should assume subcommand is run and the first positional arg is foo.yaml.

But if you're saying that boost program options doesn't let you easily express this then it does seem reasonable to only allow one positional option and have that be the subcommand and to give all other parameters via options e.g. `genny run --workload-file foo.yaml`. I've seen some tools do this so it's not totally unheard-of, but tools like git and aws-cli handle multiple positional arguments and shift them with ease.

(ultimately users interact with the genny binary so infrequently this is edging closer to bikeshedding so please just do what you feel is right - biggest thing for me is having one distinct runmode option that all the various list-actors/normal/evaluate/* all fall under.  Whether it's a --option or a positional argument and whether subsequent things are positional or not doesn't end up affecting us too much imho.

src/driver/src/DefaultDriver.cpp
rtimmons: Bump on this

src/driver/src/DefaultDriver.cpp
```diff
@@ -77,27 +75,40 @@ void runActor(Actor&& actor,
...
     auto setupCtx = actorSetup.start();
 
+    if (options.runMode == DefaultDriver::RunMode::kListActors) {
+        globalCast().streamProducersTo(std::cout);
+        setupCtx.success();
+        return DefaultDriver::OutcomeCode::kSuccess;
+    }
+
+    if (options.workloadSource.empty()) {
+        std::cout << "Must specify a workload YAML file" << std::endl;
```
rtimmons: I think this wants to be `std::cerr`

src/driver/src/DefaultDriver.cpp
guoyr: I went with option 2 by removing the existing "genny version" and "confirmed logging through boost" expect for option parsing errors and `--help`.

Happy to put them back and go with option 1 if you prefer.

guoyr:
I moved some logging around to improve error messages when no subcommand or an unknown subcommand is specified:
```
➜  build git:(tig1542) ./src/driver/genny
ERROR: missing subcommand

🧞 Genny Version 0.0.1 💝🐹🌇⛔

🧞 Usage:
    ./src/driver/genny <subcommand> [options] <workload-file>

🧞 Subcommands:‍
    run          Run the workload normally
    dry-run      Exit before the run step -- this may still make network
                 connections during workload initialization
    evaluate     Print the evaluated YAML workload file with minimal validation
    list-actors  List all actors available for use

🧞 Options:
  --subcommand arg                      1st positional argument
  -h [ --help ]                         Show help message
  -m [ --metrics-format ] arg (=csv)    Metrics format to use
  -o [ --metrics-output-file ] arg (=/dev/stdout)
                                        Save metrics data to this file. Use `-`
                                        or `/dev/stdout` for stdout.
  -w [ --workload-file ] arg            Path to workload configuration yaml
                                        file. Paths are relative to the
                                        program's cwd. Can also specify as the
                                        last positional argument.
  -u [ --mongo-uri ] arg (=mongodb://localhost:27017)
                                        Mongo URI to use for the default
                                        connection-pool.

➜  build git:(tig1542) ./src/driver/genny asdf
ERROR: Unexpected subcommand asdf

🧞 Genny Version 0.0.1 💝🐹🌇⛔

🧞 Usage:
    ./src/driver/genny <subcommand> [options] <workload-file>

🧞 Subcommands:‍
    run          Run the workload normally
    dry-run      Exit before the run step -- this may still make network
                 connections during workload initialization
    evaluate     Print the evaluated YAML workload file with minimal validation
    list-actors  List all actors available for use

🧞 Options:
  --subcommand arg                      1st positional argument
  -h [ --help ]                         Show help message
  -m [ --metrics-format ] arg (=csv)    Metrics format to use
  -o [ --metrics-output-file ] arg (=/dev/stdout)
                                        Save metrics data to this file. Use `-`
                                        or `/dev/stdout` for stdout.
  -w [ --workload-file ] arg            Path to workload configuration yaml
                                        file. Paths are relative to the
                                        program's cwd. Can also specify as the
                                        last positional argument.
  -u [ --mongo-uri ] arg (=mongodb://localhost:27017)
                                        Mongo URI to use for the default
                                        connection-pool.
```

src/driver/include/driver/DefaultDriver.hpp
```diff
@@ -47,18 +63,8 @@ class DefaultDriver {
...
-
-
-    enum class OutcomeCode {
-        kSuccess = 0,
-        kStandardException = 1,
-        kBoostException = 2,
-        kInternalException = 3,
-        kUnknownException = 10,
+        bool showHelp = false;
+        DefaultDriver::RunMode runMode = RunMode::kNormal;
```
rtimmons: sorry to keep iterating but does `showHelp` also want to be a run-mode?

src/driver/src/DefaultDriver.cpp
rtimmons: No I don't think the genny version thing is still valuable - it was just a smoke-test that things were compiling with boost etc I think it's cruft at this point (although I smile every time I see the little string of emojis it prints)

src/driver/include/driver/DefaultDriver.hpp
guoyr: Yeah that makes sense. I didn't know you could do `git help` or `brew help`

guoyr:
This PR is related to https://github.com/10gen/dsi/pull/347

PR #197, opened 2019-03-29
TIG-1535 Add sanity check for ^Parameter parsing

guoyr:


src/value_generators/test/expression_test.cpp
```diff
@@ -47,6 +47,10 @@ TEST_CASE("Expression::parseExpression error cases") {
 
         auto expr = Expression::parseExpression(yaml, rng);
         REQUIRE(expr != nullptr);
+
+        yaml = YAML::Load(R"({^Parameter: {}})");
+
+        REQUIRE_THROWS_AS(Expression::parseExpression(yaml, rng), InvalidValueGeneratorSyntax);
```
visemet: The way these test cases are organization, test cases in the "valid syntax" section shouldn't be throwing exceptions. Since you're wanting to make assertions about the behavior of a new `Expression` subclass, I'd recommend adding a new "Expression parsing with Parameter always errors" test case.

Aren't `"Name"` and `"Default"` required keys? I realize that's being enforced at a different layer but there's no reason for the tests to assume that. Regarding ←, it may also be worth noting that the arguments to all the other `Expression` subclasses are lowercase. I'm not sure if they're intended to be camelCase, snake_case, or "multiple words" because all of the existing ones are single words.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -27,12 +27,23 @@
...
 namespace {
 
+class ParameterExpression {
+public:
+    static UniqueExpression parse(YAML::Node node, DefaultRandom& rng);
+};
+
+UniqueExpression ParameterExpression::parse(YAML::Node node, genny::DefaultRandom& rng) {
+    throw InvalidValueGeneratorSyntax(
+        "Encountered unexpected '^Parameter'. This indicates a bug in genny's parameter parsing "
```
visemet: What is "genny's parameter parsing code"? Can we name a function here? Should the concept of `"ExternalPhaseConfig"` be mentioned here?

src/value_generators/test/expression_test.cpp
guoyr: lol. I stopped reading after the test case said "error cases" and completely overlooked the section name. 

> I'm not sure if they're intended to be camelCase, snake_case, or "multiple words" because all of the existing ones are single words.

`^Parameter` isn't parsed as an expression so I'm inclined to keep it as PascalCase.

I think of expression args at atomic because genny won't further decompose them. But `^Parameter` could theoretically accept arbitrary YAML objects containing more PascalCase nodes. It would seem weird to go from PascalCase to camelCase/snake_cake then back to PascalCase.

E.g. `{^Parameter: {default: {Document: {^RandomInt: {min: 0, max: 1}}}}`

src/value_generators/src/DocumentGenerator.cpp
guoyr: Good point. This error doesn't necessarily mean a bug in genny.

src/value_generators/test/expression_test.cpp
dalyd: This is my least favorite kind of technical debate. I'd like things to be "consistent" and simple, and for someone to tell me what the scheme is. If the current form is because of me, then feel free to change it and let me know what the answer is :)

If it is because of me, I suspect I was trying to match benchRun and mgenerate. All the named arguments to benchrun are lowercase (and single word I believe) https://github.com/mongodb/mongo/blob/258f72a671492853501b551cd3135a6dbf5208d4/src/mongo/shell/bench.cpp#L385. Scratch that -- the arguments are camelCase in benchRun. 

Everything is lower case in mgenerate: https://www.npmjs.com/package/mgeneratejs . We seem to have gone away from that here already.

PR #199, opened 2019-04-02
TIG-1579 Always remove toolchain before compilation

guoyr:


evergreen.yml
```diff
@@ -362,6 +349,11 @@ functions:
       script: |
         set -eo pipefail
 
+        # Hosts may have incomplete or invalid toolchain directories from
```
visemet: "3. Put comments in yaml not in inline shell."

evergreen.yml
```diff
@@ -119,10 +119,6 @@ tasks:
       short_name: genny
       install_dir: dist
 
-- name: t_rm_toolchain_dir_static_host
```
visemet: Seems like https://github.com/mongodb/genny/blob/336a7cca9525b066a3425d7775381668f6fb11e3/evergreen.yml#L221 needs to also be removed. It's probably worth keeping the `tg_compile_static_host` task group for consistency and ease of adding new tasks back.

guoyr: Yep.

I also changed all hosts to run compile in a task group to be consistent with other hosts

evergreen.yml
guoyr: Can't argue with that!

PR #201, opened 2019-04-04
TIG-1577 update to latest poplar report schema

guoyr:
Depends on https://github.com/10gen/dsi/pull/346

### DIFF
--- a/src/python/genny/cedar_report.py
+++ b/src/python/genny/cedar_report.py
@@ -17,7 +17,7 @@ CedarBucketConfig = namedtuple('CedarBucketConfig', [
     'api_token',
     'region',
     'name',
-    'permission',
+    'permissions',
     'prefix'
 ])
 
@@ -134,7 +134,7 @@ def build_report(config):
         api_token=None,
         region=config.cloud_region,
         name=config.cloud_bucket,
-        permission='public-read',
+        permissions='public-read',
         prefix=bucket_prefix
     )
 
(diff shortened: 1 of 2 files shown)


src/python/genny/cedar_report.py
```diff
@@ -105,7 +106,7 @@ def build_report(config):
             tags=[],
             local_path=path,
             created_at=config.created_at,
-            is_uncompressed=True
+            convert_bson_to_ftdc=True
```
rtimmons: personal vendetta against booleans - can this be a `conversion_mode` or something that's an enum e.g. `bson_to_ftdc` or `no_conversion`?

(I realize it was already a boolean and this just gives it a better name so feel free to punt on this)

src/python/genny/cedar_report.py
```diff
@@ -133,6 +134,7 @@ def build_report(config):
         api_token=None,
         region=config.cloud_region,
         name=config.cloud_bucket,
+        permission='public-read',
```
rtimmons: can it...not be? I understand it's not sensitive information but usually I see "public-read" and think the permissions are just poorly-defined and we're resorting to ¯\\_(ツ)\_/¯

src/python/genny/cedar_report.py
guoyr: I think you'd want to have poplar do this. The format of the `namedtuple`s in this file just mirror the schema defined in poplar, it seems weird to make an exception for this one field.

src/python/genny/cedar_report.py
guoyr: Sadly this is something poplar should do as well :(

I filed https://jira.mongodb.org/browse/MAKE-717 to have the bucket policy be set to disallow public reads.

PR #202, opened 2019-04-05
PERF-1748 Separate CommitLatency configurations into individual actors

henrikingo:
Genny reports metrics per Actor (multiple phases aggregate together), so each configuration needs to be its own actor instance.


henrikingo:
Ah, I thought github would rebase automatically. I will close this and open a new one.

PR #203, opened 2019-04-05
PERF-1748 Separate CommitLatency configurations into individual actors

henrikingo:
Genny reports metrics per Actor (multiple phases aggregate together), so each configuration needs to be its own actor instance.

### DIFF
--- a/src/workloads/networking/CommitLatency.yml
+++ b/src/workloads/networking/CommitLatency.yml
@@ -66,50 +66,26 @@ Actors:
 # All of the below is really just the same actor, but as Genny reports statistics using the actor
 # name, we need a different actor instance for each configuration. All of the below run in their
 # own slot, nothing is concurrent.
+# Since Drop+Reload is in phases 0,1,-,3,4,-... then the test itself must always be in a slot at
+# 2,5,8... = 3n-1 = 3n+2
 - Name: CommitLatency_w0_local
   Type: CommitLatency
   Threads: 1
   Repeat: 500
   Database: test
   Phases:
-  - *Nop # Load
-  - *Nop
+  - Phase: 0..1
+    Operation:
+      OperationName: Nop
   - WriteConcern:
       Level: 0
     ReadConcern:
       Level: local
     ReadPreference:
       ReadMode: primary
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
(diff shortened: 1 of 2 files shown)


henrikingo:
Sys-perf patch build fails on some cedar error, but you can verify that the genny-perf.csv files do contain metrics separated by actor.
https://evergreen.mongodb.com/task/sys_perf_linux_3_node_replSet_single_threaded_patch_9cecff85b28bc0f7279b31f10ff809947fbdaee4_5ca63c0be3c33119ed322d47_19_04_04_17_24_46

I would like to commit this change to Genny, even if the cedar bug prevents from adding this to sys-perf.

src/workloads/networking/CommitLatency.yml
```diff
@@ -137,6 +420,46 @@ Actors:
...
+  - *Nop
+  - *Nop
+
+- Name: CommitLatency_wmajority_majority_session
+  Type: CommitLatency
+  Threads: 1
+  Repeat: 500
+  Database: test
+  Phases:
+  - *Nop
```
rtimmons: FWIW you can use

```yaml
Phases:
- Phase: 0..N
  Operation: Nop
```

henrikingo: Ah cool. None of the existing workloads have that, so I assumed it wasn't implemented.

Is there also an open ticket for just doing

    Phases:
        - Phase: N

...and allowing to leave out the Nop phases completely? (Ie Nop would be default.)

henrikingo:
Ok, done with this. Assuming the evergreen builds succeed, can you merge this please.

rtimmons:
Sorry forgot to circle back here @henrikingo 

> Is there also an open ticket for just doing
> 
>     Phases:
>         - Phase: N
> 
> ...and allowing to leave out the Nop phases completely? (Ie Nop would be default.)

There isn't a ticket for this because it's still something I'm...not sure of. I'm afraid it could hide mis-configurations. 

I'm in the process of creating a Google document that lists features/questions/etc that we haven't immediately prioritized. I think this will be a better conversation venue than a Jira dumping-ground. Not *quite* ready to share this doc with a wider audience but I've put this concern in there as "GI-010: Default Omitted Phase Numbers to Nops". I'll share it widely later this week or early next week.

PR #204, opened 2019-04-05
TIG-1343 Implement composable large scale workloads

guoyr:
There's some noise from pulling in TIG-1543.

`LargeScaleSerial.yml` and `LargeScaleParallel.yml` are the two files that matter.


src/phases/scale/DesignDocWorkloadPhases.yml
```diff
@@ -0,0 +1,105 @@
+Document: &Doc # Documents are approximately 1 KiB in size
```
rtimmons: 💅 missing SchemaVersion here

src/phases/scale/DesignDocWorkloadPhases.yml
```diff
@@ -0,0 +1,105 @@
...
+  Filter: {y: {^RandomInt: {distribution: poisson, mean: 100}}}
+  Limit: 20
+
+LargeQuery:
+  Database: *DB
+  CollectionCount: {^Parameter: {Default: *DefaultCollectionCount, Name: "CollectionCount"}}
+  Filter: {z: { "$gt": { ^RandomInt: {min: 0, max: 2147483647} } } }
+  # Documents are 1KB so we want to read 1024 documents over 2 minutes.
+  Limit: {^Parameter: {Default: 0, Name: "Limit"}}
+#  Transactions: true
```
rtimmons: I realize this is a "nomerge" pr but I actually think this may be worth merging but that's your call.
If you do end up merging, let's add a comment and/or a TODO with a ticket number for these features

src/phases/scale/DesignDocWorkloadPhases.yml
```diff
@@ -0,0 +1,105 @@
...
+#  ReadConcern: Snapshot
+  #  MinDelay: 117 milliseconds  # Sleep between next()
+  #  StaggerThreadStart: 1 minute # Second thread starts after 60 seconds
+  #  HardLimits:
+  #    Next: # Note, this refers to the next() operation of a cursor
+  #      Latency: 500 milliseconds
+  #      SLA: 99.5
```
rtimmons: I'm still a bit dubious about this `HardLimits` syntax but I think I grok the `Latency,SLA` pair and how that maps nicely to [TIG-1271](https://jira.mongodb.org/browse/TIG-1271).

src/workloads/scale/LargeScaleSerial.yml
```diff
@@ -0,0 +1,82 @@
+# Load phase of ../BigWorkload.yml
```
rtimmons: this comment isn't accurate anymore 😄

src/driver/test/Driver_test.cpp
```diff
@@ -155,6 +156,7 @@ std::pair<DefaultDriver::OutcomeCode, DefaultDriver::ProgramOptions> outcome(
 
 
 TEST_CASE("Various Actor Behaviors") {
+    boost::filesystem::current_path(genny::findRepoRoot());
```
rtimmons: don't think you need this line

src/driver/test/Driver_test.cpp
```diff
@@ -326,4 +328,124 @@ TEST_CASE("Various Actor Behaviors") {
...
+
+    SECTION("Load External Config Default Parameter") {
+        auto [code, opts] = outcome(R"(
+        SchemaVersion: 2018-07-01
+        Actors:
+          - Type: Fails
+            Threads: 1
+            Phases:
+            - ExternalPhaseConfig:
+                Path: "src/testlib/phases/Good.yml"
```
rtimmons: 💅 this doesn't need to be quoted

src/gennylib/src/context.cpp
```diff
@@ -37,10 +38,15 @@ WorkloadContext::WorkloadContext(const YAML::Node& node,
       _rateLimiters{10},
       _poolManager{mongoUri, apmCallback} {
 
+    std::set<std::string> validSchemaVersions{"2018-07-01", "2019-04-01"};
```
rtimmons: so actually `2019-04-01` isn't a valid schema version for anything except "inner phases" and in fact it's the same as `2018-07-01` but just a subset of it - I wonder if a name like `2018-07-01.Phase` or something might make more sense?

src/phases/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Install all the phase config snippets.
+install(DIRECTORY .
+    DESTINATION etc/genny/phase_configs
+    COMPONENT PhaseConfigs
+    FILES_MATCHING
+    PATTERN "*.yml"
+)
```
rtimmons: newline

src/phases/scale/DesignDocWorkloadPhases.yml
guoyr: Will merge. I'm just going to leave the workloads as is. I don't think the TODOs are finalized at this point so I'll leave it entirely up to the WT team to take over these workloads. The code here serves more as a template.

I'll make this fact clear in my message to Alex C and Luke P clear in my Slack message.

src/phases/scale/DesignDocWorkloadPhases.yml
guoyr: I hear ya. If we don't end up needing to run for more than an hour or so, we can not use this feature.

PR #209, opened 2019-04-08
Dry-run-workloads.sh can guess where the genny binary is

rtimmons:
I got cranky always having to set `GENNY=` before running this script. This lets the script "guess" where it is based on common locations within `build` and `dist`. I printed out which `genny` is being used too. I also killed `set -x` since it really isn't useful.

Tested with manually setting `GENNY` var and building via `lamp -b ninja` and `lamp -b make` and `./scripts/dry-run-workloads.sh` always finds the correct genny.


scripts/dry-run-workloads.sh
```diff
@@ -15,14 +15,25 @@
...
-set -x
 
 SCRIPTS_DIR="$(dirname "${BASH_SOURCE[0]}")"
 ROOT_DIR="$(cd "${SCRIPTS_DIR}/.." && pwd)"
 BUILD_DIR="${ROOT_DIR}/build"
 
+if [[ -z "${GENNY+x}" ]]; then
+    if [ -e "$ROOT_DIR/build/src/driver/genny" ]; then
+        GENNY="$ROOT_DIR/build/src/driver/genny"
+    elif [ -e "$ROOT_DIR/dist/bin/genny" ]; then
```
guoyr: [question] Should the installed location come first?

PR #210, opened 2019-04-08
TIG-1518: Quote lamp args

rtimmons:
This isn't the full scope of the fix for TIG-1518 but it is a step there.


guoyr:
Just ran into this problem when playing with it on Windows so I merged it.

PR #211, opened 2019-04-09
TIG-1271 Allow Specifying Threshold to metrics operations

guoyr:
Ready for preliminary review. This approach was easier for me to implement than templating all the things. But I'm open to other suggestions if the code looks ugly.

The error message from exceeding the threshold still needs to be fleshed out.

### DIFF
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -19,6 +19,8 @@
 #include <type_traits>
 #include <unordered_map>
 
+#include <gennylib/conventions.hpp>
+
 #include <metrics/operation.hpp>
 #include <metrics/passkey.hpp>
 
@@ -86,6 +88,18 @@ public:
         auto& opsByType = this->_ops[actorName];
         auto& opsByThread = opsByType[opName];
         auto opIt = opsByThread.try_emplace(actorId, std::move(actorName), std::move(opName)).first;
+        return OperationT{opIt->second};
+    }
+
+    OperationT<ClockSource> operation(std::string actorName,
+                                      std::string opName,
+                                      ActorId actorId,
+                                      genny::TimeSpec threshold,
+                                      double_t percentage) {
+        auto& opsByType = this->_ops[actorName];
+        auto& opsByThread = opsByType[opName];
+        auto opIt =
+            opsByThread.try_emplace(actorId, std::move(actorName), std::move(opName), {threshold, percentage}).first;
         return OperationT{opIt->second};
     }
 
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -99,7 +99,10 @@ public:
         auto& opsByType = this->_ops[actorName];
         auto& opsByThread = opsByType[opName];
         auto opIt =
-            opsByThread.try_emplace(actorId, std::move(actorName), std::move(opName), {threshold, percentage}).first;
+            opsByThread
+                .try_emplace(
+                    actorId, std::move(actorName), std::move(opName), {threshold, percentage})
+                .first;
         return OperationT{opIt->second};
     }
 
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -16,6 +16,7 @@
 #define HEADER_058638D3_7069_42DC_809F_5DB533FCFBA3_INCLUDED
 
 #include <chrono>
+#include <optional>
 #include <type_traits>
 #include <unordered_map>
 
@@ -101,7 +102,11 @@ public:
         auto opIt =
             opsByThread
                 .try_emplace(
-                    actorId, std::move(actorName), std::move(opName), {threshold, percentage})
+                    actorId,
+                    std::move(actorName),
+                    std::move(opName),
+                    std::make_optional<typename OperationImpl<ClockSource>::OperationThreshold>(
+                        threshold, percentage))
                 .first;
         return OperationT{opIt->second};
     }
--- a/src/metrics/include/metrics/operation.hpp
+++ b/src/metrics/include/metrics/operation.hpp
@@ -91,33 +91,40 @@ struct OperationEvent final {
     OutcomeType outcome = OutcomeType::kUnknown;  // corresponds to the 'outcome' field in Cedar
 };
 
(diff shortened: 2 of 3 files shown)


src/metrics/include/metrics/operation.hpp
```diff
@@ -123,13 +147,24 @@ class OperationImpl final : private boost::noncopyable {
         return _events;
     }
 
-    void reportAt(time_point finished, OperationEvent<ClockSource>&& event) {
+    void reportAt(time_point started, time_point finished, OperationEvent<ClockSource>&& event) {
+        if (_threshold) {
```
guoyr: This is the only extra `if` statement that happens at runtime. I thought it was better to take this hit than to template all the way up to the metrics reporter.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -89,6 +92,25 @@ class RegistryT final {
...
+            opsByThread
+                .try_emplace(
+                    actorId,
+                    std::move(actorName),
+                    std::move(opName),
+                    std::make_optional<typename OperationImpl<ClockSource>::OperationThreshold>(
+                        threshold, percentage))
+                .first;
+        return OperationT{opIt->second};
+    }
```
rtimmons: 1. this is a little copy/pasta from above is there a way to DRY it up?
2. the spacing looks almost a  bit lisp-y is this really how clang-format leaves it?

src/metrics/include/metrics/operation.hpp
```diff
@@ -92,15 +91,40 @@ struct OperationEvent final {
     OutcomeType outcome = OutcomeType::kUnknown;  // corresponds to the 'outcome' field in Cedar
 };
 
+class OperationThresholdExceededException : public virtual boost::exception,
```
rtimmons: please document what this means (although the name is a bit self-explanatory)

src/metrics/include/metrics/operation.hpp
```diff
@@ -92,15 +91,40 @@ struct OperationEvent final {
...
 public:
     using time_point = typename ClockSource::time_point;
     using EventSeries = TimeSeries<ClockSource, OperationEvent<ClockSource>>;
 
-    OperationImpl(std::string actorName, std::string opName)
-        : _actorName(std::move(actorName)), _opName(std::move(opName)) {}
+    struct OperationThreshold {
+        std::chrono::nanoseconds duration;
+        double_t failedPercentageThreshold;
+        OperationCount counter;
```
rtimmons: 1. Rename `duration` to `maxDuration`?
1. Rename counter to `occurrences` or `seenEvents`?
1. Rename `failedPercentageThreshold` to `percentAllowedToExceed` or something?

src/metrics/include/metrics/operation.hpp
```diff
@@ -123,13 +147,24 @@ class OperationImpl final : private boost::noncopyable {
...
+        if (_threshold) {
+            _threshold->counter.total++;
+            if ((finished - started) > _threshold->duration) {
+                _threshold->counter.failed++;
+            }
+
+            if (_threshold->counter.failedPercentage() > _threshold->failedPercentageThreshold) {
+                throw OperationThresholdExceededException();
+            }
+        }
```
rtimmons: imho this logic makes more sense as a method on the `OperationThreshold`:

```c++
if (threshold) { threshold->check(finished,started); }
```

This may also make it easier to test the `OperationThreshold` impl.

src/metrics/test/metrics_test.cpp
```diff
@@ -369,5 +369,27 @@ TEST_CASE("Genny.ActiveActors metric") {
     }
 }
 
+TEST_CASE("Operation with threshold") {
+    RegistryClockSourceStub::reset();
+    auto metrics = v1::RegistryT<RegistryClockSourceStub>{};
+
+    genny::metrics::v1::ReporterT{metrics};
+
+    // Create an actor where 50% of the operations cannot exceed 10ns.
```
rtimmons: maybe also a test where threshold is 0 or 100?

src/metrics/include/metrics/metrics.hpp
vrachev: Agree with both points, although I'm not sure if you can DRY it up cleanly. Would be nice if you could either have one method with default params or use method chaining, but that might require having making the new params be a single `std::Optional<genny::TimeSpec, double_t>` which would clash with your  `std::optional<OperationThreshold>` type. Seems like it might be most legible to keep it as it is unless there's some other c++ voodoo you know of?

src/metrics/include/metrics/operation.hpp
```diff
@@ -123,13 +147,24 @@ class OperationImpl final : private boost::noncopyable {
...
-    void reportAt(time_point finished, OperationEvent<ClockSource>&& event) {
+    void reportAt(time_point started, time_point finished, OperationEvent<ClockSource>&& event) {
+        if (_threshold) {
+            _threshold->counter.total++;
+            if ((finished - started) > _threshold->duration) {
+                _threshold->counter.failed++;
+            }
+
+            if (_threshold->counter.failedPercentage() > _threshold->failedPercentageThreshold) {
+                throw OperationThresholdExceededException();
```
vrachev: Do I understand correctly that the threshold is checked every time an operation is reported?  Could this be problematic when operations first start running? For any threshold, if the first operation exceeds the `maxDuration` it would fail. I think it would make sense to gather more data before checked if the failPercentage is too great, to prevent giving a huge weight to the first operations.

src/metrics/include/metrics/operation.hpp
vrachev: +1

src/metrics/test/metrics_test.cpp
vrachev: Could also add a test with the first operations being slow ones. This would throw or not throw depending on what you think about my other comment about checking `> failedPercentageThreshold`.

src/metrics/include/metrics/operation.hpp
vrachev: +1 for `maxDuration`. 

I like `counter` or `opCounter` better. 

+1 on `percentAllowedToExceed` or `maxPercentAllowedToExceed`

src/metrics/include/metrics/operation.hpp
vrachev: 👍

PR #214, opened 2019-04-11
TIG-1601: Kill non-idiomatic config namespace

rtimmons:
Few grab-bag of things here that were driving me nuts:

1. Rename `ExecutionStrategy` to `RetryStrategy` (this shows as the bulk of the diff)
1. Don't have `genny::config::X` types separate from `genny::X`. This applied to `RunCommandOptions` and `ExecutionStrategyOptions` (no renamed to `RetryStrategyOptions`). This was an idea originally borrowed from the C++ driver which has `mongocxx::options` but it seems silly and it only increases the number of places where functionality lives
1. Remove non-idomatic `ExecutionStrategyOptions::getOptionsFrom(Context)` in favor of just `context.get<Options,false>(key).value_or(....)`. This is a bit more verbose but removes a non-idomatic way of extracting config values
1. Scaffolded `create-new-actor` Actors don't use RetryStrategy. I think its interface is a bit tough for new Actor-authors to absorb (I've received this feedback from a few people).


scripts/create-new-actor.sh
```diff
@@ -137,35 +139,31 @@ create_impl_text() {
...
-    echo "                    // TODO: main logic"
-    echo "                    auto document = config->documentExpr();"
-    echo "                    BOOST_LOG_TRIVIAL(info) << \" ${actor_name} Inserting \""
-    echo "                                            << bsoncxx::to_json(document.view());"
-    echo "                    config->collection.insert_one(document.view());"
-    echo "                    ctx.addDocuments(1);"
-    echo "                    ctx.addBytes(document.view().length());"
-    echo "                },"
-    echo "                config->options);"
+    echo "            auto inserts = _inserts.start();"
```
guoyr: Do you think `s/_inserts/_totalInserts` would be more intuitive? I fear the lack of anything mentioning "metrics" in this file could lead people down a rat hole of poking around in `context.operation()`.

PR #215, opened 2019-04-12
TIG-1397: Crud Actor 💅

rtimmons:
Few light housekeeping things before I add more functionality:

1. switch to `BOOST_THROW_EXCEPTION`
2. `std::move` in operation ctor params
3. Record documents and bytes metrics where appropriate

Sys-perf patch: https://evergreen.mongodb.com/version/5cb0d1c50ae606414132d48b

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -617,8 +617,10 @@ struct CountDocumentsOperation : public BaseOperation {
     void run(mongocxx::client_session& session) override {
         auto filter = _filterExpr();
         auto ctx = _operation.start();
-        (_onSession) ? _collection.count_documents(session, std::move(filter), _options)
-                     : _collection.count_documents(std::move(filter), _options);
+        auto count = (_onSession)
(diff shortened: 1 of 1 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -1,3 +1,5 @@
+#include <utility>
```
visemet: It looks like this include got added to the wrong place.

rtimmons:
Thanks for review.

> Should we call ctx.addDocuments() with the return value of _collection.count_documents()?

Was this in reference to a particular line? GH didn't attach it anywhere....

visemet:
> > Should we call ctx.addDocuments() with the return value of _collection.count_documents()?
> 
> Was this in reference to a particular line? GH didn't attach it anywhere....

It's in reference to `CountDocumentsOperation::run()`. GitHub only allows you to comment on a few lines surrounding the diff. You didn't change it so I couldn't comment on it 🙃

rtimmons:
Updated per feedback - ready for 👀 when you can

PR #217, opened 2019-04-15
[nomerge] Move CrudActor tests to yaml

rtimmons:
Still WIP but wanted feedback before I continue with this. I estimate it'll be another 2-3 hours of work but it'll pay off immensely.

I've moved 11 out of about 30 tests over to yaml and each test goes from about 50 lines to about 15 and the yaml is much easier to read and the error messages are much nicer.

Any feedback welcome but in particular I'd like input on the yaml syntax which I based lightly off the driver specs but deviated from for reasons that I think make sense.


src/testlib/include/testlib/helpers.hpp
```diff
@@ -40,6 +43,21 @@ inline std::string toString(const YAML::Node& node) {
     return std::string{out.c_str()};
 }
 
+inline std::string toString(int i) {
+    return std::to_string(i);
+}
+
+// TODO: this is duped with MongoTestFixture
```
guoyr: `MongoTestFixture` didn't work out very well. The mongo client really wants to be part of the actor and not the fixture. We should probably kill it and move `static makeApmCallback()` and `static connectionUri()` to a helper library.

Helper libraries seem to play nicer since we're using catch2's `TEST_CASE` instead of `TEST_CASE_METHOD` style of testing.

src/cast_core/test/CrudActorYamlTests.yaml
```diff
@@ -0,0 +1,185 @@
...
+              Filter: { a: { $gte: 2 } }
+              Update: { $set: { a: 8 } }
+          Options:
+            WriteConcern:
+              Level: majority
+              Timeout: 6000 milliseconds
+    OutcomeCounts:
+      - Filter: {a: 8}
+        Count: 1
+    ExpectedEvents:
```
guoyr: Should this be a list of all events or just that there exists an "event" in all events where "event" is in the list of  `ExpectedEvents`?

src/testlib/include/testlib/helpers.hpp
rtimmons: Ya I've done that in the real review - coming up!

src/cast_core/test/CrudActorYamlTests.yaml
rtimmons: bug in my own understanding - tests want to assert facts about _all_ callback events. Latest version has it.

PR #218, opened 2019-04-16
TIG-1098 Add standalone benchmark binary to measure genny perf impact

guoyr:
Most of what's described in TIG-1098 is covered. `db.ping()` is missing, but I'm planning to do it as part of TIG-1337 now that the ticket has been reopened. I'll probably also file a ticket to integrate the canaries with DSI. 

Right now the output is in a human readable format:

```bash
➜  genny git:(tig1098) ./dist/bin/genny-canaries "simple-loop" "phase-loop" "metrics-loop" "real-loop" -p false
simple-loop:	2143012ns
phase-loop:	50859249ns
metrics-loop:	192317833ns
real-loop:	219668218ns
➜  genny git:(tig1098) ./dist/bin/genny-canaries "simple-loop" "phase-loop" "metrics-loop" "real-loop" -p false
simple-loop:	2108709ns
phase-loop:	51095903ns
metrics-loop:	194208899ns
real-loop:	220928299ns
➜  genny git:(tig1098) ./dist/bin/genny-canaries "simple-loop" "phase-loop" "metrics-loop" "real-loop" -p false
simple-loop:	2282825ns
phase-loop:	51424344ns
metrics-loop:	192888125ns
real-loop:	219086235ns
```

### DIFF
--- a/src/canaries/include/canaries/Loops.hpp
+++ b/src/canaries/include/canaries/Loops.hpp
@@ -15,11 +15,8 @@
 #ifndef HEADER_C8D12C92_7C08_4B7F_9857_8B25F7258BB9_INCLUDED
 #define HEADER_C8D12C92_7C08_4B7F_9857_8B25F7258BB9_INCLUDED
 
-#include <iostream>
 
-#include <gennylib/PhaseLoop.hpp>
-#include <gennylib/context.hpp>
-
+#include <cstdint>
 
 namespace genny::canaries {
 
--- a/src/canaries/include/canaries/Loops.hpp
+++ b/src/canaries/include/canaries/Loops.hpp
@@ -1,0 +1,36 @@
+// Copyright 2019-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+
+
+namespace genny::canaries {
+
+template <bool WithPing>
+class Loops {
+public:
+    Loops() = default;
+
+    int64_t simpleLoop(int64_t iterations);
+
+    int64_t phaseLoop(int64_t iterations);
+
+    int64_t metricsLoop(int64_t iterations);
+
+    int64_t metricsPhaseLoop(int64_t iterations);
+};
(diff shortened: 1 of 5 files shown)


src/canaries/src/Loops.cpp
```diff
@@ -0,0 +1,160 @@
...
+    return after - before;
+}
+
+template <bool WithPing>
+int64_t Loops<WithPing>::metricsLoop() {
+
+    auto metrics = genny::metrics::Registry{};
+    metrics::Reporter reporter(metrics);
+
+    auto dummyOp = metrics.operation("metricsLoop", WithPing ? "db.ping()" : "Nop", 0u);
```
guoyr: There's a requirement in the ticket to report the result of this metric:
> Report the added mean and variance latency added within #2, #3, and #4

I'm wondering if this is still something we need.

We can get a sense of the mean by comparing the `phase-loop` and `real-loop` benchmarks. `variance` is likely very high for no-op loops because of CPU scheduling. It's likely more consistent but noisy for heavy loops like `db.ping()`, because there's no way to distinguish if the variance is coming from `db.ping()` or metrics internals.

src/canaries/include/canaries/Loops.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Run a basic for-loop and record one timer metric per loop.
+     */
+    int64_t metricsLoop();
+
+    /**
+     * Run the phase loop and record one timer metric per loop.
+     */
+    int64_t metricsPhaseLoop();
```
rtimmons: please document what the return value is for all of these

src/canaries/include/canaries/Loops.hpp
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Class for benchmarking Genny internals. It's recommended to run
+ * these benchmarks as canaries before running Genny workloads.
+ *
+ * @tparam WithPing whether to issue db.ping() to a mongo cluster or just
+ *                  do a simple increment. The former is useful for gauging
+ *                  the overhead of Genny and the latter is useful for seeing
+ *                  how this overhead changes over time.
+ */
+template <bool WithPing>
```
rtimmons: bikeshedding here but we could also maybe make `iterations` a template-parameter - I think we can know at compile-time how many to do?

src/canaries/src/Loops.cpp
rtimmons: Seems fine to drop this requirement for now

src/canaries/include/canaries/Loops.hpp
guoyr: typedef'd to `Nanosecond`

guoyr:
> This looks good but how do you envision this being used?

Run them as part of all genny DSI tasks? I think having these output to the same metrics format as genny so the metrics parsing scripts can be reused.

> Does it become another test type in dsi?

Hopefully the existing `GennyRunner` in DSI can be modified to run the canaries as well.

guoyr:
I filed https://jira.mongodb.org/browse/TIG-1614 for followup work to 

1. report the canaries in Genny's csv2 format
2. integrate with DSI.

Let me know if you have concerns or other ideas.

PR #219, opened 2019-04-16
TIG-1397: Test CrudActor with YAML

rtimmons:
Based on #217 (nomerge). Converted all the boilerplate C++ Crud Actor tests to yaml.


src/cast_core/test/CrudActorYamlTestRunner.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+    return false;
+}
+
+// This isn't ideal. We want to assert only a subset of the keys in the ApmEvents
+// that we receive. Doing a subsetCompare(YAML::Node, bsoncxx::document::view) is nontrivial.
+// (Mostly fighting the compiler with yaml scalars versus the many numeric bson types.)
+// For now just assert a subset
+void requireEvent(ApmEvent& event, YAML::Node requirements) {
+    if (auto w = requirements["writeConcern"]["w"]; w) {
```
guoyr: I think ideally we'd remove the keys we've read and assert that there are no keys remaining in `requirements`. Not sure if usability matters that much though since this is all internal code.

src/testlib/include/testlib/MongoTestFixture.hpp
```diff
@@ -20,6 +20,24 @@
 
 namespace genny::testing {
 
+// TODO: move to separate class/helper
+// This was copied from
+// github.com/mongodb/mongo-cxx-driver/blob/r3.4.0/src/mongocxx/test/client_session.cpp#L285
+class ApmEvent {
```
guoyr: 👍

src/testlib/include/testlib/helpers.hpp
```diff
@@ -40,6 +43,21 @@ inline std::string toString(const YAML::Node& node) {
     return std::string{out.c_str()};
 }
 
+inline std::string toString(int i) {
+    return std::to_string(i);
+}
+
+// TODO: this is duped with MongoTestFixture
+inline void dropAllDatabases(mongocxx::pool::entry& client) {
```
guoyr: Could we have the fixture's function just call this?

src/testlib/include/testlib/yamlTest.hpp
```diff
@@ -18,8 +18,13 @@
 #include <string>
 #include <vector>
 
+#include <boost/exception/diagnostic_information.hpp>
+
 #include <yaml-cpp/yaml.h>
 
+#include <testlib/findRepoRoot.hpp>
```
guoyr: Are these two headers needed?

src/cast_core/test/CrudActorYamlTestRunner.cpp
rtimmons: Ya I considered that but I didn't want to modify the input node (YAML::Node is basically a shared_ptr).

src/testlib/include/testlib/helpers.hpp
rtimmons: Good call - took a bit of template magic because they're of different types but it ended up working.

src/testlib/include/testlib/yamlTest.hpp
rtimmons: Ya the functions they export are used further down - didn't catch it before because the valgen yaml tests were including them on their own (but hey now I can remove those includes from there so did that as well)

PR #220, opened 2019-04-16
 TIG-1611 allow per-phase metrics and convert the commitlatency actor to use it

guoyr:


src/workloads/networking/CommitLatency.yml
```diff
@@ -63,214 +63,98 @@ Actors:
...
       Level: 0
     ReadConcern:
       Level: local
     ReadPreference:
       ReadMode: primary
-  - Phase: 3..32
-    Operation:
-      OperationName: Nop
-
-- Name: CommitLatency_w1_local
```
guoyr: I think we should allow phases to have names instead of appending just the phase number. In this case, the phase number can be `w1_local`. The metric name would then be: `[actor-name] << "." << [phase-name]`. Phases with the same name can still have their metrics be grouped together.

src/gennylib/include/gennylib/context.hpp
```diff
@@ -485,10 +487,11 @@ class ActorContext final : public v1::ConfigNode {
  * Represents each `Phase:` block in the YAML configuration.
  */
 class PhaseContext final : public v1::ConfigNode {
-
 public:
-    PhaseContext(const YAML::Node& node, ActorContext& actorContext)
-        : ConfigNode(node, std::addressof(actorContext)), _actor{std::addressof(actorContext)} {}
+    PhaseContext(const YAML::Node& node, int phaseNumber, ActorContext& actorContext)
```
rtimmons: `PhaseNumber` instead of `int`

src/workloads/networking/CommitLatency.yml
rtimmons: > I think we should allow phases to have names instead of appending just the phase number. 

Feel free to add to [Genny Ideas](https://docs.google.com/document/d/1K0SlHoKkIACIXhkNP_KPHvb3IV6r53ReqsHqEYTnXDA/edit#) but each Actor may think of each Phase differently. 

I think the recommendation is for an Actor to be configurable with the metrics name for the Phase:

```yaml
Actors:
- Type: MyActor
  Phases:
  - Phase: 0
    MetricsName: Insert
```

src/gennylib/include/gennylib/context.hpp
```diff
@@ -518,11 +521,26 @@ class PhaseContext final : public v1::ConfigNode {
         return _actor->workload();
     }
 
+    /**
+     * Convenience method for creating a metrics::Operation that's unique for this phase and thread.
+     *
+     * @param operationName the name of the operation being run.
+     * @param id the id of this Actor, if any.
```
rtimmons: I think kill the `if any` since it's not optional

src/gennylib/include/gennylib/context.hpp
```diff
@@ -518,11 +521,26 @@ class PhaseContext final : public v1::ConfigNode {
         return _actor->workload();
     }
 
+    /**
+     * Convenience method for creating a metrics::Operation that's unique for this phase and thread.
```
rtimmons: may be worth noting that if the `Phase:` block has a `Name:` param that will be used else it will inherit?

src/gennylib/include/gennylib/context.hpp
```diff
@@ -518,11 +521,26 @@ class PhaseContext final : public v1::ConfigNode {
...
+            this->get<std::string>("Name"), stm.str(), id);
+    }
+
+
 private:
     bool _isNop() const;
 
 private:
     ActorContext* _actor;
+    const int _phaseNumber;
```
rtimmons: `PhaseNumber`

src/workloads/networking/CommitLatency.yml
```diff
@@ -63,214 +63,98 @@ Actors:
 
 
 
-# All of the below is really just the same actor, but as Genny reports statistics using the actor
-# name, we need a different actor instance for each configuration. All of the below run in their
-# own slot, nothing is concurrent.
-# Since Drop+Reload is in phases 0,1,-,3,4,-... then the test itself must always be in a slot at
-# 2,5,8... = 3n-1 = 3n+2
-- Name: CommitLatency_w0_local
+- Name: CommitLatency
```
rtimmons: I got kinda dizzy trying to reason if this workload yaml ends up being equivalent to the existing one so I'm going to trust you there (or perhaps ask Henrik to take a peek).

src/cast_core/src/CommitLatency.cpp
```diff
@@ -110,80 +109,76 @@ void CommitLatency::run() {
...
-                            ctx.addErrors(1);
-                        }
-                    }
-
-                    if (config->useTransaction) {
-                        _session->commit_transaction();
-                    }
-                },
-                config->options);
+            auto ctx = config->op.start();
```
rtimmons: this is so much more obvious!

guoyr:
@henrikingo I added per-phase metrics collection in this PR and `git revert`ed #203. It was a clean revert. I also added the names of the metrics under a new key `MetricsName`.

Here're the updated workload files:
`CommitLatency.yml`: https://github.com/mongodb/genny/blob/8c1c9df862fe1856041b813168f7812faee43478/src/workloads/networking/CommitLatency.yml
`CommitLatencySingleUpdate.yml`: https://github.com/mongodb/genny/blob/8c1c9df862fe1856041b813168f7812faee43478/src/workloads/networking/CommitLatencySingleUpdate.yml

If you want to look them over and have any concerns/suggestions/questions. Let me know!

PR #221, opened 2019-04-16
TIG-1613 remove RetryStrategy

guoyr:


src/cast_core/test/CrudActor_test.cpp
```diff
@@ -45,8 +45,6 @@ TEST_CASE_METHOD(MongoTestFixture,
       - Name: CrudActor
         Type: CrudActor
         Database: mydb
-        RetryStrategy:
```
guoyr: There will be merge conflicts with #219. I can take of them. (Just need to bulk remove lines containing `RetryStrategy` and `ThrowOnFailure` in this file.

rtimmons: May the odds be ever in your favor
(But if you end up merging first that's fine it's an easy conflict)

guoyr: Well it was pretty much the easiest merge conflict. `RetryStategy` doesn't appear at all in the Yaml test so I just had to remove it once from the runner test CPP file.

guoyr:
sys-perf patch build: https://evergreen.mongodb.com/build/sys_perf_linux_3_node_replSet_patch_19b622ebfb42a525f38e278c09f440eb47b12f1e_5cb54131d1fe0745edb51604_19_04_16_02_44_45

There's a universal slowdown of about 20% starting around April 15th. But is unrelated to this PR.

PR #222, opened 2019-04-17
TIG-1612 Add an option to convert workloads for smoke testing

guoyr:


src/driver/include/driver/DefaultDriver.hpp
```diff
@@ -64,6 +64,7 @@ class DefaultDriver {
         std::string metricsOutputFileName;
         std::string mongoUri;
         std::string description;
+        bool isSmokeTest;
```
guoyr: I decided not to make smoke test a run mode so people can use `evaluate` to see what the smoke test does.

It's possible for the smoke test to still take a lot of time if a single iteration of a phase takes a long time to run; so people may still want to manually modify their workload configs to run the smoke test.

src/driver/include/driver/WorkloadParser.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+    YAML::Node loadFile(const std::string& path);
+    YAML::Node recursiveParse(YAML::Node, ParseMode parseMode);
+
+    // Handle external configuration files.
+    void convertExternal(std::string key, YAML::Node value, YAML::Node& out);
+    YAML::Node parseExternal(YAML::Node);
+    YAML::Node replaceParam(YAML::Node);
+
+    // Handle smoke test mode.
```
guoyr: These two groups would probably be organized as two subclasses in retrospect 🤕  (I was hoping to have one iteration of `parse()` do both conversions but didn't work out)

src/phases/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 # Install all the phase config snippets.
 install(DIRECTORY .
-    DESTINATION etc/genny/phase_configs
+    DESTINATION etc/genny/phases
```
guoyr: fixing this bug from a previous ticket.

src/workloads/docs/HelloWorld.yml
```diff
@@ -2,19 +2,19 @@ SchemaVersion: 2018-07-01
...
-      # SleepBefore: 11 milliseconds
-      # SleepAfter: 17 microseconds
-    - Message: Hello Phase 1 👬
-      Repeat: 100
-    - ExternalPhaseConfig:
-        Path: ../../phases/HelloWorld/ExamplePhase2.yml
-        Key: UseMe  # Only load the YAML structure from this top-level key.
-        Parameters:
-          Repeat: 2
+- Name: HelloWorld
```
guoyr: Drive by change to make it consistent with Genny's YAML style guide.

src/driver/include/driver/WorkloadParser.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <yaml-cpp/yaml.h>
+
+#include <driver/DefaultDriver.hpp>
+
+namespace genny::driver {
+
+using YamlParameters = std::map<std::string, YAML::Node>;
+namespace fs = boost::filesystem;
+
+class WorkloadParser {
```
guoyr: 🤓 The first commit in this PR is to move the YAML parsing code over to this file.

src/driver/include/driver/DefaultDriver.hpp
rtimmons: > I decided not to make smoke test a run mode so people can use evaluate to see what the smoke test does.

Should it instead be a flag to `genny evaluate`? E.g. `genny evaluate --smoke-test`?

That seems reasonable but I think we'd want to change the current dry-run CI task to actually do `smoke-test` (or perhaps keep it and add an additional task). The one gotcha would be that workload YAMLs would need to specify their topology requirements but that's probably a good thing to have anyway. 

So it's fine to leave it off as a "run mode" but hopefully it remains a simple shell operation to `genny dry-run > workload.yml && genny ./workload.yml`. I think it does but just to be sure.

src/driver/include/driver/WorkloadParser.hpp
rtimmons: 👍 This is a good separation-of-concerns. If/when #225 lands this may make sense in a separate `config` CMake module (project?). Fine to leave in driver for now.

src/driver/include/driver/WorkloadParser.hpp
```diff
@@ -0,0 +1,62 @@
...
+public:
+    explicit WorkloadParser(fs::path& phaseConfigPath, bool isSmokeTest = false)
+        : _isSmokeTest{isSmokeTest}, _phaseConfigPath{phaseConfigPath} {};
+
+    YAML::Node parse(const std::string& source,
+                     DefaultDriver::ProgramOptions::YamlSource =
+                         DefaultDriver::ProgramOptions::YamlSource::kFile);
+
```
rtimmons: since these aren't in `v1` they need documentation.

src/driver/include/driver/WorkloadParser.hpp
```diff
@@ -0,0 +1,62 @@
...
+    const bool _isSmokeTest;
+
+    YAML::Node loadFile(const std::string& path);
+    YAML::Node recursiveParse(YAML::Node, ParseMode parseMode);
+
+    // Handle external configuration files.
+    void convertExternal(std::string key, YAML::Node value, YAML::Node& out);
+    YAML::Node parseExternal(YAML::Node);
+    YAML::Node replaceParam(YAML::Node);
```
rtimmons: if these are private do they actually need to be members of the class at all or can they just be within the .cpp?

src/driver/include/driver/WorkloadParser.hpp
```diff
@@ -0,0 +1,62 @@
...
+namespace fs = boost::filesystem;
+
+class WorkloadParser {
+public:
+    explicit WorkloadParser(fs::path& phaseConfigPath, bool isSmokeTest = false)
+        : _isSmokeTest{isSmokeTest}, _phaseConfigPath{phaseConfigPath} {};
+
+    YAML::Node parse(const std::string& source,
+                     DefaultDriver::ProgramOptions::YamlSource =
+                         DefaultDriver::ProgramOptions::YamlSource::kFile);
```
rtimmons: this can prolly be const

src/driver/include/driver/WorkloadParser.hpp
rtimmons: this seems fine as long as there's some documentation :trollface:

src/driver/src/DefaultDriver.cpp
```diff
@@ -241,9 +100,9 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
         phaseConfigSource = fs::path(options.workloadSource).parent_path();
     }
 
-    YamlParameters params;
-    auto config = loadConfig(options.workloadSource, options.workloadSourceType);
-    auto yaml = recursiveParse(config, params, phaseConfigSource);
+    WorkloadParser parser{phaseConfigSource, options.isSmokeTest};
+
+    auto yaml = parser.parse(options.workloadSource, options.workloadSourceType);
```
rtimmons: instead of peeling off 2 things from `options` does `parser.parse()` want to just take in the whole `options`?

src/driver/include/driver/WorkloadParser.hpp
rtimmons: at the risk of over-engineering it feels funny that the parser knows about things from the `DefaultDriver`. It seems like `Parser` wants its own `SourceType` enum that `DefaultDriver` can convert to. This is intentionally adding some more layers of indirection so it's subjective - feel free to disregard or add a comment to maybe do it later.

src/driver/src/WorkloadParser.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+void WorkloadParser::convertToSmokeTest(std::string key, YAML::Node value, YAML::Node& out) {
+    if (key == "Duration" || key == "Repeat") {
+        out["Repeat"] = 1;
+    } else if (key == "Rate" || key == "SleepBefore" || key == "SleepAfter") {
+        // Ignore those keys in smoke tests.
+    } else {
+        out[key] = recursiveParse(value, ParseMode::kSmokeTest);
```
rtimmons: Since this applies at all depths of the traversal, would this prevent something like this?

```yaml
Actors:
- Type: MyActor
  Phases:
  - Document: { Duration: 7 }
```

↑ is contrived but not unreasonable...I think the `convertToSmokeTest` may need to only be applied to top-level, `Actor`, and `Phase` blocks or something.

It also doesn't seem unreasonable to require these "reserved" keys to be placed inside `{^Verbatim}` blocks and `recursiveParse` could change the `ParseMode` if it sees that. The syntax would be funny:

```yaml
Actors:
- Type: MyActor
  Phases:
  - Document: {^Verbatim: { Duration: 7 }}
  # or maybe this?
  - Document: {Duration: {^Verbatim: 7}}
```

but we'd also want a way to document or enforce that somehow I think.

(Also tbh this isn't a *huge* deal since dry-run is really just to catch bizarro mis-configurations and the likelihood of an Actor using these "keywords" in its own configuration is admittedly low.)

src/driver/test/WorkloadParser_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const auto input = (R"(
+Actors:
+- Name: WorkloadParserTest
+  Type: NonExistent
+  Threads: 2.718281828   # This field is ignored for the purpose of this test.
+  Phases:
+  - Duration: 4 scores            # Removed
+    Repeat: 1e999                 # Replaced with "1"
+    Rate: 1 per 2 megannum        # Removed
+    SleepBefore: 1 per 2 planks   # Removed
```
rtimmons: lol nice

src/phases/CMakeLists.txt
rtimmons: (separately I was thinking that `phases` may want to be a subdir of `workloads`...I added to the ideas doc)

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -0,0 +1,63 @@
...
+public:
+    explicit WorkloadParser(fs::path& phaseConfigPath)
+        : _phaseConfigPath{phaseConfigPath} {};
+
+    YAML::Node parse(const std::string& source,
+                     const DefaultDriver::ProgramOptions::YamlSource = DefaultDriver::ProgramOptions::YamlSource::kFile);
+
+private:
+    YamlParameters _params;
+    const fs::path _phaseConfigPath;
```
rtimmons: can this be a ref and/or should ctor param be const-ref?

src/driver/src/DefaultDriver.cpp
rtimmons: 👍 please put this comment in there or somesuch - `// consider passing in whole options struct if we pass in more than 2-3 fields`

src/driver/src/DefaultDriver.cpp
```diff
@@ -241,9 +100,13 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
...
 
-    YamlParameters params;
-    auto config = loadConfig(options.workloadSource, options.workloadSourceType);
-    auto yaml = recursiveParse(config, params, phaseConfigSource);
+    v1::WorkloadParser parser{phaseConfigSource};
+    auto yaml = parser.parse(options.workloadSource, options.workloadSourceType);
+
+    if (options.isSmokeTest) {
+        yaml = v1::SmokeTestConverter::convert(yaml);
+    }
```
rtimmons: hm...seems like the parser should be able to know how to do all the parsing - e.g. pass in `options.workloadSource, options.workloadSourceType, options.isSmokeTest` and have it return the fully-parsed thing?

src/driver/src/workload_parsers.cpp
```diff
@@ -0,0 +1,209 @@
...
+        out[key] = recursiveParse(value);
+    }
+}
+
+YAML::Node WorkloadParser::parseExternal(YAML::Node external) {
+    int keysSeen = 0;
+
+    if (!external["Path"]) {
+        throw InvalidConfigurationException(
+            "Missing the `Path` to-level key in your external phase configuration");
```
rtimmons: maybe also dump the `external` node to help debugability

src/driver/test/workload_parsers_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+)");
+
+    const auto expected = (R"(Actors:
+  - Name: WorkloadParserTest
+    Type: NonExistent
+    Threads: 2.718281828
+    Phases:
+      - Repeat: 1)");
+
+    auto cwd = boost::filesystem::current_path();
```
rtimmons: plz also add a test that nested `Repeat`/`Duration`/etc keys aren't replaced:

E.g. this shouldn't be modified: 

```yaml
Actors:
- Type: Foo
  Phases:
  - Foo: { Duration: 100 }
```

src/driver/include/driver/workload_parsers.hpp
guoyr: I feel like this class should own its path so I'm changing the param to a const ref, (but actually a value + `std::move` because clang tidy suggested it, probably for the simpler function signature?)

src/driver/src/DefaultDriver.cpp
guoyr: Yeah good idea.

src/driver/test/workload_parsers_test.cpp
guoyr: I reused the existing test with an additional key on the actor and on the phase

PR #223, opened 2019-04-17
TIG-1397: Remaining CrudActor functionality (*)

rtimmons:
This PR gets about 80% of the way through TIG-1397 and will punt the other 20% to the ideas doc or another ticket.

- try/catch is now handled gracefully and each operation reports a meaningful error message as much as it can.
- (Had to create a custom Catch2 Matcher to support asserting the error message now because they're multiline and apparently `std::regex` doesn't have a multiline mode lol)
- added high-level testing for all existing operations including start/commit transaction. I don't assert any pathological cases just assert that calling them doesn't barf.
- I added `find_one_and_update` and working on `find_one_and_delete`, `find_one_and_replace`, and `estimated_document_count` now (they're simple).

I decided *against* adding the ability to set the `mongocxx::options:*` types via yaml. There are a number of reasons:

1. It's a pretty tedious and error-prone `convert<>` impl for each one. I have some perl that automates it kinda but even then...

2. Many of the `options` types take `view_or_value` as fields. (e.g. even [`create_view`](http://mongocxx.org/api/mongocxx-3.4.0/classmongocxx_1_1options_1_1create__view.html) wants a `view_or_value` for its `collation`). We do have a `yamlToBson` thingy but that's only for testing (I don't think it's production-worthy).

3. I think the more correct thing to do is allow all of the fields to be set using ValueGenerators.

    E.g.

    ```yaml
    - Operations:
      - OperationName: createView
        OperationOptions:
          collation: 
            locale: {^RandomString: ...}
          write_concern:
            journal: {^RandomBool}
    ```

    This will require 

    1. adding scalar and array value-generators *and* 
    2. some intermediary classes that can be constructed from a `(YAML::Node,PhaseContext)` that convert to these `options*` type so the `options*` types are constructed at workload-time as are the documents/filters/etc.

If no objections to this, I'll create a ticket to support randomly-generated `mongocxx::options*` types on CrudActor. I left the existing conversions there.

I also decided against having a nested `withTransaction`-style syntax. It got kinda messy and I think if you want nested blocks like this you probably just want to write your own Actor C++. I'll add this to the Genny Ideas doc as wontfix. Similar for setting read concern at the Phase level.


src/cast_core/src/CrudActor.cpp
```diff
@@ -239,16 +272,58 @@ auto createGenerator(YAML::Node source,
 
 }  // namespace
 
+struct CreateIndexOperation : public BaseOperation {
+    mongocxx::collection _collection;
+    metrics::Operation _operation;
+    DocumentGenerator _keysGenerator;
+    DocumentGenerator _indexOptionsGenerator;
+    mongocxx::options::index_view _operationOptions;
+    // TODO: convert from yaml
```
guoyr: Remove this TODO?

src/cast_core/src/CrudActor.cpp
```diff
@@ -239,16 +272,58 @@ auto createGenerator(YAML::Node source,
...
+                         bool onSession,
+                         mongocxx::collection collection,
+                         metrics::Operation operation,
+                         PhaseContext& context,
+                         ActorId id)
+        : BaseOperation(context, opNode),
+          _collection(std::move(collection)),
+          _operation{operation},
+          _onSession{onSession},
+          _keysGenerator{createGenerator(opNode, "createIndex", "Keys", context, id)},
```
guoyr: This might just be me being petulant.... `createGenerator` makes me thinking we're generating the "create" command because `keysGenerator` generates... "keys" XD. Would `createDocumentGenerator` be a better name?

src/cast_core/src/CrudActor.cpp
```diff
@@ -239,16 +272,58 @@ auto createGenerator(YAML::Node source,
 
 }  // namespace
 
+struct CreateIndexOperation : public BaseOperation {
+    mongocxx::collection _collection;
```
guoyr: Can we push these guys up to `BaseOperation`, maybe as a variadic template to save more typing? I think every CRUD operation needs `collection`, `operation (for metrics)` and `onSession`.

guoyr: Also, technically... `CreateIndexOperation` isn't CRUD. Feel free to leave it in here but add a comment. Alternatively, to force an error, you can insert two documents with the same `_id` field since the `_id` index is automatically created and is unique.

src/cast_core/src/CrudActor.cpp
```diff
@@ -305,14 +379,16 @@ struct UpdateOneOperation : public WriteOperation {
...
-        auto result = (_onSession)
-            ? _collection.update_one(session, std::move(filter), std::move(update), _options)
-            : _collection.update_one(std::move(filter), std::move(update), _options);
-        if (result) {
-            ctx.addDocuments(result->modified_count());
-        }
-        ctx.success();
+        this->doBlock(_operation, [&](metrics::OperationContext& ctx) {
+            auto result = (_onSession)
+                ? _collection.update_one(session, filter.view(), update.view(), _options)
```
guoyr: Henrik pointed out you could pass in a reference to `nullptr` for `session` to have it not use a session. It's how the `CommitLatency` workload works.

guoyr: Probably not worth doing it here though

src/cast_core/src/CrudActor.cpp
```diff
@@ -881,16 +1049,14 @@ struct CrudActor::PhaseConfig {
...
 
     PhaseConfig(PhaseContext& phaseContext,
                 const mongocxx::database& db,
                 ActorContext& actorContext,
                 ActorId id)
         : collection{db[phaseContext.get<std::string>("Collection")]},
-          strategy{actorContext.operation("Crud", id)},
-          options{phaseContext.get<RetryStrategy::Options, false>("RetryStrategy")
-                      .value_or(RetryStrategy::Options{})} {
+          metrics{actorContext.operation("Crud", id)} {
```
guoyr: Should this operation have a longer name to distinguish it from the operations for the individual operations?

src/testlib/include/testlib/helpers.hpp
```diff
@@ -61,6 +63,41 @@ inline void dropAllDatabases(Client& client) {
...
+class MultiLineRegexMatch : public Catch::MatcherBase<std::string> {
+    std::string regex;
+
+public:
+    explicit MultiLineRegexMatch(std::string regex) : regex{regex} {}
+
+    // Performs the test for this matcher
+    virtual bool match(const std::string& matchee) const override {
+        try {
+            boost::regex reg(
```
guoyr: 👍

src/testlib/include/testlib/helpers.hpp
```diff
@@ -19,6 +19,8 @@
 
 #include <string>
 
+#include <boost/regex.hpp>
```
guoyr: Add `boost::regex` explicitly to `CMakeLists.txt`?

src/testlib/test/helpers_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+[genny::ServerResponse*] = { "nInserted" : 0, "nMatched" : 0, "nModified" : 0, "nRemoved" : 0, "nUpserted" : 0, "writeErrors" : [ { "index" : 0, "code" : 11000, "errmsg" : "E11000 duplicate key error collection: mydb.test index: a_1 dup key: { : 100 }" } ] }
+[genny::Message*] =
+)";
+    SECTION("Simple match") {
+        std::string rex = ".*duplicate key error.*";
+        REQUIRE_THAT(actual, genny::MultiLineRegexMatch(rex));
+    }
+}
+
+}  // namespace
```
guoyr: 💅

guoyr:
> * try/catch is now handled gracefully and each operation reports a meaningful error message as much as it can.

Big win!

> * I added `find_one_and_update` and working on `find_one_and_delete`, `find_one_and_replace`, and `estimated_document_count` now (they're simple).


> I decided _against_ adding the ability to set the `mongocxx::options:*` types via yaml. There are a number of reasons:

 
> 1. It's a pretty tedious and error-prone `convert<>` impl for each one. I have some perl that automates it kinda but even then...

+1000

> 3. I think the more correct thing to do is allow all of the fields to be set using ValueGenerators.
>    E.g.
>    ```yaml
>    - Operations:
>      - OperationName: createView
>        OperationOptions:
>          collation: 
>            locale: {^RandomString: ...}
>          write_concern:
>            journal: {^RandomBool}
>    ```
>    
>    
>    This will require
>    
>    1. adding scalar and array value-generators _and_

Yeah, array generators would be great.I think it'd also be useful to give it native support in CrudActor operations' `run()` method for operations that accept an array. E.g. xx_many_operations.

>    2. some intermediary classes that can be constructed from a `(YAML::Node,PhaseContext)` that convert to these `options*` type so the `options*` types are constructed at workload-time as are the documents/filters/etc.

Hm, I didn't quite follow this. Is there an example of a raw mongocxx::option that can't be constructed at workload time?

> If no objections to this, I'll create a ticket to support randomly-generated `mongocxx::options*` types on CrudActor. I left the existing conversions there.

Yeah definitely agree we shouldn't be manually converting `mongocxx::options`

> I also decided against having a nested `withTransaction`-style syntax. It got kinda messy and I think if you want nested blocks like this you probably just want to write your own Actor C++. I'll add this to the Genny Ideas doc as wontfix. Similar for setting read concern at the Phase level.

SGTM. I've found startTransaction and endTransaction operations sufficient for most use cases.

src/cast_core/src/CrudActor.cpp
```diff
@@ -97,14 +88,7 @@ template <>
...
     static Node encode(const BulkWriteOptions& rhs) {
-        Node node;
-        auto bypassDocValidation = rhs.bypass_document_validation();
-        if (bypassDocValidation) {
-            node["BypassDocumentValidation"] = *bypassDocValidation;
-        }
-        auto isOrdered = rhs.ordered();
-        node["Ordered"] = isOrdered;
-        return node;
+        return {};
```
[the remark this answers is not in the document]
rtimmons: Ya there's never (afaict) reason to go back from `T -> YAML::Node` for pretty much any of our types so keeping `encode` impls around feels unnecessary. I was tempted to just remove them to force a compiler error if we ever tried it but that felt extreme.

src/cast_core/src/CrudActor.cpp
rtimmons: lol ya good call  - renamed

src/cast_core/src/CrudActor.cpp
rtimmons: Cool....that would save an `if` statement in pretty much every operation but it feels like dangerously undocumented behavior..so I'll leave it alone for now 😸

src/cast_core/src/CrudActor.cpp
rtimmons: I would like that but it would be a change in the metrics names and would unnecessarily break history

src/cast_core/src/CrudActor.cpp
rtimmons: > Can we push these guys up to BaseOperation, maybe as a variadic template to save more typing?

Somehow I missed this comment before I added the remainder of the operations. I'm going to punt on this for now but will add a comment that it could likely be refactored in the future.

> Also, technically... CreateIndexOperation isn't CRUD. Feel free to leave it in here but add a comment....

Oh..ya..nice..I wish I'd thought of the same `_id` thing before implementing createindex. But it's here now so I'm gonna leave it. I added a comment per your suggestion.

PR #225, opened 2019-04-18
[nomerge] Don't expose YAML::Node to Actors

rtimmons:
This turned out to be much simpler than I thought.

1. This replaces `v1::ConfigNode`, and we would kill the various `get<>` methods in favor of regular ol' chained `operator[]`. 
2. It allows the `T` in `node.as<T>` to access the `WorkloadContext` whereas `YAML::Node.as<T>` only exposes the `YAML::Node` itself which doesn't let you pass in things like an RNG or additional per-Actor/per-Phase info.
3. The fancier `node.from<T,Args...>()` lets you pass in arbitrary ctor args for `T` kinda like the `PhaseLoop` ctor.
4. Built-in `.as<>` types work without fuss so `node.as<int>()` works by passing to `YAML::convert<>` without having to write boilerplate.
5. "Inheriting" values from node-to-node is actually pretty trivial now. And as a bonus you can use `..` to explicitly access parent values.

(C++17 and constexpr-if is really what makes this possible this would be a nightmare with SFINAE.)

This code is a prototype but I think proves that this isn't that hard. Would still need to support:

1. Better error-messaging and general error-handling. Since we keep track of the parent node via a ptr we can easily expose the "path" up to a missing config e.g. if you access `node["foo"]["bar"]` on yaml `{foo:{baz:7}}` we can say `Path '/foo' doesn't have key 'bar' did you mean one of {'baz'}?`.
2. Better `static_assert`s to prevent YAML::Node SFINAE messages from littering error messages.
3. Iterating `for(auto&& [k,v] : node)`.
4. Better `optional` semantics (probably just via `operator bool()`) or maybe an `asIfExists()` or something

Open questions:

1. How to expose this in `*Context` classes? Should it be `workloadContext.config["foo"]` or should we forward args to support `workloadContext["foo"]`?
2. Should the `T` type in `NodeT<T>` always be `WorkloadContext`? That makes the impl a lot easier but it makes it a bit contrived to do `docGen = actorContext.config["Document"].from<X,*this>()` if the `X` ctor takes in an `ActorContext&`.
3. Do we want to expose the "fallback" syntax `confg["foo"].as<int>(7)`? I think it's clearer as a different method `asUnless` or something (I'm terrible at naming today).

If this doesn't seem terrible to others I'll cut this as a ticket for a future iteration. I estimate 3-4 days to address above questions and remaining items and another couple days to replace all uses of `YAML::Node` in Actors and the like with the new API.


rtimmons:
Closing this in favor of [TIG-1626](https://jira.mongodb.org/browse/TIG-1626) which I'm hoping to prioritize in our current or next iteration.

PR #224, opened 2019-04-18
 TIG-1337 add remaining genny canaries

guoyr:
Add genny canaries for CPU, L2 cache, L3 cache, and sleep/context switching.

PR #227, opened 2019-04-19
 TIG-1614 Add benchmark tests for genny canaries

guoyr:
Please review jusst the TIG-1614 commit. Everything prior is #224


guoyr:
The arch linux EC2 shared instances we use are terrible for performance testing (20% variance between runs). I'm going to create a new variable to run genny benchmarks

src/canaries/benchmark/canaries_benchmark.cpp
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    SECTION("nop") {
+        // Run NopTask 1 million times. The total time was ~40ms for the "phase" loop
+        // version and ~200ms for the "real" loop. Replace "ms" with "ns" to get the
+        // average time per loop.
+        auto nopRes = runTest<NopTask>(loopNames, 1e6);
+
+        // Don't validate the results of nop loops, just print the results.
+        printRes(nopRes, "nop");
```
rtimmons: justify why not validating it - because it shouldn't involve any actual overhead / it's always a baseline.

src/gennylib/benchmark/orchestrator_benchmark.cpp
```diff
@@ -1,101 +0,0 @@
-// Copyright 2019-present MongoDB Inc.
```
rtimmons: 💥 👍 
i think we can also remove `PhaseLoop_benchmark.cpp`?

guoyr: I was planning on leaving it in there because it does multi-threaded runs, which this benchmark doesn't do. It's also a very stable test and we haven't needed to update the threshold ever.

src/canaries/benchmark/canaries_benchmark.cpp
guoyr: It's only going to test the overhead of looping and metrics collection, which is definitely not going to be 2% going from native for loops to phase loops. I'll update the comment to clarify.

PR #228, opened 2019-04-19
[no-merge] Config API with CRTP

guoyr:


src/config/include/config/config.hpp
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+            if (!_parent) {
+                return std::nullopt;
+            }
+            return _parent->yamlGet(key);
+        }
+    }
+
+    template <typename K>
+    NodeT get(const K& key) {  // TODO: const version
```
guoyr: I got it to compile by having the contexts inherit NodeT's constructor and making two changes mentioned below.

The first one is to remove the ".." key because the parent isn't always define. It seems like the ".." is only going to be used in code (and not static YAML files)? So maybe it could be replaced with ->parent()?

src/config/include/config/config.hpp
```diff
@@ -0,0 +1,110 @@
...
+//            if (key == "..") {
+//                if (!_parent) {
+//                    throw std::logic_error("TODO");  // TODO: better messaging
+//                }
+//                return *_parent;
+//            }
+//        }
+        std::optional<YAML::Node> yaml = this->yamlGet(key);
+        if (yaml) {
+            return NodeT{*yaml, _parent};
```
guoyr: The second change is having the parent of the returned value be _parent instead of this. Personally it feels more intuitive to have contexts be explicitly defined instead of potentially be user-defined YAML nodes.

PR #229, opened 2019-04-19
Add more docs to create-new-actor skeleton actor, fix resmoke

rtimmons:
Ran full demo end-to-end with clean machine (no genny checkout or toolchain). Hopefully these docs don't become too stale too quickly or aren't too overwhelming. Feedback on what to eliminate or add of course welcome.


scripts/create-new-actor.sh
```diff
@@ -255,6 +469,13 @@ namespace {
 using namespace genny::testing;
 namespace bson_stream = bsoncxx::builder::stream;
 
+//
+// There is a "known" failure that
```
guoyr: Increase line length?

scripts/create-new-actor.sh
```diff
@@ -332,15 +555,20 @@ cat << EOF
 Build and test ${actor_name} with the following command:
 
     ./scripts/lamp
-    ./scripts/lamp resmoke-test --suites src/resmokeconfig/genny_standalone.yml
     ./scripts/lamp cmake-test
+    ./scripts/get-mongo-source.sh
+    ./scripts/lamp resmoke-test --suites src/resmokeconfig/genny_standalone.yml
+
+The resmoke-test will fail because there is a "hidden" bug in the generated
```
guoyr: lol. "hidden"

scripts/create-new-actor.sh
```diff
@@ -151,26 +320,67 @@ struct ${actor_name}::PhaseConfig {
...
-            inserts.addDocuments(1);
-            inserts.addBytes(document.view().length());
-            inserts.success();
+
+            //
+            // If your Actor throws any uncaught exceptions, the whole Workload will
+            // attempt to end as quickly as possible. Every time this inner loop loops
+            // around it checks for any other Actors' exceptions and will stop iterating
+            // if it sees any. Such failed workloads are considered "programmer error"
+            // and will mark the workload task as a system-failure.
```
guoyr: "Evergreen" system failure? I don't know if "system-failure" by itself is a thing

scripts/create-new-actor.sh
```diff
@@ -125,24 +198,120 @@ cat << EOF
...
+// The ${q}PhaseLoop<PhaseConfig>${q} type constructs one ${q}PhaseConfig${q} instance
+// for each ${q}Phase:${q} block in your Actor's YAML. We do this at
+// Actor/Workload setup time and before we start recording real metrics.
+// This lets you do any complicated or costly per-Phase setup operations
+// (such as interacting with YAML or doing syntax validations) without
+// impacting your metrics.
+//
+// Imagine you have the following YAML:
+//
+//     SchemaVersion: 2017-07-01
```
guoyr: I feel like a visual representation  workload YAML runs and how it hooks up to the actor would have been useful for me when I first started reading workload files. Dan Pasette expressed a similar sentiment last week in that the phases that are defined next to each other do not run alongside each other IRL.

I'm going use Illustrator to draw a deconstructed diagram when I get back and see if we want to add it to the readme.

scripts/create-new-actor.sh
```diff
@@ -151,26 +320,67 @@ struct ${actor_name}::PhaseConfig {
...
+            // Evaluate the DocumentGenerator template:
+            //
             auto document = config->documentExpr();
+
+            //
+            // The clock starts running only when you call ${q}.start()${q}. The return value
+            // from this lets you record how many bytes, documents, and actual
+            // iterations (usually 1) your Actor completes while the clock is running. You then
+            // must stop the clock by calling either ${q}.success()${q} if everything is okay
+            // or ${q}.fail()${q} if there was a problem. It's undefined behavior what will
```
guoyr: Is it worth adding a small number of emojis prefixes to mark different types of comments similar to the genny ideas doc? Some way of breaking up these paragraphs might improve readability.

E.g.


👓The clock starts running only when you call ${q}.start()${q}.... # Newbie Guide of basic concepts you have to know

⚠️ It's undefined behavior what will... # Warning!

🍪 MongoException lets you include a "causing" bson document ... # Non-essential tips and tricks to improve your happiness

scripts/create-new-actor.sh
```diff
@@ -17,14 +17,16 @@
 set -eou pipefail
 
 year="$(date '+%Y')"
+q='`'
```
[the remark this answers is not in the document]
rtimmons: Unfortunately that prevents the `${actor_name}` and `${year}` vars from being interpolated 😞

scripts/create-new-actor.sh
```diff
@@ -67,25 +69,96 @@ namespace genny::actor {
...
  *
+ * Indicate what the Actor does and give an example yaml configuration.
+ * Markdown is supported in all docstrings so you could list an example here:
+ *
+ * ${q}${q}${q}yaml
+ * SchemaVersion: 2017-07-01
+ * Actors:
+ * - Name: ${actor_name}
+ *   Type: ${actor_name}
+ *   Phases:
```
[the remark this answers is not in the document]
rtimmons: Yup I agree the right documentation isn't in the right places at the moment and fixing this would definitely help the onboarding. I've added this to the genny ideas inbox.

scripts/create-new-actor.sh
rtimmons: For sure this can seem a bit "backwards" at first. Genny Ideas doc asks if we should have a "Phase-first syntax" but I think for now just more docs would help. You've inspired me to write up an extended example using multiple instances of HelloWorld with different Repeat/Duration/"background"-ness. A diagram would also be lovely if you manage to put one together!

scripts/create-new-actor.sh
rtimmons: I like the idea...I tried going through with it and it kinda felt contrived about what was what, though, so I just added ⚠️ wherever it seemed most appropriate.

PR #232, opened 2019-04-22
TIG-1566 improve metrics parsing perf

guoyr:
I [left a comment on the ticket](https://jira.mongodb.org/browse/TIG-1566) with a patch build showing a roughly 40% perf improvement from sorting 10GB of csv in memory at once. Is that sufficient or do we want to pursue more aggressive optimizations?

### DIFF
--- a/src/canaries/src/main.cpp
+++ b/src/canaries/src/main.cpp
@@ -94,7 +94,7 @@ struct ProgramOptions {
         ("metrics-output-file,o",
                 po::value<std::string>(),
                 "Write output to file in addition to stdout. The format ouf the output"
-                "file is [task-name],[loop-type],[average_duration_in_picoseconds]");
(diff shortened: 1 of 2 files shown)


rtimmons:
> Is that sufficient or do we want to pursue more aggressive optimizations?

I don't think we're very sensitive to how long this takes and a 40% improvement definitely gets us by. The "right" solution is probably to not produce such huge CSVs (don't repeat op name so many times (have a names-to-ids-section), report only clock offsets, and a few other low-hanging fruit things could drastically reduce the file-size). Or of course do more of the "ftdc" logic in genny...but we need to think about how this plays with large-scale workloads. TLDR: this seems like a good-enough stopgap and there's some more room for quick-fixes if we need something more in the short-term.

PR #235, opened 2019-04-30
TIG-1618 Add nop actor and (unreleated) consolidate nop phase syntax

guoyr:


src/gennylib/test/PhaseLoop_test.cpp
```diff
@@ -380,26 +376,6 @@ TEST_CASE("Actual Actor Example") {
                     {94, 3}});
     }
 
-    SECTION("Throws with other keywords in Nop phases") {
```
guoyr: My intention with deleting this test is to defer YAML validation to the "strict mode" ticket.

src/gennylib/src/context.cpp
```diff
@@ -154,57 +154,7 @@ std::unordered_map<PhaseNumber, std::unique_ptr<PhaseContext>> ActorContext::con
...
-    if (isNop && _node.size() > 1) {
-        if (_node.size() != 2 || !_node["Phase"]) {
-            throw InvalidConfigurationException(
-                "'Nop' cannot be used with any other keywords except 'Phase'. Check YML "
-                "configuration.");
-        }
-    }
-
-    return isNop;
+    return get<bool, false>("Nop").value_or(false);
```
guoyr: I took out operation/operationName because they're not Genny concepts outside of nop. Specific actors may choose to use them.

Also, I picked the `{Nop: true}` syntax so it's easier to validate when we implement strict mode. If "Nop" were a value like {OperationName: "Nop"}, we will have still traversed the key even if the value is mistakenly set to something else, causing the previous version of `isNop()` to return false.

src/cast_core/src/NopMetrics.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <boost/log/trivial.hpp>
+
+#include <gennylib/Cast.hpp>
+
+namespace genny::actor {
+
+/** @private */
+struct NopMetrics::PhaseConfig {
+    std::string message;
```
rtimmons: ☠️

src/workloads/docs/InsertWithNop.yml
```diff
@@ -11,16 +11,16 @@ Actors:
...
   - Collection: "inserts"
     Document: {"b": {^RandomString: {length: {^RandomInt: {min: 5, max: 15}}, alphabet: abcde}}}
   - Collection: "inserts"
     Document: {"a": {^RandomInt: {min: 63, max: 70}}}
-  - Operation: Nop
+  - *Nop
   - Collection: "inserts"
     Document: {"b": {^RandomString: {length: {^RandomInt: {min: 15, max: 25}}, alphabet: abcde}}}
-  - Operation: Nop
+  - *Nop
```
rtimmons: Personally I'd rather avoid this rather silly anchor and just put `Nop: true` in these lines

src/workloads/selftests/GennyOverhead.yml
```diff
@@ -0,0 +1,63 @@
...
+
+Actors:
+- Name: GennyOverhead100T
+  Type: NopMetrics
+  Threads: 100
+  Phases:
+  - MetricsName: LoopOnlyFixedTime
+    Duration: 10 seconds
+  - MetricsName: LoopOnly
+    Repeat: 1000000  # 1 million iterations.
```
rtimmons: I think use `1e6` here

guoyr:
Made a couple of followup changes.

1. Lowered the test duration from 10s to 10ms bc we generate 20GB of metrics per second 🐼. I'm not sure what the noise looks like yet for the new test but will keep an eye out. It might not be a big deal even if it's noise because we still have the numbers from the genny canaries task as a backstop.

2. I moved the metrics to outside the inner loop so we can measure inter-loop durations. This is useful for sleepbefore/sleepafter/ratelimiter. Again there might be high noise and we can adjust the numbers as needed. The rate limiter numbers are a little off. I'm not sure what the reason is at the moment.

PR #236, opened 2019-04-30
TIG-1626: Hide YAML::Node Behind genny::Node

rtimmons:
Sorry for big diff but this is mostly boilerplate to bind cleanly with `YAML::Node`. And this is > 50% tests. 

This is woefully lacking in documentation but didn't want to add docs before consensus on the API.

Before:

```c++
readModeString{phaseContext.get("ReadPreference")["ReadMode"].as<std::string>()},
useSession{phaseContext.get<bool, false>("Session").value_or(false)},
useTransaction{phaseContext.get<bool, false>("Transaction").value_or(false)},
threads{phaseContext.get<IntegerSpec>("Threads")
```

After:

```c++
// can chain [] calls without dipping into YAML::Node. YAML::Node is never exposed directly
// not even by iterators.
readModeString{phaseContext["ReadPreference"]["ReadMode"].to<std::string>()},
// .value_or is built in with fallback semantics (genny::Node::value_or behaves like nullopt::value_or()
useSession{phaseContext["Session"].value_or(false)},
// Renamed conversion operation to .to<> instead of .as<> to hopefully make it clearer
threads{phaseContext["Threads"].to<IntegerSpec>()
```

This change just _introduces_ `genny::Node` it doesn't swap out for `YAML::Node` yet or actually *use* the code anywhere yet. That's step 2 .

/cc @bcaimano since I think you had a precursor to this idea a long time ago. No expectation of review unless you feel inclined 😸


src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    template <typename O = Node, typename... Args>
+    std::optional<O> maybe(Args&&... args) const {
+        // TODO: tests of this
+        static_assert(!std::is_same_v<std::decay_t<YAML::Node>, std::decay_t<O>>, "🙈 YAML::Node");
+        static_assert(
+            std::is_same_v<decltype(_maybeImpl<O, Args...>(std::forward<Args>(args)...)),
+                           std::optional<O>>,
+            "Destination type must satisfy at least one of the following:\n"
```
guoyr: 🤘

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+
+    template <
+        typename O,
+        typename... Args,
+        typename = std::enable_if_t<
+            !isNodeConstructible<O, Args...>() &&
+            std::is_same_v<decltype(YAML::convert<O>::encode(std::declval<O>())), YAML::Node>>,
+        typename = void,
+        typename = void>
+    std::optional<O> _maybeImpl(Args&&... args) const {
```
guoyr: [question] How do you decide if a method goes in the header or the cpp file?

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+std::vector<T> Node::getPlural(const std::string& singular,
+                               const std::string& plural,
+                               F&& f) const {
+    std::vector<T> out;
+    auto pluralValue = (*this)[plural];
+    auto singValue = (*this)[singular];
+    if (pluralValue && singValue) {
+        std::stringstream str;
+        str << "Can't have both '" << singular << "' and '" << plural << "'. ";
+        str << "Path: '" << this->path() << "$plural(" << singular << "," << plural << ")'";
```
guoyr: Does "$plural' have a special meaning here?

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+public:
+    // jump through immense hoops to avoid knowing anything about the actual yaml iterator other
+    // than its pair form is a pair of {YAML::Node, YAML::Node}
+    template <typename ITVal>
+    IteratorValue(const Node* parent, ITVal itVal, size_t index)
+        : NodePair{std::make_pair(
+              Node{itVal.first,
+                   parent,
+                   itVal.first,
+                   itVal.first ? (itVal.first.template as<std::string>() + "$key") : ""},
```
guoyr: Could `_key` be made optional or be set to an empty string here? It seems like the first item in the pair really wants to be a scalar and doesn't want to key.

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+
+public:
+    // jump through immense hoops to avoid knowing anything about the actual yaml iterator other
+    // than its pair form is a pair of {YAML::Node, YAML::Node}
+    template <typename ITVal>
+    IteratorValue(const Node* parent, ITVal itVal, size_t index)
+        : NodePair{std::make_pair(
+              Node{itVal.first,
+                   parent,
+                   itVal.first,
```
guoyr: Can `isNull()` replace `_valid`?

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,444 @@
...
+#include <variant>
+#include <vector>
+
+#include <boost/throw_exception.hpp>
+
+#include <yaml-cpp/yaml.h>
+
+namespace genny {
+
+namespace v1 {
```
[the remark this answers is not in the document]
rtimmons: it's actually only this `toString()` that's in `v1` the rest is in regular `genny`

src/config/include/config/Node.hpp
rtimmons: so as a general rule anything templated pretty much has to go in the header....I moved whatever else I could and what was more than 1-2 lines of boilerplate.

src/config/include/config/Node.hpp
rtimmons: There's a difference in `YAML::Node` between a null node and a node that doesn't exist:

```c++
    {
        YAML::Node yaml = YAML::Load("foo: null");

        REQUIRE(yaml["foo"].IsDefined() == true);
        REQUIRE(yaml["foo"].IsNull() == true);
        REQUIRE(bool(yaml["foo"]) == true);

        REQUIRE(yaml["bar"].IsDefined() == false);
        REQUIRE(yaml["bar"].IsNull() == false);
        REQUIRE(bool(yaml["bar"]) == false);
    }
```

src/config/include/config/Node.hpp
rtimmons: We can't really make the key optional because we don't know if we're iterating over a sequence or a map.
This does set the key to the empty-string if we're iterating over a sequence rather than a map. This gets around the `YAML::Node` oddity that the iterator for all nodes is both a node and a kvp of nodes but the k in the kvp form for arrays is basically undefined and it's not really obvious what the `kvp.first.path()` should be when iterating over maps (hence the `$key` suffix).

But I actually just changed this to 

1. add some docs because it's not super clear what this logic is intending to do
2. make the keys for these weird situations a bit more logical
3. test the behavior of `genny::Node` in these situations versus the behavior of `YAML::Node`.

Hopefully the changed version is a bit clearer!

src/config/include/config/Node.hpp
guoyr: Derp. My question came out wrong.

⏪(rewind)

I had wanted to ask if _isValid is necessary or not.

I think whether a node exists or not is a YAML concept because a YAML file may or may not have a node; but for this genny::Node class, there should be a legitimate use case for every node that we create. If we don't expect a genny::Node to exist,  maybe() should be called instead of always getting back an empty Node.

There are two existing usages of `_valid`. One is in `operator bool()`, the other is in `get()`. Following my reasoning above, I think the former can be replaced with just `_yaml`, and the latter should throw an error and ask to either add a thing to the YAML or change the c++ call to user maybe().

Feel free to ignore the above if you disagree. I just wanted to see if we can simplify how we construct the iterator value. There might be legitimate use cases that need `_valid`.

src/config/include/config/Node.hpp
rtimmons: nah this is just to hopefully make it clear what the code is looking for:

```c++
    {
        Node node{"{Foo: 8, Foos: [1,2]}", ""};
        REQUIRE_THROWS_WITH([&]() { node.getPlural<int>("Foo", "Foos"); }(),
                            Catch::Matches("Can't have both 'Foo' and 'Foos'. Path: '\\$plural\\(Foo,Foos\\)'"));
    }
```

src/config/include/config/Node.hpp
guoyr: Gotcha! `_key` is for traversing back up to the parent. 

I should've read the unit tests.

src/config/include/config/Node.hpp
rtimmons: Oh I see - I like the intuition I didn't think about that at first.

Unfortunately it didn't pan out. I played with it and it made things worse. The problem is that we need a `YAML::Node _yaml` class-member, but in the cases where the path is invalid I just default-construct one, and unfortunately the default-constructed node doesn't behave the way you want or expect:

```c++
    {
        // This is why we need to have Node::_valid
        // We can't rely on our default-constructed nodes.
        auto defaultConstructed = YAML::Node{};
        REQUIRE(bool(defaultConstructed) == true);
        REQUIRE(defaultConstructed.Type() == YAML::NodeType::Null);
    }
```

I've added this to a test-case for posterity.

We could make the `YAML::Node` into a `std::optional` but that seems less clear than the `_valid` check

visemet: It's weird to me that yaml-cpp has an internal notion of `ZombieNode` for this concept but doesn't let anyone use it.

https://github.com/jbeder/yaml-cpp/blob/yaml-cpp-0.6.2/include/yaml-cpp/node/impl.h#L18-L45

src/config/src/Node.cpp
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <config/Node.hpp>
+
+namespace genny {
+
+void Node::appendKey(std::ostringstream& out) const {
+    if (_parent) {
```
visemet: It seems weird to me that `appendKey()` is also responsible for appending the parent's key and not just its own. It makes me think a "build path" kind of name is more appropriate.

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,670 @@
...
+    template <typename K>
+    const Node get(const K& key) const {
+        const std::string keyStr = v1::toString(key);
+        if (!_valid) {
+            return Node{YAML::Node{}, this, false, keyStr};
+        }
+        if constexpr (std::is_convertible_v<K, std::string>) {
+            if (key == "..") {
+                // this is...not the most succinct business-logic ever....
+                std::stringstream childKey;
```
visemet: 1. [style nit] Standardize on either using `std::stringstream` or `std::ostringstream` but not both in these files.
2. [nit] This key is actually an alias for the parent so calling it `childKey` is confusing.
3. The following would be a bit more compact if you're willing to use `+` to do one of the string concatenations.

   ```cpp
   if (!_parent) {
       return Node{YAML::Node{}, nullptr, false, _key + "/.."};
   }

   std::ostringstream parentKey;
   parentKey << _parent->_key << "/" << _key << "/..";
   return Node{_parent->_yaml, _parent->_parent, _parent->_valid, parentKey.str()};
   ```

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,670 @@
...
+     *
+     * ```c++
+     * auto x = node.value_or(7); // int
+     * auto y = node.value_or(std::string{"foo"}); // std::string
+     *
+     * // or specify it the hard way
+     * auto z = node.value_or<std::string>("foo");
+     * ```
+     *
+     * Like `operator[]` this will "fall-back" to the parent node.
```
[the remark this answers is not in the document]
rtimmons: Whether or not the node "exists" depends on whether the node could be found in the parent:

```c++
    Node node{"seven: 7", ""};
    REQUIRE(node["foo"]["bar"][0]["seven"].value_or(8) == 7);
```

src/config/include/config/Node.hpp
rtimmons: 1. Good call changed to `stringstream`
2. Took your suggestion for (3) another good suggestion.

rtimmons:
Thanks for review @guoyr and @visemet. I've addressed all feedback and I feel this is in good shape now and the behavior is well-defined and consistent with how YAML::Node behaves. Ready for another look whenever you guys have time.

src/config/include/config/Node.hpp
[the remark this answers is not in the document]
guoyr: I am also kinda confused about the "fallback" behavior. The double fallback behavior to both parent and value_or() is a little bit confusing for generic configNodes (i.e. nodes other than the old actor/phase/workload contexts).

This behavior means that actor authors will have to not reuse any names used by other actors. Otherwise if a workload uses two actors that share the same config name and the author specifies the config in a higher-up context intended for one actor, the other actor is going to inadvertently pick it up.

Fortunately, workloads almost never share configs across actors. But, I do still think genny should internally distinguish between the two types of fallbacks verbally. I think we should use "fallback" to always refer to the value passed to `value_or()` and "inherit" to refer to the parent nodes' values.

E.g. this comment can be rephrased as something like: Like `operation[]`, this will check its chain of parent nodes and inherit the value if it exists anywhere in the chain. If the value is not specified in any parent node, it'll fallback to the value passed to `value_or(8)` (i.e. 8).

Ditto throughout for all usages of fallback/fall-back/"fall back"

src/config/include/config/Node.hpp
```diff
@@ -242,95 +606,114 @@ class Node {
...
+        }
+        return _parent->yamlGet(key);
+    }
+
+    // Forward to _yaml[key] if we're a valid node else forward to the parent
     template <typename K>
     std::optional<const YAML::Node> yamlGet(const K& key) const {
         if (!_valid) {
-            return std::nullopt;
+            return parentGet(key);
```
guoyr: ah good fix

src/config/include/config/Node.hpp
```diff
@@ -458,95 +606,114 @@ class Node {
...
-        auto yaml = this->yamlGet(key);
-        if (yaml) {
-            return Node{*yaml, this, true, keyStr};
-        } else {
-            return Node{YAML::Node{}, this, false, keyStr};
+        try {
+            auto yaml = this->yamlGet(key);
+            return yaml ? Node{*yaml, this, true, keyStr} : Node{YAML::Node{}, this, false, keyStr};
+        } catch (const YAML::Exception& x) {
+            // YAML::Node is inconsitent about where it throws exceptions for `node[0]` versus
```
visemet: [typo] inconsitent -> inconsistent.

src/config/include/config/Node.hpp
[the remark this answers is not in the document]
visemet: The other thought I had about this was whether `.value_or()` should behave like `.get_noinherit()` where it won't try to look up the key in the parent if it isn't found in the innermost child.

> Whether or not the node "exists" depends on whether the node could be found in the parent:
> 
> ```c++
>     Node node{"seven: 7", ""};
>     REQUIRE(node["foo"]["bar"][0]["seven"].value_or(8) == 7);
> ```

I find this behavior mystifying - maybe because I've gotten so used to thinking about how path traversal works in MongoDB. What utility do you see in offering this API as opposed to having `node["foo"]["bar"][0]["seven"].value_or(8) == 8`?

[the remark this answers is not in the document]
rtimmons: I don't know if there's a use-case for `get_noinherit` given that you can explicitly ask for `node[".."][key]`. 

> I've gotten so used to thinking about how path traversal works in MongoDB

Can you give an example? I'm only familiar with it in `find()` expressions using `.` to traverse child docs where there is no notion of inheritance.

But I think this issue is only being confused by `value_or` so I removed it. It was intended to just be syntax sugar:

```c++
auto coll = node["Collection"].value_or("coll");
// equivalent to:
auto coll = node["Collection"].maybe<std::string>().value_or("coll");
```

This doesn't address the potential confusion around inheritance though since the same behavior exists without the syntax-sugar:

```c++
    Node node{"seven: 7", ""};
    REQUIRE(node["foo"]["bar"][0]["seven"].maybe<int>().value_or(8) == 7);
```

For in-person review during today's standup:

Given the following YAML:

```yaml
Coll: Test
Phases:
- Doc: foo
- Coll: Bar
- Another:
  - Nested: {Coll: Baz}
```

What should the following return?
What should `node[...].value_or("Or")` return?

```c++
node["Coll"] => Test, Test
node["Phases"]["Coll"] => ?
node["Phases"][0]["Coll"] => 
node["Phases"][1]["Coll"] => Bar, Bar
node["Phases"][2]["Coll"] => ?
node["Phases"][2]["Another"]["Coll"] => ?
node["Phases"][2]["Another"]["Nested"]["Coll"] => Baz, Baz
```

For now I've just killed `value_or` in favor of using vanilla `.maybe<T>().value_or()`.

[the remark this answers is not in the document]
rtimmons: We discussed briefly after standup today. Only thing that I'm deviating from is the decision to have `node["Phases"]["Coll"]` do inheritance to return `Test` rather than barfing due to attempting to access a sequence like a map - this is consistent with what `YAML::Node` does.

rtimmons:
Okay hammered out the inheritance/fallback behavior after convos with the team. 

@vrachev suggested that maybe we just don't do inheritance at all since it may be more confusing than it's worth. I'm not opposed to revisiting this if it ends up causing confusion for users. Many (all?) of our existing workload yamls rely on inheritance, however, so it would be a larger change.

Open to additional feedback else I plan to merge at the end of the day today.

src/config/include/config/Node.hpp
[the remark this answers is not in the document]
visemet: > I don't know if there's a use-case for `get_noinherit` given that you can explicitly ask for `node[".."][key]`.

But `.get_noinherit()` gave you a definitive answer as to whether the value was specified at that level rather than a higher-level parent node.

> But I think this issue is only being confused by `value_or` so I removed it. It was intended to just be syntax sugar:
> This doesn't address the potential confusion around inheritance though since the same behavior exists without the syntax-sugar:

That's fair. I was thinking `.value_or()` would make people think of `std::optional` and what's being presented is a different kind of API where "missing" means "missing from all levels" and not just the innermost one.

> Can you give an example? I'm only familiar with it in `find()` expressions using `.` to traverse child docs where there is no notion of inheritance.

Hence my confusion. It wasn't specific to the `fallback` parameter per-se and just the inheritance concept I had momentarily blanked out on. Once I reframed it as being tied to the inheritance concept, the rules about the behavior made sense to me.

src/config/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+project(config VERSION 0.0.1 LANGUAGES CXX)
+
+CreateGennyTargets(
+    NAME    config
+    TYPE    STATIC
+    DEPENDS
+        MongoCxx::bsoncxx
```
guoyr: unused now?

src/config/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+        yaml-cpp
+        Boost::boost
```
guoyr: Heh, this is a thing?

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,664 @@
...
+    /**
+     * @return
+     *   The root node that represents the whole YAML file/document.
+     */
+    const class Node& root() const& {
+        return *_root;
+    }
+
+    // disallow `NodeSource{"",""}.root();`
+    void root() const&& = delete;
```
guoyr: 👍

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,664 @@
...
+ */
+class NodeSource final {
+public:
+    ~NodeSource();
+
+    /**
+     * @return
+     *   The root node that represents the whole YAML file/document.
+     */
+    const class Node& root() const& {
```
guoyr: TIL ref-qualification

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,664 @@
...
+     *   The full yaml document
+     * @param path
+     *   Path information. Used in error messages.
+     *   Likely a file-path from where the yaml was loaded.
+     */
+    NodeSource(std::string yaml, std::string path);
+
+private:
+    const YAML::Node _yaml;
+    const std::string _path;
```
guoyr: Did we want to use this somewhere?

src/config/include/config/Node.hpp
```diff
@@ -0,0 +1,664 @@
...
+private:
+    friend Node;
+
+    NodeIterator(const class NodeImpl*, bool end);
+
+    const std::unique_ptr<class IteratorImpl> _impl;
+};
+
+
+// A bulk of this could probably be moved to the PiMPL version
```
guoyr: This makes sense. Add a thing to the GI doc?

src/config/src/Node.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+bool Node::isScalar() const {
+    return _impl->isScalar();
+}
+
+Node::Type Node::type() const {
+    return _impl->type();
+}
+
+bool Node::isNull() const {
```
guoyr: Maybe not the right thing to do but if you just check the _impl's type directly, you won't have to write the `isXXXX` functions on the impl again.

src/config/src/Node.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+bool NodeIterator::operator==(const NodeIterator& rhs) const {
+    return _impl->equal(*rhs._impl);
+}
+
+const NodeIteratorValue NodeIterator::operator*() const {
+    return _impl->getValue();
+}
+
+}  // namespace genny
```
guoyr: NL

src/config/test/Node_test.cpp
```diff
@@ -0,0 +1,1161 @@
...
+        auto count = 0;
+        for (auto& [k, v] : sil) {
+            REQUIRE(v.to<int>() == 37);
+            ++count;
+        }
+        REQUIRE(count == 1);
+    }
+}
+
+}  // namespace genny
```
guoyr: NL

src/config/CMakeLists.txt
rtimmons: Ya nice catch - I don't think `config` needs to exist as a separate cmake project to just moved `Node.hpp` and friends into `gennylib`.

src/config/include/config/Node.hpp
rtimmons: 👍 GI-067

src/config/src/Node.cpp
rtimmons: I considered that but I wanted the `class Node` functions to basically be "dumb" to keep all the logic in `NodeImpl`. It turned out better that way when I added "strict mode".

PR #238, opened 2019-05-09
TIG-1656 rename Rate to GlobalRate and use unique name by default

guoyr:
Going to put up sys perf patch shortly


src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -102,17 +102,21 @@ class IterationChecker final {
...
-        const auto rateSpec = phaseContext.get<RateSpec, false>("Rate");
+                           phaseContext.get<RateSpec, false>("GlobalRate")) {
+        const auto rateSpec = phaseContext.get<RateSpec, false>("GlobalRate");
 
+
+        std::ostringstream defaultRLName;
+        defaultRLName << phaseContext.get<std::string>("Name") << phaseContext.getPhaseNumber();
         const auto rateLimiterName =
-            phaseContext.get<std::string, false>("RateLimiterName").value_or("defaultRateLimiter");
+            phaseContext.get<std::string, false>("RateLimiterName").value_or(defaultRLName.str());
```
rtimmons: 1. let's prefix this with something to prevent collisions?
2. Do we really need it unique per Phase? the limiters for any phase other than the current one will always be unused
2. do we want it to be unique per thread rather per actor name? As it is, they'll be shared by all instances of the same Actor which is not what I thought the intent was.

guoyr: > let's prefix this with something to prevent collisions?

I was expecting the name of actors to be unique. What kind of collisions did you have in mind?

> Do we really need it unique per Phase? the limiters for any phase other than the current one will always be unused

I'm pretty sure we do. The "serial" version of Cris's workload could have had both `Rate` keywords on the same actor but in different phases. I moved the above code to be after the `if (rateSpec)` statement to avoid creating unused names. However, the rate limiter itself is only constructed for phases that define rateSpec, nothing goes to waste!

> do we want it to be unique per thread rather per actor name? As it is, they'll be shared by all instances of the same Actor which is not what I thought the intent was.

Hm... it isn't unique per thread? The actor name and phase number used to construct the rate limiter name should be the same for all threads in an actor. I'm not including the actorId in the name.

guoyr:
I'm going to push this without a sys perf patch build. There's [something wrong](https://evergreen.mongodb.com/task/sys_perf_linux_3_node_replSet_service_architecture_workloads_patch_107dd43e99f191c541c0c0faab40ff0d9bc47550_5ce53e1e2a60ed1135f6c288_19_05_22_12_18_57##%257B%2522compare%2522%253A%255B%257B%2522hash%2522%253A%2522107dd43e99f191c541c0c0faab40ff0d9bc47550%2522%257D%255D%257D) with the analysis script that's preventing patch builds from finishing.

PR #246, opened 2019-05-22
TIG-1686 use bson.int64.Int64 for all fields to avoid pymongo using a…

guoyr:
… mixture of int32 and int64 by default

### DIFF
--- a/src/python/genny/parsers/cedar.py
+++ b/src/python/genny/parsers/cedar.py
@@ -22,6 +22,7 @@ from datetime import datetime
 from os.path import join as pjoin
 
 from bson import BSON
+from bson.int64 import Int64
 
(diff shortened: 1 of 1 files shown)


src/python/genny/parsers/cedar.py
```diff
@@ -22,6 +22,7 @@
 from os.path import join as pjoin
 
 from bson import BSON
+from bson.int64 import Int64
```
visemet: FWIW, https://github.com/mongodb/mongo-python-driver/blob/3.7.2/bson/__init__.py#L89 should mean you could just write `bson.Int64` instead of needing to import it separately.

guoyr: GTK.. Although I don't know how anyone could discover that :P

The documentation uses the long import path: https://api.mongodb.com/python/current/api/bson/index.html

PR #250, opened 2019-06-21
Insert big docs

ldennis:
This adds a new workload that inserts big documents.

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -36,7 +36,7 @@ setup(name='genny',
           'console_scripts': [
               'genny-metrics-report = genny.cedar_report:main__cedar_report',
               'genny-metrics-legacy-report = genny.legacy_report:main__legacy_report',
-              'lint-yaml = genny.workload_linter:main'
+              'lint-yaml = genny.yaml_linter:main'
           ]
       },
       )
--- a/src/workloads/scale/InsertBigDocs.yml
+++ b/src/workloads/scale/InsertBigDocs.yml
@@ -17,5 +17,4 @@ Actors:
     BatchSize: 1
     Document:
       x: {^RandomInt: {min: 0, max: 2147483647}}
-      string0: {^RandomString: {length: 15000000}}
-  - &Nop {Nop: true}
+      string0: {^FastRandomString: {length: 15000000}}
--- a/src/workloads/scale/InsertBigDocs.yml
+++ b/src/workloads/scale/InsertBigDocs.yml
@@ -1,0 +1,21 @@
+# This workload loads 500 rows into a collection, each of which is approximately 15MB in size.
+
+SchemaVersion: 2018-07-01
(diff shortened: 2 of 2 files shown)


src/workloads/scale/InsertBigDocs.yml
```diff
@@ -0,0 +1,21 @@
...
+  - Repeat: 1
+    Database: &DB test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500
+    BatchSize: 1
+    Document:
+      x: {^RandomInt: {min: 0, max: 2147483647}}
+      string0: {^RandomString: {length: 15000000}}
+  - &Nop {Nop: true}
```
rtimmons: You don't need this `- &Nop {Nop: true}` line since there is only one Actor in the worklodad

src/workloads/scale/InsertBigDocs.yml
```diff
@@ -0,0 +1,21 @@
...
+  Phases:
+  - Repeat: 1
+    Database: &DB test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500
+    BatchSize: 1
+    Document:
+      x: {^RandomInt: {min: 0, max: 2147483647}}
+      string0: {^RandomString: {length: 15000000}}
```
rtimmons: 1. I would suggest using `string0: {^FastRandomString: {length: 15000000}}`. The tradeoff is *slightly* less entropy but much faster generation of the strings. I made this change on my laptop and genny was able to produce writes of around 80mb/sec versus only 25 mb/sec with regular `RandomString`.

2. Consider dividing up the load between a number of threads. I.e. change `Threads: 1` to something like `Threads: 10` and then divide `DocumentCount:` by the new threadcount to maintain the same number of documents. This will of course change what the workload is doing - doing multiple writes in parallel so may not be what you're aiming for. But since this is single-threaded as it is you're going to be waiting between each `insert_one` invocation for genny to produce the next document.

src/python/setup.py
```diff
@@ -36,7 +36,7 @@
           'console_scripts': [
               'genny-metrics-report = genny.cedar_report:main__cedar_report',
               'genny-metrics-legacy-report = genny.legacy_report:main__legacy_report',
-              'lint-yaml = genny.workload_linter:main'
+              'lint-yaml = genny.yaml_linter:main'
```
rtimmons: Thanks!

src/workloads/scale/InsertBigDocs.yml
guoyr: 1. Yeah good call. FastRandomString is perfect for this use case
2. I'm pretty sure we want just 1 thread for this particular workload. The 10 thread case may be interesting but can be done as future work (in the form of another phase)

guoyr:
Stealing this PR 😛  since I promised @ldennis to get some DSI results today.

PR #253, opened 2019-06-25
PERF-1871 Create index build performance workloads for all index types

zachyam:
https://jira.mongodb.org/browse/PERF-1871

Each of these workloads create documents and indexes them for each index type. Their index creation time will be monitored for performance testing.


rtimmons:
It looks like some of the CI tests are failing due to invalid yaml:

```
[2019/06/25 14:58:46.638] [2019-06-25 18:58:46.638565] [0x00007ff50da2b740] [error]   Error loading yaml from /data/mci/18fc4b25182a8b39bc1b0c189477d7f0/src/src/workloads/execution/geoHaystackIndexWorkload.yml: yaml-cpp: error at line 36, column 27: illegal map value
```

You can run this locally yourself with the `scripts/dry-run-workloads.sh` script.

Sorry I didn't get to this today I'll take a closer look tomorrow.

src/workloads/execution/geoHaystackIndexWorkload.yml
```diff
@@ -0,0 +1,38 @@
+# This workload inserts ~1GB of documents in the first phase, and creates a geoHaystack index on them in the second.
```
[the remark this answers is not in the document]
zachyam: I'm a purist so `2DIndex.yml` it is

zachyam:
> I wonder if you get anything out of having these as separate workloads versus having one workload where the first phase is to have a couple Loader invocations that create the 2-3 different collections that you then use in subsequent Phases - create one type of index in each subsequent Phase. Just an idea since it'll mean fewer evergreen tasks and less yaml but at the expense of not being able to run just one particular index-creation without having to wait for others.


Thanks for the comments. I will let @louiswilliams chime in over here regarding one type of index per phase vs splitting them into different workloads. I will go ahead and fix the other issues you mentioned.

src/workloads/execution/2dIndexWorkload.yml
```diff
@@ -0,0 +1,39 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            a: "2d"
+            min: -180
+            max: 180
+            bits: 32
+          name: 2dIndex
```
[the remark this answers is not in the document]
zachyam: Makes sense to have it repeated multiple times. If we were to merge the separate workloads into one workload with separate phases (as you suggested in this PR), and have the actor drop and re-index the collection for each type of index after, would this cause the evergreen task to run much slower?

@louiswilliams

louiswilliams:
Looks good, but I have a few general comments:
- I think we should go ahead an put all of these into one workload. It seems there's quite a bit of repeated data generation, and I think it would be best to have fewer tasks.
- I don't think we should use 1000 byte random strings for this many keys. That's a very long key, and it means that we'll end up spending more time in memcmp and less time doing the other things we want to measure.
- Double check the way you're constructing your createIndex command, because in many of the workloads it looks like you're passing fields that are intended to be index options as part of the actual key.

src/workloads/execution/2DIndex.yml
```diff
@@ -0,0 +1,40 @@
...
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
+    BatchSize: 10000
+    Document:
+      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
+  - &Nop {Nop: true}
+
+- Name: IndexCollection
```
dalyd: How long does this take? Do you want to do it repeatedly?

src/workloads/execution/2DSphereIndex.yml
```diff
@@ -0,0 +1,62 @@
...
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
+    BatchSize: 10000
+    Document:
+      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}]
+      name: {^RandomString: {length: 1000}}
+  - Repeat: 1
+    Database: &db test
```
dalyd: You're redefining this anchor here.

dalyd: I also feel like these two loader phases can probably be combined somehow. I think they are in the same collection. Are you explicitly doing one after the other to order the data on disk?

src/workloads/execution/GeoHaystackIndex.yml
```diff
@@ -0,0 +1,39 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
+    BatchSize: 10000
+    Document:
+      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
```
dalyd: I wonder if we can pull some of this common code out using some of the Genny helpers added for this. @rtimmons thoughts?

src/workloads/execution/HashedIndex.yml
```diff
@@ -0,0 +1,37 @@
...
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
```
dalyd: I don't think you have 1 GB of data here. I'm guessing it's about 500MB since the string length is 1000.

src/workloads/execution/TextIndex.yml
```diff
@@ -0,0 +1,42 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            a: "text"
+            b: "text"
+          name: textIndex
+          weights:
+            a: {^RandomInt: {distribution: geometric, p: 0.1}}
```
dalyd: Whoa -- I had to look this one up: https://docs.mongodb.com/manual/core/index-text/#specify-weights

Although, if this is doing one index collection call, I think you might be better off putting in actual numbers here. Super cool though if you do the index creation multiple times.

src/workloads/execution/WildCardIndex.yml
```diff
@@ -0,0 +1,38 @@
...
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
+    BatchSize: 10000
+    Document:
+      "type": "book"
+      "title": "The Red Book"
+      "attributes": { "color": "red", "size":" large", "inside": { "bookmark":1, "postitnote":2 }, "outside": { "dustcover": "worn" }}
```
dalyd: Do you really want identical documents for this one?

louiswilliams:
I discussed with @zachyam offline, but I think we want to lower the total number of documents (maybe 100k?), and increase the repetitions up to 5 (don't forget the drop the index between repetitions).

The number of documents should be high enough so we spend the most amount of time in the expensive part of index builds (sorting), but not so high so that the build takes forever. I also think more than 5 repetitions would be excessive. It would require building small indexes, which is not a use case we're trying to optimize.

dalyd:
FWIW, this workload would probably be a perfect use for GI-038 from The Genny ideas doc: https://docs.google.com/document/d/1K0SlHoKkIACIXhkNP_KPHvb3IV6r53ReqsHqEYTnXDA/edit#

GI-038: Add genny value generator to match mgenerate $choose. Moved from https://jira.mongodb.org/browse/TIG-1260. Just copy more from mgenerate in general it's nice. Also look into https://github.com/mangatmodi/RandomJson 

https://www.npmjs.com/package/mgeneratejs#choose

I'm not advocating for its use here, but just bringing it up so if others get excited they can advocate for adding the support.

src/workloads/execution/2DIndex.yml
```diff
@@ -0,0 +1,40 @@
...
+
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
```
guoyr: FYI, you can define phases in a separate YAML file to avoid repetition. E.g. for this phase, instead of 

```
  - Repeat: 1
    Database: &db test
    Threads: 1
    CollectionCount: 1
    DocumentCount: 500000
    BatchSize: 10000
    Document:
      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
```

you can write:

```
   - ExternalPhaseConfig:
      Path: ../../phases/execution/IndexWorkloadPhases.yml
      Key: InsertPhase
      Parameters:
        MyDocument: {a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]}
```

Then create a `src/phases/execution/IndexWorkloadPhases.yml` that looks like:

```
InsertPhase:
  Repeat: 1
  Database: &db test
  Threads: 1
  CollectionCount: 1
  DocumentCount: 500000
  BatchSize: 10000
  Document: {^Parameter: {Name: "MyDocument", Default: {}}}
```

The YAML structure in the external phase config will replace the YAML structure starting with `ExternalPhaseConfig` in the workload file. You can also change the `MyDocument` field in `Parameters` to insert different documents for different workloads.

src/workloads/execution/UniqueIndex.yml
```diff
@@ -0,0 +1,43 @@
...
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 500000
+    BatchSize: 10000
+    Document:
+      a: {^RandomString: {length: 500}}
+      b: {^RandomInt: {distribution: geometric, p: 0.1}}
+      c: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}]
```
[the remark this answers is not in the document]
zachyam: My understanding is that `{^RandomString: {length: 500}}` generates a random string as a value for the key `a`. Do we need values to be unique?

src/workloads/execution/2DSphereIndex.yml
zachyam: Thanks for all the comments / feedback. I will be collapsing all the separate workloads into one and indexing on the collection for each type of index

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
+# This workload inserts documents in the first phase containing all types of indexes. The second phase creates indexes for each of them.
```
louiswilliams: Update comment

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 150000
+    BatchSize: 10000
+    Document:
+      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
```
louiswilliams: I think this workload would be easier to understand if the field names were less opaque and representative of their types. e.g. "point", instead of "a".

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: Create2dIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            a: "2d"
+            min: -180
```
louiswilliams: This still looks like "min", "max", and "bits" are part of the key, and not the index options.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    Operations:
+    - OperationMetricsName: Create2dSphereIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key:
+            b: "2dsphere"
+            c: "2dsphere"
+            name: 1
```
louiswilliams: This looks like you're indexing a field called "name", which doesn't exist in any of the inserted documents.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: CreateGeoHaystackIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            a: "geoHaystack"
+            d: 1
```
louiswilliams: Remove the compound index key here

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+  - Repeat: 5
+    Database: *db
+    Operations:
+    - OperationMetricsName: CreateTextIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            a: "text"
```
louiswilliams: It doesn't look like "a" and "b" aren't strings, so this index wouldn't make sense. Also, don't create a compound index.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: Create2dSphereIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key:
+            b: "2dsphere"
+            c: "2dsphere"
```
louiswilliams: I would remove the compound index key here so we can minimize the coverage of each workload. i.e. only index one field per build and separate them if needed.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: CreateUniqueIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            d: 1
+            e: 1
```
louiswilliams: Remove compound index key. Ideally, I don't think we should test "2dsphere" in this workload to keep it simple and have less coverage.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+    CollectionCount: 1
+    DocumentCount: 150000
+    BatchSize: 10000
+    Document:
+      a: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
+      b: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}]
+      c: { type: "Point", coordinates: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}] }
+      d: {^RandomString: {length: 50}}
+      e: {^RandomInt: {min: -10000000, max: 10000000}}
+      "attributes": { "color": "red", "size":" large", "inside": { "bookmark": 1, "postitnote":2 }, "outside": { "dustcover": "worn" }}
```
louiswilliams: Randomize these values if possible.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,163 @@
...
+  - &Nop {Nop: true}
+  - &Nop {Nop: true}
```
dalyd: I don't think you want to repeat your anchor here. Lines 26 on can be 
- *Nop

src/workloads/execution/CreateIndex.yml
zachyam: For GeoHaystack indexes, I need to provide an additional key. Otherwise I get this error:

`[genny::ServerResponse*] = { "ok" : 0.0, "errmsg" : "no non-geo fields specified", "code" : 16774, "codeName" : "Location16774" }`

zachyam:
@louiswilliams @dalyd 

I ran it locally, and the whole indexing took about 11 mins.

There are a total of 8 indexes, and each type of index is repeated 5 times

src/workloads/execution/CreateIndex.yml
louiswilliams: Acknowledge

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,172 @@
...
+    CollectionCount: 1
+    DocumentCount: 150000
+    BatchSize: 10000
+    Document:
+      loc2d: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -180, max: 180}}]
+      loc2dSphere: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}]
+      point: { type: "Point", coordinates: [{^RandomInt: {min: -180, max: 180}}, {^RandomInt: {min: -90, max: 90}}] }
+      randomString: {^RandomString: {length: 50}}
+      randomInt: {^RandomInt: {min: -10000000, max: 10000000}}
+      "attributes": { randomString : randomString, randomString : randomString, randomString : { randomString :  randomInt, randomString :  randomInt }, randomString : { randomString : randomString }}
```
louiswilliams: I don't think this does what you think it does. You probably need to do the value randomization like above.

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,172 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: CreateGeoHaystackIndexesCmd
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            loc2d: "geoHaystack"
+            a: 1
```
louiswilliams: "a" doesn't exist anymore

src/workloads/execution/CreateIndex.yml
louiswilliams: I don't think "attributes" needs to be in quotes

src/workloads/execution/CreateIndex.yml
```diff
@@ -0,0 +1,172 @@
...
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key: 
+            loc2d: "2d"
+          min: -180
+          max: 180
+          bits: 32
+          name: 2dIndex
+    - OperationMetricsName: DropCreated2dIndexesCmd
```
louiswilliams: I wonder if it would be better to move all of this identical index building and dropping code in a separate YAML file and parameterized as Robert suggested?

dalyd:
> @dalyd do you have any comments on the length of the workload?

So it sounds like on average each test takes a bit over a minute. WT checkpoints run every 60 seconds, so there's approximately one checkpoint per test. I'd be interested to see the raw data and see how much the results vary.

For sys-perf we would like to get tests under 90 minutes, 60 preferred. It sounds like we're well under that here even after considering the cluster setup time and canaries we should run with it.

src/workloads/execution/CreateIndex.yml
dalyd: FYI: you cannot generate a random "key" value, only a random "value" value. It looks like you want to generate random keys here. I double checked with @guoyr on generating keys.

src/workloads/execution/CreateIndex.yml
zachyam: I'm not too sure about that since the build and drops are different for each index. I think Robert's suggestion was before I merged all the workloads and documents into one.

src/workloads/execution/HashedIndex.yml
zachyam: I changed the comment so that it's more generic

louiswilliams:
LGTM, nice work

As you work on https://jira.mongodb.org/browse/SERVER-42003, you'll probably get a better idea how to tune this workload for length and repetitions.

PR #256, opened 2019-07-12
TIG-1825 Fix rate limiter termination condition

guoyr:


src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -145,7 +146,7 @@ class IterationChecker final {
             }
             while (true) {
                 auto success = _rateLimiter->consumeIfWithinRate(SteadyClock::now());
-                if (!success && (o.currentPhase() == inPhase)) {
+                if (!success && !isDone(referenceStartingPoint, currentIteration)) {
```
guoyr: There should arguably be a test for this...

rtimmons:
I'm working on a test to add to this.

But I'm not convinced this solution is correct.

The intuitive behavior of the following configuration, assuming the operation itself happens instantaneously:

```yaml
  SchemaVersion: 2018-07-01
  ...
  Threads: 5
  Phases:
  - GlobalRate: 1 per 50 milliseconds
    Duration: 115 milliseconds
  - Phase: 1
```

- Phase 0 starts at 0ms
- first instance of operation runs at 0ms
- another at 50ms
- final one at 100ms
- phase 0 ends and phase 1 starts at 115 ms, **not** at 150ms

And I still think there's an edge-case when the rate-limited action may take longer than the timespec in the ratespec.

Imagine you have an operation that takes *on average* 10 minutes, and you want to run it once per 10 minutes and allow a second thread to run the operation concurrently if the previously-running thread takes longer than 10 minutes. To ensure a consistent rate a user would probably specify `Threads: N, Rate: 1 per 10 minutes` (for some `N`). Our current code would not support this. Similarly, if you specify `5 per 10 minutes` the rate-limiter's assumption is that one thread will be able to perform all 5 operations easily within the 10 minutes.

(The `5 per 10 minutes` behavior is kinda dubious but intuitively I would want each thread to claim 1 of those slots and those slots.)

Fixing this may require switching to a scheduling algorithm (Actors await a cv and a scheduler thread `notify_one`s Actors when there is rate-capacity to consume).

But it's 4:30 on a friday and I'm tired so we'll revisit this after skunkworks next week.

guoyr:
Wrapped up my skunkworks so I'm doing through pRs. Feel free to not read this until next week.

> Imagine you have an operation that takes on average 10 minutes, and you want to run it once per 10 minutes and allow a second thread to run the operation concurrently if the previously-running thread takes longer than 10 minutes.

I'm not sure I follow what the problem is here. If you specify a rate of 1 per 10 min, the one-thread-per-10-min behavior is exactly what you get. i.e. one thread runs every 10 min. if the first thread finishes within 10min, it might get scheduled again at the 10min mark; if it doesn't finish, a different thread would be scheduled. Allowing threads to not block each other is a mandatary requirement, otherwise the "global" part of the "global rate limiter" doesn't mean anything.

rtimmons:
Let me try again...

Imagine we have the following YAML:

```yaml
SchemaVersion: 2018-07-01
Actors:
- Type: HelloWorld
  Name: HelloWorld
  Threads: 5
  Phases:
  - Message: P0
    Duration: 55 seconds
    GlobalRate: 1 per 10 seconds
  - Message: P1
    Repeat: 1
```

Running this with this PR's changes gives the following logs (I've removed extra details):

```
[00:00:27.48]  Actor #3 says: P0
[00:00:27.48]  Actor #4 says: P0   (0s)
[00:00:27.48]  Actor #0 says: P0   (0s)
[00:00:27.48]  Actor #2 says: P0   (0s)
[00:00:27.48]  Actor #1 says: P0   (0s)
[00:00:37.51]  Actor #1 says: P0   (10s)
[00:00:47.62]  Actor #2 says: P0   (10s)
[00:00:57.73]  Actor #0 says: P0   (10s)
[00:00:07.53]  Actor #2 says: P0   (10s)
[00:00:17.96]  Actor #2 says: P0   (10s)
[00:00:28.17]  Actor #1 says: P1   (10s)
[00:00:28.17]  Actor #2 says: P1
[00:00:28.17]  Actor #0 says: P1
[00:00:28.17]  Actor #4 says: P1
[00:00:28.17]  Actor #3 says: P1
```

There are two dubious things here:

1. The rate-limit isn't respected at the start of the Phase. At `00:00:27` all 5 threads do one iteration.
2. There is a 10-second gap 00:17 to 00:28 when advancing to the next Phase. Avoiding this may be hard implementation-wise but it's at least worth thinking on how it could be avoided.

Pardon the crude drawing, but here's a situation that we need to support:

```
        ┌──────────────────────────────┐
        │Threads:    5                 │
        │GlobalRate: 1 per 10 seconds  │
        │Duration:   55 seconds        │
        └──────────────────────────────┘

Expected Timeline:

  0     10     20     30     40     50   │
 t1├──────┴──┬───┼──────┼──────┼──────┼──┼────▶
   └─────────┘                           │
 t2├──────┼──────┴──┬───┼──────┼──────┼──┼────▶
          └─────────┘                    │
 t3├──────┼──────┼──────┴──┬───┼──────┼──┼────▶
                 └─────────┘             │
 t4├──────┼──────┼──────┼──────┴──┬───┼──┼────▶
                        └─────────┘      │
 t5├──────┼──────┼──────┼──────┼──────┴──┼────▶
                               └─────────┤
   ├──────┼──────┼──────┼──────┼──────┼──┼────▶
  0     10     20     30     40     50   Phase Ends
```

Basically there's an operation that takes N seconds and we rate-limit to 1 per (N-epsilon) seconds. Since there are multiple threads we should get one hit every N-epsilon seconds.

(Whether the Phase should end at that line isn't well-defined but it seems like we should be able to check in the `==end()` case if we're past our `Duration` and signal at that point...)

The 11-second operation can be mimicked by adding an 11-second sleep in the HelloWorld Actor. Excluding #1 above we can get *close* to this today.

```
[00:03:28.27]  Actor #3 says: P0   (0s)
[00:03:28.27]  Actor #2 says: P0   (0s)
[00:03:28.27]  Actor #0 says: P0   (0s)
[00:03:28.27]  Actor #1 says: P0   (0s)
[00:03:28.27]  Actor #4 says: P0   (0s)
[00:03:49.31]  Actor #4 says: P0   (21s)
[00:03:59.42]  Actor #1 says: P0   (10s)
[00:04:09.73]  Actor #0 says: P0   (10s)
[00:04:20.45]  Actor #1 says: P0   (10s)
[00:04:31.45]  Actor #1 says: P1   (11s)
[00:04:31.45]  Actor #3 says: P1
[00:04:31.45]  Actor #2 says: P1
[00:04:31.45]  Actor #0 says: P1
[00:04:31.45]  Actor #4 says: P1
```

Dubious things here:

1. Same #1 from above with the first iteration.
2. There's a 21-second gap from 03:28 to 03:49 (the sleep plus the rate-limiting) where you would only expect 10 seconds.

The 11s gap from 04:20 to 04:31 makes sense. At this point only 51 seconds had elapsed so Actor#1 goes again, sleeps for 11 seconds, and then the Phase is ended.

I don't know the solution to this without moving to a full token-bucket implementation and a scheduler. Once we fix the initial burst issue (#1 above) then we can at least make some hand-wavy justifications for other behavior.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -88,19 +88,20 @@ class BaseGlobalRateLimiter {
         int64_t curEmptiedTime = _lastEmptiedTimeNS.load();
 
         // The time the bucket was emptied after this consumeIfWithinRate() call.
-        int64_t newEmptiedTime;
```
rtimmons: I think the comment is inaccurate now?

src/gennylib/src/context.cpp
```diff
@@ -66,6 +66,12 @@ WorkloadContext::WorkloadContext(const Node& node,
             _actors.push_back(std::move(actor));
         }
     }
+
+    orchestrator.addPrePhaseStartHook([this](Orchestrator* o) {
+        for (auto&& rl : this->_rateLimiters) {
+            rl.second->resetLastEmptied();
+        }
+    });
```
rtimmons: The limiters won’t change after startup so I think better to iterate outside and call addHook once for each limiter

src/gennylib/test/GlobalRateLimiter_test.cpp
```diff
@@ -73,43 +73,60 @@ TEST_CASE("Global rate limiter") {
         REQUIRE(grl.consumeIfWithinRate(now));
         REQUIRE(!grl.consumeIfWithinRate(now));
     }
+
+    SECTION("Can math") {
+        // int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX))))
+    }
```
rtimmons: Kill

guoyr:
I'm going to comment out the new rate limiter test for now. It segfaults when run in conjunction with the previous tests. I don't think there's anything wrong with the test case, but something lower in the stack doesn't like to be constructed twice.

rtimmons:
>  It segfaults when run in conjunction with the previous tests

I looked into this. A missing set curly brackets caused a segfault and a failing `REQUIRE()` before a `Thread::join()` caused another. (ubsan saved my sanity here 😹). I fixed both and added a couple 💅 changes as well after reverting your latest commit. Please feel to merge once you've taken a  look and agree or push back wherever makes sense.

guoyr:
sys perf patch: https://evergreen.mongodb.com/version/5d373455c9ec446d08a805b8

PR #259, opened 2019-07-17
Unique indexes

dalyd:
I made this workload for https://jira.mongodb.org/browse/BF-10848. We will probably add a microbenchmark instead of a sys-perf benchmark for it ultimately, so this probably doesn't get used in production (unless we use this instead of a mongo-perf in there). 

That said, do you guys want to accumulate workloads like this in the workloads directory, or only mainly get the ones that go into production? Let me know. For now I'm putting it up here for discussion. 

Also, it was interesting to make -- I have a lot of actors to keep the results straight. I would have preferred to have 2 actors total, and differentiate the results by phase. I suspect that are things that could make this workload a lot more concise.

### DIFF
--- a/src/workloads/scale/UniqueIndexStress.yml
+++ b/src/workloads/scale/UniqueIndexStress.yml
@@ -1,9 +1,14 @@
 # This workload is developed as ... FILL ME IN
+# This workload tests insert performance with unique indexes. Each pair of phases first, drops the
+# database, then creates 7 indexes, before inserting documents as fast as it can. The difference
+# between the different phases is how many of the indexes are unique. It first does 0 unique
+# secondary indexes, then 1, 2, up to 7.
 
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 
 string_length: &string_length 10
+insert_threads: &insert_threads 1
 
 Actors:
 - Name: Setup
@@ -22,7 +27,18 @@ Actors:
         indexes:
         - key: {a: 1}
           name: a
-          unique: true
+        - key: {b: 1}
+          name: b
+        - key: {c: 1}
+          name: c
+        - key: {d: 1}
+          name: d
+        - key: {e: 1}
+          name: e
+        - key: {f: 1}
+          name: f
+        - key: {g: 1}
+          name: g
   - &Nop {Nop: true}
   - Repeat: 1
     Database: &db test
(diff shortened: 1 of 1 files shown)


guoyr:
I chatted with David offline about the large number of Insert actors; it looks like merging is not part of the YAML spec and is unsupported in yaml-cpp: https://github.com/jbeder/yaml-cpp/issues/41

We can use ExternalPhaseConfig to get around the problem; something like:
```
# Create src/phases/scale/InsertPhase.yml
InsertMany: # Feel free to change this name of this key.
    Collection: {^Parameter: {Name: "CollName"}}
    Repeat: 10000
    Operations:
    - OperationName: {^Parameter: {Name: "MetricName"}}
      OperationCommand:
        Documents:
        - &doc
          a: &rs {^RandomString: {length: *string_length}}
          b: *rs
          c: *rs
          d: *rs
          e: *rs
          f: *rs
          g: *rs
          h: *rs
          i: *rs
          j: *rs
        - *doc
        - *doc
        - *doc
        - *doc
        - *doc
        - *doc
        - *doc
        - *doc
        - *doc
```

Then in the workload file:
```
- Name: UniqueIndexes
  Type: CrudActor
  Database: *db
  Threads: *insert_threads
  Phases:
  - *Nop
  - ExternalPhaseConfig:
      Path: ../../phases/scale/InsertPhase.yml
      Key: InsertMany
      Parameters:
        CollName: *coll
        MetricName: Insert1
  - *Nop
  - ExternalPhaseConfig:
      Path: ../../phases/scale/InsertPhase.yml
      Key: InsertMany
      Parameters:
        CollName: *coll
        MetricName: Insert2
  # ... and so on
```

dalyd:
I'll give that a try. Thanks.

dalyd:
Doh. I reconfigured it as Robert suggested, and realized that the Crud Actor doesn't support OperationMetricsName. I took a very quick look, and it wasn't instantly obvious to me for stuff to copy from one actor to the other to copy the functionality over.

guoyr:
> Doh. I reconfigured it as Robert suggested, and realized that the Crud Actor doesn't support OperationMetricsName. I took a very quick look, and it wasn't instantly obvious to me for stuff to copy from one actor to the other to copy the functionality over.

Ah good catch. I think this was an oversight when implementing the crud actor since the RunCommand actor does support custom metric names. Could you make the following two line change to add support for `OperationMetricsName`? I verified that it works manually with `CrudActor.yml`

```
diff --git a/src/cast_core/src/CrudActor.cpp b/src/cast_core/src/CrudActor.cpp
index 36b6a14..a75149f 100644
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1195,7 +1195,9 @@ struct CrudActor::PhaseConfig {
           metrics{actorContext.operation("Crud", id)} {
         auto addOperation = [&](const Node& node) -> std::unique_ptr<BaseOperation> {
             auto& yamlCommand = node["OperationCommand"];
             auto opName = node["OperationName"].to<std::string>();
+            auto metricName = node["OperationMetricsName"].maybe<std::string>().value_or(opName);
             auto onSession = false;
             if (yamlCommand) {
                 onSession = yamlCommand["OnSession"] && yamlCommand["OnSession"].to<bool>();
@@ -1211,7 +1213,7 @@ struct CrudActor::PhaseConfig {
             return createOperation(yamlCommand,
                                    onSession,
                                    collection,
-                                   actorContext.operation(opName, id),
+                                   actorContext.operation(metricName, id),
                                    phaseContext,
                                    id);
         };
```

rtimmons:
LGTM.

You could use the PhaseRange syntax for all the nops:

```yaml
Phases:
- Phase: 0..7
  Nop: true
```

imho this ends up making it clearer than the external config alternative.

>  We will probably add a microbenchmark instead of a sys-perf benchmark for it ultimately, so this probably doesn't get used in production (unless we use this instead of a mongo-perf in there).

That's fine. FWIW genny runs on microbenchmarks...

> do you guys want to accumulate workloads like this in the workloads directory

I'd like them to be in the workloads directory if they're "good" workloads that we'd like to be used as examples. Plus keeping them in the workloads dirs ensures (at least a bit) that Actors don't break them since we dry-run all workloads as part of CI.

This said if it's not going to be used in production I'd advocate to either put them in `workloads/docs` or to mention in the comments that the workload isn't in production and indicate _why_ it's not in production so somebody doesn't use them as an end-to-end example of what a good production workload is.

guoyr:
> You could use the PhaseRange syntax for all the nops:

+1 to using the nop syntax.

PR #260, opened 2019-07-23
TIG-1847 global rate limiter should burst across all available threads

guoyr:
I think the overall approach is correct but I need to fix up the perf tests at least and maybe some edge cases.

(FYI I'm putting up this PR after Ryan showed me Luke's WT caching actor and we determined that he needed this functionality)

### DIFF
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -138,11 +138,6 @@ public:
         }
     }
 
-    constexpr bool shouldLimitRate(int64_t currentIteration) const {
-        // Only rate limit if the current iteration is a multiple of the burst size.
-        return _rateLimiter && (currentIteration % _rateLimiter->getBurstSize() == 0);
-    }
-
     constexpr void limitRate(const SteadyClock::time_point referenceStartingPoint,
                              const int64_t currentIteration,
                              const Orchestrator& o,
@@ -151,7 +146,7 @@ public:
         // first iteration. This means the number of completed operations is always
         // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
         // _burstSize. `m` here is the number of threads using the rate limiter.
-        if (shouldLimitRate(currentIteration)) {
+        if (_rateLimiter) {
             while (true) {
                 const auto now = SteadyClock::now();
                 auto success = _rateLimiter->consumeIfWithinRate(now);
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -140,7 +140,6 @@ public:
 
     constexpr void limitRate(const SteadyClock::time_point referenceStartingPoint,
                              const int64_t currentIteration,
-                             const Orchestrator& o,
                              const PhaseNumber inPhase) {
         // This function is called after each iteration, so we never rate limit the
         // first iteration. This means the number of completed operations is always
@@ -260,7 +259,7 @@ public:
         if (_iterationCheck) {
             _iterationCheck->sleepBefore(*_orchestrator, _inPhase);
             _iterationCheck->limitRate(
-                _referenceStartingPoint, _currentIteration, *_orchestrator, _inPhase);
+                _referenceStartingPoint, _currentIteration, _inPhase);
         }
         // clang-format off
         return
--- a/src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
+++ b/src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
@@ -40,10 +40,10 @@ namespace genny::v1 {
  * a subset of threads. Coordinating across multiple global rate limiters
  * is currently not supported.
  *
- * 2. The caller is expected to be "nice" and respect the burst size. The
- * rate limiter itself does not know anything about the rate-limited functionality
- * or enforce any behavior about it. This is different to the non-global RateLimiter,
- * which invokes the rate-limited callback function itself.
+ * 2. The burst size should either be 1, or roughly equal to the number of
+ * actors using this rate limiter. If you have a large number of threads
+ * (using this rate limiter) but a small burst size and a high frequency rate,
+ * you may experience bad performance.
  *
  * Inspired by
  * https://github.com/facebook/folly/blob/7c6897aa18e71964e097fc238c93b3efa98b2c61/folly/TokenBucket.h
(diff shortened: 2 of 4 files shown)


src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -84,14 +84,28 @@ class BaseGlobalRateLimiter {
      * appropriate back-off strategy if this function returns false.
      */
     bool consumeIfWithinRate(const typename ClockT::time_point& now) {
+
+        int64_t optimisticallyConsume;
+        if (_burstSize > 1) {
+            optimisticallyConsume = _curBurstCount++;
+            if (optimisticallyConsume % _burstSize == 0) {
```
rtimmons: 🤔 why `%`?

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -138,11 +138,6 @@ class IterationChecker final {
...
     }
 
-    constexpr bool shouldLimitRate(int64_t currentIteration) const {
-        // Only rate limit if the current iteration is a multiple of the burst size.
-        return _rateLimiter && (currentIteration % _rateLimiter->getBurstSize() == 0);
-    }
-
     constexpr void limitRate(const SteadyClock::time_point referenceStartingPoint,
                              const int64_t currentIteration,
                              const Orchestrator& o,
```
rtimmons: `Orchestrator` is no longer needed 🎉

rtimmons:
This checks out with real but simple HelloWorld test-cases.

(I thought I found 2 issues but they seem mutually exclusive..the first issue is only sporadic and turning off boost logging makes the second go away entirely...keeping text here for posterity but no need to read)

---------------------------------------- 

Reference YAML:

```yaml
SchemaVersion: 2018-07-01
Owner: "@mongodb/stm"

Actors:
- Name: HelloWorld
  Type: HelloWorld
  Threads: 10
  Phases:
  - Message: P0
    Duration: 55 seconds
    GlobalRate: 5 per 10 seconds
  - Message: P1
    Repeat: 1
```

**First issue**
 
There is a funny "stutter" of about a second on issuing tokens in iterations after the first iteration. Is this the "Don't sleep for more than 1 second (1e9 nanoseconds)" condition coming in where it shouldn't be?

```
[00:15:30.798] [T1] Actor #9 says: P0
[00:15:30.798] [T2] Actor #4 says: P0
[00:15:30.798] [T3] Actor #1 says: P0
[00:15:30.798] [T4] Actor #2 says: P0
[00:15:30.798] [T5] Actor #3 says: P0

[00:15:40.808] [T6] Actor #5 says: P0
[00:15:40.917] [T7] Actor #8 says: P0
[00:15:41.734] [T8] Actor #0 says: P0 (*)
[00:15:41.758] [T9] Actor #7 says: P0
[00:15:41.785] [T0] Actor #6 says: P0

[00:15:50.968] [T4] Actor #2 says: P0
[00:15:51.015] [T2] Actor #4 says: P0 (*)
[00:15:51.739] [T3] Actor #1 says: P0
[00:15:51.750] [T5] Actor #3 says: P0
[00:15:51.802] [T1] Actor #9 says: P0

[00:16:00.871] [T7] Actor #8 says: P0
[00:16:01.654] [T8] Actor #0 says: P0 (*)
[00:16:01.766] [T9] Actor #7 says: P0
[00:16:01.795] [T6] Actor #5 says: P0
[00:16:01.827] [T0] Actor #6 says: P0

[00:16:10.819] [T1] Actor #9 says: P0
[00:16:10.857] [T2] Actor #4 says: P0
[00:16:11.042] [T4] Actor #2 says: P0 (*)
[00:16:11.685] [T5] Actor #3 says: P0
[00:16:11.711] [T3] Actor #1 says: P0

[00:16:20.803] [T8] Actor #0 says: P0
[00:16:20.875] [T6] Actor #5 says: P0
[00:16:20.985] [T0] Actor #6 says: P0
[00:16:21.063] [T7] Actor #8 says: P0 (*)
[00:16:21.818] [T9] Actor #7 says: P0

[00:16:32.822] [T3] Actor #1 says: P1
....
```

**Second Issue**

It seems the bucket isn't full (er..empty) at the start of the first phase.

I added the following to `IterationChecker::limitRate`:

```c
const auto sleepDur = std::chrono::nanoseconds(
    int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX))));
BOOST_LOG_TRIVIAL(info) << "Sleeping " << sleepDur.count() << "ns to wait for tokens";
```

And this is how it reliably starts:

```
[00:28:12.802] [T0]  Sleeping 1025560532ns to wait for tokens
[00:28:12.802] [T1]  Sleeping 995865013ns to wait for tokens
[00:28:12.802] [T2]  Actor #9 says: P0
[00:28:12.803] [T3]  Actor #4 says: P0
[00:28:12.803] [T4]  Actor #2 says: P0
[00:28:12.803] [T5]  Sleeping 1003276723ns to wait for tokens
[00:28:12.803] [T6]  Sleeping 963153778ns to wait for tokens
[00:28:12.803] [T7]  Actor #6 says: P0
[00:28:12.803] [T8]  Sleeping 950000782ns to wait for tokens
[00:28:12.803] [T9]  Actor #0 says: P0
```

guoyr:
Thanks for the quick reply 😄 

Re your last comment:

> There is a funny "stutter" of about a second . Is this the "Don't sleep for more than 1 second (1e9 nanoseconds)" condition.

I think it's related; it's a combination of the 1sec sleep and the 5% jitter. If you compute the stutter for the lines you marked with asterisks, the gap isn't always around 1sec. They are, from top to bottom, in seconds: 0.817, 0.047, 0.173, 0.185, 0.078.

These numbers are all within (5% * number of iterations) of the timespec in the rate (10sec).

> It seems the bucket isn't full

We allow the number part of the global rate to run once at the start, which is 5. So I think what you're seeing is expected?

rtimmons:
> We allow the number part of the global rate to run once at the start, which is 5. So I think what you're seeing is expected?

I was seeing "Sleeping 1025560532ns to wait for tokens" as the very first thing any Actor did

(and fwiw I'm okay to ship this as-is - that behavior wasn't super repeatable and all but went away when I removed logging...)

guoyr:
> I was seeing "Sleeping 1025560532ns to wait for tokens" as the very first thing any Actor did

I'm reliably seeing the results you pasted above with the counter immediately jumping to 5. Maybe you're seeing the "Sleeping..." message that gets printed at the beginning of the second iteration by the 5 actors that ran the first iteration?

Here's my git diff
```
diff --git a/src/gennylib/include/gennylib/PhaseLoop.hpp b/src/gennylib/include/gennylib/PhaseLoop.hpp
index 53f7933..0fa6450 100644
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -138,7 +138,7 @@ public:
         }
     }
 
-    constexpr void limitRate(const SteadyClock::time_point referenceStartingPoint,
+    const void limitRate(const SteadyClock::time_point referenceStartingPoint,
                              const int64_t currentIteration,
                              const Orchestrator& o,
                              const PhaseNumber inPhase) {
@@ -158,8 +158,10 @@ public:
                     const auto rate = _rateLimiter->getRate() > 1e9 ? 1e9 : _rateLimiter->getRate();
 
                     // Add ±5% jitter to avoid threads waking up at once.
-                    std::this_thread::sleep_for(std::chrono::nanoseconds(
-                        int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX)))));
+                    const auto sleepDur = std::chrono::nanoseconds(
+                        int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX))));
+                    BOOST_LOG_TRIVIAL(info) << "Sleeping " << sleepDur.count() << "ns to wait for tokens";
+                    std::this_thread::sleep_for(std::chrono::nanoseconds(sleepDur));
                     continue;
                 }
                 break;
```

PR #261, opened 2019-07-29
TIG-1745

luke-pearson:
Initial PR, feel free to tear it to pieces.

I think the workload is still subject to change as it's not been tested in evergreen.

Cheers,
Luke


src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+        collectionNames = CollectionScanner::getCollectionNames(collectionCount, threads, actorId);
+    }
+};
+
+void CollectionScanner::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            if (config->skipFirstLoop) {
+                config->skipFirstLoop = false;
+            } else {
```
rtimmons: consider

```cpp
if (skip) { 
  skip = true;
  continue;
}
```
this would eliminate the `else` block indentation

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+                try {
+                    _runningActorCounter ++;
+                    BOOST_LOG_TRIVIAL(info) << "Starting collection scanner id:" << config->actorId;
```
rtimmons: consider adding a "Ending collection scanner id" log message

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+                    // Count over all collections this thread has been tasked with scanning each.
+                    auto statTracker = config->scanOperation.start();
+                    for (int i = 0; i < config->collectionNames.size(); ++i){
+                        statTracker.addDocuments(config->database[config->collectionNames[i]].count_documents({}));
+                    }
+                    _runningActorCounter --;
+                    statTracker.success();
+                } catch(mongocxx::operation_exception& e) {
```
rtimmons: if the `count_documents` throws this would leave teh `statTracker` in an undefined state - if there's an exception we want to record that data.

```c++
auto statTracker = start();
try {
  for (...) {...}
  statTracker.success();
} catch() { statTracker.failure(); }
```

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+            } else {
+                try {
+                    _runningActorCounter ++;
+                    BOOST_LOG_TRIVIAL(info) << "Starting collection scanner id:" << config->actorId;
+                    // Count over all collections this thread has been tasked with scanning each.
+                    auto statTracker = config->scanOperation.start();
+                    for (int i = 0; i < config->collectionNames.size(); ++i){
+                        statTracker.addDocuments(config->database[config->collectionNames[i]].count_documents({}));
+                    }
+                    _runningActorCounter --;
```
rtimmons: I think you want this outside of your `try` loop so an exception inside `count_documents` doesn't cause the counter to get off-kilter

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+            if (config->skipFirstLoop) {
+                config->skipFirstLoop = false;
+            } else {
+                try {
+                    _runningActorCounter ++;
+                    BOOST_LOG_TRIVIAL(info) << "Starting collection scanner id:" << config->actorId;
+                    // Count over all collections this thread has been tasked with scanning each.
+                    auto statTracker = config->scanOperation.start();
+                    for (int i = 0; i < config->collectionNames.size(); ++i){
+                        statTracker.addDocuments(config->database[config->collectionNames[i]].count_documents({}));
```
rtimmons: 1. You can use range-based loops here

    ```cpp
    for (const auto&& collectionName : this->config->collectionNames)`
    ```

2. You may also consider having a `std::vector<collection>` rather than `std::vector<string>` and doing the `database[collectionName]` operation at Actor-setup time. This would save the time it takes to allocate and deallocate the `mongocxx::collection` type. Not a huge deal but probably an easy change nonetheless.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    PhaseConfig(PhaseContext& context, const mongocxx::database& db, ActorId id, int collectionCount,
+        int threads, ActorCounter &counter)
+        : database{db}, readOperation{context.operation("Read", id)},
+        readWithScanOperation{context.operation("ReadWithScan", id)} {
+        int actorId = counter;
+        counter ++;
+        // Distribute the collections among the actors.
+        collectionNames = CollectionScanner::getCollectionNames(collectionCount, threads, actorId);
+        // Setup the int distribution.
+        integerDistribution = std::uniform_int_distribution(0, (int)collectionNames.size() - 1);
```
rtimmons: use `boost::random::uniform_int_distribution` since it gives the same values on all platforms.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+};
+
+void RandomSampler::run() {
+    for (auto&& config : _loop) {
+        // Construct basic pipeline for retrieving one random record.
+        mongocxx::pipeline pipeline{};
+        pipeline.sample(10);
```
rtimmons: This could be done in the PhaseContext ctor

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+        for (const auto&& _ : config) {
+            try {
+                auto statTracker = _collectionScannerCounter > 0 ? config->readWithScanOperation.start() :
+                  config->readOperation.start();
+                int index = config->collectionNames.size() > 1 ? config->integerDistribution(config->random) : 0;
+                auto cursor = config->database[config->collectionNames[index]].aggregate(pipeline,
+                    mongocxx::options::aggregate{});
+                for (auto doc : cursor){
+                  ;
```
rtimmons: consider adding the number of docs and/or bytes you read to the statTracker.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+}
+
+RandomSampler::RandomSampler(genny::ActorContext& context)
+    : Actor{context},
+      _client{context.client()},
+      _actorCounter{WorkloadContext::getActorSharedState<RandomSampler, ActorCounter>()},
+      _collectionScannerCounter{WorkloadContext::getActorSharedState<CollectionScanner,
+        CollectionScanner::RunningActorCounter>()},
```
rtimmons: consider var name `_activeCollectionScannerInstances` instead of `_collectionScannerCounter` and consider having it be the same name in the `CollectionScanner` for consistency.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <gennylib/context.hpp>
+
+#include <value_generators/DocumentGenerator.hpp>
+
+namespace genny::actor {
+
+struct RandomSampler::PhaseConfig {
+    mongocxx::database database;
+    std::vector<std::string> collectionNames;
+    genny::DefaultRandom random;
```
rtimmons: Here you'll always initialize the rng with the default seed so all instances/phases will produce the exact same sequence of values. I think you probably want to pipe through the result of `WorkloadContext::getRNGForThread(ActorId)`. Since Actors are single-threaded the `genny::DefaultRandom` probably wants to live on the Actor type rather than ont he `PhaseConfig` which makes this a bit easier since you can just initialize it in your Actor's ctor.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+    std::uniform_int_distribution<> integerDistribution;
+    metrics::Operation readOperation;
+    metrics::Operation readWithScanOperation;
+    PhaseConfig(PhaseContext& context, const mongocxx::database& db, ActorId id, int collectionCount,
+        int threads, ActorCounter &counter)
+        : database{db}, readOperation{context.operation("Read", id)},
+        readWithScanOperation{context.operation("ReadWithScan", id)} {
+        int actorId = counter;
+        counter ++;
```
rtimmons: I think it's worth some comments on how you use these variables in particular how the two different `Operation` instances are used in conjunction with `_collectionScannerCounter`

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                int index = config->collectionNames.size() > 1 ? config->integerDistribution(config->random) : 0;
+                auto cursor = config->database[config->collectionNames[index]].aggregate(pipeline,
+                    mongocxx::options::aggregate{});
+                for (auto doc : cursor){
+                  ;
+                }
+                statTracker.success();
+            } catch(mongocxx::operation_exception& e) {
+                BOOST_LOG_TRIVIAL(error) << boost::diagnostic_information(e);
+            }
```
rtimmons: similar comment to other actor about the `try/catch` block and the tracker calling `.failure()`

src/cast_core/src/RandomSampler.cpp
```diff
@@ -0,0 +1,99 @@
...
+                  config->readOperation.start();
+                int index = config->collectionNames.size() > 1 ? config->integerDistribution(config->random) : 0;
+                auto cursor = config->database[config->collectionNames[index]].aggregate(pipeline,
+                    mongocxx::options::aggregate{});
+                for (auto doc : cursor){
+                  ;
+                }
+                statTracker.success();
+            } catch(mongocxx::operation_exception& e) {
+                BOOST_LOG_TRIVIAL(error) << boost::diagnostic_information(e);
```
rtimmons: I don't know enough about the driver to know when this exception would be thrown but it seems like it may be just a normal course of business (the exception is expected) in which case I don't know if logging it is all that helpful.

I'd either

1. log it at debug level (which is off by default iirc) or
2. re-throw it (after calling `tracker.failure()`) if you're not expecting this exception to happen in the normal flow of the workload. Rethrowing it will cause the workload to end asap and will report the workload as a system failure in evergreen. (Syntax to re-throw is just `throw;`..I learned that only somewhat recently lol)

And you can catch by const-ref `catch(const mongocxx::operation_exception& e)`.

I think the same likely applies to the other Actor as well.

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include <metrics/metrics.hpp>
+
+namespace genny::actor {
+
+/**
+ * This actor will scan all collections it is tasked with.
+ *
+ * Owner: Storage Engines
+ */
```
rtimmons: We don't have strong guidance here but an example of the yaml would be helpful. Alternatively you could create an example yaml in the `src/workloads/docs` directory that uses both of your new Actors and just mention that yaml in the comments. Your call here if you think these Actors would get use in other workloads.

src/cast_core/include/cast_core/actors/RandomSampler.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <gennylib/Actor.hpp>
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+
+#include <metrics/metrics.hpp>
+
+namespace genny::actor {
+
+/**
+ * This actor will sample 10 documents from the collections it is tasked with
```
rtimmons: is it always 10?

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
luke-pearson: Added in comment pointing to workload/docs

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+    struct RunningActorCounter : genny::WorkloadContext::ShareableState<std::atomic_int> {};
+
+    static std::vector<std::string> getCollectionNames(int collectionCount, int threadCount, int actorId) throw (){
+        //We always want a fair division of collections to actors
+        std::vector<std::string> collectionNames{};
+        if ((threadCount > collectionCount && threadCount % collectionCount != 0) || collectionCount % threadCount != 0){
+            throw std::invalid_argument("Thread count must be mutliple of database collection count");
+        }
+        int collectionsPerActor = threadCount > collectionCount ? 1 : collectionCount / threadCount;
+        int collectionIndexStart = (actorId % collectionCount) * collectionsPerActor;
```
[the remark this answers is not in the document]
luke-pearson: Double allocating is fine.

src/cast_core/include/cast_core/actors/RandomSampler.hpp
luke-pearson: Yes currently not planning on making it configurable.

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <value_generators/DocumentGenerator.hpp>
+
+
+namespace genny::actor {
+
+struct CollectionScanner::PhaseConfig {
+    mongocxx::database database;
+    std::vector<std::string> collectionNames;
+    bool skipFirstLoop = false;
+    metrics::Operation scanOperation;
```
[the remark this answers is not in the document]
luke-pearson: Actually do prefer keeping Operation in the name, because if you have line like:
config->readOperation.start();
vs
config->read.start();

Its a little clearer with operation in the name what is actually happening.

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * This actor will scan all collections it is tasked with.
+ *
+ * Owner: Storage Engines
+ */
+class CollectionScanner : public Actor {
+    struct ActorCounter : genny::WorkloadContext::ShareableState<std::atomic_int> {};
+public:
+    struct RunningActorCounter : genny::WorkloadContext::ShareableState<std::atomic_int> {};
+
+    static std::vector<std::string> getCollectionNames(int collectionCount, int threadCount, int actorId) throw (){
```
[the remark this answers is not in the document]
luke-pearson: I tried moving this to the .cpp file but RandomSampler didn't seem to like it so I moved it back to CollectionScanner.hpp for the time being.

src/cast_core/src/CollectionScanner.cpp
luke-pearson: 💯💯💯💯💯💯💯👍

src/cast_core/src/CollectionScanner.cpp
luke-pearson: ✔

luke-pearson:
I've fixed the majority of your suggestions and I've run clang format. Currently patch testing isn't passing for something that appears unrelated. I'll try rerunning it. @rtimmons Could you please re-review if necessary and then merge?

PR #263, opened 2019-08-14
Fix use of shared state in CollectionScanner and RandomSampler

rtimmons:
The shared-state counter was being incremented for each PhaseConfig constructor call not for each Actor ctor call. I think this fix also resolves the weird `.2` suffix thing.


luke-pearson:
evergreen merge

luke-pearson:
Thanks for fixing this up for me! Definitely appreciate it.

PR #267, opened 2019-08-23
TIG-1894 scanner

luke-pearson:
This PR improves the CollectionScanner to meet the criteria of the LSWA. I've added a few configurations, the index specification is met using the Filter configuration. I think the read at a snapshot configuration is met by specifying a begin transaction with read concern majority but I'm not 100% certain on that.


luke-pearson:
Closing as I'm removing unnecessary commits.

PR #270, opened 2019-08-23
TIG-1888 TIG-1889 TIG-1890 Implement LongLived Actors using existing Actor types

ddanderson:
I started to write my own Creator class (which was a good exercise), but ultimately decided that the standard classes probably suffice.

None of these implement "Prefix" from the design document - which was a way to prefix collection names so that multiple copies of these can run simultaneously and not overlap, but it looks like the same intention can be achieved by using different database names.

And the design document lists an "Index" parameter to specify that the Reader uses an index.  I just created different actors that query based on an index name vs. an _id.

These seemed like reasonable simplifications given the benefits of using existing actors.

Note that LongLivedCreator is identical between the three files, I don't know if there is a way to "include" or otherwise share that code.

### DIFF
--- a/src/workloads/docs/LongLivedCreator.yml
+++ b/src/workloads/docs/LongLivedCreator.yml
@@ -1,0 +1,43 @@
+SchemaVersion: 2018-07-01
+Owner: Storage Engines
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 400
+
+Actors:
+- Name: LongLivedCreator
+  Type: Loader
+  Threads: 10
+  Phases:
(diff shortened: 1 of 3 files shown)


ddanderson:
@luke-pearson, these are lightly tested; I'm putting this up so you might have a chance to look and comment before the weekend.

I originally started to write my own Creator class (which was a good exercise), but ultimately decided that the standard classes probably suffice.

luke-pearson:
Sorry I didn't get a chance to look at this on friday, looks good to me. I also agree with removing the prefix, and assuming it'll use indexes if specified correctly.

PR #273, opened 2019-08-28
TIG-1888, TIG-1889, TIG-1890, TIG-1892, TIG-1893, TIG-1894, Large scaleworkload automation initial workload actors

luke-pearson:
@rtimmons We're keen to get a basic workload into production with the loader actor and maybe the collection scanner. As such I've clang formatted the branch and created this pull request so that @ddanderson can begin creating patch builds with the new workload. We'll still keep this branch around though for future merges.


luke-pearson:
@rtimmons I've fixed up the three comments although I think I might be doing the exceptions caught metric wrong? Additionally I've fixed the file permissions on a bunch of files in genny that were being set incorrectly. I think we're going to want to squash this branch in a with a fairly meaningful commit name, I'm hoping that won't make future merges ugly?

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
```diff
@@ -50,16 +50,17 @@ class CollectionScanner : public Actor {
     static std::string_view defaultName() {
         return "CollectionScanner";
     }
+    struct PhaseConfig;
```
rtimmons: I doubt this needs to be public?

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -34,31 +31,127 @@
...
+                 distributeCollectionNames(collectionCount, threads, actor->_index)) {
+                collections.push_back(db[collectionName]);
+            }
+        } else {
+            queryCollectionList = true;
         }
     }
 };
 
+void collectionScan(genny::v1::ActorPhase<CollectionScanner::PhaseConfig>& config,
```
rtimmons: hrm...having to spell out a `v1` type is definitely not something we wanted you to have to do. I *think* you can instead have this be a `CollectionScanner::PhasConfig&` and pass it by calling `collectionScan(*config, collections)`.

(Similar comment applies to other functions doing the same thing)

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -34,31 +31,127 @@
...
+                    scanFinished = true;
+                    break;
+                }
+            }
+            if (scanFinished) {
+                break;
+            }
+        } catch (const mongocxx::operation_exception& e) {
+            auto exceptionsCaught = config->exceptionsCaught.start();
+            exceptionsCaught.addDocuments(1);
```
rtimmons: I think start `exceptionsCaught` at the same time you start `statTracker` and just call `addDocuments(1)` inside this `catch` block. This way you know how long it took for the start of the iteration to occur and how many (in total) exceptions you got.

(Same thing for the other places where this is being done)

src/cast_core/src/Deleter.cpp
```diff
@@ -0,0 +1,71 @@
...
+void Deleter::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto statTracker = config->deleteOperation.start();
+            /*
+             * This will delete the oldest document as .find() returns documents
+             * in order of _id.
+             */
+            config->collection.find_one_and_delete({});
+            statTracker.success();
```
rtimmons: `addDocuments(1)` as well
and while you're there maybe only do that if `find_one_and_delete` actually did its thing?

```c++
const auto result = config->find_one_and_delete({});
if (result) {
    ctx.addDocuments(1);
    ctx.addBytes(result->view().length());
}
/* do you want to else { statTracker.failure() } here? idk. your call */ }
statTracker.success();
```

luke-pearson:
evergreen retry

src/cast_core/include/cast_core/actors/CollectionScanner.hpp
luke-pearson: it does otherwise you can't access it in the genny::actor scope.

src/cast_core/src/CollectionScanner.cpp
luke-pearson: Seems to blow things up?

luke-pearson:
@rtimmons it looks like the evergreen testing is failing incorrectly again, we'll have to leave it to you to squash and merge.

rtimmons:
@luke-pearson actually the test was failing due to an undefined-behavior bug (dtors for `RunOperation` weren't virtual). I've fixed that. Once CI comes back clean here I'll merge it.

PR #274, opened 2019-08-29
PERF-1911 Test million documents transaction

BlakeIsBlake:
https://evergreen.mongodb.com/version/5d5efc8fa4cf471f304cdc9f (Patch name is under the SERVER ticket, not the PERF ticket I later created)

### DIFF
--- a/src/workloads/scale/UpdateMillionDocsInTxn.yml
+++ b/src/workloads/scale/UpdateMillionDocsInTxn.yml
@@ -47,7 +47,7 @@ Actors:
   Phases:
   - *Nop
   - MetricsName: TransactionUpdate
-    Repeat: 1
+    Repeat: 5
     Collection: &Coll Collection0 # This is the default collection populated by the Loader.
     Operations:
     - OperationName: startTransaction
--- a/src/workloads/scale/UpdateMillionDocsInTxn.yml
+++ b/src/workloads/scale/UpdateMillionDocsInTxn.yml
@@ -1,0 +1,78 @@
+# This workload is developed to test the amount of time it takes to update a million documents in
+# a single replica set transaction. At the moment, the average time taken raises above the default
+# limit, so until we add a way to manually increase a transaction's lifetime, we must raise the
+# lifetime of all transactions.
+
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+
+Actors:
+- Name: Loader
+  Type: Loader
+  Threads: 1
+  Phases:
+  - ExternalPhaseConfig:
+      Path: &PhasePath ../../phases/scale/DesignDocWorkloadPhases.yml
+      Key: Loader
+      Parameters:
+        CollectionCount: 1
+        DocumentCount: 1e6
+        BatchSize: 100000
+    Repeat: 1
+    Threads: 1
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: RaiseTransactionLifetimeLimit
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Operation:
+      OperationName: RunCommand
+      OperationCommand:
+        setParameter: 1
+        transactionLifetimeLimitSeconds: 120
+  - *Nop
+  - Repeat: 1
+    Operation:
+      OperationName: RunCommand
+      OperationCommand:
+        setParameter: 1
+        transactionLifetimeLimitSeconds: 60
+
+- Name: UpdateDocumentsInTransaction
+  Type: CrudActor
+  Database: test
+  Phases:
+  - *Nop
+  - MetricsName: TransactionUpdate
+    Repeat: 1
+    Collection: &Coll Collection0 # This is the default collection populated by the Loader.
+    Operations:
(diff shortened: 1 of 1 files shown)


src/workloads/scale/UpdateMillionDocsInTxn.yml
```diff
@@ -0,0 +1,78 @@
...
+      OperationCommand:
+        setParameter: 1
+        transactionLifetimeLimitSeconds: 120
+  - *Nop
+  - Repeat: 1
+    Operation:
+      OperationName: RunCommand
+      OperationCommand:
+        setParameter: 1
+        transactionLifetimeLimitSeconds: 60
```
rtimmons: this 1 operation will happen concurrently with the `UpdateDocumentsInTransaction` in Phase index 1. Do you want this to happen before you start that Phase? If so you could have a separate Phase that just does this `setParameter` operation.

src/workloads/scale/UpdateMillionDocsInTxn.yml
```diff
@@ -0,0 +1,78 @@
...
+
+- Name: UpdateDocumentsInTransaction
+  Type: CrudActor
+  Database: test
+  Phases:
+  - *Nop
+  - MetricsName: TransactionUpdate
+    Repeat: 1
```
rtimmons: Are you sure you only want to repeat this 1 time? I'd be worried about how noisy this may be. More typically we'll see at least 100-1k repeats. It's okay if each iteration takes a while we want good data. Your call here but @dalyd may have more input.

dalyd: How long does the transaction take, and how repeatable are the results?

BlakeIsBlake: Each transaction takes about 90 seconds to run... not sure how many times would be acceptable for this test.

src/workloads/scale/UpdateMillionDocsInTxn.yml
```diff
@@ -0,0 +1,78 @@
...
+      Parameters:
+        CollectionCount: 1
+        DocumentCount: 1e6
+        BatchSize: 100000
+    Repeat: 1
+    Threads: 1
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: RaiseTransactionLifetimeLimit
```
dalyd: Blake, is this likely to go away in the future based on Server changes? If so, I would consider taking this step out and putting it as a pre-set in DSI in the test config part. (pre task or pre test, or just mongodb_setup). I think that will be easier to control. It also feels like that is about the system under test, as opposed to the actual test you are running. We could evens script that to do something different based on version numbers. 

That said, if this is always to be needed, then no harm having it here.

src/workloads/scale/UpdateMillionDocsInTxn.yml
dalyd: Let's do a few of them (that's purposefully vague). Then we can get a distribution also. We want to keep tasks under 1 hour. We should be well below that now. If the test runs for a 5-10 minutes that seems good to me. Note that there are some processes at a 60 second granularity in the server, so doing a few minutes helps average that out.

src/workloads/scale/UpdateMillionDocsInTxn.yml
BlakeIsBlake: It seems to me like the first phase is just the loader and the admin command, while the `UpdateDocumentsInTransaction` is a no-op. Am I reading this wrong?

src/workloads/scale/UpdateMillionDocsInTxn.yml
BlakeIsBlake: Setting it to repeat five times.

src/workloads/scale/UpdateMillionDocsInTxn.yml
BlakeIsBlake: The current view is that re offline conversation about this ticket

https://jira.mongodb.org/browse/SERVER-42976

that we may not actually do it, so I'd prefer leaving this here, because it's specific to the transaction we're trying to run, and if we reach a transaction timeout in other tests, that should count as a failure.

BlakeIsBlake:
Pushed the only update so far, changing the transaction to repeat five times

src/workloads/scale/UpdateMillionDocsInTxn.yml
dalyd: That sounds reasonable. Do we need to make sure to unset the parameter afterwards? Will it survive a reboot if we run another test afterwards (like canaries).

BlakeIsBlake: We set the parameter back to the default value at the end of this task, in the third phase. And no, this would not survive a reboot. It's sourced from a config file or command line parameter on startup.

rtimmons:
@BlakeIsBlake (still) LGTM here - lmk when you're ready to merge and I'll press the button (not sure if you have write/merge permissions)

BlakeIsBlake:
LGTM to merge, thanks @rtimmons

PR #275, opened 2019-09-03
Metrics progress-reporting

rtimmons:
Still a bit wip as I tune this to find the right intervals.
Attempting to spit out log messages during metrics processing so it doesn't look like Genny's hanging.

Patch-build [here](https://evergreen.mongodb.com/version/5d6e9aeec9ec44233765fe26)


src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -295,6 +305,12 @@ class ReporterT final {
                         out << event.second.ops << ",";
                         out << event.second.errors << ",";
                         out << event.second.size << std::endl;
+
+                        if (++iter % 1000000 == 0 && logMode != LogMode::kNone) {
```
guoyr: Can we try adding another two 0s here? I would expect this loop to be at least a few dozen times faster than the loop that [writes out metrics to a file](https://github.com/mongodb/genny/blob/0b30c1585fd4b31dc5232e2bfa8e79473dbc3b7a/src/python/genny/parsers/cedar.py#L206) where we use 1e6

src/driver/src/DefaultDriver.cpp
```diff
@@ -178,7 +178,10 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
         std::ofstream metricsOutput;
         metricsOutput.open(options.metricsOutputFileName,
                            std::ofstream::out | std::ofstream::trunc);
-        reporter.report(metricsOutput, options.metricsFormat);
+        auto logMode = options.metricsOutputFileName == "/dev/stdout"
+            ? genny::metrics::LogMode::kNone
+            : genny::metrics::LogMode::kNormal;
```
guoyr: Could we call it `kCI` to further distinguish it from kNone?

PR #278, opened 2019-09-13
TIG-1891 Create simple workload for long lived actors for Large Scale Workload Automation

ddanderson:


ddanderson:
Closing, will redo this against master

PR #279, opened 2019-09-16
TIG-1940 tune workload

luke-pearson:
I've had a look at the perf.jsons in the relevant patch builds and they're looking good. As such I think this PR is good to merge pretty much straight away. 

The original workload was not performing as expected in prod causing it to have extremely varied throughputs.

This also fixes a bug where a sleep was in an incorrect location causing the out of cache scanner threads to act out of order.


luke-pearson:
evergreen merge

PR #281, opened 2019-09-20
TIG-1964 support collecting overlapping metrics

guoyr:
The metrics parsing code assumes each thread runs one op at a time. This is not always the case as we collect "nested" metrics. E.g. metrics for a single op in a bulk op, as well as the total for the entire bulk op.

There's a side effect of this change: since we don't know which metrics are nested, every time genny encounters a new metric, it needs to assume the duration equals the total, instead of using the latest timestamp on that thread from some previous operation

Will run a sys-perf patch once PR is approved.

### DIFF
--- a/src/python/genny/parsers/cedar.py
+++ b/src/python/genny/parsers/cedar.py
@@ -82,8 +82,8 @@ class IntermediateCSVReader:
         # dict(operation -> total_duration)
         self.total_for_op = defaultdict(int)
 
-        # dict(thread -> prev_op_ts))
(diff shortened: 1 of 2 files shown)


guoyr:
> I'm not sure I'd call this "nested" metrics

Makes sense! s/nested/overlapping

guoyr:
sys-perf PB: https://evergreen.mongodb.com/version/5d84f3eb57e85a721a5086a4

guoyr:
Sadness, GH didn't use the name of the PR as the squashed commit.

PR #283, opened 2019-09-24
Ability to directly record computed/synthetic metrics

rtimmons:
Still WIP - need to add some testing and re-jigger some types/aliases to be consistent - but wanted to get this out there for feedback on the API.

cc @luke-pearson


guoyr:
Merging to unblock Luke. API looks great. 

I created [TIG-1984](https://jira.mongodb.org/browse/TIG-1984) to add tests and "rejigger names"

PR #284, opened 2019-09-24
Add logging and make log-level configurable by command-line opts

rtimmons:
Spent a fair amount of time debugging a thing with @luke-pearson that could have been avoided if genny were more verbose with its logging at key lifecycle points.

- add begin/end logging around MetricsReporter::report
- do the same periodic output-logging when reporting `--metrics-format=csv` as in `--metrics-format=cedar-csv`
- add begin/end logging for each Phase
- make log level configurable by passing in the `-v $level` flag

### DIFF
--- a/src/metrics/include/metrics/MetricsReporter.hpp
+++ b/src/metrics/include/metrics/MetricsReporter.hpp
@@ -24,11 +24,6 @@
 
 namespace genny::metrics {
 
-enum class LogMode {
-    kCI,
-    kNone,
-};
-
 /**
  * @namespace genny::metrics::v1 this namespace is private and only intended to be used by genny's
  * own internals. No types from the genny::metrics::v1 namespace should ever be typed directly into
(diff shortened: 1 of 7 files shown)


src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -164,6 +166,13 @@ class ReporterT final {
                         out << ",";
                         out << getter(event.second);
                         out << std::endl;
+
+                        // Log progress every 100e6 iterations
+                        if (++iter % (100 * 1000 * 1000) == 0) {
+                            BOOST_LOG_TRIVIAL(info)
```
luke-pearson: My only real comment is why did you chose to make the other messages debug level and this one info? I'd say that phase level should be debug and overall level should be info.

rtimmons: We want to see this output on evergreen but we don't want debug logging on evergreen so info it is 😄

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -307,7 +315,7 @@ class ReporterT final {
                         out << event.second.size << std::endl;
 
                         // Log progress every 100e6 iterations
-                        if (++iter % (100 * 1000 * 1000) == 0 && logMode != LogMode::kNone) {
+                        if (++iter % (100 * 1000 * 1000)) {
                             BOOST_LOG_TRIVIAL(info)
```
guoyr: DRY the progress log here and on line 169? 🤷‍♂

PR #285, opened 2019-09-27
TIG-1891 Large Scale Workload Automation: Simple workload for long lived actors

ddanderson:


src/workloads/scale/LargeScaleLongLived.yml
```diff
@@ -0,0 +1,82 @@
...
+
+Actors:
+- Name: LongLivedCreator
+  Type: Loader
+  Threads: 10
+  Phases:
+  - Repeat: 1
+    Database: &DB longlived
+    CollectionCount: &CollectionCount 1000
+    Threads: 10
```
dalyd: Should threads be here in the phase and in the higher level actor level?

src/workloads/scale/LargeScaleLongLived.yml
```diff
@@ -0,0 +1,82 @@
...
+    DocumentCount: &DocumentCount 1000
+    BatchSize: 1000
+    Document:
+      # Each document ranges in size from about 90 to 150 bytes (average 120)
+      x0: &rand_10k_int {^RandomInt: {min: 0, max: 10000}}
+      x1: &rand4int {^RandomInt: {min: 0, max: 2147483647}}
```
dalyd: Why the string rand4int? It's a 32bit integer, yes? Is the 4 for 4 bytes?
It really doesn't matter, just confused me for a moment.

src/workloads/scale/LargeScaleLongLived.yml
```diff
@@ -0,0 +1,82 @@
...
+  Type: MultiCollectionQuery
+  Threads: 100
+  GlobalRate: 10000 per 1 second
+  Phases:
+  - {Nop: true}
+  - Duration: 10 minutes
+    Database: *DB
+    CollectionCount: *CollectionCount
+    DocumentCount: *DocumentCount
+    Threads: 10
```
dalyd: This thread count differs with the one on line 46. How should I be reading it?

dalyd:
I can't add other reviewers (it seems). I think we should include STM. I think there' s an option to include the group. Otherwise Robert. 

Also, I meant to add I like the use of the document generators. I assume those are used throughout and re-evaluated on each operation in these actors. Nice.

src/workloads/scale/LargeScaleLongLived.yml
ddanderson: I'm renaming this to rand_4b_int to be clearer.  Yes, 4 for 4 bytes.

src/workloads/scale/LargeScaleLongLived.yml
ddanderson: This was a mistake.  Having a "Threads" parameter in the actor phase only has meaning for the Loader actor (it may have something to do with how the loading work is broken up...?).  Anyway having a "Threads" parameter in the actor phase for a MultiCollection{Query,Update} is ignored, and is confusing.  I'm removing it in the 3 places I did that and doing another evergreen run to make sure it doesn't affect anything.

ddanderson:
Thanks @dalyd, I don't know what you mean by document generators (is that the "*" operation?)

dalyd:
> Thanks @dalyd, I don't know what you mean by document generators (is that the "*" operation?)

Eh, value generators to be more correct. Things like "^RandomInt" for generating random integers.

ddanderson:
Ah, yes.  It's a nice technique that we borrowed from another script!

ddanderson:
I just pushed modifications to address the points raised by @dalyd.  I ran one round of evergreen sys-perf tests to make sure that the changes did not change performance, and they are in line with previous runs.

ddanderson:
@guoyr, could you please review?

src/workloads/scale/LargeScaleLongLived.yml
dalyd: Oh yeah -- this is a bug/limitation of the loader if I remember correctly. Not sure if we fixed or not. I'll leave that for @guoyr to address.

src/workloads/scale/LargeScaleLongLived.yml
dalyd: Thanks

rtimmons:
@ddanderson this is good to merge whenever you're ready to do so. I've changed the repo permissions to allow anyone in the mongo kernel org to merge approved PRs. In general STM is not a gate-keeper for workloads (yamls) or actors (cast_core/*) so no need to wait for us.

PR #288, opened 2019-10-03
Modify ValidateCmd workload to work on MongoDB 4.0

GWlodarek:
Tested this locally with Genny on a 4.0 binary as Evergreen is in a bit of a weird situation with Genny for 4.0.


GeertBosch:
LGTM

PR #293, opened 2019-10-23
TIG-2082: Simple Actor that periodically logs a message

rtimmons:


src/cast_core/src/LoggingActor.cpp
```diff
@@ -0,0 +1,75 @@
...
+    explicit PhaseConfig(PhaseContext& phaseContext)
+        : logEvery{phaseContext["LogEvery"].to<TimeSpec>()}, started{metrics::clock::now()} {
+        if (phaseContext["Blocking"].to<std::string>() != "None") {
+            BOOST_THROW_EXCEPTION(
+                InvalidConfigurationException("LoggingActor must have Blocking:None"));
+        }
+    }
+    void report() {
+        // Avoid calling now() on most iterations
+        if (++iteration % 10000 != 0) {
```
guoyr: It should be possible to add a `__builtin_expect` compiler hint here if you want to be extra-fancy, but probs not worth it.

src/cast_core/test/LoggingActor_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+)",
+                          ""};
+        ActorHelper ah(config.root(), 2, MongoTestFixture::connectionUri().to_string());
+        ah.run();
+        // Don't actually assert anything because it's hard to assert interactions with logging.
+        // If you're running this test manually you should see exactly 3 log messages from
+        // the LoggingActor.
+    }
+}
+}  // namespace
```
guoyr: 💅

src/workloads/docs/LoggingActor.yml
```diff
@@ -0,0 +1,12 @@
+SchemaVersion: 2018-07-01
```
guoyr: `LoggingActorExample.yml`? So people know it's not meant to be run as a standalone workload?

src/cast_core/src/LoggingActor.cpp
rtimmons: Ya so far we've stayed pretty much entirely on non-vendor-specific C++17 so I'm hesitant to venture into extensions for something as simple as this.

PR #298, opened 2019-10-29
TIG 2096: Add --workloads option to genny_auto_tasks

danieler15:
- Add --workloads as an option in the ArgumentParser
- Branch logic in main to use --workloads specified if present, else use the result of `modified_workload_files`
- Add function `validate_user_workloads` to validate the user input
- Add a test for the above
- Update the README with info on `genny_auto_tasks`

### DIFF
--- a/src/python/genny/genny_auto_tasks.py
+++ b/src/python/genny/genny_auto_tasks.py
@@ -62,15 +62,17 @@ def validate_user_workloads(workloads):
     if len(workloads) == 0:
         return 'No workloads specified'
 
+    errors = []
     genny_root = get_project_root()
     for w in workloads:
         workload_path = '{}/src/workloads/{}'.format(genny_root, w)
         if not os.path.isfile(workload_path):
-            return 'no file at path {}'.format(workload_path)
+            errors.append('no file at path {}'.format(workload_path))
+            continue
         if not workload_path.endswith('.yml'):
-            return '{} is not a .yml workload file'.format(workload_path)
+            errors.append('{} is not a .yml workload file'.format(workload_path))
 
-    return None
+    return errors
 
 
 def construct_task_json(workloads, variants):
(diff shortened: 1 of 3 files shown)


src/python/genny/genny_auto_tasks.py
```diff
@@ -52,12 +52,27 @@ def get_project_root():
...
+
+    genny_root = get_project_root()
+    for w in workloads:
+        workload_path = '{}/src/workloads/{}'.format(genny_root, w)
+        if not os.path.isfile(workload_path):
+            return 'no file at path {}'.format(workload_path)
+        if not workload_path.endswith('.yml'):
+            return '{} is not a .yml workload file'.format(workload_path)
+
+    return None
```
rtimmons: if the user passes in multiple invalid workloads it would be nice to know of all of them rather than just the first one. I think return an array of error messages (empty array if no errors) and print all of them and exit if non-empty from the caller.

src/python/genny/genny_auto_tasks.py
```diff
@@ -102,20 +117,33 @@ def main():
         description="Generates json that can be used as input to evergreen's generate.tasks, representing genny workloads to be run")
 
     parser.add_argument('--variants', nargs='+', required=True, help='buildvariants that workloads should run on')
-    parser.add_argument('-o', '--output', default='build/generated_tasks.json')
+    parser.add_argument('--workloads', nargs='+', help='paths of workloads to run, relative to genny/src/workloads/')
```
rtimmons: ``relative to `src/workloads` in the genny repository root`` because we don't know that genny is checked out into a dir named `genny` 😸

src/python/genny/genny_auto_tasks.py

src/python/genny/genny_auto_tasks.py
```diff
@@ -52,12 +52,29 @@ def get_project_root():
...
-        print(e.output)
+        print(e.output, file=sys.stderr)
         raise e
 
     return out.decode().strip()
 
 
+def validate_user_workloads(workloads):
+    if len(workloads) == 0:
+        return 'No workloads specified'
```
rtimmons: this needs to be an array else you'll iterate over it as a string 😄

danieler15:
Looks like this is ready to be merged @guoyr @rtimmons

PR #299, opened 2019-10-29
TIG-2106 Oplog tailer actor improvements for large scale workload automation

ddanderson:
Details:
Have the tailer's run function use an "await" cursor, so that the cursor
waits for the next entry, rather than poll. This generally gives tighter
results for lag time.  Needed to change how we "catch up" at the start
of the run.

Track lag times per each minute and report best/worst/average over the minute.

Catch any exceptions and report them.

Some refactoring into functions and a helper struct.

Modified test case to run a bit longer to show the minute reporting.

### DIFF
--- a/src/cast_core/src/RollingCollections.cpp
+++ b/src/cast_core/src/RollingCollections.cpp
@@ -256,6 +256,32 @@ private:
 };
 
 struct OplogTailer : public RunOperation {
+    // Used only within this class, this tracks best/worst/average lag times.
+    struct LagTrack {
+        uint64_t best = UINT64_MAX;
+        uint64_t worst = 0;
+        uint64_t total = 0;
+        int count = 0;
+
+        void addLag(uint64_t lag) {
+            if (lag < best) {
+                best = lag;
+            }
+            if (lag > worst) {
+                worst = lag;
+            }
+            total += lag;
+            count++;
+        }
+
+        void clear() {
+            best = UINT64_MAX;
+            worst = 0;
+            total = 0;
+            count = 0;
+        }
+    };
+
     OplogTailer(PhaseContext& phaseContext,
                 mongocxx::database db,
                 ActorId id,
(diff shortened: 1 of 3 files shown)


ddanderson:
@rtimmons and @luke-pearson, can you please look at these changes?  This is a largish change from the original version - it seemed better to go with an "await" cursor, even though it meant the catch up code is a little more complex.  Once things got complex, I refactored.

src/cast_core/src/RollingCollections.cpp
```diff
@@ -269,50 +295,127 @@ struct OplogTailer : public RunOperation {
...
-                            std::chrono::duration_cast<metrics::clock::time_point::duration>(
-                                std::chrono::milliseconds{std::stol(timeSplit[2])})};
-                        _oplogLagOperation.report(
-                            now,
-                            std::chrono::duration_cast<std::chrono::microseconds>(now - started),
-                            metrics::OutcomeType::kSuccess);
+        return (_caughtUp);
+    }
+
+    // Called when we see a create of a rolling collection.
```
luke-pearson: Possibly better to use block comments here?

src/cast_core/src/RollingCollections.cpp
```diff
@@ -269,50 +295,127 @@ struct OplogTailer : public RunOperation {
...
+    void trackRollingCreate(uint64_t rollingMillis, LagTrack &lagTrack) {
+        auto now = metrics::clock::now();
+        auto started = metrics::clock::time_point{
+            std::chrono::duration_cast<metrics::clock::time_point::duration>(
+              std::chrono::milliseconds{rollingMillis})};
+        auto lag = std::chrono::duration_cast<std::chrono::microseconds>(now - started);
+        auto lagns = lag.count();
+
+        if (caughtUp(lagns)) {
+            // BOOST_LOG_TRIVIAL(info) << "Oplog tailer lag: " << lagns << "ns";
```
luke-pearson: Maybe delete this commented out line

src/cast_core/src/RollingCollections.cpp
```diff
@@ -269,50 +295,127 @@ struct OplogTailer : public RunOperation {
...
+        mongocxx::options::find opts{};
+        opts.cursor_type(mongocxx::cursor::type::k_tailable_await);
+        _cursor = std::optional<mongocxx::cursor>(database["oplog.rs"].find({}, opts));
+
+        // Track the best, worst, average lag times, we'll display them
+        // periodically to the output.
+        LagTrack lagTrack;
+
+        try {
+            // If the cursor is idle for more than 10 seconds (that is 10
```
luke-pearson: I'm not really a fan on this, I think the oplog tailer should stay alive regardless. It should persist as long as is specified in the duration field in the yml. If we do this we may miss some critical timeout due to replication or otherwise? Possibly miss cache stuck issues?

ddanderson: Point taken about cache stuck issues.  The problem is that genny will not quit while there is a live actor (at least that's what I've seen), so I'll need a timeout of some kind.  I'll look into running at least as long as the duration.

src/cast_core/src/RollingCollections.cpp
ddanderson: I think this is a C++-ism.  Looking at the genny source code, it looks like it primarily uses // for block commenting, and /** ... */ style for "javadoc style" function comments.  @rtimmons, any opinion on comment styles?

src/cast_core/src/RollingCollections.cpp
rtimmons: If you set an Actor to be `Blocking: None` (and omit `Repeat` and `Duration`) the Actor will run as long as there are other Actors running in the Phase but it won't keep the Phase (or Workload) from completing on its own.

src/cast_core/src/RollingCollections.cpp
ddanderson: Will do.

src/cast_core/src/RollingCollections.cpp
rtimmons: Example/explanation in the `HelloWorld-MultiplePhases` workload:

https://github.com/mongodb/genny/blob/869dca937569cad3f04e6ee03c1992bdbd96e819/src/workloads/docs/HelloWorld-MultiplePhases.yml#L72-L79

ddanderson: Ah, okay.  I resolved this a different way, simply returning to genny when the oplog cursor returns no data.  Genny will start the actor again (or not if the time has expired).

src/cast_core/src/RollingCollections.cpp
```diff
@@ -269,50 +295,125 @@ struct OplogTailer : public RunOperation {
         }
     }
 
+    bool isRollingOplogEntry(const bsoncxx::document::view &doc, long *rollingMillis)
```
rtimmons: c++isms would be to have this either

1. return `std::optional<long>` - return `make_optional(timesplit[2])` if found else `return std::nullopt` or
2. pass `rollingMillis` by non-const ref and just use regular assignment

src/cast_core/src/RollingCollections.cpp
rtimmons: no opinion here - double-asterisk is for javadoc-style as you indicated. I personally don' like 

```
/*
foo
*/
```

style comments but that's totally personal-preference.

src/cast_core/src/RollingCollections.cpp
ddanderson: Resolved using std::optional.  Nice feature!

src/cast_core/src/RollingCollections.cpp
ddanderson: I'm going to leave multiline comments using '//' to be closer with other genny/mongodb c++ style I've seen.

ddanderson:
@rtimmons, I would be totally up for writing and automated test for oplog tailing.  The two challenges I think are: 1) needs to be part of a replication set for oplog to work. It can be just 1 member, replication needs to be started, though.  Is that supported in the test harness?  2) it needs to be multithreaded, I think to do a reasonable test - we need at least a thread that is creating file names, and then the oplog tailer as well.  Ideally, we would use 4 actors - Setup in phase 1, and a Writer, Manage, and OplogTailer in phase 2.  The test examples I saw seem to be set up with a single actor.  Is this reasonable to do in the test framework?

rtimmons:
@ddanderson cool - ya it's totally reasonable to have multiple Actors and run only on certain topologies.

You can specify which topology the test should run against by adding Catch2 tags - see the CommitLatency Actor test as an example:

https://github.com/mongodb/genny/blob/554da2804793c286891e01c65e4402e864b2acab/src/cast_core/test/CommitLatency_test.cpp#L31-L30

The PR testing is run through resmoke.py - you can run it locally pretty easily as well. That roughly looks like the following:

```sh
cd genny
./scripts/lamp
pushd build
git clone git@github.com:mongodb/mongo.git
virtualenv venv
source ./venv/bin/activate
pip install -r ./mongo/buildscripts/requirements.txt
popd
python ./build/mongo/buildscripts/resmoke.py --configDir ./build/mongo/buildscripts/resmokeconfig --suite ./src/resmokeconfig/genny_three_node_replset.yml --mongod mongod --mongo mongo --mongos mongos
```

You can modify the tags run by the  `genny_three_node_replset.yml` (line 16); just add a new tag to your test e.g. `[just_for_don]` and use that tag there. Another potential gotcha is that when you setup the `ActorHelper` as in the following:

```c++
genny::ActorHelper ah(config.root(), 1, MongoTestFixture::connectionUri().to_string());
```

The `1` (second arg) is the total number of threads across all Actors - we haven't gotten around to auto-setting this param or barfing gracefully if it's set ncorrectly.

(The `./scripts/lamp` helper is supposed to help with this process but I think it broke for local development after resmoke changed around recently...)

Happy to help further just ping me on slack or reply here.

ddanderson:
@rtimmons, thanks!  This looks a little more involved, so I split out the automated testing into TIG-2119.  I'd like to get the rest of this resolved and merged as there is other work waiting on it (and as a side effect, will test oplog tailing in a real-world way).

I see the macos evergreen test for this did not succeed - looks like some sort of configuration problem?  Any clues on how to address that?

rtimmons:
@ddanderson 

> I see the macos evergreen test for this did not succeed

Ya that's nothing to do with your changes - see [BUILD-9216](https://jira.mongodb.org/browse/BUILD-9216).

Let me know when you're ready to merge and I'll press the button (commit-queue is coming soon!)

ddanderson:
@rtimmons, yes this is ready to merge.  Thanks!

ddanderson:
@rtimmons, oops, hold on, I'll resolve the conflicts.

PR #300, opened 2019-10-31
TIG 2077: Add --autorun option to genny_auto_tasks

danieler15:
When `--autorun` is set, we'll now generate tasks based on workloads' AutoRun section and bootstrap.yml/runtime.yml files in the working directory, as described in [the ticket](https://jira.mongodb.org/browse/TIG-2077).

Main work:
- Added functions in `genny_auto_tasks.py` to find all files in the `workloads` directory that have an `AutoRun` section that is satisfied by the environment values from `bootstrap.yml` and `runtime.yml` in the current directory.
- Added a test of the function `workload_should_autorun`. Because there were many test cases to consider, I put them in a new fixtures file.
- Modified `genny_auto_tasks.sh` to not change the working directory, because the python script relies on having `bootstrap.yml` etc in the cwd. As a result, I modified the functions that use `git` to temporarily cd into the `genny` repo then switch back

One thing to flag: the original ticket proposes $in-like syntax for workloads that want to run for multiple values of specific fields. I decided to use `platform: linux | windows | osx` syntax instead for simplicity but am open to alternatives.


src/python/genny/genny_auto_tasks.py
```diff
@@ -113,36 +213,51 @@ def construct_task_json(workloads, variants):
...
 
-    if args.workloads is not None:
-        errs = validate_user_workloads(args.workloads)
-        if len(errs) > 0:
-            for e in errs:
-                print('invalid workload: {}'.format(err), file=sys.stderr)
+    if args.autorun:
+        env_dict = make_env_dict()
+        if env_dict is None:
+            print('fatal error: bootstrap.yml and runtime.yml files not found in current directory, cannot AutoRun workloads')
```
guoyr: I think it might be helpful to say something like "autorun is intended to be used in Evergreen, if your using genny locally, please just run the workload directly"

src/python/genny/genny_auto_tasks.py
```diff
@@ -22,19 +23,53 @@ def to_snake_case(str):
     s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', str)
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
 
+def cd_python_dir():
```
guoyr: Per offline discussion. You can go to the repo root instead to avoid having to encode an implicit relative location between this file and the python root; this file could get moved around.

guoyr: Also, you can just change to the repo root once at the start of this script. the current working directory that you specify is only going to apply to this file, not its caller.

src/python/genny/genny_auto_tasks.py
```diff
@@ -45,17 +80,83 @@ def modified_workload_files():
...
+    """
+    env_files = ['bootstrap.yml', 'runtime.yml']
+    env_dict = {}
+    for fname in env_files:
+        if not os.path.isfile(fname):
+            return None
+        with open(fname, 'r') as handle:
+            config = yaml.safe_load(handle)
+            module = os.path.basename(fname).split('.yml')[0]
+            env_dict[module] = config
```
guoyr: It's better to use an explicitly typed class (or at least a namedtuple) over a dictionary whenever possible. E.g. for `env_dict`

src/python/genny/genny_auto_tasks.py
```diff
@@ -45,17 +80,83 @@ def modified_workload_files():
...
     """
-    try:
-        out = subprocess.check_output('git rev-parse --show-toplevel', shell=True)
-    except subprocess.CalledProcessError as e:
-        print(e.output, file=sys.stderr)
-        raise e
 
-    return out.decode().strip()
+    # First check that the workload has a proper AutoRun section.
+    if 'AutoRun' not in workload_dict or not isinstance(workload_dict['AutoRun'], dict):
```
guoyr: Similar to my comment above, I think we should go the extra mile of making an object representation of the AutoRun section of the YAML file. This way we can avoid typing strings and make it easy to rename fields. Talk to @aggrand if you want examples :)

src/python/genny/genny_auto_tasks.py
```diff
@@ -45,17 +80,83 @@ def modified_workload_files():
...
+
+    for module, required_config in workload_dict['AutoRun']['Requires'].items():
+        if module not in env_dict:
+            return False
+        if not isinstance(required_config, dict):
+            return False
+
+        # True if set of config key-vaue pairs is subset of env_dict key-value pairs
+        # This will be false if the AutoRun yaml uses a logical or (i.e. branch_name: master | prod), but it is efficient so we use it for a first-pass.
+        if not required_config.items() <= env_dict[module].items():
```
guoyr: I'm debating between using "|" or a list for multiple environment variables. Any preference @rtimmons?

guoyr: I chatted with @rtimmons in passing yesterday and I think he prefers lists. String parsing is always a bit magical and the "|" convention may not seem intuitive for Windows folks

src/python/genny/genny_auto_tasks.py
```diff
@@ -6,13 +6,42 @@
...
-import textwrap
+import yaml
+import glob
 
 from shrub.config import Configuration
 from shrub.command import CommandDefinition
 from shrub.variant import TaskSpec
 
 
+class AutoRunSpec():
```
guoyr: 🎆

src/python/genny/genny_auto_tasks.py
```diff
@@ -113,36 +226,58 @@ def construct_task_json(workloads, variants):
...
+                note: --autorun is intended to be called from within Evergreen. If using genny locally, please run the workload directly.',
+                file=sys.stderr)
+            print(os.getcwd(), file=sys.stderr)
             return
-        workloads = args.workloads
-    else:
+
+        workloads = autorun_workload_files(env_dict)
+        if len(workloads) == 0:
+            print('No AutoRun workloads found matching environment, generating no tasks.')
```
guoyr: Hm, could we throw an error in this case? I think one use case is for people adding new sys-perf build variants. They will want to run Genny workloads on them, but they may not know to change workload YAMLs to support the new build variant. Errorring and telling people to double check the workload YAML files could be a safeguard, though it's not foul-proof.

src/python/genny/genny_auto_tasks.py
danieler15: I agree with that logic and changed the code accordingly

src/python/genny/genny_auto_tasks.py
danieler15: Done. I did the same if no workloads are generated with the `--modified` flag because I imagine that someone would only use it if they expected it to spawn some workloads.

guoyr:
Still LGTM!

danieler15:
In order to fix the issue of `generate.tasks` redefining the same tasks for different variants, `genny_{auto, patch}_tasks` are now dependent on a generate_all_tasks task. Consequently, I just added a `--generate-all-tasks` option which it uses to generate JSON representing tasks for *all* generated workloads. 

`genny_{auto, patch}_tasks` now calls `construct_variant_json` which only generate JSON representing variants that run the appropriate tasks (which must have already been generated).

README.md
```diff
@@ -254,12 +254,33 @@ cd genny
...
-In the browser window, select the workloads you wish to run. Good
-examples are Linux Standalone / `big_update` and Linux Standalone /
-`insert_remove`.
+In the browser window, select either `genny_patch_tasks` or `genny_auto_tasks`.
+`genny_patch_tasks` will run any workloads that you have added or modified locally
+(based on your git history). This is useful if you want to test only the workload(s)
+you've been working on. 
+
+`genny_auto_tasks` automatically runs workloads based on the evergreen environment
+(variables from `bootstrap.yml` and `runtime.yml`) and an optional AutoRun section
```
guoyr: Suggest adding "in DSI" at the end of the parenthetical.

README.md
```diff
@@ -254,12 +254,33 @@ cd genny
...
+
+In this case, `test_workload` would be run whenever `bootstrap.yml`'s `mongodb_setup`
+variable has a value of `replica` or `single-replica`. In practice, the workload
+AutoRun sections are setup so that you can use `genny_auto_tasks` to run all relevant
+workloads on a specific buildvariant.
 
-The task will compile mongodb and will then run your workloads. Expect to
-wait around 25 minutes.
+Both `genny_patch_tasks` and `genny_auto_tasks` will compile mongodb and then run
+the relevant workloads. Expect to wait around 25 minutes.
```
guoyr: I would remove the last sentence that mentions "Expect to wait around 25min". cc @rtimmons

src/python/genny/genny_auto_tasks.py
```diff
@@ -230,67 +253,82 @@ def construct_task_json(workloads, variants):
...
                                 help='if set, the script will generate tasks for workloads that have been added/modifed locally, relative to origin/master')
-    workload_group.add_argument('--workloads', nargs='+',
-                                help='paths of workloads to run, relative to src/workloads/ in the genny repository root')
+    parser.add_argument('--forced-workloads', nargs='+',
+                        help='paths of workloads to run, relative to src/workloads/ in the genny repository root')
+    parser.add_argument('-o', '--output', default='build/generated_tasks.json',
+                        help='path of file to output result json to, relative to the genny root directory')
 
     args = parser.parse_args(sys.argv[1:])
+    if args.generate_all_tasks and (args.autorun or args.modified):
```
guoyr: 👍

src/python/genny/genny_auto_tasks.py
```diff
@@ -230,67 +253,82 @@ def construct_task_json(workloads, variants):
 
 def main():
     """
-    Main Function: parses args, writes to file evergreen tasks in json format for workloads that are specified in args.
+    Main Function: parses args, writes a JSON file representing *either* evergreen tasks or buildvariants that can be passed to generate.tasks.
+    If run with --generate-all-tasks, this program outputs a JSON file of task definitions for all workloads in the /src/workloads genny directory.
```
guoyr: Maybe also add a phrase to say this is to get around limitations of generate.tasks

src/workloads/scale/InsertBigDocs.yml
```diff
@@ -18,3 +18,10 @@ Actors:
     Document:
       x: {^RandomInt: {min: 0, max: 2147483647}}
       string0: {^FastRandomString: {length: 15000000}}
+
+AutoRun:
+  Requires:
+    bootstrap:
+      mongodb_setup: 
+        - replica
+        - single-replica
```
guoyr: 💅

src/workloads/selftests/GennyOverhead.yml
```diff
@@ -61,3 +61,9 @@ Actors:
   - MetricsName: LoopNSleepAfter
     Repeat: 1e4
     SleepAfter: 1 millisecond
+
+AutoRun:
+  Requires:
+    bootstrap:
+      mongodb_setup: 
+        - standalone
```
guoyr: 💅

README.md
danieler15: Done.

README.md
danieler15: Done.

src/python/genny/genny_auto_tasks.py

danieler15:
evergreen merge

guoyr:
evergreen merge

PR #301, opened 2019-11-08
Make CrudActor respect MetricsName in Phase: YAML Blocks

rtimmons:
1. We were previously recording all data on one `metrics::Operation` per `Operation` name, so e.g. all `bulkWrite` operations would be recorded together even if they were on different phases or had very different semantics.

    After this Actor was written we introduced the `MetricsName:` concept in a `Phase` block, but we never brought it in.

2. Fixed a bug where we could not specify insert-options and added a test

3. Also did some drive-by 💅.

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1072,10 +1072,6 @@ struct SetReadConcernOperation : public BaseOperation {
         : BaseOperation(context, opNode),
           _collection{std::move(collection)},
           _operation{operation} {
-        if (!opNode["ReadConcern"]) {
-            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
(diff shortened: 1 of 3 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,22 +1179,30 @@ struct CrudActor::PhaseConfig {
...
+            // Node is convertible to bool but only explicitly so need to do the odd-looking
+            // `? true : false` thing.
+            const bool perPhaseMetrics = phaseContext["MetricsName"] ? true : false;
+
             auto createOperation = op->second;
             return createOperation(yamlCommand,
                                    onSession,
                                    collection,
-                                   phaseContext.actor().operation(opName, id),
+                                   perPhaseMetrics ? phaseContext.operation(opName, id)
```
guoyr: 👌

rtimmons:
evergreen merge

PR #303, opened 2019-11-13
TIG-2077 Don't throw exception if no autorun tasks are generated, fix small bug

danieler15:
As I discussed with @guoyr offline, this change would make it **not** an error if there are no tasks that should be autorun. 

There are some buildvariants on Evergreen that do not have any associated genny workloads. We could get rid of `genny_auto_tasks` for these variants, but then, if someone added a new genny workload that should run on one of these variants, they would have to edit `system_perf.yml` which defeats the purpose of this ticket.

This PR also fixes an unexpected issue that would occur if `bootstrap.yml` or `runtime.yml` is empty.


danieler15:
evergreen merge

rtimmons:
evergreen merge

PR #304, opened 2019-11-13
TIG-2017 Mixed Write Workload with Mixed Delays

rtimmons:
Depends on https://github.com/10gen/dsi/pull/501 and a bit of 💅 work from @danieler15 to support passing in the `setup` param to the generated `"prepare environment"` evergreen invocation.

Patch-test [here](https://evergreen.mongodb.com/version/5dcac8e07742ae10faff2915). I will send an updated patch once more pre-reqs are done.


guoyr:
evergreen merge

PR #306, opened 2019-11-14
TIG-1896 Large Scale Workload Automation: Create a basic model workload

ddanderson:


ddanderson:
@rtimmons, can you review this please?

src/workloads/scale/LargeScaleModel.yml
```diff
@@ -0,0 +1,469 @@
...
+  RollingDocument: &RollingDocument {
+    a: *rand_30b_string,
+    b: *rand_30b_string,
+    c: *rand_30b_string,
+    d: *rand_30b_string,
+    e: *rand_30b_string,
+    f: *rand_30b_string,
+    g: *rand_4b_int,
+    h: *rand_4b_int,
+    i: *rand_4b_int }
```
rtimmons: you don't need the curly brackets or the commas:

```yaml
Rolling: &Rolling
  a: *rand
  b: *rand
```

src/workloads/scale/LargeScaleModel.yml
```diff
@@ -0,0 +1,469 @@
...
+#  Database: *ScannerColdDB
+#  CollectionCount: 60
+#  Phases:
+#  - {Nop: true}
+#  - {Nop: true}
+#  - Duration: *Duration
+#    # Optionally skip the first iteration of the collection scanner
+#    SkipFirstLoop: true
+#    GlobalRate: 1 per 10 seconds
+#    ScanType: count
```
rtimmons: I'd either remove these actors or add an additional comment saying why they're commented-out / link to a ticket/TODO to uncomment them.

src/workloads/scale/LargeScaleModel.yml
```diff
@@ -0,0 +1,469 @@
+SchemaVersion: 2018-07-01
+Owner: Storage Engines
```
rtimmons: Using the new "auto-run" functionality you can have this be scheduled as part of the sys-perf waterfall by adding something like this:

```yaml
AutoRun:
  Requires:
    bootstrap:
      mongodb_setup: single-replica-15gbwtcache
```

Where `single-replica-15gbwtcache` corresponds to which mongodb_setup is required.

src/workloads/scale/LargeScaleModel.yml
ddanderson: I'll remove them (I just had them there to remind me how to do these things).

src/workloads/scale/LargeScaleModel.yml
ddanderson: Great, fixed.

src/workloads/scale/LargeScaleModel.yml
ddanderson: That sounds very nice.  But in its current state, it will probably fail because of genny OOM, or timeout.  I'll defer that change for a later iteration.

ddanderson:
@rtimmons, can you please merge?  I've resolved the issues and don't have merge priv. The single evergreen failure is a compile failure with arch linux.  Doesn't seem related to anything in this change.  Thanks.

ddanderson:
@guoyr, can you please merge when you have a chance?  I've resolved the issues from the review and don't have merge priv. The single evergreen failure is a compile failure with arch linux. Doesn't seem related to anything in this change. Thanks.

danieler15:
evergreen merge

guoyr:
@ddanderson Done! (I asked Daniel to do it to check if the permissions are correct)

You should be able to type `evergreen merge` in a comment to merge future PRs after Ryan turned on the commit queue a couple of days ago.

bsamek:
@danieler15 It looks like a task failed in the build https://evergreen.mongodb.com/version/5dd54fd957e85a0dc22ce6d6. (I happened to look at this because it looks like it caused the queue to be stuck somehow. We're looking into it. If you think the compile should succeed, you can re-add it to the queue.)

ddanderson:
evergreen merge

guoyr:
I merged this PR manually because there was a change to the Arch distro that caused all Genny tasks to fail. The problem was that the new Arch image contains a new libicu version than what the mongo driver supports.

There's no easy fix in the short or long run because the Arch base image is temperamental and maintained by an individual. Even if we did explicit dependency management we'd still need to figure out why the driver isn't compiling, which is a day or so of slogging through CMakeLists. I'm not convinced it's worth the effort for STM, which is certainly not to see we shouldn't pin base images.

So instead, I'm going to remove Arch from genny's Evergreen project while still compiling gennytoolchain on it.

ddanderson:
Thanks, @guoyr, for that explanation.  I was wondering what was happening there.

PR #307, opened 2019-11-22
update toolchain to fix compile on arch

guoyr:


guoyr:
Updated toolchain still is broken. Closing this PR in favor of removing Arch from Genny's EVG project

PR #310, opened 2019-11-26
Don't raise exception for no modified workloads if there are --forced-workloads

danieler15:
Small fix to increase the usability of genny_patch_tasks: currently we always raise an exception if there are no locally modified tasks. This PR would stop throwing an exception if the user also supplies workloads to run via the `--forced-workloads` flag. This way, genny_patch_tasks can be used to just run a one-off task via `--forced-workloads`, even if there are no locally modified tasks.

### DIFF
--- a/src/python/genny/genny_auto_tasks.py
+++ b/src/python/genny/genny_auto_tasks.py
@@ -329,10 +329,13 @@ def main():
         elif args.modified:
             workloads = modified_workload_files()
             if len(workloads) == 0:
-                raise Exception(
-                    'No modified workloads found, generating no tasks.\n\
+                err_msg = 'No modified workloads found.\n\
                     No results from command: git diff --name-only --diff-filter=AMR $(git merge-base HEAD origin/master) -- ../workloads/\n\
-                    Ensure that any added/modified workloads have been committed locally.')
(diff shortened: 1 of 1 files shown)


src/python/genny/genny_auto_tasks.py
```diff
@@ -329,10 +329,13 @@ def main():
...
-                raise Exception(
-                    'No modified workloads found, generating no tasks.\n\
+                err_msg = 'No modified workloads found.\n\
                     No results from command: git diff --name-only --diff-filter=AMR $(git merge-base HEAD origin/master) -- ../workloads/\n\
-                    Ensure that any added/modified workloads have been committed locally.')
+                    Ensure that any added/modified workloads have been committed locally.'
+                if args.forced_workloads is None:
+                    raise Exception(err_msg)
+                else:
+                    print(err_msg, file=sys.stderr)
```
guoyr: I think this else statement can be removed. It could trick people reading logs into thinking this is an error.

guoyr:
Going to merge this on behalf of Daniel since he's rotated off and Don needs it.

guoyr:
evergreen merge

PR #318, opened 2019-12-09
TIG-2137 Compile Poplar gRPC Client

rtimmons:
The actual generated client is already in the codebase (didn't send for review because it wasn't interesting). This just enables other parts of genny to actually link to it via, e.g., `#include <poplarlib/collector.grpc.pb.h>`.


cmake/GetGRpc.cmake
```diff
@@ -12,6 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-## gRPC is not always available.
```
aggrand: So was gRPC previously sometimes unavailable but now we should always be able to link it?

rtimmons: gRPC was sometimes a pain to compile because of SSL but now that we have the genny toolchain (vcpkg) it's a nop.

rtimmons:
evergreen merge

PR #320, opened 2019-12-11
PERF-1948 Add performance workload for creating an index on a sharded collection

cheahuychou:
Added a workload similar to CreateIndex.yml for running against a sharded collection. The only difference is that I got rid of `mongodb_setup: single-replica` and used `Duration` instead of `Repeat` to make it run more operations.

### DIFF
--- a/src/workloads/execution/CreateIndexSharded.yml
+++ b/src/workloads/execution/CreateIndexSharded.yml
@@ -2,65 +2,18 @@ SchemaVersion: 2018-07-01
 Owner: "@mongodb/sharding"
 
 Actors:
-- Name: InsertData
-  Type: Loader
-  Threads: 1
-  Phases:
-  - ExternalPhaseConfig:
-      Path: ../../phases/execution/CreateIndexPhase.yml
-      Key: InsertData
-      Parameters:
-        db: &db test
-  - &Nop {Nop: true}
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-
 - Name: EnableSharding
   Type: AdminCommand
   Threads: 1
   Phases:
-  - *Nop
   - Repeat: 1
     Database: admin
     Operations:
     - OperationMetricsName: EnableSharding
       OperationName: AdminCommand
       OperationCommand:
-        enableSharding: *db
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-  - *Nop
-
-- Name: CreateShardKeyIndex
-  Type: RunCommand
-  Threads: 1
-  Phases:
-  - *Nop
-  - *Nop
-  - Repeat: 1
-    Database: *db
-    Operations:
-    - OperationMetricsName: CreateShardKeyIndex
-      OperationName: RunCommand
-      OperationCommand:
-        createIndexes: Collection0 # This is the default collection populated by the Loader.
-        indexes:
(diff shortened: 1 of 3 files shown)


cheahuychou:
I have a question. Should `mongodb_setup` still be `single-replica`? I thought it had to be different because I want the workload in a sharded cluster.

cheahuychou:
I refactored the workloads and created a Phase as suggested. I also reduced the `Duration` for `CreateIndexSharded` to `5 minutes` because I realized the test was taking too long. [Here](https://evergreen.mongodb.com/version/5df2bee79ccd4e195f65f9c7) is the new evergreen patch with no changes to DSI.

rtimmons:
This is really great, @cheahuychou. I hope the external Phase syntax didn't give you much trouble.

> I have a question. Should mongodb_setup still be single-replica? I thought it had to be different because I want the workload in a sharded cluster.

That value corresponds to which [mongodb_set yaml file](https://github.com/10gen/dsi/tree/master/configurations/mongodb_setup) you want this to be run with. We have a couple different sharded topologies:

- The "biggest" is [a 3-shard cluster](https://github.com/10gen/dsi/blob/master/configurations/mongodb_setup/mongodb_setup.shard.yml) with each shard having a 3-node replset.  You would just have `mongodb_setup: shard`. 
- There's also [shard lite](https://github.com/10gen/dsi/blob/master/configurations/mongodb_setup/mongodb_setup.shard-lite.yml) which is a 2-shard cluster each with 2 nodes. You would put `mongodb_setup: shard-lite`
- other sharded topologies are of course possible but would require some changes in DSI and potentially in the server repo to add the necessary configs

Let me know if this is what you're after. Otherwise you're free to just type `evergreen merge` and this will be merged and your workload will start running on the waterfall.

src/workloads/execution/CreateIndexSharded.yml
```diff
@@ -53,11 +132,6 @@ Actors:
       Key: CreateTextIndexesCmd
       Parameters:
         Duration: 5 minutes
-  - ExternalPhaseConfig:
```
cheahuychou: We cannot create unique indexes for an already-sharded collection. https://docs.mongodb.com/manual/core/index-unique/

cheahuychou:
Thanks @jsmulrow for pointing out that I need additional phases in CreateIndexSharded.yml for sharding the collection. [Here](https://evergreen.mongodb.com/version/5df589d732f417667c58f8a3) is the new evergreen patch. I'm not sure why the commit is failing evergreen/amazon2 but it should be unrelated to my changes.

src/workloads/execution/CreateIndexSharded.yml
```diff
@@ -0,0 +1,146 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+
+Actors:
+- Name: InsertData
```
jsmulrow: Can you put this stage after creating the hashed index and sharding the collection? Otherwise the data will all start on one shard and need to be balanced while the indexes are being built, which isn't really what this workload was for. That should also speed up building the hashed index, since there won't be any data yet.

src/workloads/execution/CreateIndexSharded.yml
```diff
@@ -44,23 +23,20 @@ Actors:
   - *Nop
   - *Nop
 
-- Name: CreateShardKeyIndex
```
cheahuychou: Since the collection is empty when we shard the collection, we don't need a phase for creating the index.

cheahuychou:
evergreen merge

cheahuychou:
evergreen merge

rtimmons:
evergreen merge

PR #321, opened 2019-12-11
PERF-1933: Run parallel_insert workload with slow network

ldennis:
This is for the project "Use Exhaust Cusor for Oplog Fetching".

### DIFF
--- a/src/workloads/docs/ParallelInsert.yml
+++ b/src/workloads/docs/ParallelInsert.yml
@@ -187,3 +187,7 @@ AutoRun:
         - replica
         - single-replica
         - replica-noflowcontrol
+  PrepareEnvironmentWith:
+    setup:
+    - replica-delay-mixed
+    - replica


src/workloads/docs/ParallelInsert.yml
```diff
@@ -187,3 +187,7 @@ AutoRun:
         - replica
         - single-replica
         - replica-noflowcontrol
+  PrepareEnvironmentWith:
+    setup:
+    - replica-delay-mixed
+    - replica
```
ldennis: There is a problem for the task naming here as the existing parallel_insert workload is named parallel_insert_replica after this. So I think we will be losing some history data for comparisons. I am not sure if we should care.

rtimmons: Oh ya good point - that's an unfortunate side-effect we didn't consider when devising this autorun syntax and task-gen strategy. Since this is arguably a bug in the design I'm fine to prioritize a fix to make it possible to specify the generated task name or something - let me know it's not a big ask. Otherwise imho it's up to you if you care about losing history here.

ldennis: I am ok with it. But is there a plan to fix this? Maybe we could somehow skip appending the suffix if the environment parameter is the same as one of the mongo_setups in the bootstrap section?

ldennis:
evergreen merge

ldennis:
Why was this not merged?

rtimmons:
Github permissions are confusing - sorry about that. I'll merge now.

PR #322, opened 2019-12-27
PERF-1897 Sys-perf tests for external sort

jseyster:
This workload has three benchmark phases, which each run a sort operation. For initial testing, I chose to run 20 trials of each operation.

Looking at one execution of the test, I only need to look at the elapsed time of the first 7 trials for the half width of the 95% confidence interval to be less than 1% of the sample mean. I think that would make 7 a reasonable choice for the number of trials for each operation. We could also round up to 10. Each trial takes about 2.2s, so we don't have to worry about setting this too high.

### DIFF
--- a/src/workloads/execution/ExternalSort.yml
+++ b/src/workloads/execution/ExternalSort.yml
@@ -1,0 +1,103 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 100000
+    BatchSize: 10000
+    Document:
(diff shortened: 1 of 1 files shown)


rtimmons:
LGTM - can you paste a link to the patch-build showing how this runs?

jseyster:
Sure!
https://evergreen.mongodb.com/version/5e00ffcbd6d80a5c1a472925

The relevant variant is "external_sort." As per the "new workload" template, I'll set up a run of 5 patch builds later today.

jseyster:
I ran a series of 5 patch builds of this test and put the results in this spreedsheet (under the "Nose/Robustness testing" tab):
https://docs.google.com/spreadsheets/d/1FKKaPcmDYFZNGZmHUGVmA-H2AsfgxYTyiO9LAEneh1U/edit?usp=sharing

Results are very consistent between runs.

rtimmons:
Those numbers look good to me. Whenever you're ready to merge just type `evergreen merge` 💯

jseyster:
evergreen merge

PR #323, opened 2020-01-03
BF-15755 Remove single-replica from ParallelInsert.yml

rtimmons:
The current autorun section of this workload generates a task that tries to have the following in bootstrap.yml:

```yaml
infrastructure_provisioning: single
mongodb_setup: replica
```

This [fails](https://evergreen.mongodb.com/task/sys_perf_linux_1_node_replSet_parallel_insert_replica_8ea803b42abc5b9b653fe7931e90f00f0e8c8043_19_12_18_21_48_52/1) because `mongodb_setup.replica.yml` assumes an `infrastructure_provisioning` of at least 3 nodes.

I suspect this was introduced by TIG-2077 and we accidentally added this line.

This is theoretically possible to catch prior to committing, so I added to GI-075 the [Genny Ideas Doc](https://docs.google.com/document/d/1K0SlHoKkIACIXhkNP_KPHvb3IV6r53ReqsHqEYTnXDA/edit#) to consider adding code to prevent something like this from happening in the future.

(fyi @ldennis - nothing you need to do just a heads-up.)

### DIFF
--- a/src/workloads/docs/ParallelInsert.yml
+++ b/src/workloads/docs/ParallelInsert.yml
@@ -185,7 +185,6 @@ AutoRun:
     bootstrap:
       mongodb_setup:
         - replica
-        - single-replica
         - replica-noflowcontrol
   PrepareEnvironmentWith:
     setup:


ldennis:
Just curious, then currently, is it still possible to run a workload under both replica and single-replica?

guoyr:
>  Is it still possible to run a workload under both replica and single-replica?

Thanks for pointing this out Dennis, it is indeed a problem and I believe it's not possible to run a workload with a custom prepare environment config under both `replica` and `single-replica` at the moment.

It's not a problem for InsertRemove because there's a check for [matching environment config](https://github.com/mongodb/genny/blob/b40576e1afa40c691521abdac0dd7fe5f8e05928/src/python/genny/genny_auto_tasks.py#L133).

@rtimmons I think it's a problem in DSI to tightly couple the infrastructure provisioning and the mongodb setup configs. The only non-hacky fix I think of is to rewrite mongodb setup in Python and leave only the topology config and the server parameters in YAML, but this is a non-trivial change. Open to other ideas.

rtimmons:
evergreen merge

rtimmons:
For some reason the commit-queue is failing at compile but [a no-change patch-build succeeds](https://evergreen.mongodb.com/version/5e1de2ca0ae606653228cb2b). Merging the old-fashioned way 🥃

PR #324, opened 2020-02-06
TIG-2323 runCommand actor should measure failure latency

bcaimano:


bcaimano:
My motivation here is I want to see when a replSetStepDown occurred in the genny-perf.csv. This allows me to analyze CRUD behavior pre and post stepdown.

rtimmons:
Hey @bcaimano anything you need from me/STM here?

PR #325, opened 2020-02-10
TIG-2260 Make metrics::Registry Aware of Actor-counts and Phase-Numbers

aggrand:
Right now the actor counts and phase numbers aren't used for anything, but they'll be passed to poplar to used. I'm thinking the OperationImpl makes sense as the location where the two forms of metrics co-reside, because I think we can mostly create a new version of EventSeries to hold it.

Lotsa feedback welcome 😄

### DIFF
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -127,8 +127,8 @@ public:
         return ClockSource::now();
     }
 
-    std::size_t threadCount(const std::string& actorName, const std::string& opName) const {
-        return _ops[actorName][opName].size();
+    std::size_t getWorkerCount(const std::string& actorName, const std::string& opName) const {
+        return (_ops.at(actorName).at(opName)).size();
     }
 
 private:
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -127,6 +127,10 @@ public:
         return ClockSource::now();
     }
 
+    std::size_t threadCount(const std::string& actorName, const std::string& opName) const {
+        return _ops[actorName][opName].size();
+    }
+
 private:
     OperationsMap _ops;
 };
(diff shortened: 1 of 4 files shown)


src/metrics/include/metrics/metrics.hpp
```diff
@@ -34,6 +34,9 @@ namespace genny::metrics {
  */
 namespace v1 {
 
+template <typename Clocksource>
```
aggrand: A forward declaration was necessary since metrics.hpp and operation.hpp import each other.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -119,6 +127,10 @@ class RegistryT final {
         return ClockSource::now();
     }
 
+    std::size_t getWorkerCount(const std::string& actorName, const std::string& opName) const {
```
aggrand: Wasn't sure if we wanted to account for current phase when figuring the worker count.

rtimmons: I think for now we'll make a blanket statement that the workerCount is constant across phases for a given (actor,operation), but that may not always be the case. 

Let's add a comment saying basically...that 😄

src/metrics/include/metrics/operation.hpp
```diff
@@ -33,6 +33,8 @@ namespace genny::metrics {
 
 using count_type = long long;
 
+using actor_count_t = uint32_t;
```
rtimmons: `size_t`?

src/metrics/test/metrics_test.cpp
```diff
@@ -482,5 +483,24 @@ TEST_CASE("Phases can set metrics") {
...
+    metrics.operation("actor1", "op1", 3u);
+
+    metrics.operation("actor2", "op1", 4u);
+    metrics.operation("actor2", "op1", 5u);
+
+    std::size_t expected1 = 3;
+    std::size_t expected2 = 2;
+
+    REQUIRE(metrics.getWorkerCount("actor1", "op1") == expected1);
+    REQUIRE(metrics.getWorkerCount("actor2", "op1") == expected2);
```
rtimmons: just inline expecteds? `REQUIRE(count == 3)`?

aggrand:
evergreen merge

PR #327, opened 2020-02-18
TIG-2349 Start Poplar gRPC in Genny

aggrand:
Things this PR does:

- Refactor the `toolchain.py` downloader to into an ABC with child toolchain and curator downloaders and download curator.
- The `lamp` setup script will install a virtualenv unless the `-g` or `--run-global` option is passed in.
- Create a `genny_runner.py` script that will start poplar, run genny, then close poplar. This will be installed by `setup.py`
- Rename the core binary to `genny_core` so that `setup.py` can install `genny_runner.py` as `genny`. 
- Rename the `genny` python package to `gennylib` so it's not conflated with the `genny` executable.
- Let `lamp` run `setup.py` to install the python packages and `genny` executable.
- Created a ./scripts/genny script that can automatically source the virtualenv and run genny.

Still need to update the README docs to reflect these changes for running things. I'll put up a corresponding PR for DSI to use the new entry point (TIG-2378) and we can merge them simultaneously. 

Patch build: https://evergreen.mongodb.com/version/5e4c523461837d060a43f3ee


src/driver/src/DefaultDriver.cpp
```diff
@@ -249,14 +249,15 @@ boost::log::trivial::severity_level parseVerbosity(const std::string& level) {
...
 
+const std::string RUNNER_NAME = "genny";
 
 DefaultDriver::ProgramOptions::ProgramOptions(int argc, char** argv) {
     namespace po = boost::program_options;
 
     std::ostringstream progDescStream;
     // Section headers are prefaced with new lines.
     progDescStream << u8"\n🧞 Usage:\n";
-    progDescStream << "    " << argv[0] << " <subcommand> [options] <workload-file>\n";
```
aggrand: This is mostly to make the wrappers invisible to users.

src/lamplib/tasks/__init__.py
```diff
@@ -46,12 +47,16 @@ def compile_all(context, env):
...
-
 def install(context, env):
+    
     install_cmd = [context.BUILD_SYSTEM, '-C', 'build', 'install']
     logging.info('Running install: %s', ' '.join(install_cmd))
     subprocess.run(install_cmd, env=env)
 
+    # Install python scripts
+    python_cmd = ['python3', 'setup.py', 'develop']
+    subprocess.run(python_cmd, cwd='src/python')
```
aggrand: We have to use a subprocess call instead of using the .py file more natively because setuptools makes use of the program's CLI arguments.

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,126 @@
...
+    def _get_url(self):
+        raise NotImplementedError
+
+    def _can_ignore(self):
+        raise NotImplementedError
+
+class ToolchainDownloader(Downloader):
+
+    def __init__(self, os_family, distro):
+        super().__init__(os_family, distro, Context.TOOLCHAIN_ROOT, "gennytoolchain")
```
guoyr: 💅  Consider moving the Context.*TOOLCHAIN* options here so Context doesn't become too bloated. Dittor for curator options stored on the context

src/python/gennylib/genny_runner.py
```diff
@@ -0,0 +1,48 @@
...
+    args[0] = genny_core
+    return args
+
+def get_poplar_args():
+    """
+    Returns the argument list used to create the Poplar gRPC process.
+    """
+    return ['/data/mci/curator/curator', 'poplar', 'grpc']
+
+@contextmanager
```
guoyr: Oh very clever!

evergreen.yml
```diff
@@ -368,28 +368,29 @@ functions:
         set -eo pipefail
 
         rm -rf /data/mci/gennytoolchain*
+        rm -rf /data/mci/curator*
```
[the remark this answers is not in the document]
aggrand: Makes sense; I'll modify execution of curator to look in either the CWD or, if that fails, look in the path.

evergreen.yml
```diff
@@ -368,28 +368,29 @@ functions:
...
 
         export DEVELOPER_DIR="/Applications/Xcode10.1.app"
 
         ./scripts/lamp \
           -d=${distro|not-linux} \
           -DENABLE_JASPER:BOOL=ON \
-          -DGENNY_INSTALL_DIR:PATH=${workdir}/dist \
+          -DGENNY_INSTALL_DIR:PATH=${workdir}/src/dist \
```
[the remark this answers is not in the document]
aggrand: The genny script looks in `dist` in the root of the genny repo for the `genny_core` binary (since that's where it gets built by default). Note that the `src` directory mentioned above isn't the `src` directory inside the repo. `src` is the genny repository which seems kinda confusingly named

scripts/lamp
```diff
@@ -1,7 +1,36 @@
...
+
 # for mongodbtoolchain python3
 export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
 
+if [[ "$*" != *-g* ]]
+then
+    if [[ ! -d genny_venv ]]
+    then
+        echo "Installing virtual environment."
+        echo "To run without a virtual environment, use the -g or --run-global option."
```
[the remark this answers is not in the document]
aggrand: That's fair, but what about when lamp gets run by automated scripts? Maybe just have a `-y` option that will mean "yes install it"?

scripts/lamp
```diff
@@ -1,7 +1,36 @@
...
+        python_path=/opt/mongodbtoolchain/v3/bin/python3
+        if [[ ! -e "$python_path" ]]; then
+            echo "${python_path} does not exist"
+            python_path="$(command -v python3.7)"
+        fi
+        echo "creating new env with python: ${python_path}"
+        _pip install virtualenv
```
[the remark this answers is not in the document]
aggrand: Oof good to know

scripts/lamp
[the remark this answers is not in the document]
guoyr: Oh good point. Ignore my comment then. What you have LGTM

evergreen.yml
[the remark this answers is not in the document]
guoyr: > `src` is the genny repository itself

Right, because we set it to be `src` on line 376.

I think using workdir was wrong to start with. `PATH=$PWD/dist` is probably what we want to do here.

src/python/gennylib/genny_runner.py
```diff
@@ -0,0 +1,60 @@
+import sys
+import os
+import subprocess
+from contextlib import contextmanager
+
+def get_genny_args():
+    """
+    Returns the argument list used to create the Genny process.
+
+    If we are in the root of the genny repo, use the local executable.
```
guoyr: 👍

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def _check_toolchain_githash(self):
+        res = subprocess.run(['git', 'rev-parse', 'HEAD'], cwd=self.result_dir, capture_output=True, text=True)
+        return res.stdout.strip() == ToolchainDownloader.TOOLCHAIN_GIT_HASH
+
+
+class CuratorDownloader(Downloader):
+    CURATOR_VERSION = '9fee6c2020c3d85bbe8fffa03e1d0e224c9652f5'
+    CURATOR_BUILD_DATE = '20_02_12_21_20_42'
+    CURATOR_ROOT = os.getcwd()
```
guoyr: 👍

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,134 @@
...
+        if self._os_family == 'Darwin':
+            self._distro = 'macos'
+
+        if 'ubuntu' in self._distro:
+            self._distro = 'ubuntu1604'
+
+        if self._distro in ('amazon2', 'rhel8', 'rhel62'):
+            self._distro = 'rhel70'
+
+        return 'https://s3.amazonaws.com/mciuploads/curator/' \
```
guoyr: Forgot to mention that you should get curator from the push task, that way you have a shorter URL. E.g. https://s3.amazonaws.com/boxes.10gen.com/build/curator/curator-dist-rhel70-6f7afe441326c1a6fa726e341d092839369b9d95.tar.gz

This way you won't need to record the build date

scripts/lamp
```diff
@@ -1,7 +1,32 @@
...
+    if [[ ! -d venv ]]
+    then
+        echo "Installing virtual environment."
+        echo "To run without a virtual environment, use the -g or --run-global option."
+        python_path=/opt/mongodbtoolchain/v3/bin/python3
+        if [[ ! -e "$python_path" ]]; then
+            python_path="$(command -v python3)"
+        fi
+        echo "creating new env with python: ${python_path}"
+        /usr/bin/env virtualenv -q "${DIR}/venv" --python="${python_path}"
```
guoyr: Couple of comments.
1. I don't think you need `/usr/bin/env` is necessary outside of `#!`, because it just looks at the first thing in your path anyway.
2. Python3's virtualenv is called `venv` and is 90% similar to virtualenv, you'll want to use the following syntax here:
`python3 -m venv venv`
3. related to the above point, you can remove the code related to python_path in this file. Instead, just add `/opt/mongodbtoolchain/v3/bin/python3` to the front of $PATH and call `python3` directly

src/python/gennylib/genny_runner.py
```diff
@@ -0,0 +1,68 @@
...
+        yield poplar
+    finally:
+        try:
+            poplar.terminate()
+            exit_code = poplar.wait(timeout=10)
+            if exit_code not in (0, -15): # Termination or exit.
+                raise OSError("Poplar exited with code: {code}.".format(code=(exit_code)))
+
+        except:
+            # If Poplar doesn't die then future runs can be broken.
```
guoyr: This is... alarming. What error is poplar throwing? We should file a MAKE ticket if we can't get poplar to shutdown.

guoyr: Per offline discussion, we are just cleaning up resources here. We expect terminate() to work in the happy case.

evergreen.yml
```diff
@@ -385,7 +385,7 @@ functions:
         ./scripts/lamp \
           -d=${distro|not-linux} \
           -DENABLE_JASPER:BOOL=ON \
-          -DGENNY_INSTALL_DIR:PATH=${workdir}/dist \
+          -DGENNY_INSTALL_DIR:PATH=$PWD/dist \
```
rtimmons: quoting - `-DGENNY_INSTALL_DIR="$PWD"/dist` (this was latent before but may as well fix now)

evergreen.yml
```diff
@@ -395,12 +395,12 @@ functions:
...
     params:
-      working_dir: .
+      working_dir: src
       binary: /bin/bash
       args:
-      - ${workdir}/src/scripts/dry-run-workloads.sh
+      - scripts/dry-run-workloads.sh
       env:
-        GENNY: ${workdir}/dist/bin/genny
+        GENNY: scripts/genny
```
rtimmons: hm maybe `dry-run-workloads.sh` should know about `scripts/genny` now?

scripts/genny
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+# Runs the genny python wrapper in the venv which runs the binary.
```
rtimmons: `set -eou pipefail` at the top and put `VIRTUAL_ENV_DISABLE_PROMPT=1` before the `source` to prevent errors about `$PS1` being undefined.

scripts/lamp
```diff
@@ -1,7 +1,28 @@
...
 
+set -eou pipefail
+
+DIR=$(pwd)
+
 # for mongodbtoolchain python3
 export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
 
+if [[ "$*" != *-g* ]]
+then
```
rtimmons: 💅 the scripts tend to use `if [[ foo ]]; then` (one line)

scripts/lamp
```diff
@@ -1,7 +1,28 @@
...
 # Because genies come from lamps.
 
+set -eou pipefail
+
+DIR=$(pwd)
+
 # for mongodbtoolchain python3
 export PATH=/opt/mongodbtoolchain/v3/bin:$PATH
 
+if [[ "$*" != *-g* ]]
```
rtimmons: this syntax is nice but mystical - suggest a comment `# Check if called with -g flag`

src/lamplib/__main__.py
```diff
@@ -7,22 +7,27 @@
...
     res.check_returncode()
     sys.exit(0)
 
-
 def validate_environment():
     # Check Python version
     if not sys.version_info >= (3, 7):
-        logging.error('Please run this script with Python 3.7 or newer')
+        logging.error('Detected Python version less than 3.7. Please delete '
+                      'the virtualenv and run lamp again.')
```
rtimmons: if feasible/easy, let's include `sys.version_info` in the message e.g. "Detected Python version (3.6) less than 3.7....`

src/lamplib/__main__.py
```diff
@@ -54,11 +58,20 @@ def main():
     if args.subcommand == 'self-test':
         run_self_test()
 
-    url = get_toolchain_url(os_family, args.linux_distro)
-    toolchain_dir = fetch_and_install_toolchain(url, Context.TOOLCHAIN_ROOT)
+    toolchain_downloader = ToolchainDownloader(os_family, args.linux_distro)
+    if not toolchain_downloader.fetch_and_install():
+        sys.exit(1)
```
rtimmons: let's just raise here - in general it's bad karma to sys.exit anywhere except in the catch block of a top-most-level try/catch

src/lamplib/__main__.py
```diff
@@ -68,8 +81,8 @@ def main():
         tasks.clean(context, compile_env)
     else:
         tasks.compile_all(context, compile_env)
-
         if args.subcommand == 'install':
+            check_venv(args)
```
rtimmons: hm why not just always `check_venv` at the top of the method?

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,132 @@
+import logging
+import os
+import shutil
+import subprocess
+import sys
+import urllib.request
+
+from context import Context
+
+class Downloader():
```
rtimmons: should we be using [abc](https://docs.python.org/3/library/abc.html) here? (I actually don't know what the best-practices are these days). Either way we're python3.7+ so we don't need the parens just `class Downloader` 💅

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    Abstract base class for things that can be downloaded.
+
+    :return: The directory the target was installed to.
+    """
+
+    def __init__(self, os_family, distro, install_dir, name):
+        self._os_family = os_family
```
rtimmons: plz doc at least the types of these (all strings)

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        Download the target and install to its directory.
+
+        :return: Whether the operation succeeded.
+        """
+        if not os.path.exists(self._install_dir):
+            logging.critical(
+                'Please create the parent directory for %s: '
+                '`sudo mkdir -p %s; sudo chown $USER %s`', self._name, self._install_dir, self._install_dir)
+            return False
```
rtimmons: let's (at least try to) quote this properly for the user
`sudo mkdir -p "%s"; sudo chown "$USER" "%s"`

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,132 @@
...
+
+        if not os.access(self._install_dir, os.W_OK):
```
rtimmons: would this catch if the "install dir" were actually a file and not a directory?

src/lamplib/tasks/download.py
```diff
@@ -0,0 +1,132 @@
...
+class ToolchainDownloader(Downloader):
+    TOOLCHAIN_BUILD_ID = 'cd5a4031b1dc93e47d598ad41521fd2e8aa865a0_20_02_12_20_07_30'
+    TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split('_')[0]
+    TOOLCHAIN_ROOT = '/data/mci'  # TODO BUILD-7624 change this to /opt.
+
+    def __init__(self, os_family, distro):
+        super().__init__(os_family, distro, ToolchainDownloader.TOOLCHAIN_ROOT, "gennytoolchain")
+
+    def _get_url(self):
+        # These build IDs are from the genny-toolchain Evergreen task.
```
rtimmons: I think this comment makes most sense above `TOOLCHAIN_BUILD_ID`

src/python/gennylib/genny_runner.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Returns the argument list used to create the Genny process.
+
+    If we are in the root of the genny repo, use the local executable. 
+    Otherwise we search the PATH.
+    """
+    args = sys.argv
+    genny_core = 'genny_core'
+
+    local_core = './dist/bin/genny_core'
```
rtimmons: Personally I've never been much of a fan of the `dist` directory since if you run `ninja` or `make` again inside `build` after you make changes it doesn't update the stuff inside `dist`. The only way to update `dist` it seems is to run `lamp` again (which unfortunately is hard to make idempotent). Not sure what to suggest other than to start a small fire and slowly homer-simpson-hedge-wall away

src/lamplib/__main__.py
aggrand: Running cmake tests and such don't require the venv so I didn't want to make people enter the venv. This was before it started being done automatically so now I'm fine with always doing it

scripts/genny
aggrand: Ah didn't know about the latter trick!

evergreen.yml
aggrand: Question: What does the quoting accomplish here?

rtimmons: You don't want to break tokenization if `$PWD` has any spaces.
E.g. if your cwd is literally `;rm -rf /;` (which is allowed on macos since it allows any unicode code-points in filenames) you could end up...messing a couple things up 😸

evergreen.yml
aggrand: It does and uses that in the default case! Figured I'd keep setting the variable to be explicit, let me know if you disagree.

aggrand:
evergreen merge

PR #329, opened 2020-03-04
TIG-2261 Implement genny::metrics API using Poplar gRPC

aggrand:
All the changes to actually use poplar!

Most of the additions are in `event.hpp` which has the `EventStream` class that completely wraps the new metrics system.

Still need to add unit tests, had made some progress with that but had trouble without thorough mocking, may need some tips there, since we'll want to assert that the right calls are being made to stub mocks. Figured they'd be brittle if changes are recommended anyway so went ahead and put this up. Did some end-to-end testing instead and we're getting valid-seeming ftdc files!

Patch: https://evergreen.mongodb.com/version/5e6029a461837d36674cd325

### DIFF
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -80,19 +80,19 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     // setup logging as the first thing we do.
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= options.logVerbosity);
 
-    genny::metrics::Registry metrics;
-
     const auto workloadName = fs::path(options.workloadSource).stem().string();
-    auto start_time = metrics::clock::now();
+    auto start_time = genny::metrics::Registry::clock::now();
 
     if (options.runMode == DefaultDriver::RunMode::kListActors) {
         globalCast().streamProducersTo(std::cout);
+        genny::metrics::Registry metrics;
         reportMetrics(metrics, workloadName, true, start_time);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
     if (options.workloadSource.empty()) {
         std::cerr << "Must specify a workload YAML file" << std::endl;
+        genny::metrics::Registry metrics;
         reportMetrics(metrics, workloadName, false, start_time);
         return DefaultDriver::OutcomeCode::kUserException;
     }
@@ -125,6 +125,7 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
 
     if (options.runMode == DefaultDriver::RunMode::kEvaluate) {
         std::cout << YAML::Dump(yaml) << std::endl;
+        genny::metrics::Registry metrics;
         reportMetrics(metrics, workloadName, true, start_time);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
@@ -135,8 +136,11 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
                               ? options.workloadSource
                               : "inline-yaml"};
 
+
     auto workloadContext =
-        WorkloadContext{nodeSource.root(), metrics, orchestrator, options.mongoUri, globalCast()};
+        WorkloadContext{nodeSource.root(), orchestrator, options.mongoUri, globalCast()};
+
+    genny::metrics::Registry& metrics = workloadContext.getMetrics();
 
     if (options.runMode == DefaultDriver::RunMode::kDryRun) {
         std::cout << "Workload context constructed without errors." << std::endl;
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -184,7 +184,7 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     for (auto& thread : threads)
         thread.join();
 
-    if (metrics.getFormat().use_csv()) {
+    if (metrics.getFormat().useCsv()) {
         const auto reporter = genny::metrics::Reporter{metrics};
 
         {
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -65,6 +65,14 @@ void runActor(Actor&& actor,
     }
 }
 
+void reportMetrics(genny::metrics::Registry& metrics, const std::string& workloadName, bool success, metrics::clock::time_point start_time) {
+    auto finish_time = metrics::clock::now();
+    auto actorSetup = metrics.operation(workloadName, "Setup", 0u);
+    auto outcome = success ? metrics::OutcomeType::kSuccess : metrics::OutcomeType::kFailure;
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(finish_time - start_time);
+    actorSetup.report(std::move(finish_time), std::move(duration), std::move(outcome));
+}
+
 DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& options) {
     // setup logging as the first thing we do.
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= options.logVerbosity);
@@ -72,18 +80,17 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     genny::metrics::Registry metrics;
 
     const auto workloadName = fs::path(options.workloadSource).stem().string();
-    auto actorSetup = metrics.operation(workloadName, "Setup", 0u);
-    auto setupCtx = actorSetup.start();
+    auto start_time = metrics::clock::now();
 
     if (options.runMode == DefaultDriver::RunMode::kListActors) {
         globalCast().streamProducersTo(std::cout);
-        setupCtx.success();
+        reportMetrics(metrics, workloadName, true, start_time);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
     if (options.workloadSource.empty()) {
         std::cerr << "Must specify a workload YAML file" << std::endl;
-        setupCtx.failure();
+        reportMetrics(metrics, workloadName, false, start_time);
         return DefaultDriver::OutcomeCode::kUserException;
     }
 
@@ -115,7 +122,7 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
 
     if (options.runMode == DefaultDriver::RunMode::kEvaluate) {
         std::cout << YAML::Dump(yaml) << std::endl;
-        setupCtx.success();
+        reportMetrics(metrics, workloadName, true, start_time);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
@@ -130,14 +137,14 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
 
     if (options.runMode == DefaultDriver::RunMode::kDryRun) {
         std::cout << "Workload context constructed without errors." << std::endl;
-        setupCtx.success();
+        reportMetrics(metrics, workloadName, true, start_time);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
     orchestrator.addRequiredTokens(
         int(std::distance(workloadContext.actors().begin(), workloadContext.actors().end())));
 
-    setupCtx.success();
+    reportMetrics(metrics, workloadName, true, start_time);
 
     auto startedActors = metrics.operation(workloadName, "ActorStarted", 0u);
     auto finishedActors = metrics.operation(workloadName, "ActorFinished", 0u);
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -69,7 +69,7 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     // setup logging as the first thing we do.
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= options.logVerbosity);
 
-    genny::metrics::Registry metrics(options.metricsFormat, options.metricsPathPrefix);
+    genny::metrics::Registry metrics;
 
     const auto workloadName = fs::path(options.workloadSource).stem().string();
     auto actorSetup = metrics.operation(workloadName, "Setup", 0u);
@@ -174,14 +174,14 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     for (auto& thread : threads)
         thread.join();
 
-    if (options.metricsFormat.use_csv()) {
+    if (metrics.getFormat().use_csv()) {
         const auto reporter = genny::metrics::Reporter{metrics};
 
         {
             std::ofstream metricsOutput;
-            metricsOutput.open(options.metricsOutputFileName,
+            metricsOutput.open(metrics.getPathPrefix() + ".csv",
                                std::ofstream::out | std::ofstream::trunc);
-            reporter.report(metricsOutput, options.metricsFormat);
+            reporter.report(metricsOutput, metrics.getFormat());
         }
     }
 
@@ -278,15 +278,6 @@ DefaultDriver::ProgramOptions::ProgramOptions(int argc, char** argv) {
             ("subcommand", po::value<std::string>(), "1st positional argument")
             ("help,h",
              "Show help message")
-            ("metrics-format,m",
-             po::value<std::string>()->default_value("csv"),
-             "Metrics format to use")
-            ("metrics-output-file,o",
-             po::value<std::string>()->default_value("/dev/stdout"),
-             "Save metrics data to this file when using csv format. Use `-` or `/dev/stdout` for stdout.")
-            ("metrics-path-prefix,p",
-             po::value<std::string>()->default_value("."),
-             "Save metrics data to this directory when using ftdc format. Defaults to current directory.")
 
             ("workload-file,w",
              po::value<std::string>(),
@@ -349,10 +340,7 @@ DefaultDriver::ProgramOptions::ProgramOptions(int argc, char** argv) {
         this->runMode = RunMode::kHelp;
 
     this->logVerbosity = parseVerbosity(vm["verbosity"].as<std::string>());
-    this->metricsFormat = metrics::MetricsFormat(vm["metrics-format"].as<std::string>());
     this->isSmokeTest = vm["smoke-test"].as<bool>();
-    this->metricsOutputFileName = normalizeOutputFile(vm["metrics-output-file"].as<std::string>());
-    this->metricsPathPrefix = normalizeOutputFile(vm["metrics-path-prefix"].as<std::string>());
     this->mongoUri = vm["mongo-uri"].as<std::string>();
 
     if (vm.count("workload-file") > 0) {
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -284,7 +284,7 @@ DefaultDriver::ProgramOptions::ProgramOptions(int argc, char** argv) {
             ("metrics-output-file,o",
              po::value<std::string>()->default_value("/dev/stdout"),
              "Save metrics data to this file when using csv format. Use `-` or `/dev/stdout` for stdout.")
-            ("metrics-path-prefix,o",
+            ("metrics-path-prefix,p",
              po::value<std::string>()->default_value("."),
              "Save metrics data to this directory when using ftdc format. Defaults to current directory.")
 
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -68,12 +68,12 @@ void runActor(Actor&& actor,
 void reportMetrics(genny::metrics::Registry& metrics,
                    const std::string& workloadName,
                    bool success,
-                   metrics::clock::time_point start_time) {
-    auto finish_time = metrics::clock::now();
+                   metrics::clock::time_point startTime) {
+    auto finishTime = metrics::clock::now();
     auto actorSetup = metrics.operation(workloadName, "Setup", 0u);
     auto outcome = success ? metrics::OutcomeType::kSuccess : metrics::OutcomeType::kFailure;
-    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(finish_time - start_time);
-    actorSetup.report(std::move(finish_time), std::move(duration), std::move(outcome));
+    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(finishTime - startTime);
+    actorSetup.report(std::move(finishTime), std::move(duration), std::move(outcome));
 }
 
 DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& options) {
@@ -81,19 +81,19 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     boost::log::core::get()->set_filter(boost::log::trivial::severity >= options.logVerbosity);
 
     const auto workloadName = fs::path(options.workloadSource).stem().string();
-    auto start_time = genny::metrics::Registry::clock::now();
+    auto startTime = genny::metrics::Registry::clock::now();
 
     if (options.runMode == DefaultDriver::RunMode::kListActors) {
         globalCast().streamProducersTo(std::cout);
         genny::metrics::Registry metrics;
-        reportMetrics(metrics, workloadName, true, start_time);
+        reportMetrics(metrics, workloadName, true, startTime);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
     if (options.workloadSource.empty()) {
         std::cerr << "Must specify a workload YAML file" << std::endl;
         genny::metrics::Registry metrics;
-        reportMetrics(metrics, workloadName, false, start_time);
+        reportMetrics(metrics, workloadName, false, startTime);
         return DefaultDriver::OutcomeCode::kUserException;
     }
 
@@ -126,7 +126,7 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     if (options.runMode == DefaultDriver::RunMode::kEvaluate) {
         std::cout << YAML::Dump(yaml) << std::endl;
         genny::metrics::Registry metrics;
-        reportMetrics(metrics, workloadName, true, start_time);
+        reportMetrics(metrics, workloadName, true, startTime);
         return DefaultDriver::OutcomeCode::kSuccess;
     }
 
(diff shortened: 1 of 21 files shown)


src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,279 @@
...
+
+    std::string _name;
+    poplar::PoplarID _id;
+};
+
+/**
+ * Class that keeps track of the duration and totals for an operation across all threads.
+ */
+template <typename ClockSource>
+struct DurationCounter {
```
aggrand: This doesn't actually track duration across all threads, just for the current thread. I think I'll cut another ticket to make it actually track across all threads because I'm not sure of the best way to architecture that yet.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,279 @@
...
+ * Manages the stream of poplar EventMetrics.
+ */
+class StreamInterface {
+public:
+    StreamInterface(CollectorStubInterface& stub)
+        : _options{}, _response{}, _context{}, _stream{stub->StreamEvents(&_context, &_response)} {
+        _options.set_no_compression().set_buffer_hint();
+    }
+
+    void write(const poplar::EventMetrics& event) {
```
aggrand: After about 30 minutes, writes start failing because the grpc server kicks us out due to too many pings. I think a thread-safe global channel refresh might be necessary, unless we can figure out channel options to prevent that. The default C++ channel options are compatible between client and server, but the language-agnostic documentation doesn't say if that's necessarily the case between the Go server defaults and the C++ client defaults.

rtimmons: What do you mean by "too many pings" and "global channel refresh"? I had a hard time finding good docs on the client stuff so if you've found a good source I wanna catch up.

src/metrics/include/metrics/v2/event.hpp
rtimmons: We shouldn't ever need cross-thread duration since the grpc client is supposed to do that for us (albeit it has to lock because...go...)

rtimmons:
🌶  I'm impressed this only ended up being 500 lines! Nicely done. I think I want a demo/walkthrough of this tomorrow though.

src/metrics/include/metrics/v2/event.hpp
aggrand: I haven't found that many details, I just noticed that after 30 minutes grpc gives me a a`GOAWAY` with error code `ENHANCE_YOUR_CALM` due to `too_many_pings`. Apparently that can happen depending on the ping settings of the client and the settings of the server ([go](https://godoc.org/google.golang.org/grpc/keepalive) doc, [C++](https://github.com/grpc/grpc/blob/master/doc/keepalive.md) doc). Might ping cedar people tomorrow to see if there are any weird settings for server enforcement.

By "global channel refresh" I mean taking the class-static channel variable, acquiring a lock, and replacing it with a new one. (I recognize that we don't want those explicit locks, but if it's brief and every 30 minutes maybe it's not so bad?) That's a heavy-handed approach though and I'd rather not need to re-connect. There might be another option in the API but I'm still learning there.

Only detailed references I can find beyond the gRPC tutorials are the [API reference](https://grpc.github.io/grpc/cpp/index.html) and the source repo.

src/metrics/include/metrics/v2/event.hpp
aggrand: Slightly confused what you mean by that, this is for the `duration` field in the `EventMetrics` message, which we need to set manually? Should probably chat in-person about that.

aggrand: After an in-person discussion, we learned that the cross-thread summation can be done automatically poplar-side if we have cross-thread collectors.

src/metrics/include/metrics/v2/event.hpp
aggrand: This was resolved by disabling the BDP probe.

src/metrics/test/metrics_test.cpp
```diff
@@ -502,5 +508,109 @@ TEST_CASE("Registry counts the number of workers") {
...
+    SECTION("Empty stream") {
+        RegistryClockSourceStub::reset();
+        auto stream = internals::v2::EventStream<RegistryClockSourceStub, 
+             internals::v2::MockStreamInterface>{1, "TestName", 1, "/test/prefix"};
+
+        EventVec expected;
+        compareEventsAndClear(expected);
+    }
+
+    SECTION("Two events") {
```
aggrand: This test is probably pretty brittle but it found some errors that I'd had so there's some value. If necessary, we could template-ize the Registry on the grpc stub to do tests involving it.

src/driver/src/DefaultDriver.cpp
```diff
@@ -174,13 +184,15 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
     for (auto& thread : threads)
         thread.join();
 
-    const auto reporter = genny::metrics::Reporter{metrics};
+    if (metrics.getFormat().use_csv()) {
```
rtimmons: 💅 we use camelCase in C++

src/gennylib/src/context.cpp
```diff
@@ -52,6 +53,12 @@ WorkloadContext::WorkloadContext(const Node& node,
     // Make sure we have a valid mongocxx instance happening here
     mongocxx::instance::current();
 
+    // Set the metrics format information.
+    auto format = ((*this)["Metrics"]["format"]).maybe<std::string>().value_or("cedar-csv");
```
rtimmons: 💅 Genny uses PascalCase for all keys so this would be `["Metrics"]["Format"]`

src/metrics/include/metrics/MetricsReporter.hpp
```diff
@@ -25,11 +25,11 @@
...
 /**
- * @namespace genny::metrics::v1 this namespace is private and only intended to be used by genny's
- * own internals. No types from the genny::metrics::v1 namespace should ever be typed directly into
- * the implementation of an actor.
+ * @namespace genny::metrics::internals::v1 this namespace is private and only intended to be used
+ * by genny's own internals. No types from the genny::metrics::internals::v1 namespace should ever
+ * be typed directly into the implementation of an actor.
  */
-namespace v1 {
+namespace internals::v1 {
```
rtimmons: The `v1` namespace is already there to indicate it's "internal" - I think this can safely live in just `v1`

src/gennylib/src/event.cpp
```diff
@@ -0,0 +1,7 @@
+#include <metrics/v2/event.hpp>
+
+namespace genny::metrics::internals::v2 {
+
+std::unique_ptr<poplar::PoplarEventCollector::StubInterface> CollectorStubInterface::_stub(nullptr);
+
```
rtimmons: Note to self - come back here.

1. `event.cpp` is a strange name for this imho.
2. Can this instead be done inside the impl's cpp?
3. if nothing else plz comment what this is

src/metrics/include/metrics/Period.hpp
```diff
@@ -21,6 +21,8 @@
 
 namespace genny::metrics {
 
+const auto NANO_PER_SECOND = 1000000000;
```
rtimmons: shouldn't need this - `std::chrono` can do all this math using `std::chrono::duration_cast`.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -28,11 +29,57 @@
...
  */
-namespace v1 {
+class MetricsFormat {
+public:
+    enum class Format {
+        csv = 0,
+        cedar_csv = 1,
+        ftdc = 2,
+        csv_ftdc = 3,
+    };
```
rtimmons: 1. Convention is to have `kCamelCase`
2. No need to specify enum values

```c++
enum class Format {
    kCsv,
    kCedarCsv,
...
}
```

I'd also put a comment on what cedar csv is versus csv versus csv-ftdc 😄

rtimmons: I don't think you need this if you just have `enum MetricsFormat` and you can add [a `struct NodeConvert` impl](https://github.com/mongodb/genny/blob/master/src/gennylib/include/gennylib/Node.hpp#L232-L249) so you can extract from the yaml

(but wait you're doing some logic in use_grpc and use_csv - do those need to be exposed outside of that one hpp file - can they just be `inline`d there?

src/metrics/include/metrics/MetricsReporter.hpp
aggrand: There were a couple things (like the `OperationImpl` class) that were relevant for both the old metrics system and the new one, while being internal. I tried to refactor the namespaces so that `v1` contains things that are only relevant to the old csv-based system and `v2` is things that are part of the new system, while both are `internal`.

aggrand: (so the `v1` and `v2` namespaces are both inside the `internal` namespace, but some things that are used by both systems are just inside the `internal` namespace)

src/metrics/include/metrics/metrics.hpp
aggrand: The use_grpc and use_csv get used in the registry, in the operationimpl, and in the default driver to make choices related to the new/old system. Ideally we wouldn't need these flags and in the future when we exorcise the old metrics system we could remove them.

aggrand:
> I'm curious what changes after we no longer need to support multiple metrics formats. Is there a straightforward TDLR on that? I'm thinking we may want to basically start from scratch on `genny::metrics` (and the related stuff in the python) at that point and just treat the existing corpus of code as an organ-donor. Seems cleaner than trying to reason about what each kind of format needs.

Ideally, we could convert to the new metrics system by finding any spot where `use_grpc` and `use_csv` are called, and then just removing the `use_csv` branch and related logic. That might be less clean than an implementation that was built straight from the grpc-based metrics though.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -82,6 +129,7 @@ class RegistryT final {
     // OperationsMap is a map of
     // actor name -> operation name -> actor id -> OperationImpl (time series).
     using OperationsMap = std::unordered_map<std::string, OperationsByType>;
+    using CollectorsMap = std::unordered_map<std::string, v2::Collector>;
```
rtimmons: Let's add a comment here saying that this is a map from "Actor.Operation.Phase" to collector

src/metrics/include/metrics/metrics.hpp
```diff
@@ -92,13 +140,23 @@ class RegistryT final {
...
         auto& opsByThread = opsByType[opName];
-        auto opIt =
-            opsByThread
-                .try_emplace(
-                    actorId, std::move(actorName), *this, std::move(opName), std::move(phase))
-                .first;
+        auto opIt = opsByThread
+                        .try_emplace(actorId,
+                                     actorId,
+                                     std::move(actorName),
```
rtimmons: er..I think I'm confused about how this works/why this changed - we didn't change anything under opsByThread so not sure why we now need to pass actorId twice (or really why any change to this portion was necessary)? I'm like 100% reading this wrong though I'd guess.

src/metrics/include/metrics/operation.hpp
```diff
@@ -197,14 +223,19 @@ class OperationImpl final : private boost::noncopyable {
...
+        return *_events;
     }
 
     void reportAt(time_point started, time_point finished, OperationEventT<ClockSource>&& event) {
         if (_threshold) {
             _threshold->check(started, finished);
         }
-        _events.addAt(finished, event);
+        if (_registry.getFormat().use_grpc()) {
+            _stream->addAt(finished, event, _registry.getWorkerCount(_actorName, _opName));
```
rtimmons: this never changes after we're done with setup - is there a way to pass this value once rather than continuously asking the registry?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+    poplar::PoplarEventCollector::StubInterface* operator->() {
+        return _stub.get();
+    }
+
+private:
+    // We should only have one channel across all threads.
+    static std::unique_ptr<poplar::PoplarEventCollector::StubInterface> _stub;
+
+    auto createStub() {
```
rtimmons: comment that this should only be called by `setStub()`

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+    grpc::WriteOptions _options;
+    poplar::PoplarResponse _response;
+    grpc::ClientContext _context;
+    std::unique_ptr<grpc::ClientWriterInterface<poplar::EventMetrics>> _stream;
+};
+
+/**
+ * Very basic mock for tests.
+ */
+class MockStreamInterface {
```
rtimmons: is it feasible to put this impl into the tests rather than letting test code (albeit simple test code) live in production?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+/**
+ * Manages the gRPC-side collector for each operation.
+ * Exists for construction / destruction resource management only.
+ */
+class Collector {
+public:
+    Collector(const Collector&) = delete;
+
+    explicit Collector(const std::string& name, const std::string& path_prefix)
```
rtimmons: I'd add a brief comment to these RAII classes that they're only constructed/destructed at setup & teardown time so efficiency isn't a concern and we're really trying to ensure correctness of setup and teardown operations.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+            BOOST_THROW_EXCEPTION(PoplarRequestError(os.str()));
+        }
+    }
+
+    ~Collector() {
+        grpc::ClientContext context;
+        poplar::PoplarResponse response;
+        auto status = _stub->CloseCollector(&context, _id, &response);
+        if (!status.ok()) {
+            BOOST_LOG_TRIVIAL(error) << "Couldn't close collector: " << status.error_message();
```
rtimmons: here and elsewhere, see if it's feasible to give names or more context in error messages - it may be helpful to know *which* collector fails to start/stop when debugging by looking at the state of the ftdc files dir etc.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+    void addAt(const typename ClockSource::time_point& finish,
+               const OperationEventT<ClockSource>& event,
+               size_t workerCount) {
+        _metrics.mutable_time()->set_seconds(
+            std::chrono::time_point_cast<std::chrono::seconds>(finish).time_since_epoch().count());
+        _metrics.mutable_time()->set_nanos(
+            std::chrono::time_point_cast<std::chrono::nanoseconds>(finish)
+                .time_since_epoch()
+                .count());
```
rtimmons: Here and elsewhere in this function - I **think** you need to mod by seconds - it's [only the fractional portion](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto#L133-L137)

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+        _metrics.mutable_counters()->set_number(event.iters);
+        _metrics.mutable_counters()->set_ops(event.ops);
+        _metrics.mutable_counters()->set_size(event.size);
+        _metrics.mutable_counters()->set_errors(event.errors);
+
+        _metrics.mutable_gauges()->set_failed(event.isFailure());
+        _metrics.mutable_gauges()->set_workers(workerCount);
+        if (_phase) {
+            _metrics.mutable_gauges()->set_state(*_phase);
+        }
```
rtimmons: comment here that "state" is an arbitrary value per the poplar spec and we're using it to store phase.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+        _metrics.mutable_timers()->mutable_duration()->set_nanos(
+            event.duration.getNanosecondsCount());
+
+        _metrics.mutable_timers()->mutable_total()->set_seconds(
+            Period<ClockSource>(finish - _last_finish).getSecondsCount());
+        _metrics.mutable_timers()->mutable_total()->set_nanos(
+            Period<ClockSource>(finish - _last_finish).getNanosecondsCount());
+
+        _metrics.mutable_counters()->set_number(event.iters);
+        _metrics.mutable_counters()->set_ops(event.ops);
```
rtimmons: elsewhere the `event` has names consistent with the poplar names - does `event.iters` want to be `event.number` just to keep the "glue" code as dumb as possible?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -0,0 +1,282 @@
...
+        _metrics.mutable_timers()->mutable_duration()->set_nanos(0);
+        _metrics.mutable_timers()->mutable_total()->set_nanos(0);
+        _metrics.mutable_counters()->set_errors(0);
+        _metrics.mutable_counters()->set_number(0);
+        _metrics.mutable_counters()->set_ops(0);
+        _metrics.mutable_counters()->set_size(0);
+        _metrics.mutable_gauges()->set_state(0);
+        _metrics.mutable_gauges()->set_workers(0);
+        _metrics.mutable_gauges()->set_failed(false);
+        //_metrics.mutable_time()->set_nanos(0);
```
rtimmons: 1. do we actually need this reset function?
2. remove commented-out last field?
3. no need to prefix private methods with underscore - if this is needed it can be just `void reset()`

src/metrics/test/metrics_test.cpp
```diff
@@ -502,5 +508,122 @@ TEST_CASE("Registry counts the number of workers") {
...
+TEST_CASE("Events stream to gRPC") {
+    using EventVec = std::vector<poplar::EventMetrics>;
+
+    auto compareEventsAndClear = [](const EventVec& events_in) {
+        internals::v2::MockStreamInterface interface;
+        REQUIRE(events_in.size() == interface.events.size());
+        for (int i = 0; i < events_in.size(); i++) {
+            REQUIRE(google::protobuf::util::MessageDifferencer::Equals(events_in[i],
+                                                                       interface.events[i]));
+            // REQUIRE(2 == 1);
```
rtimmons: 💅

src/metrics/test/metrics_test.cpp
```diff
@@ -502,5 +508,122 @@ TEST_CASE("Registry counts the number of workers") {
...
+
+
+        // First Event
+        RegistryClockSourceStub::advance(std::chrono::microseconds(5));
+        {
+            OperationEventT<RegistryClockSourceStub> event(
+                2,
+                3,
+                7,
+                0,
```
rtimmons: I wish C++ had lightweight named params here! Can you add inline comments about what these fields are

```
event(
  2, // duration
  3, // size
```

src/metrics/test/metrics_test.cpp
```diff
@@ -502,5 +508,122 @@ TEST_CASE("Registry counts the number of workers") {
...
+            metric.mutable_counters()->set_ops(3);
+            metric.mutable_counters()->set_size(7);
+            metric.mutable_counters()->set_errors(0);
+
+            metric.mutable_gauges()->set_failed(false);
+            metric.mutable_gauges()->set_workers(1);
+            metric.mutable_gauges()->set_state(9);
+
+            expected.push_back(metric);
+        }
```
rtimmons: Feels like we've got good happy-path testing, but personally I would be a bit afraid to make changes to the code with this testing.

Any way to assert:

1. the filenames being generated
2. what happens when two actors (threads with same Actor,Operation) send and/or get collectors?
3. behavior when the gRPC side of things barfs
4. behavior when metrics formats are other values
5. ensure that when an actor fails to call .success or .fail that we do/don't report to grpc?


It could well be that this testing is hard given the current design or testing libraries. If that's the case we can certainly punt on adding more un-happy-path tests and do them whenever we rip out csv metrics and/or when we add a better story for mocking. If that's the case, let's write a few sentences listing the cases that would be interesting to test but that are hard to test and make sure we point to those words in the places in the code that are currently untested. This will help future maintainers set their expectations about what the tests are or aren't doing.

src/metrics/test/metrics_test.cpp
rtimmons: Riffing: I'm not sure we need to templatize on the whole registry, but we can certainly re-think some of the design assumptions knowing that we can freely use virtuals (we were *very* cautious of *anything* that somebody might be able to "blame" performance anomalies on). I can imagine having OperationImpl be a pure-virtual class (interface) and similar for the grpc glue such that you could easily assert the interactions between Registry<->Operation<->GRPC. That's basically a rewrite of the API but most of the internals stay the same and the testability (probably) goes way up. This is probably *remarkably* easier once we kill csv stuff, though.

ANYWAY - I think you've templatized at a fine level of granularity, but I really think I want to spend a few days cleaning up when we have the csv junk hauled away.

src/gennylib/src/event.cpp
rtimmons: (still not convinced we need this but C++ static initialization is annoying (but getting better in C++20 I've heard). I think this would be much clearer with a comment, perhaps a different name for the cpp, and using assignment syntax `StubInterface::_stub = nullptr` rather than a [vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse).

src/metrics/include/metrics/MetricsReporter.hpp
rtimmons: okay that's reasonable. Maybe add a comment somewhere that v1 mostly goes away once we kill csv, and perhaps at that point we rename `internal::v2` to just `v1`.

src/metrics/include/metrics/metrics.hpp
aggrand: We added actorId to the constructor for OperationImpl. So now the first actorId is used as the key for the map and the second actorId is passed into the constructor. We needed to pipe in the actorId so that it can be piped into the id field in poplar's EventMetrics.

src/metrics/include/metrics/v2/event.hpp
aggrand: You're right, and I'm really disappointed that the gRPC C++ library doesn't have any automatic interoperability between std::chrono classes and its own time system.

src/metrics/include/metrics/metrics.hpp
aggrand: I ended up having a hard time getting the `NodeConvert` stuff to work right, so I ended up keeping it wrapped in the class and creating a constructor that takes a Node& so that extraction from yaml is more idiomatic. Let me know if you still think it should be an enum outside the class and I can ping you for questions about the `NodeConvert` struct.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -140,13 +206,43 @@ class RegistryT final {
         return (_ops.at(actorName).at(opName)).size();
     }
 
+    void initializeMetrics(MetricsFormat format, std::string path_prefix) {
+        _format = std::move(format);
+        _path_prefix = std::move(path_prefix);
+        if (_format.use_grpc()) {
+            boost::filesystem::create_directory(_path_prefix);
+        }
+    }
```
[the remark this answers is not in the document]
aggrand: Yeah but since the workload context is what reads the metrics format, I had to make the registry be owned by the workload context instead of the driver, and added a method to expose a reference to it to the driver.

src/gennylib/include/gennylib/context.hpp
```diff
@@ -251,6 +250,8 @@ class WorkloadContext : public v1::HasNode {
      */
     v1::GlobalRateLimiter* getRateLimiter(const std::string& name, const RateSpec& spec);
 
+    metrics::Registry& getMetrics() {return _registry;}
```
rtimmons: 💅 ensure you run clang-format prior to merging

src/metrics/include/metrics/Period.hpp
```diff
@@ -65,6 +65,19 @@ class Period final {
...
+    std::chrono::seconds::rep getSecondsCount() const {
+        return std::chrono::duration_cast<std::chrono::seconds>(_duration).count();
+    }
+
+    /**
+     * Gets the nanosecond component that wouldn't be counted in seconds.
+     */
+    std::chrono::nanoseconds::rep getNanosecondsCount() const {
+        return std::chrono::duration_cast<std::chrono::nanoseconds>(_duration).count() -
+            std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::duration_cast<std::chrono::seconds>(_duration)).count();
```
rtimmons: yikes - your call consider `using namespace std::chrono` at the top of this function 😄

src/metrics/include/metrics/metrics.hpp
```diff
@@ -15,24 +15,75 @@
...
+            _format == Format::kCsvFtdc;
+    }
+
+    Format get() const {
+        return _format;
+    }
+
+private:
+    Format _format;
+    Format str_to_enum(const std::string& to_convert) {
```
rtimmons: `strToEnum` (won't call out others but plz ensure it's camelCase throughout)

src/metrics/include/metrics/metrics.hpp
```diff
@@ -82,23 +133,44 @@ class RegistryT final {
...
+    using CollectorsMap = std::unordered_map<std::string, v2::Collector>;
 
 public:
     using clock = ClockSource;
 
     explicit RegistryT() = default;
 
+    explicit RegistryT(MetricsFormat format, std::string path_prefix) {
+        _format = std::move(format);
+        _path_prefix = std::move(path_prefix);
```
rtimmons: you can do this in an initializer block

```c++
explicit Registry(MetricsFormat format, std::string prefi)
: _fomat{std::move(format)}, _prefix{std::move(prefix)} {
  if (format...
}
```

src/metrics/include/metrics/metrics.hpp
```diff
@@ -82,23 +133,44 @@ class RegistryT final {
...
-                .first;
+        auto opIt = opsByThread
+                        .try_emplace(actorId,
+                                     actorId,
+                                     std::move(actorName),
+                                     *this,
+                                     std::move(opName),
+                                     std::move(phase),
+                                     _path_prefix,
+                                     name)
```
rtimmons: looks like `name` may be default-constructed here?

src/metrics/include/metrics/operation.hpp
rtimmons: bump here - the workerCount shouldn't ever change post cosntruciton

src/metrics/include/metrics/v2/event.hpp
rtimmons: bump on this

src/metrics/test/metrics_test.cpp
rtimmons: plz add a note about cases that aren't covered (maybe including some of my examples here). We'll want to circle back to add them once we remove CSV

src/metrics/include/metrics/operation.hpp
aggrand: Ahh we'd figured out in the last ticket (where the `getWorkerCount` function was added), that since the actors are constructed in sequence then the earlier `OperationImpl` constructions don't know what the final number of workers added to the phase will be.

src/metrics/include/metrics/metrics.hpp
aggrand: Oof thanks for catching!

src/metrics/include/metrics/metrics.hpp
aggrand: Yeah should only be used in gRPC situations, made it optional to signify that.

src/metrics/test/metrics_test.cpp
aggrand: Added a comment above the `TEST_CASE` explaining that, hopefully exhaustive enough.

aggrand:
Scheduled a pb with genny tasks on all the variants to look for regressions: https://evergreen.mongodb.com/version/5e72a0087742ae405ff7274a

aggrand:
evergreen merge

PR #331, opened 2020-03-19
TIG-2461 Document the new genny invocation

aggrand:
I had previously forgotten to update the README to reflect the new way of invoking Genny. I also went ahead and added a FTDC-based HelloWorld workload.


src/workloads/docs/HelloWorldFtdc.yml
```diff
@@ -0,0 +1,25 @@
...
+    # MetricsName: 🐳Message
+  - Message: Hello Phase 1 👬
+    Repeat: 100
+  - ExternalPhaseConfig:
+      Path: ../../phases/HelloWorld/ExamplePhase2.yml
+      Key: UseMe  # Only load the YAML structure from this top-level key.
+      Parameters:
+        Repeat: 2
+
+Metrics:
```
guoyr: Since the distinction between different metric formats is subtle enough, my vote is to not introduce the notion at all to new workload authors, who would be the target audience to helloworld workloads. Instead, suggest adding this section to the existing hello world workload in a commented-out format and add a comment above to say that people can use the ftdc metrics format instead of the default csv (I wouldn't say "Cedar" here since it could confuse the reader) format if they expect to record more than XYZ (10M?) data points.

README.md
```diff
@@ -176,14 +176,14 @@ And then run a workload:
...
-    --metrics-format      csv                                       \
-    --metrics-output-file build/genny-metrics.csv                   \
     --mongo-uri           'mongodb://localhost:27017'
 ```
 
-Logging currently goes to stdout, and time-series metrics data is
-written to the file indicated by the `-o` flag (`./genny-metrics.csv`
-in the above example).
+Logging currently goes to stdout, and in this example, metrics data is
+written to `./build/genny-metrics.csv`. The metrics format and output
```
guoyr: Similar to my comment below, I think of the metrics format as something for advanced users, so personally I would leave its description in the HelloWorld workload file instead of the README. Whatdoyathink?

aggrand: Yeah that sounds reasonable; I'll just modify the README to say where the results will be outputted.

aggrand:
evergreen merge

PR #333, opened 2020-03-25
TIG-2386 Add connection establishment latency test

sgolemon:
This diff creates a simple workload to invoke runCommand({replSetTestEgress: 1}) against replset-auth-delay configuration.

See DSI PR 592 for a detailed description of the goals of this project.  TL;DR - Track connection establishment time.

### DIFF
--- a/src/workloads/networking/TransportLayerConnectTiming.yml
+++ b/src/workloads/networking/TransportLayerConnectTiming.yml
@@ -1,0 +1,33 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/security"
+Description: >-
+    Invoke replSetTestEgress command on replica set members.
+    Command synchronously establishes temporary connections between cluster nodes.
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 10
+      socketTimeoutMS: 5000
+      connectTimeoutMS: 5000
+Actors:
+- Name: TestEgress
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Phase: 0
+    Duration: 5 minutes
+    Database: admin
+    ThrowOnFailure: false
+    Operation:
+      OperationMetricsName: Connection
+      OperationName: RunCommand
+      OperationIsQuiet: true
+      OperationCommand:
+        replSetTestEgress: 1
+
+AutoRun:
+  Requires:
+    bootstrap:
+      mongodb_setup: 
+        - replica-auth-cluster-delay


src/workloads/networking/TransportLayerConnectTiming.yml
```diff
@@ -0,0 +1,31 @@
...
+      maxPoolSize: 10
+      socketTimeoutMS: 5000
+      connectTimeoutMS: 5000
+Actors:
+- Name: TestEgress
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Phase: 0
+    Duration: 180 seconds
```
rtimmons: 💅 this can be `3 minutes` if you want 😄

src/workloads/networking/TransportLayerConnectTiming.yml
```diff
@@ -0,0 +1,31 @@
...
+    GlobalRate: 1 per 1 microsecond
+    Database: admin
+    ThrowOnFailure: false
+    Operation:
+      OperationMetricsName: "Connection"
```
rtimmons: 💅 can remove the quotes

src/workloads/networking/TransportLayerConnectTiming.yml
```diff
@@ -0,0 +1,31 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/security"
+
```
rtimmons: Consider adding `Description: this workload does....` for future maintainers / build-barons.

src/workloads/networking/TransportLayerConnectTiming.yml
sgolemon: Done

src/workloads/networking/TransportLayerConnectTiming.yml
sgolemon: Done

src/workloads/networking/TransportLayerConnectTiming.yml
sgolemon: Done

sgolemon:
> You may want to consider running for longer than 3 minutes - it probably takes at least 5-10 mins to spin up the cluster so may as well gather data for longer. (Or was this the thing that you wanted to run longer but Genny's barfing on too much data at the moment?)

I copied this duration from ServiceArchitectureWorkloads (along with the bulk of the file).  I'll give 5min a try and see how it fares.

dalyd:
Oh, apologies from me on this one -- it fell off my radar (and I thought I was FYI). I'll take a look today.

src/workloads/networking/TransportLayerConnectTiming.yml
```diff
@@ -0,0 +1,34 @@
...
+      socketTimeoutMS: 5000
+      connectTimeoutMS: 5000
+Actors:
+- Name: TestEgress
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Phase: 0
+    Duration: 5 minutes
+    GlobalRate: 1 per 1 microsecond
```
dalyd: How long does a connection take to create? More specifically is it greater than or less than a microsecond. If it's less we don't need to rate limit it at all and could let the thread run as fast as it can.

sgolemon: At minimum 200ms using the configuration we have (with induced latency), but even without that extra latency I'd be amazed if these connection establishments took less than 1 MILLIsecond, let alone a microsecond.  May I assume your use of "If it's less" was a typo, and you're saying that in this case I can remove the line completely and just let 'er ride?

dalyd: > At minimum 200ms using the configuration we have (with induced latency), but even without that extra latency I'd be amazed if these connection establishments took less than 1 MILLIsecond, let alone a microsecond. May I assume your use of "If it's less" was a typo, and you're saying that in this case I can remove the line completely and just let 'er ride?

Yeah -- just delete line 20 and let her ride. It shouldn't be doing anything with those times. If your intent was to put delay between connection creation, that would be a different thing. 

@rtimmons can you just confirm I haven't said something crazy? I haven't touched Genny in a bit.

sgolemon:
evergreen merge

PR #334, opened 2020-03-25
TIG-2478 FTDC outputs for setup operation have negative totals

aggrand:
This just make sure that, if the _last_finished var is after the finish (i.e. if the EventStream was constructed and _last_finish was initialized to now after the finish time was recorded), we just use the duration as the total this time around, next recorded operation will work as normal since _last_finish will be updated.

If we expect a lot of edge cases like this, it might make sense to expose a synthetic_report equivalent for the gRPC stream.


aggrand:
evergreen merge

PR #336, opened 2020-03-30
TIG-2470 Change AutoRun syntax

rtimmons:
TLDR: Slight change to the syntax for `AutoRun`.

This is the first part that will enable reading the state of the environment from `expansions.yml`.

Currently we an `AutoRun` section looks like this:

```yaml
AutoRun:
  Requires:
    bootstrap:
      mongodb_setup: replica
  PrepareEnvironmentWith:
    setup:
    - replica-delay-mixed
    - replica
```

This says "run when `${bootstrap.mongodb_setup} == 'standalone'`". But we want to be able to do this without the presence of `bootstrap.yml`. And the keys are actually unique so there's not `bootstrap.mongodb_setup` **and** `infrastructure_provisioning.mongodb_setup`; there's no reason to "namespace" the requires entries.

After state:

```yaml
AutoRun:
  Requires:
    mongodb_setup: replica
  PrepareEnvironmentWith:
    mongodb_setup:
    - replica-delay-mixed
    - replica
```

Also 

1. reduced some complexity by not allowing scalars and instead always require lists.
2. renamed the `setup` param for `PrepareEnvironmentWith` to be `mongodb_setup`

### DIFF
--- a/src/python/gennylib/genny_auto_tasks.py
+++ b/src/python/gennylib/genny_auto_tasks.py
@@ -40,10 +40,6 @@ class AutoRunSpec():
         required_dict = None
         if 'Requires' in workload_yaml['AutoRun'] and isinstance(autorun['Requires'], dict):
             required_dict = autorun['Requires']
-            for module, config in required_dict.items():
-                if not isinstance(config, dict):
-                    required_dict = None
-                    break
 
         prepare_environment_with = None
         if 'PrepareEnvironmentWith' in autorun and isinstance(autorun['PrepareEnvironmentWith'], dict):
@@ -71,6 +67,7 @@ class AutoRunSpec():
             curr.update(self.prepare_environment_with)
             prepare_environment_vars.append(curr)
         return prepare_environment_vars
+
 
 def to_snake_case(str):
     """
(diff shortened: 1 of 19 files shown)


guoyr:
```
AutoRun:
  Requires:
    mongodb_setup: replica
  PrepareEnvironmentWith:
    mongodb_setup:
    - replica-delay-mixed
    - replica
```

Interestingly, the simplification has surfaced a latent issue, in that there's `mongodb_setup` and there's`mongodb_setup`.

it seems confusing that we expect mongodb_setup to be replica but we're setting it to replica-delay-mixed. I think really what we want to say with the `mongodb_setup` in the `Requires` section is to run it on a certain task, in this case `linux-3-node-replSet` (maybe also `wtdevelop-3-node-replSet`, which has the same mongodb_setup).

Let me know if the above makes sense

src/python/gennylib/genny_auto_tasks.py
```diff
@@ -140,28 +158,15 @@ def workload_should_autorun(autorun_spec, env_dict):
...
-    for module, required_config in autorun_spec.required_dict.items():
-        if module not in env_dict:
+    simplified = _simplified_env_dict(env_dict)
+    for key, values in autorun_spec.required_dict.items():
+        if not isinstance(values, list):
+            raise Exception(f"Must give a list for key {key}")
+        if key not in simplified:
             return False
-        if not isinstance(required_config, dict):
+        if not any([simplified[key] == value for value in values]):
```
guoyr, line 167: 👍

PR #338, opened 2020-04-01
Run black and require black formatting in CI

rtimmons:
This depends on #336.

Only "real" changes here are in evergreen.yml and setup.py; rest are just auto-formatting from `black`.


rtimmons:
evergreen merge

PR #339, opened 2020-04-02
PERF-1959 Add replica variant to unionWith perf tests

kaywux:
Add 3 Node ReplSet variant to UnionWith workload.


kaywux:
evergreen merge

PR #341, opened 2020-04-08
[nomerge] Rewrite Auto-Tasks

rtimmons:
- Wanted some eyes before I add additional testing.
- Tested locally using old-style invocations from system_perf.yml and new-style (proposed new syntax that I'll incorporate into DSI in the future)
- Once everything's on new-style any references to "legacy" in this can be removed (quick POC of this in 4d9da76)
- I don't plan to add tests for the legacy stuff since it's going away, and the new-style invocations are much easier to test anyway


src/python/gennylib/auto_tasks.py
```diff
@@ -0,0 +1,361 @@
...
+    def _load(self):
+        if self.conts is not None:
+            return
+        conts = _yaml_load(
+            [os.path.join(self.cwd, f"{b}.yml") for b in {"bootstrap", "runtime", "expansions"}]
+        )
+        if "expansions" in conts:
+            self.use_expansions_yml = True
+            conts = conts["expansions"]
+        else:
```
guoyr, line 144: Will the else-clause go away eventually?

src/python/gennylib/auto_tasks.py
```diff
@@ -0,0 +1,361 @@
...
+        return "./src/genny"
+
+    @property
+    def output_file(self) -> str:
+        if self.is_legacy:
+            return os.path.join(self.repo_root, self.output_file_suffix)
+        return self.output_file_suffix
+
+    @staticmethod
+    def parse(argv: List[str]) -> "CLIOperation":
```
guoyr, line 91: argparse?

src/python/gennylib/auto_tasks.py
```diff
@@ -0,0 +1,361 @@
...
+    return out
+
+
+class DirectoryStructure:
+    def __init__(self, repo_root: str):
+        self.repo_root = repo_root
+
+    def all_workload_files(self) -> Set[str]:
+        pattern = os.path.join(self.repo_root, "src", "workloads", "*", "*.yml")
+        return {*glob.glob(pattern)}
```
guoyr, line 51: 👍

src/python/gennylib/auto_tasks.py
```diff
@@ -0,0 +1,361 @@
...
+    def all_tasks(self) -> List[GeneratedTask]:
+        base = self._to_snake_case(self.file_base_name())
+        if self.setups is None:
+            return [GeneratedTask(base, None, self)]
+        return [
+            GeneratedTask(f"{base}_{self._to_snake_case(setup)}", setup, self)
+            for setup in self.setups
+        ]
+
+    def variant_tasks(self, runtime: Runtime) -> List[GeneratedTask]:
```
guoyr, line 212: I didn't quite get how the 3 `variant_tasks` function fit together when skimming this file. Asking bc I'm curious how we can use this to get to the cheap/expensive buildvariant world you described the other day.

src/python/gennylib/auto_tasks.py
rtimmons: Yessir! We'll just assert `expansions.yml` exists.

src/python/gennylib/auto_tasks.py
rtimmons: Debated it, but new-world will not have arguments just a single argv subcommand. Debatable if argparse is worth it but punting that for now.

src/python/gennylib/auto_tasks.py
rtimmons: One is on the `Workload` that indicates what tasks (if any) the workload yaml file contributes to the tasks for the given variant. Another is on the `Repo` which basically flatmaps over all its workloads `variant_tasks` values. The final is on the writer which translates the Repo's value to the shrub config.

It seemed simpler to me in code than in writing it just now, so I'll ensure this is clear by names or comments. Good thing to point out.

> how we can use this to get to the cheap/expensive buildvariant world

I'm not sure to be honest - not sure how we list the various combinations of bootstrap inputs if we don't do it by build-variant. Maybe each workload lists the runs it wants to have:

```yaml
AutoRun:
  Expensive:
  - mongodb_setup: replica
    infrastructure_provisioning: replica
```

but that may be too much tying Genny workloads to DSI stuff.

Thankfully I think I've got the right level of factoring in here to allow such a change to be contained to just the `Workload` class 😸

guoyr: SGTM

rtimmons:
Thanks for review @guoyr  - going to close this for now and re-open once ready for production review.

PR #343, opened 2020-04-10
TIG-2497 Use non-blocking gRPC calls in Genny metrics

aggrand:
gRPC can only have one call "in-flight" per stream, so what this does is for every operation that writes to the stream, check if there's already an in-flight operation. If so, block until it completes then do the write.

Hopefully this corrects the performance problem. If not we might have to do something more complicated like adding buffers or more streams per thread.

[Patch-build](https://evergreen.mongodb.com/version/5e90c155a4cf47259c6d318b) where I modified InsertRemove to use FTDC outputs.

### DIFF
--- a/src/metrics/include/metrics/v2/event.hpp
+++ b/src/metrics/include/metrics/v2/event.hpp
@@ -19,8 +19,8 @@
 #include <atomic>
 #include <cstdlib>
 
-#include <boost/log/trivial.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/log/trivial.hpp>
 
 #include <grpcpp/client_context.h>
 #include <grpcpp/create_channel.h>
@@ -103,19 +103,19 @@ public:
           _response{},
           _context{},
           _cq{},
-          _tag{(void*) 1},
+          _tag{(void*)1},
           _stream{_stub->AsyncStreamEvents(&_context, &_response, &_cq, _tag)} {
         _options.set_no_compression().set_buffer_hint();
-        finish_call();
+        finish_call();  // We expect a response from the initial construction.
     }
 
     void write(const poplar::EventMetrics& event) {
         if (!finish_call()) {
-                std::ostringstream os;
-                os << "Failed to write to stream for operation name " << _name << " and actor ID "
-                   << _actorId << ". EventMetrics object: " << event.ShortDebugString();
+            std::ostringstream os;
+            os << "Failed to write to stream for operation name " << _name << " and actor ID "
+               << _actorId << ". EventMetrics object: " << event.ShortDebugString();
 
-                BOOST_THROW_EXCEPTION(PoplarRequestError(os.str()));
+            BOOST_THROW_EXCEPTION(PoplarRequestError(os.str()));
         }
 
         _stream->Write(event, _options, _tag);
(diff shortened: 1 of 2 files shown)


src/metrics/include/metrics/v2/event.hpp
```diff
@@ -98,23 +98,28 @@ class StreamInterfaceImpl {
     StreamInterfaceImpl(const std::string& name, const ActorId& actorId)
         : _name{name},
           _actorId{actorId},
+          _in_flight{true},
           _options{},
           _response{},
           _context{},
-          _stream{_stub->StreamEvents(&_context, &_response)} {
+          _cq{},
+          _tag{(void*)1},
```
rtimmons: this seems archaic - can you document why this value?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -137,13 +155,28 @@ class StreamInterfaceImpl {
     }
 
 private:
+    bool finish_call() {
+        if (_in_flight) {
+            void* got_tag;
```
rtimmons, line 160: 💅 CamelCase

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -137,13 +155,28 @@ class StreamInterfaceImpl {
...
     std::string _name;
     ActorId _actorId;
+    bool _in_flight;
     CollectorStubInterface _stub;
     grpc::WriteOptions _options;
     poplar::PoplarResponse _response;
     grpc::ClientContext _context;
-    std::unique_ptr<grpc::ClientWriterInterface<poplar::EventMetrics>> _stream;
+    grpc::CompletionQueue _cq;
+    void* _tag;
```
rtimmons, line 178: while not incorrect this name isn't very..obvious. Instead maybe `_grpcInFlight`? I think it can also be `const` but actually just realizing the rest of this isn't const so prolly best to not shake it up.

src/metrics/include/metrics/v2/event.hpp
aggrand: Yeah I think I updated with a comment right as you wrote this haha. It's basically a meaningless tag associated with each call to identify them, but is still required syntactically. Maybe useful in situations where we don't know which message is which? Which isn't here.

rtimmons: lol - I think some variation on what you just said would be helpful here - I'm sure I'll forget what this is in approximately 5 minutes 😸 

```
// This is used by the gRPC client to indicate which message is being completed in calls to XYZ. We only ever have 1 message in flight at a time, so the value doesn't really matter but it is required by the API.
```

I'm also curious if we should fail if the recieved tag isn't `==(void*)1`? Would that indicate that there's murder afoot? If unsure let's add a comment **or** (my preference) barf and we can remove it if/when we actually hits us.

aggrand: Good point, will comment. We actually are checking that they're equal, that's what the `gotTag == _grpcTag` part of the `finishCall()` private method does. I'll comment to clarify that though since it's not obvious without digging that the pointer passed to `_cq.Next` gets modified with the return

aggrand: Didn't want to throw in `finishCall` itself so we could use it in the destructor. IMHO though that sort of C-style pointer manipulation feels really out of place in a C++ context.

aggrand:
evergreen merge

PR #344, opened 2020-04-13
TIG-2464 Create a FTDC-based canary in Genny

aggrand:
Will have a corresponding DSI change with the new invocation.

Patch-build: https://evergreen.mongodb.com/version/5e936cde2a60ed7ca1ae6d1f


src/canaries/src/Loops.cpp
```diff
@@ -140,6 +166,45 @@ Nanosecond Loops<Task, Args...>::metricsPhaseLoop(Args&&... args) {
...
+                                               std::make_optional(IntegerSpec(_iterations)),
+                                               false,
+                                               0_ts,
+                                               0_ts,
+                                               std::nullopt),
+        1};
+    auto task = Task(std::forward<Args>(args)...);
+
+    boost::filesystem::path ph =
+            boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
```
guoyr, line 186: 👍

aggrand:
@guoyr regarding the numbers, running the benchmarks on my machine I get roughly a factor of 5 difference for the nop between the metrics loop and the metricsFtdc loop, and similar for the real loops. For the more expensive tasks that aren't just a nop the difference between FTDC and non-FTDC metrics options is only about 2% to 5%, presumably because the metrics are a smaller fraction of the operation.

guoyr:
> The difference between FTDC and non-FTDC metrics options is only about 2% to 5%

Very nice. Merge away!

aggrand:
evergreen merge

PR #345, opened 2020-04-13
TIG-2494 Set timeouts in genny auto tasks on a per-task level

aggrand:
Sets the timeouts using `timeout.update` since setting it at the top level doesn't work when generating tasks.


aggrand:
Patch build actually running long: https://evergreen.mongodb.com/version/5e93542f7742ae582b499d67

src/python/gennylib/genny_auto_tasks.py
```diff
@@ -279,12 +278,15 @@ def construct_all_tasks_json():
         task_name = to_snake_case(base_parts[0])
 
         prepare_environment_vars = get_prepare_environment_vars(task_name, fname)
+        timeout_params = {"exec_timeout_secs": 86400, # 24 hours
+                          "timeout_secs": 86400}
```
guoyr, line 282: This is the timeout for no output, often incorrectly referred to as the "test timeout". We should probably lower it to 2 hours or something. If test runs for more than 2 hours, it needs to provide progress updates.

src/python/gennylib/genny_auto_tasks.py
```diff
@@ -263,7 +263,6 @@ def construct_all_tasks_json():
     :return: json representation of tasks for all workloads in the /src/workloads directory relative to the genny root.
     """
     c = Configuration()
-    c.exec_timeout(64800)  # 18 hours
```
guoyr, line 266: Did you get a chance to file the ticket to fix shrub.py?

src/python/gennylib/genny_auto_tasks.py
aggrand: 2 hours is actually the default in evergreen; I had to set it higher to run InsertRemove since it doesn't output anything. If we want to require progress updates though I'm cool with that.

src/python/gennylib/genny_auto_tasks.py
aggrand: Oof spent some time with a local fix before deciding to just use a command, but I'll do that now.

aggrand:
evergreen merge

PR #347, opened 2020-04-15
TIG-2505 Update poplar version in genny

aggrand:
Not running locally since the MacOS version of the curator waterfall hasn't run yet, but if it passes the evergreen tests we should be good.

Will run a patch to see if this actually fixes MAKE-1256 but either way we should still merge to fix MAKE-1257.


aggrand:
Putting up a DSI-side PR that'll have the patch-build.

aggrand:
evergreen merge

PR #355, opened 2020-05-20
TIG-2534 Don't assume lamp has run

richardsamuels:
This is pretty much the same as Ryan's previous PR, but setup has been moved over to the bash scripts.


richardsamuels:
Build failure is in head

scripts/env.sh
```diff
@@ -0,0 +1,25 @@
...
+            python3 -m venv venv
+        popd >/dev/null
+        touch "$SCRIPTS_DIR/venv/.setup-complete"
+    fi
+
+    set +u
+        source "${SCRIPTS_DIR}/venv/bin/activate"
+        pushd "$SCRIPTS_DIR/../src/python" >/dev/null
+            python3 setup.py --quiet develop
+        popd >/dev/null
```
rtimmons, line 23: this has the down-side of calling setup every time the `./scripts/genny` call is made. I think we can do his inside of the "if not already setup" block right before touching the setup-complete file.

src/lamplib/tasks/download.py
```diff
@@ -44,6 +45,18 @@ def fetch_and_install(self):
...
+                release_triplet = platform.mac_ver()[0].split(".")
+                minor_ver = int(release_triplet[1])
+                if minor_ver >= 12:
+                    logging.info(
+                        "On MacOS El Capitan and later, you have to disable System Integrity Protection. "
+                        "See https://apple.stackexchange.com/a/208481 for instructions")
+                if minor_ver >= 16:
+                    logging.info(
+                        "On MacOS Catalina and later, you have to remount the system root as rw: "
+                        "'sudo mount -uw /'")
```
rtimmons, line 59: this is ♨️♨️♨️

scripts/env.sh
richardsamuels: My only fear here was I can just see someone getting hurt by an out of date environment after someone changes setup.py, and wasting quite a bit of time trying to investigate.

 I'm guessing this was also a risk beforehand as well? I'm gonna defer to you here.

rtimmons: Ya good point, but I think that's fine. If somebody's explicitly modifying setup.py I don't think it's unreasonable to require them to run it directly (and they're likely going to have to understand this shell-python-shell triple-indirection anyway 🙄😸)

src/lamplib/tasks/download.py
guoyr: I've personally been using `synthetic.conf` instead of disabling SIP as it seems to be more of a Tim Cook sanctioned approach. https://apple.stackexchange.com/a/372500

guoyr:
Failure in PR check is caused by artifacts being reaped after 1 month. #364 will address the problem; I'll give you a ping after that goes in.

scripts/env.sh
```diff
@@ -0,0 +1,25 @@
...
+        pushd "$SCRIPTS_DIR" >/dev/null
+            python3 -m venv venv
+        popd >/dev/null
+    fi
+
+    set +u
+        source "${SCRIPTS_DIR}/venv/bin/activate"
+    set -u
+
+    if [[ ! -e "$SCRIPTS_DIR/venv/.setup-complete" ]]; then
```
rtimmons, line 19: you've already done the `.setup-complete` check so no need to do it again 💅

richardsamuels: I really hate the way this logic is expressed, but the second half of the logic finishes the setup after the venv is sourced.

I think it might be clearer to just copy and paste the venv activation into the setup block and early return.

richardsamuels:
Patch: https://evergreen.mongodb.com/version/5ec7e4af9ccd4e743f86ae79

scripts/env.sh
```diff
@@ -0,0 +1,34 @@
...
+        python3 -m pip install -e "$SCRIPTS_DIR/../src/python"
+        touch "$SCRIPTS_DIR/venv/.setup-complete"
+    fi
+fi
```
rtimmons, line 34: I'd recommend running this script through `shellcheck` - either `brew install shellcheck` or copy/paste [online](https://www.shellcheck.net/). We don't lint against it, but I'd like to. I consider myself pretty good at shell-games, but shellcheck always beats me.

richardsamuels:
evergreen merge

guoyr:
Going to merge manually since integration tests are still failing

PR #356, opened 2020-05-28
Use Genny venv for resmoke

guoyr:


src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,7 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
         # Default download location for MongoDB binaries.
         env["PATH"] += ":" + os.path.join(mongo_dir, "bin") + ":" + mongo_dir
 
-    evg_venv_dir = os.path.join(workdir, "build", "venv")
+    evg_venv_dir = os.path.join(os.environ['SCRIPTS_DIR'], "venv")
```
richardsamuels, line 113: Thanks to the -g flag, there is no guarantee that this venv exists which may affect people not using the automatically downloaded genny toolchain on weird platforms. Should we do anything about that?

If we're going to handwave this issue, then let's add a clean error message in the cases where the env variable is empty string/undefined. Also, can we rename it to LAMP_VENV_DIR since we're exporting it. You can leave all the other genny scripts alone.

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
         # Default download location for MongoDB binaries.
         env["PATH"] += ":" + os.path.join(mongo_dir, "bin") + ":" + mongo_dir
 
-    evg_venv_dir = os.path.join(os.environ['SCRIPTS_DIR'], "venv")
+    evg_venv_dir = os.path.join(os.environ['LAMP_VENV_DIR'], "venv")
```
richardsamuels, line 113: Oh i'm sorry, I didn't understand how os.environ worked. This will raise a KeyError exception if LAMP_VENV_DIR isn't set. Let's try/catch this., and raise a prettier error message

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
...
+
+    if not os.path.exists(evg_venv_dir,):
```
richardsamuels, line 115: I'm assuming that comma was a typo?

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
...
+        raise ValueError("The venv directory is required for resmoke and does not exist, "
+                         "please ensure you're running Genny in a virtualenv")
```
richardsamuels, line 117: I would change this error message to "please ensure you're not running Genny with the -g/--run-global flag." However, now I"m afraid we're basically just undoing the work from this ticket. Or are they completely separate worlds?

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
         # Default download location for MongoDB binaries.
         env["PATH"] += ":" + os.path.join(mongo_dir, "bin") + ":" + mongo_dir
 
-    evg_venv_dir = os.path.join(workdir, "build", "venv")
+    if 'LAMP_VENV_DIR' not in os.environ:
```
richardsamuels, line 113: Do you care about being 'pythonic'? The pythonic way of doing things is, "do it and handle the exception" instead of checking if something will avoid generating an exception in the first place

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
...
+        raise ValueError("The venv directory is required for resmoke and does not exist, "
```
richardsamuels, line 114: This error message is wrong and should say "The LAMP_VENV_DIR environment variable must be set to the parent directory of the path which contains the python virtual environment". Reuse this error message in the os.path.exists call mentioned elsewhere.

src/lamplib/tasks/run_tests.py
```diff
@@ -110,7 +110,11 @@ def resmoke_test(env, suites, mongo_dir, is_cnats):
...
+                         "please ensure you're not running Genny with the --run-global flag")
+
+    evg_venv_dir = os.path.join(os.environ['LAMP_VENV_DIR'], "venv")
```
richardsamuels, line 117: Right after this, an `os.path.exists(evg_venv_dir)` check is probably a good call.

src/lamplib/tasks/run_tests.py
guoyr: Ah the age old EAFP vs LBYL debate. For this check, I think of it as conceptually belonging to input validation, rather than the more generic theme of error handling. Even in Python, user input validation's main value as I see it is usability, where we can ensure the error message shows up early on to be visible, and we can capture arbitrary information that may exceed the length of an error message.

src/lamplib/tasks/run_tests.py
guoyr: LAMP_VENV_DIR is an internal concept hidden to genny users. Are you sure we want to mention it in the error message? :)

src/lamplib/tasks/run_tests.py
guoyr: Similar to above, the lack of a valid virtualenv directory is an internal genny error and should be something we're able to catch in our tests, not something the user has to see. The purpose of the error handling here is to capture user mistakes.

guoyr:
evergreen merge

PR #361, opened 2020-06-23
PERF-2001: InCacheSnapshotReads & OutOfCacheSnapshotReads

ldennis:
One of the phase (PointInTimeScanner-Scan-History) in the OutOfCacheSnapshotReads workload is pretty noisy at the moment (see the workload doc in the ticket). But I would like to open a PR first to have some feedbacks while I continue tuning some parameters for the workloads.
This PR extends the existing CollectionScanner actor to read at a point in time using {level: "snapshot", atClusterTime: <TS>}. Since the driver doesn't support reading at a point in time at the moment, I had to write my own find and getMore commands to do the scan.

### DIFF
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -18,6 +18,13 @@
 #include <chrono>
 #include <memory>
 
+#include <bsoncxx/array/view.hpp>
+#include <bsoncxx/builder/basic/array.hpp>
+#include <bsoncxx/builder/basic/document.hpp>
+#include <bsoncxx/builder/basic/kvp.hpp>
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <bsoncxx/types.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/collection.hpp>
 #include <mongocxx/database.hpp>
@@ -33,11 +40,14 @@
 #include <value_generators/DocumentGenerator.hpp>
 
 namespace genny::actor {
-enum class ScanType { kCount, kSnapshot, kStandard };
+enum class ScanType { kCount, kSnapshot, kStandard, kPointInTime };
 enum class SortOrderType { kSortNone, kSortForward, kSortReverse };
 // We don't need a metrics clock, as we're using this for measuring
 // the end of phase durations and scan durations.
 using SteadyClock = std::chrono::steady_clock;
+namespace bson_stream = bsoncxx::builder::stream;
+using bsoncxx::builder::basic::kvp;
+using bsoncxx::builder::basic::make_document;
 
 struct CollectionScanner::PhaseConfig {
     // We keep collection names, not collections here, because the
@@ -56,6 +66,7 @@ struct CollectionScanner::PhaseConfig {
     SortOrderType sortOrder;
     int64_t collectionSkip;
     std::optional<TimeSpec> scanDuration;
+    bool selectClusterTimeOnly = false;
     bool queryCollectionList;
     std::optional<mongocxx::options::transaction> transactionOptions;
 
@@ -73,7 +84,8 @@ struct CollectionScanner::PhaseConfig {
           documents{context["Documents"].maybe<IntegerSpec>().value_or(0)},
           scanSizeBytes{context["ScanSizeBytes"].maybe<IntegerSpec>().value_or(0)},
           collectionSkip{context["CollectionSkip"].maybe<IntegerSpec>().value_or(0)},
-          scanDuration{context["ScanDuration"].maybe<TimeSpec>()} {
+          scanDuration{context["ScanDuration"].maybe<TimeSpec>()},
+          selectClusterTimeOnly{context["selectClusterTimeOnly"].maybe<bool>().value_or(false)} {
         // The list of databases is comma separated.
         std::vector<std::string> dbnames;
         boost::split(dbnames, databaseNames, [](char c) { return (c == ','); });
@@ -87,7 +99,7 @@ struct CollectionScanner::PhaseConfig {
         }
 
         // Initialise scan type enum.
-        auto scanTypeString = context["ScanType"].to<std::string>();
+        auto scanTypeString = context["ScanType"].maybe<std::string>().value_or("");
         // Ignore case
         boost::algorithm::to_lower(scanTypeString);
         if (scanTypeString == "count") {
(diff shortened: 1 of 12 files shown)


src/cast_core/src/MultiCollectionUpdate.cpp
```diff
@@ -53,6 +54,8 @@ struct MultiCollectionUpdate::PhaseConfig {
 
     // uniform distribution random int for selecting collection
     std::uniform_int_distribution<int64_t> uniformDistribution;
+
+    metrics::Operation updateOperation;
```
ldennis, line 58: This was to support per phase metric names.

src/workloads/scale/LargeScaleModel.yml
```diff
@@ -161,6 +161,7 @@ Actors:
   - {Nop: true}
   - {Nop: true}
   - Duration: *Duration
+    MetricsName: Update
```
ldennis, line 164: These "MetricsName: Update" were my attempts to keep old metrics name for the phase because I made changes about to support per phase metric name for the MultiCollectionUpdate actor.
But I dont know why this test (and large_scale_parallel and large_scale_serial) shows up in my new build variant even without specifying "replica-1dayhistory-15gbwtcache" in AutoRun, see [this patch build](https://evergreen.mongodb.com/version/5ef23cb40305b93e1b259610). Any idea why?

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -73,7 +84,8 @@ struct CollectionScanner::PhaseConfig {
           documents{context["Documents"].maybe<IntegerSpec>().value_or(0)},
           scanSizeBytes{context["ScanSizeBytes"].maybe<IntegerSpec>().value_or(0)},
           collectionSkip{context["CollectionSkip"].maybe<IntegerSpec>().value_or(0)},
-          scanDuration{context["ScanDuration"].maybe<TimeSpec>()} {
+          scanDuration{context["ScanDuration"].maybe<TimeSpec>()},
+          selectClusterTimeOnly{context["selectClusterTimeOnly"].maybe<bool>().value_or(false)} {
```
rtimmons, line 88: 💅 `selectClusterTimeOnly` -> `SelectClusterTimeOnly`

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -227,9 +241,117 @@ void countScan(CollectionScanner::PhaseConfig* config,
...
 }
 
+void pointInTimeScan(mongocxx::pool::entry& client,
+                     CollectionScanner::PhaseConfig* config,
+                     std::optional<bsoncxx::types::b_timestamp> readClusterTime) {
+    int64_t docCount = 0;
+    int64_t scanSize = 0;
+    bool scanFinished = false;
+    mongocxx::options::client_session sessionOption;
+    sessionOption.causal_consistency(false);
```
rtimmons, line 251: you could probably construct the sessionOptions during PhaseConfig construction-time and re-use it but I'm not sure how this object's lifetime works

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -227,9 +241,117 @@ void countScan(CollectionScanner::PhaseConfig* config,
...
+    mongocxx::client_session session = client->start_session(sessionOption);
+    if (!readClusterTime) {
+        client->database("admin").run_command(session, make_document(kvp("ping", 1)));
+        readClusterTime = session.operation_time();
+    }
+    bsoncxx::document::value readConcern = bson_stream::document{}
+        << "level"
+        << "snapshot"
+        << "atClusterTime" << readClusterTime.value() << bson_stream::finalize;
```
rtimmons, line 260: here readClusterTime could be an empty optional

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -227,9 +241,117 @@ void countScan(CollectionScanner::PhaseConfig* config,
...
+             * Try-catch this as the collection may have been deleted.
+             * You can still do a find but it'll throw an exception when we iterate.
+             */
+            try {
+                bsoncxx::document::value findCmd = bson_stream::document{}
+                    << "find" << collection << "filter" << filter << "readConcern"
+                    << readConcern.view() << bson_stream::finalize;
+
+                auto findTracker = config->scanOperation.start();
+                auto res = database.run_command(session, findCmd.view());
```
rtimmons, line 282: if we barf here (like it seems like we can based on your comment above) the tracker will be held open and we'll fail to record the data-point. I think just move the `try` opening statement to below the `.start` and in your catch block call .failure.

src/workloads/scale/LargeScaleModel.yml
rtimmons: If you select genny_auto_tasks it will run this workload any time we have `mongodb_setup` set to `single`. If you only want your changed workload to run, you select `genny_patch_tasks`. Upcoming changes to parameterized builds in evergreen will hopefully make this less painful.

dalyd:
I'm not sure I'm going to have time to look today myself, but I did want t call out that this workload needs to run only on master -- it won't run on 4.4. That came up on the workload doc. I'm not sure how we do that with generated tasks. FYI @rtimmons

guoyr:
> it won't run on 4.4

That should be ok given the workloads run on a new buildvariant that only exists on master

src/cast_core/src/CollectionScanner.cpp
ldennis: This is only used in the pointInTimeScan and it is only called once per point-in-time scan phase. I think it is fine to keep it as is so the definition is closer to the code that uses it.

src/cast_core/src/CollectionScanner.cpp
ldennis: If readClusterTime is empty, we use session.operation_time() to initialize it at line 255. And that's expected to return a non-empty timestamp.

src/cast_core/src/CollectionScanner.cpp
ldennis: Good catch! Done.

src/cast_core/src/CollectionScanner.cpp
ldennis: Done.

ldennis:
Yes, the workloads specify `mongodb_setup: [replica-1dayhistory-15gbwtcache]` and the 4.4 branch doesn't have that buildvariant in system-perf.yml, I think it should be fine.

src/workloads/scale/LargeScaleModel.yml
rtimmons: Just for posterity, if you select genny_patch_tasks we'll always schedule all changed workloads for that variant even if they don't specify that variant in their autorun sections [as this test shows](https://github.com/mongodb/genny/blob/44ca5e0b99563fd1d3b261b5249f52763134c029/src/python/tests/auto_tasks_test.py#L154). I'm not convinced this is "right," but it only affects patch-builds and only downside is some wasted cycles and a false-purple. We can revisit this logic if this turns out to be a common thing.

ldennis:
Adding @luke-pearson because storage engine initially added those actors.

luke-pearson:
The changes to the actor I implemented look good to me.

ldennis:
evergreen merge

PR #363, opened 2020-07-13
PERF-2016 Create UserAcquisition workload

sgolemon:
This workload measures the amount of time required to acquire a User object from the local privilege database on a UserCache miss.

It is preliminary to work to remove the internal RoleGraph structure in favor of transactional aggregation based user document resolution.

### DIFF
--- a/src/workloads/execution/UserAcquisition.yml
+++ b/src/workloads/execution/UserAcquisition.yml
@@ -1,0 +1,177 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/security"
+Description: Measure user acquisition time on UserCache miss.
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 10
+      socketTimeoutMS: 5000
+      connectTimeoutMS: 5000
+
+Actors:
+- Name: UserAcquisitionActors
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Phase: 0 # Create a user we can pull from the privilege database
+    Repeat: 1
+    Database: admin
+    Operations:
+      # Setup a tree of roles and privileges
+      # The specific roles and privileges are unimportant
+      # so long as they are complex, but non-cyclic.
+      # Note that the yaml parser has some issues
+      # with empty arrays, but createRole must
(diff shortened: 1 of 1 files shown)


sgolemon:
Sorry, should have mentioned that I already ran this a few times. Totes McGotes green and has already yielded useful information for our next stage plans.

sgolemon:
@dalyd 

1. Sys-perf versus Performance. I don't actually know how to answer that question. The thing I'm measuring is very narrow. It doesn't need to go off-node, and the only storage necessary is the privilege database (which could be in-memory).  I think that means my answer is the latter.  What does that mean for this commit? Is there some other way I should present the workload?

2. Possibly some tweaks, and maybe a deeper role graph (I'm thinking of writing a script to generate the yaml per your note #3). The core of the test is what you see though: Invalidate user cache, acquire user, repeat.

dalyd:
> @dalyd
> 
> 1. Sys-perf versus Performance. I don't actually know how to answer that question. The thing I'm measuring is very narrow. It doesn't need to go off-node, and the only storage necessary is the privilege database (which could be in-memory).  I think that means my answer is the latter.  What does that mean for this commit? Is there some other way I should present the workload?

I'm going to lean on @rtimmons here. I know we have genny workloads running in performance, but I'm not sure precisely what's needed to make that happen. I don't think there's an auto run there, so it might require a SERVER ticket to edit perf.yml. For this code review it would mean at least removing/changing the autorun section at the bottom.

rtimmons:
>  we have genny workloads running in performance, but I'm not sure precisely what's needed to make that happen

Correct there is no autorun (yet!), so it would be a change to perf.yml. Unless you also wanted to run it on sys-perf you would remove the `AutoRun` section.

It should be as straightforward as creating a task like [the `InsertRemove` task](https://github.com/mongodb/mongo/blob/master/etc/perf.yml#L274-L282). And then adding it to the proper variants.

sgolemon:
ping @rtimmons See @dalyd's comment above.  Do we need to change this PR (and possibly add a small SERVER diff)?

sgolemon:
evergreen merge

PR #364, opened 2020-07-15
TIG-2564 Add findOne support to CrudActor

dalyd:
Added findOne support to the crud actor while working on porting over mixed workloads. I'll put that up for a separate review.

### DIFF
--- a/src/cast_core/test/CrudActorYamlTests.yml
+++ b/src/cast_core/test/CrudActorYamlTests.yml
@@ -443,7 +443,7 @@ Tests:
         OperationCommand:
           Filter: {a: 100}
 
-- Description: simple findOneAndUpdate
+  - Description: simple findOneAndUpdate
     Operations:
       - OperationName: insertOne
         OperationCommand:
--- a/src/cast_core/test/CrudActorYamlTests.yml
+++ b/src/cast_core/test/CrudActorYamlTests.yml
@@ -442,6 +442,8 @@ Tests:
       - OperationName: findOne
         OperationCommand:
           Filter: {a: 100}
+    OutcomeData:
+      - {a: 2}
 
   - Description: simple findOneAndUpdate
     Operations:
--- a/src/cast_core/test/CrudActorYamlTests.yml
+++ b/src/cast_core/test/CrudActorYamlTests.yml
@@ -434,15 +434,6 @@ Tests:
       - Filter: {a: 1}
         Count: 0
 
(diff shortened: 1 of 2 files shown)


dalyd:
I added a simple test, but now we're getting task failures and I don't know why. @rtimmons do you see what's happening?

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -434,7 +434,16 @@ Tests:
...
+  - Description: simple findOne
+    Operations:
+      - OperationName: insertOne
+        OperationCommand:
+          Document: {a: 100}
+      - OperationName: findOne
+        OperationCommand:
+          Filter: {a: 100}
+
+- Description: simple findOneAndUpdate
```
rtimmons, line 446: ugh lololol yaml - the tests are failing because this line (446) needs to be indented a couple spaces

dalyd: Thanks. I pushed a fix. My yamllint is all very unhappy with that file anyways, so I didn't see the warning on it. Have you considered a yaml lint task?

dalyd: Hm, still failing. I tried a couple of things also. Could you take a look again @rtimmons?

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -434,6 +434,17 @@ Tests:
...
+  - Description: simple findOne
+    Operations:
+      - OperationName: insertOne
+        OperationCommand:
+          Document: {a: 100}
+      - OperationName: findOne
+        OperationCommand:
+          Filter: {a: 100}
+    OutcomeData:
+      - {a: 2}
```
rtimmons, line 446: It's asserting there is a `{a: 2}` document. You can just remove the `OutcomeData` here and it should work. There's not an easy way to assert the results of the `findOne` because the Actor doesn't expose it to the outside world.

dalyd: Yeah, I originally didn't have this and this was a shot in the dark to fix the failure. It seems to have been no difference. I'll back it back out, but it won't change the result. See results here: https://github.com/mongodb/genny/pull/364/commits/089495cc4aa1c2d69c960925365fee39a45ea077

dalyd:
evergreen merge

PR #365, opened 2020-07-17
TIG-2564 Convert mixed_workloads to Genny

dalyd:
This is a port of an existing workload. Putting it up in part because I'm getting different results than the existing workload and would like some eyes on it. 

Patch build: https://spruce.mongodb.com/version/5f2978ee0ae6060c2dd1425d/tasks

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1249,10 +1249,10 @@ struct CrudActor::PhaseConfig {
 
 void CrudActor::run() {
     for (auto&& config : _loop) {
+        auto session = _client->start_session();
         for (const auto&& _ : config) {
             auto metricsContext = config->metrics.start();
 
-            auto session = _client->start_session();
             for (auto&& op : config->operations) {
                 op->run(session);
             }
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1264,7 +1264,8 @@ void CrudActor::run() {
 
 CrudActor::CrudActor(genny::ActorContext& context)
     : Actor(context),
-      _client{std::move(context.client())},
+      _client{std::move(
+          context.client(context.get("ClientName").maybe<std::string>().value_or("Default")))},
       _loop{context, _client, CrudActor::id()} {}
 
 namespace {
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -810,6 +810,40 @@ private:
     metrics::Operation _operation;
 };
 
+struct FindOneOperation : public BaseOperation {
+    FindOneOperation(const Node& opNode,
+                     bool onSession,
+                     mongocxx::collection collection,
+                     metrics::Operation operation,
+                     PhaseContext& context,
+                     ActorId id)
+        : BaseOperation(context, opNode),
+          _onSession{onSession},
+          _collection{std::move(collection)},
+          _operation{operation},
+          _filter{opNode["Filter"].to<DocumentGenerator>(context, id)} {}
+
+    void run(mongocxx::client_session& session) override {
+        auto filter = _filter();
+        this->doBlock(_operation, [&](metrics::OperationContext& ctx) {
+            auto result = (_onSession) ? _collection.find_one(session, filter.view(), _options)
+                                       : _collection.find_one(filter.view(), _options);
+            if (result) {
+                ctx.addDocuments(1);
+                ctx.addBytes(result->view().length());
+            }
+            return std::make_optional(std::move(filter));
+        });
+    }
+
+private:
+    bool _onSession;
+    mongocxx::collection _collection;
+    mongocxx::options::find _options;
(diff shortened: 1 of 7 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -810,6 +810,40 @@ struct FindOperation : public BaseOperation {
     metrics::Operation _operation;
 };
 
+struct FindOneOperation : public BaseOperation {
```
dalyd, line 813: Changes in this file are part of a separate PR that hasn't been merged yet https://github.com/mongodb/genny/pull/364

src/cast_core/src/Loader.cpp
```diff
@@ -75,6 +75,8 @@ void genny::actor::Loader::run() {
                  i++) {
                 auto collectionName = "Collection" + std::to_string(i);
                 auto collection = config->database[collectionName];
+                // drop the collection if it already exists
```
dalyd, line 78: I'm dropping this change.

src/phases/scale/MixPhases.yml
```diff
@@ -0,0 +1,49 @@
+PhaseSchemaVersion: 2018-07-01
+
+dbname: &dbname mix
+runtime: &runtime 7 minutes
+DocumentCount: &NumDocs 100000
+Filter: &filter {id: {^RandomInt: {min: 0, max: *NumDocs}}}
+string: &string {^FastRandomString: {length: 50}}
```
dalyd, line 7: This may be overkill, but the quality of random strings isn't important for this workload (I don't think), and there are a lot of them.

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -0,0 +1,354 @@
...
+  - *nop
+
+- Name: Update_1
+  Type: CrudActor
+  Database: *dbname
+  Threads: 1
+  Phases:
+  - Phase: 0..1
+    Nop: true
+  - Threads: 1
```
dalyd, line 79: Do I need this threads entry? I don't remember anymore. I have it set on the actor.

rtimmons: I don't believe this is necessary. Your call if you want to do another patch-test with it removed - probably worth it given how many times the per-phase Threads is used in this doc

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -0,0 +1,354 @@
...
+  Threads: 64
+  Phases:
+  - Phase: 0..7
+    Nop: true
+  - Threads: 64
+    ExternalPhaseConfig:
+      Path: *PhasePath
+      Key: UpdatePhase
+  - Phase: 9..10
+    Nop: true
```
rtimmons, line 127: something to think about for syntax tweaks is to allow for explicit `NopOtherPhases` so this could become:

```yaml
- Name: Update_64
  Type: CrudActor
  Database: *dbname
  Threads: 64
  NopOtherPhases: true
  Phases:
  - Phase: 8
    ExternalPhaseConfig:
      Path: *PhasePath
      Key: UpdatePhase
```

since it seems like there was probably some anguish in lining up all the phase numbers here. That would be an incremental change, but repeating phases and increasing the threads between them is more useful.

src/workloads/scale/MixedWorkloadsGenny.yml
dalyd: I'll take them out but verify it still works correctly. They may have been added in out of superstition. I don't recall.

src/workloads/scale/MixedWorkloadsGenny.yml
dalyd: Oh that would be lovely. It would make the workload easier to read, and more robust. I definitely had cases where things barfed after a while because I left off a nop phase on an actor. It found that late.

dalyd:
Updated using new connection pool feature Ryan added. This seems to saturate. 
Sys-perf patch build: https://spruce.mongodb.com/version/5f2978ee0ae6060c2dd1425d/tasks

dalyd:
One last commenting out of 128 thread level as result processing is taking too long. I think Im going to merge fter this one regardless.

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -0,0 +1,365 @@
...
+
+
+AutoRun:
+  Requires:
+    mongodb_setup:
+    - replica
+    - single-replica
+    - standalone
+    - replica-noflowcontrol
+    - replica-1dayhistory-15gbwtcache
```
guoyr, line 364: One more comment :) Could you run this on the new maintenance build variants?

guoyr: `linux-3-node-replSet-maintenance-events`

dalyd: Yeah, I was thinking about that. I'll add.

dalyd:
evergreen merge

PR #366, opened 2020-07-21
PERF-2020 Run CreateIndex and CreateBigIndex on more variants

dalyd:
I found I wish I had results on 3 node replica set for this. This PR is to add it. @louiswilliams any reason not to do it? I know we didn't originally include it.


dalyd:
evergreen merge

PR #369, opened 2020-08-03
TIG-2618 Allow CrudActor to specify which connection pool (Client) to use

rtimmons:
Does what it says on the tin.

### DIFF
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -61,6 +61,11 @@ public:
     explicit HasNode(const Node& node) : _node{node} {}
 
     template <typename... Args>
+    auto& get(Args&&... args) const {
+        return this->_node.operator[](std::forward<Args>(args)...);
+    }
+
+    template <typename... Args>
     auto& operator[](Args&&... args) const {
         return this->_node.operator[](std::forward<Args>(args)...);
     }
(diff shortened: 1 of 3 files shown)


src/gennylib/include/gennylib/context.hpp
```diff
@@ -60,6 +60,11 @@ class HasNode {
 public:
     explicit HasNode(const Node& node) : _node{node} {}
 
+    template <typename... Args>
```
dalyd, line 63: Just noting that this is just a magic incantation to me. Someone else will need to sanity check this part.

src/workloads/docs/CrudActor.yml
```diff
@@ -1,10 +1,19 @@
...
+      maxPoolSize: 10
+  SomeOtherPool:
+    QueryOptions:
+      maxPoolSize: 400
+
 Actors:
 - Name: CrudActor
   Type: CrudActor
   Database: mydb
+  ClientName: SomeOtherPool # Which connection pool to use. Must be defined in Clients section.
```
dalyd, line 16: Nice!

src/gennylib/include/gennylib/context.hpp
rtimmons: lol ya np - it's copy/paste from the lines below - saved me from having to write 

```c++
context.operator[]("ClientName")
```

😹

PR #370, opened 2020-08-13
SERVER-50133 Kill "Legacy" Auto-Tasks Logic

rtimmons:
Will depend on a couple other changes in DSI and the server


rtimmons:
evergreen merge

PR #373, opened 2020-08-25
PERF-1997 Introduce pipeline-based updates sys-perf workload

RuoxinXu:
Added a new workload that uses delta-style update oplog entries. There is also a 'phase' to run only classic updates that are equivalent to another phase which runs only pipeline-based updates. 

Sys-perf evergreen: https://spruce.mongodb.com/version/5f4516fce3c33163f25b4a75/tasks

### DIFF
--- a/src/workloads/execution/PipelineUpdate.yml
+++ b/src/workloads/execution/PipelineUpdate.yml
@@ -10,11 +10,11 @@ Actors:
     Database: &db test
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 10000
-    BatchSize: 2000
-    Document: &smallDocument    # Around 200B size.
-      t: &partitionIDSmall {^RandomInt: {min: 0, max: 1000}}
-      largeStr: &mediumString {^FastRandomString: {length: 128}}
+    DocumentCount: 5000
+    BatchSize: 1000
+    Document: &smallDocument    # Around 10KB size.
+      t: &partitionIDSmall {^RandomInt: {min: 0, max: 5000}}
+      largeStr: {^FastRandomString: {length: 10000}}
       int: &integer {^RandomInt: {min: 0, max: 100}}
       str: &string {^RandomString: {length: {^RandomInt: {min: 10, max: 20}}}}
       obj: &object {a: *integer, b: *integer, c: *string}
@@ -29,28 +29,13 @@ Actors:
     Database: *db
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 10000
-    BatchSize: 2000
-    Document: &mediumDocument   # Around 1KB size.
-      t: &partitionIDMedium {^RandomInt: {min: 1001, max: 2000}}
-      largeStr: &largeString {^FastRandomString: {length: 1024}}
+    DocumentCount: 5000
+    BatchSize: 1000
+    Document: &largeDocument    # Around 500KB Size.
+      t: &partitionIDLarge {^RandomInt: {min: 5001, max: 10000}}
       int: *integer
       str: *string
-      obj: *object
-      nestedObj: *nestedObject
-    Indexes:
-    - keys: {t: 1}
-  - Repeat: 1
-    Database: *db
-    Threads: 1
-    CollectionCount: 1
-    DocumentCount: 10000
-    BatchSize: 2000
-    Document: &largeDocument    # Around 100KB Size.
-      t: &partitionIDLarge {^RandomInt: {min: 2001, max: 3000}}
-      int: *integer
-      str: *string
-      largeStr: &giantString {^FastRandomString: {length: 100000}}
+      largeStr: {^FastRandomString: {length: 500000}}
       obj: *object
       nestedObj: *nestedObject
     Indexes:
@@ -59,17 +44,14 @@ Actors:
   - *Nop
   - *Nop
   - *Nop
-  - *Nop
-  - *Nop
 
 - Name: RunUpdates
   Type: RunCommand
-  Threads: 8
+  Threads: 16
   Phases:
   - *Nop
   - *Nop
-  - *Nop
-  - Repeat: 100    # Classic updates on small documents phase.
+  - Repeat: 500    # Classic updates on small documents phase.
     Database: *db
     Operations:
     - OperationMetricsName: ClassicUpdateAddFieldsSmallDoc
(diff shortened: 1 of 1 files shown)


src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+      arr: &array [*integer, *string, *object, *largeString]
+      nestedObj: &nestedObject {
+        int: *integer,
+        str: *string,
+        obj: *object,
+        arr: [1, 2, 3],
+        nestedArr: [*integer, *string, *object, *array],
+        largeStr: *largeString}
+    Indexes:
+    - keys: {t: 1}
```
puppyofkosh, line 31: Good call to have this index. We don't want a lot of time in the workload being spent doing collection scans.

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
```
puppyofkosh, line 12: Would it be possible to run these tests on different size documents? In my ideal world there'd be one collection of small documents (100-200 bytes) a collection of "medium" size documents (1-2kb) and collection of semi-large documents (100kb+). (The max document size is 16MB but I don't think we need to test documents that big). We could just run the same updates/operations on all three collections.

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+  Phases:
+  - *Nop
+  - Repeat: 1000    # Classic updates phase.
+    Database: *db
+    Operations:
+    - OperationMetricsName: ClassicUpdateAddFields
+      OperationName: RunCommand
+      OperationCommand:
+        update: Collection0
+        updates: [{q: {t: *partitionID}, u:
```
puppyofkosh, line 47: If I understand correctly, this will change the value of the 'int' field and add a new 'string' field. Suggest forking this into two tests: one about changing an existing field and one about adding a new field. For the one with a new field suggest giving it a really obvious name like "newStringField" or similar.

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+                     {$set: {int: *integer, string: *string}}, multi: true}]
+        writeConcern: {w: majority}
+    - OperationMetricsName: ClassicUpdateModifyArray
+      OperationName: RunCommand
+      OperationCommand:
```
puppyofkosh, line 52: Instead of re-writing the whole array I think we should just change one element or add a new element.

EDIT:
Thinking about this some more, having a perf test where we change one element may not be worth it. For context, the classic update system has an easy way to change individual array elements: {$set: {"someArray.0": "newVal"}}.

But this does not work in the pipeline update system! In aggregation pipelines field names like "0" "1" "2" are always interpreted as keys into objects rather than array indexes. In fact, there really a succinct way of changing an arbitrary array element in aggregation.

So I think we should instead have one case where we append a new element to an array and another case where we truncate the array. These should both be do-able. For appending, you can use $push in the classic update system and $concatArrays in the pipeline update system. For truncation, you can use $pop in the classic update system and $slice in pipeline updates.

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+        update: Collection0
+        updates: [{q: {t: *partitionID}, u:
+                     {$set: {nestedObj.arr: [*integer, *integer, *integer]}}, multi: true}]
+        writeConcern: {w: majority}
+    - OperationMetricsName: ClassicUpdateUnset
```
puppyofkosh, line 57: I know this wasn't mentioned in the original doc, but I think it'd be cool to test the following additional cases:

1) Deleting a single top level field
2) Deleting multiple top levels fields (two or three)
3) Deleting a nested field (like nestedObj.str)

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+        writeConcern: {w: majority}
+  - Repeat: 1000    # Pipeline-based updates phase.
+    Database: *db
+    Operations:
+    - OperationMetricsName: PipelineUpdateAddFields
+      OperationName: RunCommand
+      OperationCommand:
+        update: Collection0
+        updates: [{q: {t: *partitionID}, u: [
+          {$set: {int: *integer, string: *string}}], multi: true}]
```
puppyofkosh, line 77: Suggest using $addFields for this one :)

(also see comment above about maybe forking this into two tests)

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,102 @@
...
+      OperationCommand:
+        update: Collection0
+        updates: [{q: {t: *partitionID}, u: [
+          {$set: {nestedObj.arr: [*integer, *integer, *integer]}}], multi: true}]
+        writeConcern: {w: majority}
+    - OperationMetricsName: PipelineUpdateUnset
+      OperationName: RunCommand
+      OperationCommand:
+        update: Collection0
+        updates: [{q: {t: *partitionID}, u: [{$unset: [string, int]}], multi: true}]
```
puppyofkosh, line 90: [question] are you sure this one works? I would have thought 'string' and 'int' would need quotes around them. A lot of this stuff is magic to me though so I could be wrong.

src/workloads/execution/PipelineUpdate.yml
RuoxinXu: Hmm  Yea we can do updates on specific size of documents. but it seems difficult for this "Loader" actor to load three collections and each collection contains only documents of same size (There is a 'CrudActor' that is able to run insertions with particular documents on a named collection, not sure if the actor can be used as a loader to load docs though.)
How about we have three different sizes of documents (small, medium, and semi-large) in **one** collection, and do different types of updates (something like "UpdateSmallSizeDoc", "UpdateMediumSizeDoc")?

src/workloads/execution/PipelineUpdate.yml
RuoxinXu: yea it will quote "string" and "int" when generating commands.

You can also check the log on [Evergreen](https://evergreen.mongodb.com/task_log_raw/sys_perf_linux_3_node_replSet_pipeline_update_patch_4a95b7ebc88499c2b3b9aaa11fabbed895f40f18_5f4516fce3c33163f25b4a75_20_08_25_13_51_29/2?type=T) that gives the generated commands: 

> [2020/08/25 20:51:12.208] 20:51:12Z>  [2020-08-25 20:51:12.207972] [0x00007f37061ee700] [info]     Running command: { "update" : "Collection0", "updates" : [ { "q" : { "t" : 64 }, "u" : [ { "$unset" : [ "string", "int" ] } ], "multi" : true } ], "writeConcern" : { "w" : "majority" } } on database: test

puppyofkosh: Ah, nice. Thanks for looking into it.

src/workloads/execution/PipelineUpdate.yml
puppyofkosh: Sounds good!

RuoxinXu:
Posted another patch for extending the coverage to different size of documents as Ian suggested (100-200 bytes, 1Kb, and 100Kb). Also add a few update operations. Sadly, I did not find a good way to dedup the code. We could save the ops on "Small Doc"(200bytes doc) since the metrics are pretty close between "Small Doc" and "Medium Doc" judge from the result: 

https://evergreen.mongodb.com/task/sys_perf_linux_3_node_replSet_pipeline_update_patch_a28e071d22e8056781b03b1243c3961b054801b6_5f438eaee3c33155daf59173_20_08_24_10_02_39##comparehashes=a28e071d22e8056781b03b1243c3961b054801b6&threads=all

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -0,0 +1,352 @@
...
+  - &Nop {Nop: true}
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: RunUpdates
+  Type: RunCommand
+  Threads: 8
```
RuoxinXu, line 67: Only tried 4 and 8 threads, haven't well tuned this workload yet. Will do it once the workload is set.

puppyofkosh: Sounds good.

RuoxinXu:
Updates now are run on two different size documents, 10K and 500K, and are run for 500 times for stability. The workload seems to be not so stable when evergreen is busy. I've run many builds for the workload during the noise testing.

The metrics of the 5 patch builds for workload noise testing, https://docs.google.com/spreadsheets/d/1BllUONrbwJMuvqzKN-EBrAqAAeWcmdlpYvL0PaiuzuM/edit?usp=sharing .  Updates on 500K docs seem not as stable as updates on 10K. 
Basically the required metric((max-min)/median) is 4-5% for updates on SmallDoc, 8-9% for updates on LargeDoc.

puppyofkosh:
LGTM! It looks like we're within the 10% window, so I'd say you're good to merge this. We were also planning on running this workload against 4.4. Can you find out what it will take to make that happen? I imagine we'll have to edit some yml file somewhere.

RuoxinXu:
Thank you all for the review!
Pushing this.

RuoxinXu:
evergreen merge

PR #374, opened 2020-09-09
TIG-2653 Fix Bottleneck in Genny FTDC metrics collection

aggrand:
Patch running various genny tasks: https://evergreen.mongodb.com/version/5f59243b2fbabe37272f28c1

Patch running insert_remove and mixed_workloads_genny with ftdc output: https://evergreen.mongodb.com/version/5f59259257e85a19caf397f1


src/metrics/include/metrics/v2/event.hpp
```diff
@@ -42,11 +49,22 @@ class OperationEventT;
 
 namespace internals::v2 {
 
+const int NUM_CHANNELS = 4;
+const int BUFFER_SIZE = 1000;
+const int CLIENT_THREADS = 3;
+const int GRPC_THREAD_SLEEP_MS = 50;
+const double SWAP_BUFFER_PERCENT = .25;
```
aggrand, line 56: There's not a super motivated reason for these values other than that I ran a lot of patches and these seemed to get the best performance. I wasn't sure if we'd want to make these configurable in the workload yaml; e.g. insert_remove does a bit better with fewer client threads but mixed workloads needs them.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,267 @@ class Collector {
...
+
+        // Maybe a bit nuclear, but this draws a box around the entire grpc system
+        // and errors if it ever backs up enough to slow down an actor thread.
+        if (_draining->size() > size) {
+            std::ostringstream os;
+            os << "Metrics buffer for operation name " << name << " exceeded pre-allocated space"
+               << ". Expected size: " << size << ". Actual size: " << _draining->size()
+               << ". This may affect recorded performance.";
+
+            BOOST_THROW_EXCEPTION(MetricsError(os.str()));
```
aggrand, line 429: The bad thing about this is that the constants (at the top of the file) that get best performance don't work locally, which might mess up local development. Similarly, it fails in the canaries. Would it be worthwhile having a CLI option to turn this off?

aggrand (commented):
Had a couple questions.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -17,7 +17,14 @@
 #define HEADER_960919A5_5455_4DD2_BC68_EFBAEB228BB0_INCLUDED
```
aggrand, line 17: Now that I've made these changes, I'm pretty unsatisfied with the architecture of this whole file. In particular, I think the Operations shouldn't interact with EventStream, which behaves more similarly to the ClientContext in the grpc qps benchmark drivers. Instead, I'd like to evolve GrpcClient towards being a complete interface to the system, owning the channels and streams. Right now it feels a little shoehorned in. In that case we'd be able to actually use the EventStreams (or ClientContext) as tags in the completion queue, so the main goal of client threads would be to reap the cq(s) and take actions based on the tag returned.

That said, I think reworking that would make sense as part of the change to finally deprecate CSV metrics and streamline the whole system through grpc.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
+    std::vector<GrpcThread<ClockSource, StreamInterface>> _threads;
+    std::atomic<size_t> latest_thread = 0;
+};
+
+template <typename ClockSource>
+struct MetricsArgs {
+    MetricsArgs(const typename ClockSource::time_point& finish,
+                OperationEventT<ClockSource> event,
+                size_t workerCount)
+        : finish(finish), event(std::move(event)), workerCount(workerCount) {}
```
rtimmons, line 370: brace-initialize

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
+
+// Efficient buffer that should rarely be blocked on insertions.
+template <typename ClockSource>
+class MetricsBuffer {
+public:
+    explicit MetricsBuffer(size_t size, const std::string& name)
+        : size{size},
+          name{name},
+          _loading(new std::vector<MetricsArgs<ClockSource>>()),
+          _draining(new std::vector<MetricsArgs<ClockSource>>()) {
```
rtimmons, line 384: I think `make_unique` is generally less surprising than `new` inside paren initializers

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
+            os << "Metrics buffer for operation name " << name << " exceeded pre-allocated space"
+               << ". Expected size: " << size << ". Actual size: " << _draining->size()
+               << ". This may affect recorded performance.";
+
+            BOOST_THROW_EXCEPTION(MetricsError(os.str()));
+        }
+    }
+
+    std::unique_ptr<std::vector<MetricsArgs<ClockSource>>> _loading;
+    std::unique_ptr<std::vector<MetricsArgs<ClockSource>>> _draining;
```
rtimmons, line 436: I wonder if instead of `vector<MetricsArgs>` we should instead have `std::array<std::optional<MetricsArgs>,BUFFER_SIZE>`. This enforces that we can't over-extend and saves std::vector from resizing on us. At the very least, let's `reserve` or whatever it is the buffer size so we're not waiting on `std::vector` to `memcpy` when it grows.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
+    std::mutex _loading_mutex;
+};
 /**
  * Primary point of interaction between v2 poplar internals and the metrics system.
  */
 template <typename ClockSource, typename StreamInterface>
 class EventStream {
     using duration = typename ClockSource::duration;
     using OptionalPhaseNumber = std::optional<genny::PhaseNumber>;
+    using ClientPtr = std::shared_ptr<GrpcClient<ClockSource, StreamInterface>>;
```
rtimmons, line 446: The use of `shared_ptr` is really the only thing giving me pause about giving my LGTM here. There are valid uses for it, but I don't quite see it. If it is indeed the right decision, please add a bit of commentary on the justification.

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
 
 public:
     explicit EventStream(const ActorId& actorId,
                          const std::string& name,
                          const OptionalPhaseNumber& phase,
-                         const boost::filesystem::path& pathPrefix)
-        : _name{name}, _stream{name, actorId}, _phase{phase}, _last_finish{ClockSource::now()} {
+                         const boost::filesystem::path& pathPrefix,
+                         ClientPtr grpcClient = ClientPtr(nullptr))
```
rtimmons, line 453: shouldn't this always be provided?
Another case to consider a ref as well.

rtimmons (changes requested):
Logic looks good. Nice work.

TBH I got a bit lost in some of the ownership details but that's life with C++. Biggest concerns are `shared_ptr` and a few missing `lock_guard`s. Rest are relatively minor/stylistic. I think the ownership model would be clearer without ptrs wherever possible and instead refs. Take advantage of the fact that `RegistryT` is basically singleton and outlives any other objects.

src/metrics/include/metrics/v2/event.hpp
aggrand: Ended up making it a ref to a unique_ptr.

src/metrics/include/metrics/v2/event.hpp
aggrand: Okay didn't add a CLI option but I at least added a registry parameter to disable it for the canaries. Worst case scenario, the canaries perform a bit worse than well-tuned parameters in IRL situations.

src/metrics/include/metrics/operation.hpp
```diff
@@ -184,13 +188,16 @@ class OperationImpl final : private boost::noncopyable {
     using OptionalOperationThreshold = std::optional<OperationThreshold>;
     using OptionalPhaseNumber = std::optional<genny::PhaseNumber>;
     using stream_t = internals::v2::EventStream<ClockSource, v2::StreamInterfaceImpl>;
+    using grpcClientPtrRef =
+        std::unique_ptr<internals::v2::GrpcClient<ClockSource, v2::StreamInterfaceImpl>>&;
```
rtimmons, line 192: Consider name `UniqueGrpcClient` and let the ref be an impl detail

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +292,274 @@ class Collector {
...
+                    stillReaping = true;
+            }
+        } while (stillReaping);
+    }
+
+    std::atomic<bool> destructing = false;
+    bool _assertMetricsBuffer;
+    std::thread _thread;
+    std::set<StreamPtr> streams;
+    std::mutex _streamsMutex;
```
rtimmons, line 346: here and elsewhere: leading underscore on all these for consistency

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +285,269 @@ class Collector {
...
+    typedef EventStream<ClockSource, StreamInterface>* stream_ptr;
+
+    void registerStream(stream_ptr stream) {
+        _threads[latest_thread++ % _threads.size()].registerStream(stream);
+    }
+
+    void closeStream(stream_ptr stream) {
+        for (int i = 0; i < _threads.size(); i++) {
+            _threads[i].closeStream(stream);
+        }
```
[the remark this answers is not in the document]
rtimmons: that's odd - `for (const auto&& thread : _threads)` uses "universal references" so should never not work. If that doesn't work can you add a comment that we'd like to use it but couldn't here for whatever reason?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +292,274 @@ class Collector {
...
+    std::atomic<size_t> latest_thread = 0;
+};
+
+template <typename ClockSource>
+struct MetricsArgs {
+    MetricsArgs(const typename ClockSource::time_point& finish,
+                OperationEventT<ClockSource> event,
+                size_t workerCount)
+        : finish{finish}, event{std::move(event)}, workerCount{workerCount} {}
+    typename ClockSource::time_point finish;
```
rtimmons, line 385: can we use `using` here for consistency?

src/metrics/include/metrics/v2/event.hpp
```diff
@@ -247,70 +292,274 @@ class Collector {
...
+        }
+    }
+
+    const std::string name;
+    const size_t size;
+
+private:
+    void refresh(bool force, bool assertMetricsBuffer) {
+        if (_draining->empty()) {
+            const std::lock_guard<std::mutex> lock(_loadingMutex);
```
rtimmons, line 429: super nit-picky (too much coffee this am apparently):
I prefer to have early-return rather than the full impl inside the `if` block 

```
if (!draining) {
  return;
}
...rest of fn...
```

rtimmons (approved):
This turned out quite nice. Ownership model is clean and easy to reason about its correctness. All my comments in this review are syntactic/naming in nature; please feel free to implement or disregard at your discretion.

aggrand:
@rtimmons I know you already approved the PR but wanted to see if you had any opinions on the updates I made client-side after the poplar changes and after trying with the full-speed mixed workloads. Things changed:

- revendored curator
- new protobufs and re-compiled
- changed the stream ownership model, they as well as collectors are now owned by the GrpcClient, with operations just getting pointers to them (this was necessary so that each stream can be registered with the collector, but it streamlined things as well).
- each stream gets a separate thread; this ended up being necessary to avoid blocking server-side, but they don't have noticeable overhead. Most of the time the threads are sleeping, it's just that when a send does happen they benefit from the concurrency. I think with the server-side buffering now it might be less necessary, but I'd prefer to keep it for now just in case.
- finish streams concurrently instead of serially when destructing
- using a cv-based thread sleep so that threads can be woken up when finishing
- threads drain their streams in chunks, yielding their time-slice after a chunk to keep the server-side buffers balanced
- increased the client-side buffering a lot, to the point where for shorter workloads like InsertRemove all the metrics results end up stored in memory like with CSV metrics and sent afterwards. This is mostly to avoid locking/unlocking for super fast workloads to swap buffers, but for the full-speed mixed workloads for full duration, it turns out that even with all these optimizations we need to store the whole result in memory or else we get a performance hit (it actually exceeds the intended SLA of 100k ops/sec!) However, draining the buffer to grpc afterwards is waaay faster than the current CSV postprocessing, so that should be fine. We just need to keep an eye out for a possible performance hit if we tried to extend mixed workloads to like 20 mins per phase or something to where the buffer starts getting drained (which I don't think we will). Other workloads seem to be fine when run longer.

rtimmons (approved):
Still LGTM. Great example of ownership here.

Plz put a quick patch out to DSI to use the same curator version. I think fast-follow for this is to merge in the change to convert all workloads to FTDC and to then cut a ticket to kill the CSV logic which I'm sure will be quite cathartic.

aggrand:
Failed on mac because curator's page hadn't built it yet, should be good now. Going to squash.

PR #375, opened 2020-09-30
TIG-2696 All workloads report to FTDC

rtimmons:
Don't remove the other metrics-format code just yet. Need to let this "bake" for a while to ensure this doesn't break anyone's workflow.

PBs:
- [sys-perf](https://spruce.mongodb.com/version/5f74cde0d1fe0744082ce684/tasks)
- [microbenchmarks](https://spruce.mongodb.com/version/5f74ce701e2d1760756e80c7/tasks)


aggrand:
The change looks good. I hadn't realized how quickly we were moving forward with this (I guess I should've figured from the slack convo), but I don't think this can safely be done before TIG-2653 (Fix Bottleneck in Genny FTDC Metrics collection) is merged, otherwise we'll quite likely see throughput drops in at least `insert_remove` and `mixed_workloads_genny`, possibly others too

rtimmons:
Ah sounds right - I'll put this on hold until that is further along then. Plus gotta fix some broken tests here and it may be an unpleasant fix.

rtimmons:
evergreen merge

PR #377, opened 2020-10-15
TIG-2585 Add rate limit based on metrics feedback

aggrand:
Had a bit of difficulty managing the complexity of a rate being either a percent or the existing format, any thoughts on a way to improve that are welcome. I'd considered having the `RateSpec` be complicated enough to enclose both forms, but wasn't sure if we wanted those convention primitives to be that smart.

Patch running Insert-Remove for twice as long with a 50% rate limit: https://evergreen.mongodb.com/version/5f87a5152a60ed65c96234f6

### DIFF
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -162,7 +162,7 @@ public:
                 }
                 break;
             }
-            _rateLimiter->addIter();
+            _rateLimiter->notifyOfIteration();
         }
     }
 
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -24,6 +24,9 @@
 #include <type_traits>
 #include <unordered_map>
 #include <utility>
+#include <variant>
+#include <iostream>
+#include <iomanip>
 
 #include <boost/exception/exception.hpp>
 #include <boost/throw_exception.hpp>
@@ -117,11 +120,24 @@ public:
             throw InvalidConfigurationException(msg.str());
         }
 
-        const auto rateSpec = phaseContext["GlobalRate"].maybe<RateSpec>();
+        std::variant<std::monostate, RateSpec, PercentileRateSpec> rateSpec;
+        // First treat as a RateSpec, then try as a PercentileRateSpec.
+        try {
+            auto optionalSpec = phaseContext["GlobalRate"].maybe<RateSpec>();
+            if (optionalSpec) rateSpec = optionalSpec.value();
+        } catch (InvalidConfigurationException e) {
+            try {
+                auto optionalSpec = phaseContext["GlobalRate"].maybe<PercentileRateSpec>();
+                if (optionalSpec) rateSpec = optionalSpec.value();
+            } catch(InvalidConfigurationException e2) {
+                throw e;
+            }
+        }
+
         const auto rateLimiterName =
             phaseContext["RateLimiterName"].maybe<std::string>().value_or("defaultRateLimiter");
 
-        if (rateSpec) {
+        if (rateSpec.index() > 0) {
             std::ostringstream defaultRLName;
             defaultRLName << phaseContext.actor()["Name"] << phaseContext.getPhaseNumber();
             const auto rateLimiterName =
@@ -133,8 +149,10 @@ public:
                     "there's no guarantee the rate limited operation will run in the correct "
                     "phase");
             }
-            _rateLimiter =
-                phaseContext.workload().getRateLimiter(rateLimiterName, rateSpec.value());
+            if (auto pval = std::get_if<RateSpec>(&rateSpec)) {
+                _rateLimiter =
+                    phaseContext.workload().getRateLimiter(rateLimiterName, (*pval));
+            }
         }
     }
 
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -24,7 +24,6 @@
 #include <type_traits>
 #include <unordered_map>
 #include <utility>
-#include <variant>
 
 #include <boost/exception/exception.hpp>
 #include <boost/throw_exception.hpp>
@@ -71,7 +70,7 @@ public:
                      bool isNop,
                      TimeSpec sleepBefore,
                      TimeSpec sleepAfter,
-                     std::optional<BaseRateSpec> rateSpec)
+                     std::optional<RateSpec> rateSpec)
         : _minDuration{minDuration},
           // If it is a nop then should iterate 0 times.
           _minIterations{isNop ? IntegerSpec(0l) : minIterations},
@@ -103,7 +102,7 @@ public:
                            phaseContext.isNop(),
                            phaseContext["SleepBefore"].maybe<TimeSpec>().value_or(TimeSpec{}),
                            phaseContext["SleepAfter"].maybe<TimeSpec>().value_or(TimeSpec{}),
-                           phaseContext["GlobalRate"].maybe<BaseRateSpec>()) {
+                           phaseContext["GlobalRate"].maybe<RateSpec>()) {
         if (!phaseContext.isNop() && !phaseContext["Duration"] && !phaseContext["Repeat"] &&
             phaseContext["Blocking"].maybe<std::string>() != "None") {
             std::stringstream msg;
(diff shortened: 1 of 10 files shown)


src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -84,6 +87,30 @@ class BaseGlobalRateLimiter {
      * appropriate back-off strategy if this function returns false.
      */
     bool consumeIfWithinRate(const typename ClockT::time_point& now) {
+
+        // Logic for percentile rates.
+        bool curFullSpeed = _fullSpeed.load();
+        if (curFullSpeed) {
+            _burstCount++;
+            auto _nsSincePhase = ClockT::now().time_since_epoch().count() - _lastEmptiedTimeNS;
```
rtimmons, line 95: 1. don't need `_` for local vars, only instance vars--I was reading it as an instance var and totally confused mysefl until I saw the `auto` here.
2. consider name `nsSincePhaseStarted`

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -423,6 +441,56 @@ struct convert<genny::RateSpec> {
...
 };
 
+/**
+ * Convert between YAML and genny::PercentileRateSpec
+ *
+ * The YAML syntax accepts [genny::Integer]%
+ * The syntax is interpreted as a percentage of the max throughput.
+ */
+template <>
+struct convert<genny::PercentileRateSpec> {
```
rtimmons, line 451: Let's also update one of the `docs` workloads to use this new percentage syntax and explain how it works.

rtimmons (changes requested):
Overall this looks good.

I do think it's worth seeing about merging `PercentileRateSpec` into `RateSpec`, but really the only thing that I'm blocking on is a couple naming/💅 things.

Nicely done.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -148,27 +176,41 @@ class BaseGlobalRateLimiter {
...
 
     // Note that the rate limiter as-is doesn't use the burst size, but it is cleaner to
     // store the burst size and the rate together, since they're specified together in
     // the YAML as RateSpec.
-    const int64_t _burstSize;
-    const int64_t _rateNS;
+    int64_t _burstSize;
+    int64_t _rateNS;
+    const std::optional<int64_t> _percent;
+    std::atomic<bool> _fullSpeed;
```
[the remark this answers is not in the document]
rtimmons: also I think `burst`/`rate` can be `const` too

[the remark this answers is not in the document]
aggrand: `burst`/`rate` can't be atomic because they get modified after the full-speed run.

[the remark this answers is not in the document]
aggrand: `iters` is now `std::atomic_int64_t`

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -163,6 +182,7 @@ class IterationChecker final {
                 }
                 break;
             }
+            _rateLimiter->addIter();
```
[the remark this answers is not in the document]
aggrand: Done.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
aggrand: Good catch, done.

src/gennylib/src/context.cpp
```diff
@@ -119,6 +119,23 @@ v1::GlobalRateLimiter* WorkloadContext::getRateLimiter(const std::string& name,
     return rl;
 }
 
+v1::GlobalRateLimiter* WorkloadContext::getRateLimiter(const std::string& name,
+                                                       const PercentileRateSpec& spec) {
+    if (this->isDone()) {
+        BOOST_THROW_EXCEPTION(std::logic_error("Cannot create rate-limiters after setup"));
```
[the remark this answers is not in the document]
aggrand: Good idea, done.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -84,6 +87,30 @@ class BaseGlobalRateLimiter {
...
+
+            // 3 iterations or 1 minute, whichever is longer.
+            if (_iters >= _numUsers * 3 && _nsSincePhase >= _nsPerMinute) {
+                const auto success =
+                    _fullSpeed.compare_exchange_weak(curFullSpeed, false);
+                if (!success) {
+                    return true;
+                }
+                // Reconfigure as a "normal" rate limiter running for the first time.
+                _burstSize = _burstCount * _percent.value() / 100;
```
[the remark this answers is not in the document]
aggrand: I think that would be a bit weird; this is all doing integer division since `burstSize` is an int and to have a subzero fraction in the RateSpec we'd use a double and cast back to an int here, and I think that'd be a bit trickier to test.

src/gennylib/include/gennylib/v1/GlobalRateLimiter.hpp
```diff
@@ -84,6 +87,30 @@ class BaseGlobalRateLimiter {
      * appropriate back-off strategy if this function returns false.
      */
     bool consumeIfWithinRate(const typename ClockT::time_point& now) {
+
+        // Logic for percentile rates.
+        bool curFullSpeed = _fullSpeed.load();
+        if (curFullSpeed) {
+            _burstCount++;
```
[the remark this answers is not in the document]
aggrand: That makes sense, done.

src/gennylib/include/gennylib/conventions.hpp
aggrand: Done, added a comment in `HelloWorld.yml` explaining.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -117,11 +118,26 @@ class IterationChecker final {
             throw InvalidConfigurationException(msg.str());
         }
 
-        const auto rateSpec = phaseContext["GlobalRate"].maybe<RateSpec>();
+        std::variant<std::monostate, RateSpec, PercentileRateSpec> rateSpec;
```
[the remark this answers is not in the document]
aggrand: Following up on our offline conversation, I'd started to move more of the logic for running in multiple phases into `RateSpec`, but in practice it ended up needing a lot of actual rate limiting logic to do it (e.g. needing to know `_burstCount` and `_iters`, which started to basically turn it into `GlobalRateLimiter`. 

I'd settled for wrapping a private `std::variant` but then having accessors for its components that are `std::optional` to soften the experience of using it. And at least nothing outside of `GlobalRateLimiter` and `conventions.hpp` needs to use those.

That said, please push back if you think trying a different integration would be worthwhile.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -115,24 +115,80 @@ inline bool operator==(const TimeSpec& lhs, const TimeSpec& rhs) {
...
+        if (auto pval = std::get_if<BaseRateSpec>(&_spec)) {
+            return *pval;
+        } else {
+            return std::nullopt;
+        }
+    }
+
+    std::optional<PercentileRateSpec> getPercentileSpec() const {
+        if (auto pval = std::get_if<PercentileRateSpec>(&_spec)) {
+                return *pval;
```
rtimmons, line 176: 💅 ensure you run clang-format prior to merge

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -115,24 +115,80 @@ inline bool operator==(const TimeSpec& lhs, const TimeSpec& rhs) {
...
+        } else {
+            return std::nullopt;
+        }
+    }
+
+    inline bool operator==(const RateSpec& rhs) {
```
rtimmons, line 182: everything in a header is implicitly inline

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -417,12 +473,116 @@ struct convert<genny::RateSpec> {
...
+            msg << "Invalid value for PercentileRateSpec field, expected an integer followed by %."
+                   " Saw: "
+                << strRepr;
+            throw genny::InvalidConfigurationException(msg.str());
+        }
+
+        auto percentYaml = Load(strRepr.substr(0, percentPos));
+        auto percent = percentYaml.as<genny::IntegerSpec>();
+
+        if (percent.value > 100) {
```
rtimmons, line 517: or if <= 0.

src/gennylib/include/gennylib/conventions.hpp
```diff
@@ -417,12 +473,116 @@ struct convert<genny::RateSpec> {
...
+ * Convert between YAML and genny::RateSpec
+ *
+ * The YAML syntax accepts either [genny::Integer] per [genny::Time]
+ * or [genny::Integer]%
+ *
+ * The syntax is interpreted as operations per unit of time or
+ * percentage of max throughput.
+ */
+template <>
+struct convert<genny::RateSpec> {
```
rtimmons, line 541: 💅 can you put the `convert<>`s after the classes? It's technically possible to not have the `convert` structs if there is a ctor that takes in a `genny::Node`. No need to do that now, but putting convert next to impl will make that change easier

rtimmons (approved):
LGTM. Thanks for pushing on this. I wish we didn't have to rely on try/catch for logical flow-control, but that's what the primitives kinda force upon us. A few stylistic things please push back wherever you disagree.

src/gennylib/include/gennylib/conventions.hpp
aggrand: This one's actually kinda tricky; all the structs and classes are defined in the `genny` namespace while all the `convert<>`s are in the `YAML` namespace, so all the converts are together. I think it's a bit better to not have to open and close the namespaces for each one or to resolve the `genny` scope every time.

src/gennylib/include/gennylib/conventions.hpp
aggrand: Fixed.

src/gennylib/include/gennylib/conventions.hpp
aggrand: Done.

src/gennylib/include/gennylib/conventions.hpp
aggrand: Good catch, done.

aggrand:
evergreen merge

PR #378, opened 2020-10-19
TIG-2608 Add appName=Genny to Genny URI

MikhailShchatko:
Sys-perf patch build: https://spruce.mongodb.com/version/5f89ab98c9ec446bcf0ec04b/tasks

### DIFF
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -74,6 +74,8 @@ struct PoolFactory::Config {
         }
 
         // Hardcode Genny appName
+        // The appName is not strictly a query option and the queryOptions can be renamed if we
+        // intend to continue to use for generic key value pairs in the URI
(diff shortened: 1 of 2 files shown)


src/gennylib/src/PoolFactory.cpp
```diff
@@ -72,6 +72,9 @@ struct PoolFactory::Config {
             queryOptions[key] = value;
             i += matches.length();
         }
+
+        // Hardcode Genny appName
+        queryOptions["appName"] = "Genny";
```
guoyr, line 77: Please add a comment here saying the appName is not strictly a query option and the `queryOptions` can be renamed if we intend to continue to use for generic key value pairs in the URI.

guoyr (approved):
LGTM w/ comment!

MikhailShchatko:
evergreen merge

PR #379, opened 2020-10-23
TIG-2745 Update $unionWith workloads to not rely on explain

nzolnierzmdb:
Hey Katherine - do you mind taking a look at this PR? The idea is to move away from using explain but make sure that all returned documents can fit into a single batch, such that we can still "exhaust" the union pipelines but with the downside that we're incurring overhead for returning documents over the network. 

EVG run: https://evergreen.mongodb.com/version/5f92f1782a60ed74cffbb0ab

### DIFF
--- a/src/workloads/execution/UnionWith.yml
+++ b/src/workloads/execution/UnionWith.yml
@@ -10,12 +10,13 @@ Actors:
     Database: &db test
     Threads: 1
     CollectionCount: 20
-    DocumentCount: 10000
-    BatchSize: 1000
+    DocumentCount: 5000
+    # Choose a batchSize and document size such that all documents can fit in a single 16MB batch.
+    BatchSize: &batchSize 100000
     Document:
-      integer: &integer {^RandomInt: {min: 0, max: 10000}}
+      integer: &integer {^RandomInt: {min: 1, max: 10000}}
       double: &double {^RandomInt: {distribution: geometric, p: 0.1}}
-      string: &string {^RandomString: {length: {^RandomInt: {min: 1100, max: 2000}}}}
+      string: &string {^RandomString: {length: 16}}
       array:
       - *integer
       - *integer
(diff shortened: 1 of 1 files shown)


src/workloads/execution/UnionWith.yml
```diff
@@ -10,12 +10,13 @@ Actors:
     Database: &db test
     Threads: 1
     CollectionCount: 20
-    DocumentCount: 10000
-    BatchSize: 1000
+    DocumentCount: 5000
+    # Choose a batchSize and document size such that all documents can fit in a single 16MB batch.
+    BatchSize: &batchSize 100000
```
nzolnierzmdb, line 15: I've modified the documents to be roughly 100bytes, which with some quick math means that (20 collections x 5000 docs/collections * 100 doc size ~= 10MB). So I think we should be good to fit into a 16MB batch....does that make sense or did I forget to carry a 1 somewhere?

src/workloads/execution/UnionWith.yml
```diff
@@ -10,12 +10,13 @@ Actors:
...
     Document:
-      integer: &integer {^RandomInt: {min: 0, max: 10000}}
+      integer: &integer {^RandomInt: {min: 1, max: 10000}}
```
nzolnierzmdb, line 17: Oh yeah, I should mention that leaving a min: 0 here was causing the $divide in one of the workloads to fail. I'm not sure if this isn't caught by evergreen or if I just got unlucky.

src/workloads/execution/UnionWith.yml
kaywux: Makes sense to me! Out of curiosity how did you check the generated document size?

kaywux (approved):
This looks good to me! I'm curious to see the results of these changes and if there are significance performance drops now that we're actually sending documents over the wire.

src/workloads/execution/UnionWith.yml
nzolnierzmdb: I went through each of the fields and estimated their size :)

I likely overestimated but it should be fine for this purpose.

src/workloads/execution/UnionWith.yml
nzolnierzmdb: Thanks! Turns out the reason is that explain doesn't report any runtime errors in the command being explained, so we were silently ignoring this before.

nzolnierzmdb:
> This looks good to me! I'm curious to see the results of these changes and if there are significance performance drops now that we're actually sending documents over the wire.

Good question, the results look to be almost back to what we had before the regression. But this is deceiving, because I've substantially lowered the number of documents as well as the width of the documents. So that, in combination with moving away from explain, means that we really have to start a new baseline with this commit.

nzolnierzmdb:
evergreen merge

PR #382, opened 2020-11-12
TIG-2773 Add throw to prevent false-positive warning

rtimmons:
Older compilers don't see that the `case` statements are exhaustive.


rtimmons:
evergreen merge

PR #383, opened 2020-11-16
PERF-2061 Remove references of geoHaystackIndex from workloads testcases

mtandrei:
Evergreen: https://spruce.mongodb.com/version/5fb2daa50ae6060498b51321/tasks (the failures are not relevant; create_index_sharded failing on a standalone/repl set is expected).

This is my proposed fix for https://jira.mongodb.org/browse/PERF-2061. Another way to handle this would be to allow for geoHaystack indexes to only be created when running against version 4.4 or below, but a recent thread on slack seems to suggest that this isn't possible at the moment.


rtimmons (approved):

mtandrei:
Thanks for taking a look @rtimmons!

mtandrei:
evergreen merge

PR #384, opened 2020-11-19
TIG-2781 Reduce metrics buffer size

aggrand:
Previously the buffer would only start draining when it reached 25% capacity, requiring the buffer to be 4x larger than we need just to be safe. This PR greatly increases the grpc thread sleep and adds an actor-thread-side check of when the buffer is almost full, to then wake the grpc thread.

In most workloads the buffer drain is probably going to be determined by the actor thread check, but the grpc-side wakeup and check may still be used by workloads like `large_scale_model`.

[Patch](https://evergreen.mongodb.com/version/5fb5ae13a4cf4723df226a62) running `large_scale_model` and `mixed_workloads_genny` along with the poplar changes, showing similar `mixed_workloads_genny` throughput to a [patch](https://evergreen.mongodb.com/version/5fb54dd1a4cf4723df1e6330) with the current master (biggest difference I could find was 1%, which is probably noise).

When [poplar#61](https://github.com/evergreen-ci/poplar/pull/61) is merged and revendored into curator I'll update this with the revendored curator.

### DIFF
--- a/src/lamplib/tasks/download.py
+++ b/src/lamplib/tasks/download.py
@@ -220,7 +220,7 @@ class CuratorDownloader(Downloader):
 
     # Note that DSI also downloads Curator, the location is specified in defaults.yml.
     # Please try to keep the two versions consistent.
-    CURATOR_VERSION = "8b3ef73c55fdc45ef78383b0dcd4a84335b8eecb"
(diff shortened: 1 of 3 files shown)


rtimmons (approved):

aggrand:
evergreen merge

PR #387, opened 2020-11-25
WT-6702 Create modified LSWA model workload

tetsuo-cpp:
@sulabhM 
This is where I'm up to with WT-6702. There are a few customizations that are needed on top of simply pasting the actors and phases from `LargeScaleLongLived.yml`.

- Apply a fixed scan rate with the `SnapshotScanner5Gigabytes`. From what I can see in the code, we can't configure it to work like this so we'll need to change the underlying actor. I've added a parameter and left a comment where I think we need to do some work.
- Use a weighted distribution for the long-lived readers and writers. I think this is supported out of the box. Instead of using a default `RandomInt` which uses a uniform distribution, we can choose `binomial` which I think gives us the effect we're looking for.

### DIFF
--- a/src/phases/scale/LargeScalePhases.yml
+++ b/src/phases/scale/LargeScalePhases.yml
@@ -235,6 +235,25 @@ SnapshotScanner5GigabytesCmd:
   CollectionSortOrder: forward
   GlobalRate: 1 per 5 minutes
 
+FixedRateSnapshotScanner5GigabytesCmd:
+  Duration: *Duration
+  ScanSizeBytes: 5000000000
+  # tetsuo-cpp: We don't supply scan duration since documents and scan rate implies it.
+  #
+  # This is how it's described in the technical design but I'm thinking that scan rate
+  # should be bytes per second rather than documents per second.
+  #
+  # The reasoning is that we don't know the document size. So for example, we could scan
+  # 100 large documents and exceed the 5GB size meaning that the scan will complete within
+  # a second. Similarly, we could scan 100000 tiny documents and finish the phase without
+  # hitting 5GB in total.
+  Documents: 100000
+  ScanRate: 100 per 1 second
+  ScanType: snapshot
+  CollectionSkip: 300
+  CollectionSortOrder: forward
+  GlobalRate: 1 per 5 minutes
+
 SnapshotScannerAllCmd:
   Duration: *Duration
   Documents: 100000000000
--- a/src/phases/scale/LargeScalePhases.yml
+++ b/src/phases/scale/LargeScalePhases.yml
@@ -1,7 +1,12 @@
 PhaseSchemaVersion: 2018-07-01
 Owner: Storage Engines
 
-# This section contains shared definitions that are used in the workload.
+# This is the set of shared phases for the Large Scale Workload Automation project.
+#
+# Since the LSWA workloads contain common phases, they've been separated in this phase file and can
+# be included in each workload as needed via the `ExternalPhaseConfig` functionality.
+
+# This section contains definitions that are used in the following phases.
 GlobalDefaults:
   Random10KInt: &rand_10k_int {^RandomInt: {min: 0, max: 10000}}
   Random10KInt: &rand_1k_int {^RandomInt: {min: 0, max: 1000}}
--- a/src/phases/scale/LargeScalePhases.yml
+++ b/src/phases/scale/LargeScalePhases.yml
@@ -39,12 +39,12 @@ GlobalDefaults:
     - keys: {h: 1, i: 1}
 
   # Commonly used parameters.
-  DatabaseParam: &DatabaseParam {^Parameter: {Name: "Database", Default: db}}
-  CollectionCountParam: &CollectionCountParam {^Parameter: {Name: "CollectionCount", Default: 0}}
-  TrackProportionParam: &TrackProportionParam {^Parameter: {Name: "TrackProportion", Default: 0}}
-  DocumentCountParam: &DocumentCountParam {^Parameter: {Name: "DocumentCount", Default: 0}}
-  WritesParam: &WritesParam {^Parameter: {Name: "Writes", Default: 0}}
-  ReadsParam: &ReadsParam {^Parameter: {Name: "Reads", Default: 0}}
+  DatabaseParam: &DatabaseParam {^Parameter: {Name: "Database", Default: ""}}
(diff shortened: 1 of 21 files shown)


src/workloads/scale/ScanWithLongLived.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  LongLivedDB: &LongLivedDB longlived # name of the long lived database
+  LongLivedDocumentCount: &LongLivedDocumentCount 1000
+
+  ScannerHotDB: &ScannerHotDB rolling
+
+  LongLivedWrites: &LongLivedWrites 600 per 1 second # x5 threads
+  LongLivedReads: &LongLivedReads 20 per 1 second    # x10 threads
+
+  binomial_10k_int: &binomial_10k_int {^RandomInt: {distribution: binomial, t: 10000, p: 0.5}}
```
tetsuo-cpp, line 15: @ddanderson When you mentioned weighting ids for the long-lived readers/writers, is this what you had in mind?

src/phases/scale/LargeScalePhases.yml
```diff
@@ -235,6 +236,25 @@ SnapshotScanner5GigabytesCmd:
...
+  #
+  # This is how it's described in the technical design but I'm thinking that scan rate
+  # should be bytes per second rather than documents per second.
+  #
+  # The reasoning is that we don't know the document size. So for example, we could scan
+  # 100 large documents and exceed the 5GB size meaning that the scan will complete within
+  # a second. Similarly, we could scan 100000 tiny documents and finish the phase without
+  # hitting 5GB in total.
+  Documents: 100000
+  ScanRate: 100 per 1 second
```
tetsuo-cpp, line 252: My understanding is that this isn't supported out of the box so we'll have to tweak the underlying actor to get this to work.

tetsuo-cpp: @ddanderson 
From the technical design:

> Our modified scan will have a fixed scan rate and a fixed total number of operations to complete.  Thus, provided it can keep up, it will run for a fixed amount of time.  As a strawman example, let’s propose 100 reads per second, and 100,000 total reads.  So the scan will finish in 1000 seconds.

Am I right in thinking that the `CollectionScanner` actor can't be configured to do this and requires some changes to the underlying C++? As of now, I've scheduled some time to implement this functionality in WT-6966.

rtimmons:
Looks like this PR is still in "draft" state. STM is happy to review. Whenever this is ready, please move the PR to the Open state.

src/workloads/scale/ScanWithLongLived.yml
ddanderson: Yes, I think so.  I'm guessing this would produce a roughly 'bell shape curve' centered on 5000?  So the "hot spot" is in the middle of a key space of up to 10000.  I wonder if this (the "t" number) should match the `LongLivedDocumentCount` as well, so the key space gets entirely populated?

And it's quite possible we'll want to adjust both these numbers up - we want to get significant chunks of data to be old and potentially evicted.  Depends on the number of documents, size of documents and number of collections.

src/phases/scale/LargeScalePhases.yml
ddanderson: @tetsuo-cpp You're right.  There's a way to set the total number of bytes to scan, but I don't see a way to limit the rate.

tetsuo-cpp: Ok, we'll address this in WT-6966.

src/workloads/scale/ScanWithLongLived.yml
tetsuo-cpp: f87b253

I've adjusted the `t` value down to match the document count.

This PR is just aimed to get something merged that seems reasonable and captures the intent of the technical design. We have WT-6967 to analyze and tweak the values so the key distribution looks the way we expect.

tetsuo-cpp:
@sulabhM @ddanderson 
Ok, this is ready to look at.
The fixed-rate scanning is a decent chunk of work so I'll be following up with a separate PR. We also have follow up tickets to run the workload and observe stats to check that the workload is performing (ID distribution, scan rate, etc) as specified in the technical design.

The objective of this change is to get something in that has all the right components (aside from the fixed-rate scanning) and looks reasonable.

ddanderson (approved):
LGTM

src/workloads/scale/ScanWithLongLived.yml
```diff
@@ -0,0 +1,82 @@
...
+  LongLivedDB: &LongLivedDB longlived # name of the long lived database
+  LongLivedCollectionCount: &LongLivedCollectionCount 1000
+  LongLivedDocumentCount: &LongLivedDocumentCount 1000
+
+  ScannerHotDB: &ScannerHotDB rolling
+
+  LongLivedWrites: &LongLivedWrites 600 per 1 second # x5 threads
+  LongLivedReads: &LongLivedReads 20 per 1 second    # x10 threads
+
+  binomial_10k_int: &binomial_10k_int {^RandomInt: {distribution: binomial, t: 1000, p: 0.5}}
```
ddanderson, line 16: I wonder if `t: 1000` should be `t: *LongLivedDocumentCount`

tetsuo-cpp: These two values are related (e.g. it doesn't make sense to make the doc count much smaller than the key-range) but I don't think it's clear that they will remain equal to each other.

tetsuo-cpp:
@sulabhM Friendly ping. Can you please take a look when you get the chance?

sulabhM (approved):

tetsuo-cpp:
@mongodb/stm This is ready to look at.

tetsuo-cpp:
Friendly ping @mongodb/stm. Is this good to merge?

src/workloads/scale/ScanWithLongLived.yml
```diff
@@ -0,0 +1,82 @@
+SchemaVersion: 2018-07-01
+Owner: Storage Engines
+
```
rtimmons, line 3: Consider adding a `Description` here with a link to the ticket or the doc.

rtimmons (approved):
LGTM apologies for my delay. Please ensure this passes a patch-build but otherwise if youre' happy with it STM is happy with it.

src/workloads/scale/ScanWithLongLived.yml
tetsuo-cpp: baf9639

PR #388, opened 2020-11-25
WT-6790 Package and document eMRCf tests

keitharnoldsmith:
Adding a set of tests and scripts written to evaluate the impact
of growing the WiredTiger history store in a PSA replica set with
a failed secondary node.  The README file provides details.

### DIFF
--- a/src/workloads/contrib/historystore/eMRCf_runner.sh
+++ b/src/workloads/contrib/historystore/eMRCf_runner.sh
@@ -154,8 +154,11 @@ cd $metrics_dir
 # For each benchmark phase, we pull the relevant operations out of the CSV files and sort them
 # to generate a CDF.  
 #
-# Note that the csv files created by Genny capture per request latency in nanoseconds.
+# Note that the CSV files created by Genny capture per request latency in nanoseconds.
 # We divide by 1000 and plot them in microseconds.
+#
+# WARNING:  Future versions of Genny won't provide this CSV metrics output.  The following
+# code (and the -p option) will stop working at that point.
 
 echo "Preparing plots..."
 tmpfile=$(mktemp)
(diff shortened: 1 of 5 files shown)


src/workloads/contrib/historystore/eMRCf_runner.sh
```diff
@@ -0,0 +1,230 @@
...
+do
+    mkdir $RUN &> /dev/null
+    for PHASE in InsertPhase UpdateHeavyPhase ReadMostlyPhase DeletePhase
+    do
+        echo "    Processing $RUN/$PHASE..."
+        grep "^[0-9].*$PHASE.*One" Bench_$RUN/Bench_$RUN.csv > $tmpfile
+        count=$(wc -l $tmpfile | awk '{print $1}')
+        count=$(($count / 100))
+        cat $tmpfile | sed -e 's/,/ /g' | awk '{print $5}' | sort -n | \
+            awk "BEGIN {n = 1} {print \$1/1000, n/$count; n++}" > $RUN/$PHASE
```
rtimmons, line 172: I love me a good `awk` script. Unfortunately this analysis logic will be broken soon when we kill the CSV output format. Can you put a comment here or in the readme indicating that the plotting may be broken?

rtimmons (approved):
Thanks for contributing this. I hope it helps someone in the future.

src/workloads/contrib/historystore/eMRCf_runner.sh
keitharnoldsmith: Good point.  I'll add comments accordingly.  I assume this means that the `Format: csv-fctdc` metrics option will be going away as well?

rtimmons: Correct - there will no longer be an option to specify the output format, it will always be ftdc.

PR #390, opened 2020-12-07
PERF-2041: Add new value generators (choose, join, IP, double)

dalyd:


dalyd:
I'm sure there's better ways to do some of this, so please let me know. Also, if there's any clang like stuff that doesn't look right, let me know and I'll run it.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +569,133 @@ class GeometricInt64Generator : public Generator<int64_t> {
     const double _p;
 };
 
+// This generator allows choosing any valid generator, incuding documents. As such it cannot be used
```
dalyd, line 572: I'm open to ideas here. This allows be to choose anything, which is cool, but without a `std::string evaluate()` method, it can't be called within the JoinGenerator. I made a specific string only version of this that will work with join, but it would be nice to have it all work in one.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -392,3 +392,50 @@ Tests:
...
+    GivenTemplate:
+      a: {^Join: {array: ["Hello", "random", "person:", {^Choose: {from: ["Huey", "Duey",
+    {^FastRandomString: {length: 10}}]}}], sep: " "}}
+    ThenReturns:
+      - a: "Hello random person: Duey"
+      - a: "Hello random person: Duey"
+      - a: "Hello random person: smlCLkHjFD"
+      - a: "Hello random person: mnqojklKFs"
+
+  - Name: Double Generator
```
dalyd, line 427: I didn't test all the distributions. I'm not sure I want to, but I can if you would like.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +569,133 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+protected:
+    DefaultRandom& _rng;
+    std::vector<UniqueGenerator<std::string>> _choices;
+    std::vector<int64_t> _weights;
+};
+
+class IPGenerator : public Generator<std::string> {
+public:
+    IPGenerator(const Node& node, DefaultRandom& rng)
+        : _rng(rng), subnet_mask{std::numeric_limits<uint32_t>::max()}, prefix{} {}
```
dalyd, line 637: I ran out of steam here. This will handle sub-masking and prefixes, but just requires the parsing the arguments.

dalyd:
I started to templatize the double generators. Will do more, but wanted to get the idea up here.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -152,13 +154,308 @@ using Parser = std::function<O(const Node&, DefaultRandom&)>;
...
+                   DefaultRandom& rng,
+                   const std::map<std::string, Parser<Out>>& parsers);
 
 template <bool Verbatim>
 std::unique_ptr<DocumentGenerator::Impl> documentGenerator(const Node& node, DefaultRandom& rng);
 
 template <bool Verbatim>
 UniqueGenerator<bsoncxx::array::value> arrayGenerator(const Node& node, DefaultRandom& rng);
 
+template <typename distribution,
```
dalyd, line 171: I realized I could template all this stuff rather than having a full class for every double generator. If I could figure out how to pass in a function to the template, I would generalize to NumberGeneraotr. I need to pass in doubleGenerator or intGenerator depending on the type. I suppose I could just put some special casing in for them.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +376,132 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+    void append(const std::string& key, bsoncxx::builder::basic::document& builder) override {
+        choose().append(key, builder);
+    }
+    void append(bsoncxx::builder::basic::array& builder) override {
+        choose().append(builder);
+    }
+
+protected:
+    DefaultRandom& _rng;
+    std::vector<UniqueAppendable> _choices;
```
dalyd, line 401: My C++ skills aren't up to this -- but maybe...
Is there a way to definite a template function the exist only if it can exist? I'd like to add a stringEvaluate function to Generator<T>, but only if T can be passed as an argument to to_string http://www.cplusplus.com/reference/string/to_string/ or some similar form. Ideally we could handle the  bson view version also.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +376,132 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+        // convert to a string
+        std::ostringstream ipout;
+        ipout << octets[3] << "." << octets[2] << "." << octets[1] << "." << octets[0];
+        return (ipout.str());
+    }
+
+protected:
+    DefaultRandom& _rng;
+    uint32_t subnet_mask;
+    uint32_t prefix;
```
rtimmons, line 465: 1. use `camelCase`
2. `_` prefix

e.g.

```
unit32_t _subnetMask;
```

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +376,132 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+protected:
+    DefaultRandom& _rng;
+    std::vector<UniqueGenerator<std::string>> _choices;
+    std::vector<int64_t> _weights;
+};
+
+class IPGenerator : public Generator<std::string> {
+public:
+    IPGenerator(const Node& node, DefaultRandom& rng)
+        : _rng(rng), subnet_mask{std::numeric_limits<uint32_t>::max()}, prefix{} {}
```
rtimmons, line 444: use curlies for initializers to prevent conversions - `_rng{rng}`

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +376,132 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+        auto ipint = (distribution(_rng) /*& subnet_mask*/) + prefix;
+        int32_t octets[4];
+        for (int i = 0; i < 4; i++) {
+            octets[i] = ipint & 255;
+            ipint = ipint >> 8;
+        }
+        // convert to a string
+        std::ostringstream ipout;
+        ipout << octets[3] << "." << octets[2] << "." << octets[1] << "." << octets[0];
+        return (ipout.str());
```
rtimmons, line 459: 💅 no parens around `return`.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -256,6 +376,132 @@ class GeometricInt64Generator : public Generator<int64_t> {
...
+            }
+            output << part->evaluate();
+        }
+        return output.str();
+    }
+
+protected:
+    DefaultRandom& _rng;
+    std::vector<UniqueGenerator<std::string>> _parts;
+    std::string _sep;
```
rtimmons, line 503: `_separator`

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -561,6 +872,85 @@ UniqueGenerator<int64_t> intGenerator(const Node& node, DefaultRandom& rng) {
...
+        _choices.push_back(valueGenerator<true, UniqueAppendable>(v, rng, allParsers));
+    }
+    if (node["weights"]) {
+        for (const auto&& [k, v] : node["weights"]) {
+            _weights.push_back(v.to<int64_t>());
+        }
+    } else {
+        // If not passed in, give each choice equal weight
+        _weights.assign(_choices.size(), 1);
+    }
```
rtimmons, line 951: there's a bit of DRYing that could be done with this and the string choose logic. Not sure if worth it but at least a comment in both places would help if we have to fix/change in the future.

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: > I didn't test all the distributions.

I think just a single test that exercises all distributions would be worthwhile to ensure we don't break something in the future.

```
GivenTemplate:
  laplace: {^RandomDouble: {withDistribution: laplace ...}
  chi_squared: ....
```

(syntax is wrong but hopefully that's clear enough)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -152,13 +154,131 @@ using Parser = std::function<O(const Node&, DefaultRandom&)>;
...
+static const char weibullstr[] = "weibull";
+static const char extremestr[] = "extreme_value";
+static const char laplacestr[] = "laplace";
+static const char normalstr[] = "normal";
+static const char lognormalstr[] = "lognormal";
+static const char chisquaredstr[] = "chi_squared";
+static const char noncentralchisquaredstr[] = "non_central_chi_squared";
+static const char cauchystr[] = "cauchy";
+static const char fisherfstr[] = "fisher_f";
+static const char studenttstr[] = "student_t";
```
rtimmons, line 241: I always forget you can have `const char[]` as template params. This is quite nice and saves a lot of boilerplate.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -392,3 +392,50 @@ Tests:
...
+    ThenReturns:
+      - a: "It worked"
+
+  - Name: Join With Choose
+    GivenTemplate:
+      a: {^Join: {array: ["Hello", "random", "person:", {^Choose: {from: ["Huey", "Duey",
+    {^FastRandomString: {length: 10}}]}}], sep: " "}}
+    ThenReturns:
+      - a: "Hello random person: Duey"
+      - a: "Hello random person: Duey"
```
rtimmons, line 423: maybe another invocation or two so we can ensure we get `Huey` at least once -- prevent off-by-1 error from never selecting the last element

src/workloads/docs/Generators.yml
```diff
@@ -0,0 +1,135 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/stm"
+
+# This workload demonstrates all the current Document and Value Generators. It creates documents
```
rtimmons, line 4: Good idea!

rtimmons (approved):
Nicely done. I had a few ideas for some C++-isms and a few 💅 style things, but otherwise this looks 💯 from a correctness perspective.

Looks like you already did, but please run `clang-format` prior to merging, and please manually run the test-case binary one final time before final merge, until [TIG-2802](https://jira.mongodb.org/browse/TIG-2802) fixes it not being run in CI.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -152,13 +154,131 @@ using Parser = std::function<O(const Node&, DefaultRandom&)>;
...
+template <typename distribution, const char* diststring, const char* param1name>
+class DoubleGenerator1Parm : public Generator<double> {
+public:
+    DoubleGenerator1Parm(const Node& node, DefaultRandom& rng)
+        : _rng{rng}, _param1Gen{doubleGenerator(extract(node, param1name, diststring), _rng)} {}
+
+    double evaluate() override {
+        auto param1 = _param1Gen->evaluate();
+        auto dist = distribution{param1};
+        return dist(_rng);
```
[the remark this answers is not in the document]
dalyd: What happens to the _param2Gen{doubeGenerator... in the constructor if param2 isn't specified? Will that barf? 

I'm going to skip this one for now, but thanks for sharing -- I did like seeing what it looked like.

src/value_generators/src/DocumentGenerator.cpp
dalyd: I wish I could pass them in directly and not define them separately, but oh well.

src/value_generators/src/DocumentGenerator.cpp
dalyd: That is what I was looking for, but also scary :) One thing this would enable (elsewhere) would be joining not just string types, but also numbers easily. I think I figured out how to do this without the enable_if stuff, but also resisting the urge to do it as it's not worth the effort right now. I think I could add a generateString method to Generator<T>, and have it call to_string(generate()), and then I just need to make sure to define a to_string for every T used with Generator. 

It's not really worth it, especially after I realized I can generate random strings that look like numbers easily enough through the alphabet option.

dalyd (commented):
Okay. I think I'm good. I'm going to let the tests run and then I will try to merge.

dalyd:
evergreen merge

PR #391, opened 2020-12-09
WT-6966 Add fixed rate scanning to CollectionScanner

tetsuo-cpp:


src/cast_core/src/CollectionScanner.cpp
```diff
@@ -209,6 +212,30 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
                     scanFinished = true;
                     break;
                 }
+                // tetsuo-cpp: We don't want to do this after every insert.
+                // Figure out a better spot to put this.
+                if (config->scanRate) {
```
tetsuo-cpp, line 217: @rtimmons 
This is just hacky WIP code. I need to look at it closer and clean it up to make sure I'm not truncating/overflowing anything unintentionally.

But do you think the general idea is correct? The `GlobalRateLimiter` class looks like it does something similar to what I want but it does look like it was designed specifically for the `PhaseLoop` implementation so I wasn't sure whether it was wise for me to try to use it.

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -209,6 +212,30 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
...
+                    const auto now = SteadyClock::now();
+
+                    // Calculate the scan rate as a double for ease of use.
+                    const auto spec = config->scanRate->getBaseSpec();
+                    assert(spec);
+                    const auto secs = spec->per.count() / (1000 * 1000 * 1000);
+                    const auto elapsed = (now - start).count() / (1000 * 1000 * 1000);
+                    const auto expectedSize =
+                        (static_cast<double>(elapsed) / secs) * spec->operations;
```
tetsuo-cpp, line 226: I'm overloading ops/sec to mean bytes/sec. I need to add a bytes/sec variant to `RateSpec`.

- [ ] Add a bytes/second rate to the `RateSpec` implementation.

src/cast_core/src/CollectionScanner.cpp
rtimmons: Interesting.

One downside to what you have here is that it's not coordinated across all instances of the Actor. You could use the shared Actor state functionality in `ActorContext`, but I think you'll end up close to what `v1::GlobalRateLimiter` already provides.

I do think it's reasonable to use `v1::GlobalRateLimiter` here, and this is a good excuse to move it out of the "private" `v1::` namespace and get rid of the `@warning`
https://github.com/mongodb/genny/blob/09d104042931503994e5f845b0991d8c4331574b/src/gennylib/include/gennylib/context.hpp#L238-L243

We didn't make it public because we weren't sure how it would be used outside of `PhaseLoop` but this Actor does a fair amount of its own flow-control already so wanting the rate-limiting primitive makes sense.

You can just get `workloadContext->getRateLimiter("CollectionScanner", actorContext["ScanRate"].to<RateSpec>())` to get a ptr to the rate-limiter that would be shared by all `CollectionScanner` instances.

One slightly confusing thing is that `GlobalRateLimiter` doesn't actually..rate-limit, it just does the bookkeeping.  This code in PhaseLoop actually does the sleeping. The rate-limiting is done in `PhaseLoop`:
https://github.com/mongodb/genny/blob/1cfcdc88cb2c2eddbfafc3469a40f63ea2f1f12f/src/gennylib/include/gennylib/PhaseLoop.hpp#L141
With this said, I think making a little public member function on `GlobalRateLimiter` that does this makes sense. Something like this:

```c++
class GlobalRateLimiterOrWhateverItsCalled {
...
void simpleLimitRate() {
    while (true) {
        const auto now = SteadyClock::now();
        auto success = this->consumeIfWithinRate(now);
        if (!success) {
            // maybe the ternary ↓ goes away if this is an member function?
            const auto rate = this->getRate() > 1e9 ? 1e9 : this->getRate();
            std::this_thread::sleep_for(std::chrono::nanoseconds(
                int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX)))));
            continue;
        }
        break;
    }
    this->notifyOfIteration();
}
...
};
```

(This would be more or less duplicated from `PhaseLoop`, but we could handle the light refactor later.)

This would ideally be a PR separate from this PR. If this feels like too much surgery to Genny itself, please push back. STM or I can do this probably next sprint. You can just make this `simpleLimitRate` function a function in your Actor's cpp and use a ptr param instead of `this`.

Let me know if that makes sense. What you have here looks like it might work and if it works we're not in the business of saying no to individual Actors' code, but I think Genny itself could be more helpful.

tetsuo-cpp: Thanks @rtimmons. We've decided to go down this path so I've opened #395 to make the `GlobalRateLimiter` usable to us.

tetsuo-cpp:
This change will look different since I'm going to be building on top of #395.
I'll close this and re-open another PR with the new changeset.

PR #395, opened 2020-12-16
WT-7024 Make RateLimiter component public in gennylib

tetsuo-cpp:
As discussed in [this thread](https://github.com/mongodb/genny/pull/391#discussion_r539324595), it makes sense to move the `GlobalRateLimiter` component out of the private `v1` namespace and make it a public component in `gennylib` for use in actors.


src/gennylib/include/gennylib/GlobalRateLimiter.hpp
```diff
@@ -237,6 +240,27 @@ class BaseGlobalRateLimiter {
 
 using GlobalRateLimiter = BaseGlobalRateLimiter<std::chrono::steady_clock>;
 
-}  // namespace genny::v1
+inline void simpleLimitRate(GlobalRateLimiter& rateLimiter) {
```
tetsuo-cpp, line 243: This is a helper function to assist with the general case of using the rate limiter.

I figured that if it can be implemented without access to things `private` or `protected` to `GlobalRateLimiter`, then it ought to be a free function. But if you'd prefer this as a member, I can do that.

rtimmons: Slight preference to make it a member simply because it makes it more discoverable using IDEs and the like.

rtimmons (approved):
Thanks for doing this.

I've scheduled [a patch-build](https://spruce.mongodb.com/version/5fdb72fde3c3314368eae256/tasks) with these changes. If you give your 👍🏽 I'll merge this for you in a couple hours after I spot-check that this didn't affect any existing workloads.

tetsuo-cpp:
> Thanks for doing this.
> 
> I've scheduled [a patch-build](https://spruce.mongodb.com/version/5fdb72fde3c3314368eae256/tasks) with these changes. If you give your 👍🏽 I'll merge this for you in a couple hours after I spot-check that this didn't affect any existing workloads.

Sounds good! The build looks healthy.

I've just pushed a change to use a member function instead of a free function. After the PR build turns green, I'll merge this.

rtimmons (approved):

PR #396, opened 2020-12-17
WT-6966 Add fixed rate scanning to CollectionScanner

tetsuo-cpp:


tetsuo-cpp:
I've verified that this parses the rate spec and creates the rate limiter as required. Now I'm trying to construct a small workload to sanity check that this does what I think it's doing. I'm using this change:

```
modified   src/cast_core/src/CollectionScanner.cpp
@@ -215,6 +215,7 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
                     // Perform an iteration for each megabyte since our rate will be MB/time.
                     simpleLimitRate(*rateLimiter);
                     ++scanMegabytes;
+                    BOOST_LOG_TRIVIAL(info) << "Scanned " << scanMegabytes << " MB";
                 }
             }
             if (scanFinished) {
```

I'm hoping that if I use `ScanRateMegabytes: 1 per 1 second`, then I should see this message once per second when I run the workload.

rtimmons (approved):
LGTM. 
(I know this is still in "draft" but perhaps only because you're waiting on #395, so please lmk if you'd like an additional review if you plan additional changes here.)

PR #393, opened 2020-12-10
TIG-2805 Ignore failed writes and don't error out.

guoyr:
Temporary workaround until the underlying drivers issue gets fixed.

Raiden and I had been staring at this for a while and we're pretty confident the driver just isn't retrying where it should. We're going to file a HELP ticket to have some domain experts look at it and revert this commit.

In progress maintenance build variant patch: https://evergreen.mongodb.com/version/5fd21592d1fe071a9dcae722

PR #397, opened 2020-12-18
TIG-2805 allow specifying ignoreError at the phase level

guoyr:
Followup to #393 

This works around the C++ driver issue.

green patch: https://evergreen.mongodb.com/build/sys_perf_linux_3_node_replSet_maintenance_events_patch_e36da1b27ef763f4d7367ddf0f2f97259c0deac3_5fd21592d1fe071a9dcae722_20_12_10_12_50_50

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -230,9 +230,11 @@ ThrowMode decodeThrowMode(const Node& operation, PhaseContext& phaseContext) {
     // TODO: TIG-2805 Remove this mode once the underlying drivers issue is addressed.
     static const char* ignoreKey = "RecordFailure";
 
-    bool throwOnFailure =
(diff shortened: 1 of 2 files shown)


rtimmons (approved):
Plz update the crud actor docs yaml. Ideally we'd add a test-case but that may be hard to rig up. If adding tests for this isn't really feasible (ugh c++), please add a comment above these lines indicating that this logic isn't tested--alert somebody to potential dragons if they change this.

dalyd (approved):
Yeah -- more context on the ticket or in comments in the code would be appreciated. Change LGTM.

guoyr:
Thanks for the feedback. I added a doc comment. The changes here and in #393 will be reverted once https://jira.mongodb.org/browse/SERVER-53431 lands; it's currently scheduled for the upcoming Repl sprint.

I don't think there's a more permanent need for recording failed commands as mongo makes more and more errors transparent to the end user.

guoyr:
evergreen merge

PR #398, opened 2020-12-18
Add Now and RandomDate Generators

jimoleary:
The general ethos is to support  ISO-8601 and  "%Y-%m-%d". Other formats could be added but are not for the moment.

Parsers support is to the milliseconds (for easy cut and paste) in the input string but the actual precision is to seconds.
Timezone specifiers as the end of the input strings are not supported(they don't throw an error but they also are not applied unless they happend to be 'Z' or '+00:00').

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -19,9 +19,11 @@
 #include <limits>
 #include <map>
 #include <sstream>
+#include <iostream>
 
 #include <boost/algorithm/string/join.hpp>
 #include <boost/log/trivial.hpp>
+#include <boost/date_time.hpp>
 
 #include <bsoncxx/builder/basic/array.hpp>
 #include <bsoncxx/builder/basic/document.hpp>
@@ -648,6 +650,89 @@ private:
     std::string _actorId;
 };
 
+/** `{^Now: {}}` */
+class NowGenerator : public Generator<bsoncxx::types::b_date> {
+public:
+    NowGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_date evaluate() override {
+        return bsoncxx::types::b_date{std::chrono::system_clock::now()};
+    }
+};
+
+// The 2 formats also cover "%Y-%m-%d", timezones are not permissible,
+// %z, %Q are output only https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
+const std::locale formats[] = {
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%dT%H:%M:%S.%F")),
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%d %H:%M:%S.%F")),
+};
+const size_t formats_n = sizeof(formats)/sizeof(formats[0]);
+const static boost::posix_time::ptime epoch(boost::gregorian::date(1970,1,1));
+
+static unsigned long long parseDateAsMillis(const std::string &datetime) {
+    boost::posix_time::ptime pt;
+    auto millis =  0LL;
+
+    // Empty datetime is 0.
+    if(!datetime.empty()){
+        for(size_t i=0; i<formats_n; ++i) {
+            std::istringstream is(datetime);
+            is.imbue(formats[i]);
+            is >> pt;
+            if(pt != boost::posix_time::ptime()){
+                millis =  (pt - epoch).total_milliseconds();
+                break;
+            }
+        }
+
+        // No format matched.
+        if(pt == boost::posix_time::ptime()) {
+            auto msg = "^Date: Invalid Dateformat '" + datetime + "'";
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg));
+        }
+    }
+    return millis;
+}
+
+
+/** `{^Date: {min: "2015-01-01", max: "2015-01-01T23:59:59.999Z"}}` */
+class DateGenerator : public Generator<bsoncxx::types::b_date> {
+public:
+    DateGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _rng{generatorArgs.rng},
+          _min{parseDateAsMillis(node["min"].maybe<std::string>().value_or(""))},
+          _max{parseDateAsMillis(node["max"].maybe<std::string>().value_or(""))} {
+        if(to() < from()) {
+            auto max = node["max"].maybe<std::string>().value_or("");
+            auto min = node["min"].maybe<std::string>().value_or("");
+
+            std::string msg = "^Date: max('" + max +"') must be greater / equal than min('" + min +"')";
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg));
+        }
+    }
+
+    bsoncxx::types::b_date evaluate() override {
+        boost::random::uniform_int_distribution<unsigned long long> dist(from(), to());
+        return bsoncxx::types::b_date{std::chrono::milliseconds{dist(_rng)}};
+    }
+
+    unsigned long long from() {
+        return _min;
+    }
+
+    unsigned long long to() {
+        if (_max == 0) {
+            return std::chrono::duration_cast< std::chrono::milliseconds >(
+                std::chrono::system_clock::now().time_since_epoch()
+            ).count();
+        }
+        return _max;
+    }
+private:
+    DefaultRandom& _rng;
+    const unsigned long long _min;
+    const unsigned long long _max;
+};
+
 /** `{a: [...]}` */
 class ArrayGenerator : public Generator<bsoncxx::array::value> {
 public:
@@ -664,8 +749,7 @@ public:
     }
 
 private:
-    const ValueType _values;
-};
+    const ValueType _values;};
 
 /**
  * @tparam O
@@ -790,6 +874,14 @@ const static std::map<std::string, Parser<UniqueAppendable>> allParsers{
      }},
     {"^RandomInt", int64GeneratorBasedOnDistribution},
     {"^RandomDouble", doubleGeneratorBasedOnDistribution},
+    {"^Now",
+     [](const Node& node, GeneratorArgs generatorArgs) {
+         return std::make_unique<NowGenerator>(node, generatorArgs);
+     }},
+    {"^Date",
+     [](const Node& node, GeneratorArgs generatorArgs) {
+         return std::make_unique<DateGenerator>(node, generatorArgs);
+     }},
     {"^Verbatim",
      [](const Node& node, GeneratorArgs generatorArgs) {
          return valueGenerator<true, UniqueAppendable>(node, generatorArgs, allParsers);
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -663,7 +663,7 @@ public:
 // see https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
 // We strive to use smart pointers where possible. In this case this is not possible
 // but not a huge deal as these objects are statically allocated.
-const std::array formats = {
+const static auto formats = {
     std::locale(std::locale::classic(),
                 new boost::local_time::local_time_input_facet("%Y-%m-%dT%H:%M:%s%ZP")),
     std::locale(std::locale::classic(),
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -659,7 +659,7 @@ public:
     }
 };
 
-// The 2 formats also cover "%Y-%m-%d", timezones are not permissible,
+// The 2 formats also cover "%Y-%m-%d", timezones are not parsed,
 // %z, %Q are output only https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
 const std::locale formats[] = {
     std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%dT%H:%M:%S.%F")),
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -16,14 +16,14 @@
 
 #include <fstream>
 #include <functional>
+#include <iostream>
 #include <limits>
 #include <map>
 #include <sstream>
-#include <iostream>
 
 #include <boost/algorithm/string/join.hpp>
-#include <boost/log/trivial.hpp>
 #include <boost/date_time.hpp>
+#include <boost/log/trivial.hpp>
 
 #include <bsoncxx/builder/basic/array.hpp>
 #include <bsoncxx/builder/basic/document.hpp>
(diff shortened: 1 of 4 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
     std::string _actorId;
 };
 
+/** `{^Now: {}}` */
+class NowGenerator : public Generator<bsoncxx::types::b_date> {
+public:
+    NowGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_date evaluate() override {
+        return bsoncxx::types::b_date{std::chrono::system_clock::now()};
```
rtimmons, line 658: hrm. this starts to represent a departure from existing behavior where we run the same workload twice and it results in the same sequence of values being generated. That's probably exactly what the user expects in this case, but it makes me a bit hesitant. Maybe we should start the clock at some arbitrary time-point at the beginning of the workload and then make now() be an offset? Just a thought.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+    }
+};
+
+// The 2 formats also cover "%Y-%m-%d", timezones are not parsed,
+// %z, %Q are output only https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
+const std::locale formats[] = {
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%dT%H:%M:%S.%F")),
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%d %H:%M:%S.%F")),
```
rtimmons, line 666: 1. can these be smartptrs? `std::make_unique<time_input_face>("%Y--")`. Seeing "naked new" anywhere makes my spidey-senses perk-up. If that's not possible, plz put a comment saying as such. Not a huge deal since this is statically-allocated storage but still good to ack when we have to break modern-isms.
2. I *think* the formatter would put a space after the comma. Maybe you did already but plz double-check you ran `clang-format` prior to merge.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+public:
+    NowGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_date evaluate() override {
+        return bsoncxx::types::b_date{std::chrono::system_clock::now()};
+    }
+};
+
+// The 2 formats also cover "%Y-%m-%d", timezones are not parsed,
+// %z, %Q are output only https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
+const std::locale formats[] = {
```
rtimmons, line 664: this can be a `std::array` since you know the size at compile-time, and you can use that instead of `formats_n` as well. Moreover C++17 lets you use auto-deduction:

```c++
const std::array formats = { .... }
// or std::make_array would work
```

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+const static boost::posix_time::ptime epoch(boost::gregorian::date(1970,1,1));
+
+static unsigned long long parseDateAsMillis(const std::string &datetime) {
+    boost::posix_time::ptime pt;
+    auto millis =  0LL;
+
+    // Empty datetime is 0.
+    if(!datetime.empty()){
+        for(size_t i=0; i<formats_n; ++i) {
+            std::istringstream is(datetime);
```
rtimmons, line 678: curlies avoid implicit conversions and most-vexing-parse:

```c++
std::istringstream is{datetime};
```

(Applies to a few other places as well - anywhere you instantiate a thing see if you can use `{}`. Parens is a hard habit to break if you're used to C++-11 and earlier.)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+public:
+    DateGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _rng{generatorArgs.rng},
+          _min{parseDateAsMillis(node["min"].maybe<std::string>().value_or(""))},
+          _max{parseDateAsMillis(node["max"].maybe<std::string>().value_or(""))} {
+        if(to() < from()) {
+            auto max = node["max"].maybe<std::string>().value_or("");
+            auto min = node["min"].maybe<std::string>().value_or("");
+
+            std::string msg = "^Date: max('" + max +"') must be greater / equal than min('" + min +"')";
```
rtimmons, line 708: nice

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg));
+        }
+    }
+
+    bsoncxx::types::b_date evaluate() override {
+        boost::random::uniform_int_distribution<unsigned long long> dist(from(), to());
```
rtimmons, line 714: this may be over-optimization but can you do the `from()` and `to()` just once at ctor-time? I think we're not supposed to store distribution objects (at least the existing code doesn't) so otherwise this looks fine.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+// The 2 formats also cover "%Y-%m-%d", timezones are not parsed,
+// %z, %Q are output only https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
+const std::locale formats[] = {
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%dT%H:%M:%S.%F")),
+    std::locale(std::locale::classic(),new boost::posix_time::time_input_facet("%Y-%m-%d %H:%M:%S.%F")),
+};
+const size_t formats_n = sizeof(formats)/sizeof(formats[0]);
+const static boost::posix_time::ptime epoch(boost::gregorian::date(1970,1,1));
+
+static unsigned long long parseDateAsMillis(const std::string &datetime) {
```
rtimmons, line 671: any reason why we use `unsigned long long` versus just `std::chrono::milliseconds`? I know the `std::chrono` API is a bit awkward but losing the precision/type has bitten me a couple times and here it would save you from having to do a `duration_cast` in at least one place I think. I've not used the boost chrono stuff you're using here, though, so maybe std::chrono isn't applicable.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+    boost::posix_time::ptime pt;
+    auto millis =  0LL;
+
+    // Empty datetime is 0.
+    if(!datetime.empty()){
+        for(size_t i=0; i<formats_n; ++i) {
```
rtimmons, line 677: if you use `std::array` above this becomes:

```c++
for(const auto&& format : formats) { ... }
```

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+
+    unsigned long long from() {
+        return _min;
+    }
+
+    unsigned long long to() {
+        if (_max == 0) {
+            return std::chrono::duration_cast< std::chrono::milliseconds >(
+                std::chrono::system_clock::now().time_since_epoch()
+            ).count();
```
rtimmons, line 726: consider a fixed max value (2030? idk) so tests pass. If users want max=now() I think they could do `{^Date: {max: {^Now: {}}}`

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -498,4 +498,40 @@ Tests:
...
+  #   GivenTemplate:
+  #     date: {^Now: {}}
+  #   ThenReturns:
+  #   - { "date" : { "$date" : { "$numberLong" : "1453689676330" } } }
 
+  - Name: DateMinMax
+    GivenTemplate:
+      date: {^Date: {min: "2020-01-01", max: "2021-01-01"}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "1606971090942" } } }
```
rtimmons, line 512: ermm..not sure how to read this or if indeed it's in the range lol..plz add a comment of what that date represents

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -498,4 +498,40 @@ Tests:
...
+    GivenTemplate:
+      date: {^Date: {max: "2020-01-01", max: "2020-01-01"}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "1468895229847" } } }
+
+  - Name: DateRange
+    GivenTemplate:
+      date: {^Date: {max: "2020-01-01 00:00:00", max: "2020-06-01T12:00:00"}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "179757581005" } } }
```
rtimmons, line 537: maybe add a few items to this list to guard against regressions that could e.g. kill the distribution and instead start returning constant values.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+/** `{^Now: {}}` */
+class NowGenerator : public Generator<bsoncxx::types::b_date> {
+public:
+    NowGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_date evaluate() override {
+        return bsoncxx::types::b_date{std::chrono::system_clock::now()};
+    }
+};
+
+// The 2 formats also cover "%Y-%m-%d", timezones are not parsed,
```
rtimmons, line 662: not parsing timezones makes me sad (not quite as sad as parsing libs that don't support emojis but sad nonetheless). Is there any reason why zulu offsets aren't parsed? Perhaps with `%ZP` according to the docs?

rtimmons (approved):
The code and logic looks correct. A few rather minor C++-ism to consider.

Otherwise my biggest pause is using "real" now. I'll defer to you and David D on that, but I can imagine very annoying performance things that happen only on occasional runs of the workload where now() happens to be pathological (hey maybe prime timestamps cause us to go pear-shaped).

dalyd:
> The code and logic looks correct. A few rather minor C++-ism to consider.
> 
> Otherwise my biggest pause is using "real" now. I'll defer to you and David D on that, but I can imagine very annoying performance things that happen only on occasional runs of the workload where now() happens to be pathological (hey maybe prime timestamps cause us to go pear-shaped). That could be maddening to try to repro or investigate.

I'm looking now, but wanted to address Ryan's comment first. I hate using real dates, but I think we're almost forced to. We want these increasing across the length of the test (when using $now), and I don't think we have a reasonable way to synchronize all the nodes in the cluster to a fake datetime. I'm not sure I would want to attempt that in a real distributed system either. Possibly on a single system within a container, but otherwise seems ripe for a lot of bad things to happen.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -498,4 +498,40 @@ Tests:
     #    - { "int" : { "$numberLong" : "2000000000"}, "string" : "2000000000" }
     - "string": "2"
 
+  # Now moves with time so would always fail.
```
dalyd, line 501: Can we mock out time in C++ so this gives a deterministic value? That would need to be in the DocGenYamlTestCaseRunner, but I think catch2 has some ability for monkey patching. Maybe not. Might be time to look at google test and it's mock support https://github.com/google/googletest/blob/8779937dd05016e3a96e477d2bb1f94947486605/googlemock/docs/cook_book.md FYI @rtimmons .

dalyd: This is just an FYI and for future conversation. Not this ticket.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -498,4 +498,40 @@ Tests:
...
+  # - Name: Now
+  #   GivenTemplate:
+  #     date: {^Now: {}}
+  #   ThenReturns:
+  #   - { "date" : { "$date" : { "$numberLong" : "1453689676330" } } }
```
dalyd, line 506: Would it be worth extending the test runner to have an approximate check here? Something to make sure it returns a nonZero number?

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: Yeah, Mongo BSON dates are -- hard to read.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -498,4 +498,40 @@ Tests:
...
 
+  - Name: DateMinMax
+    GivenTemplate:
+      date: {^Date: {min: "2020-01-01", max: "2021-01-01"}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "1606971090942" } } }
+
+  # The tests with no max generate different values every time as max moves.
```
dalyd, line 514: Can we make default max something constant?

src/value_generators/src/DocumentGenerator.cpp
dalyd: Oh, I need to read more closely. I prefer Ryan's version. That fits into my challenge question above. I'm not sure a default of NOW is a reasonable answer. I think we should just pick a value to use as the default if none is given.

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: That's probably good advice on all of these.

dalyd (approved):
Alright, despite dropping a lot of comments on here, I'm going to approve for now. The code looks good. Don't let my comments make it arbitrarily large amount of work. This is really cool.

src/value_generators/src/DocumentGenerator.cpp
rtimmons: I'm not sure how much "real" time values matters. I was imagining "now" could start as `2020-12-18T00:00:00.00000` or something and then every subsequent call to `now` could increment the current clock by a random amount or something. Or perhaps it could increment the fake clock by the amount of time that has elapsed since the workload has started.

I don't know the full use-case for wanting `now()` though. We don't really need to worry about it being the same across clusters because the workload-client is the only machine that would see the value unless queries use `$currentDate` and friends.

(I'm fine with however you decide to proceed here, imho, as long as we have clear messaging _somewhere_ about the implications of repro-ability for workloads that use `^Now`.)

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: When I first added Catch2 the mocking support wasn't as advanced as it is now and I think the underlying classes had to be designed pretty carefully to support mocking. That seems less true now. It's worth considering adopting more deeply.

We do some primitive dummying of clocks for our rate-limiters tests:
https://github.com/mongodb/genny/blob/3f5b992a916cff4a5408d729b0f391f9e8be0a25/src/gennylib/test/GlobalRateLimiter_test.cpp#L61-L79
But I haven't thought deeply if the same approach works well for value-generators.

src/value_generators/src/DocumentGenerator.cpp
dalyd: I'm going to mull on this a bit more. I think we should do as built for now, but see if we can do something better. I worry about some interaction with the running system and real times coming up, but don't have a handy example so maybe what you suggest would work. If we did something like this I would like to fix the test start time, and then advance from their proportional to real time. That is, if the server got faster, there could end up more documents within a specific datetime range.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,89 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+            auto msg = "^Date: Invalid Dateformat '" + datetime + "'";
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg));
+        }
+    }
+    return millis;
+}
+
+
+/** `{^Date: {min: "2015-01-01", max: "2015-01-01T23:59:59.999Z"}}` */
+class DateGenerator : public Generator<bsoncxx::types::b_date> {
```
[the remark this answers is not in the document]
jimoleary: 1) is a new ticket IMO. A date range makes sense but it needs a new ticket and discussion of what it should do and how.
2) nesting makes sense but I would suggest raising a ticket and coming back to it.
3) The code currently uses ```uniform_int_distribution```. Other distributions make sense but should be implemented later as a follow on ticket.
4) trivially implemeneted  with std::stoull but could probably also be implemented with a format string.

src/value_generators/src/DocumentGenerator.cpp
jimoleary: Updated to show both the input and parsed value (this is useful for TZ support).

src/value_generators/src/DocumentGenerator.cpp
jimoleary: OCD on my part to have ```to``` be symmetrical with ```from```.  Unwrapped now.

This also highlighted a subtle issue related to what type of interval is defined by min / max and what the precision is. The definition is now tightened up as a half-open interval **[min, max)** with a max precision of 1ms.

src/value_generators/src/DocumentGenerator.cpp
jimoleary: default min is 1970-01-01, default max is 2030-01-01. And support for non string parameters should be added later.

src/value_generators/test/DocGenYamlTestCaseRunner.cpp
```diff
@@ -131,7 +131,7 @@ struct convert<genny::ValGenTestCase> {
 
 }  // namespace YAML
 
-TEMPLATE_TEST_CASE("YAML Tests", "", genny::ValGenTestCase) {
+TEMPLATE_TEST_CASE("YAML Tests", "[vector][template]", genny::ValGenTestCase) {
```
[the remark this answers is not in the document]
jimoleary: nothing, this was me trying to figure out how to run the value generator tests.

```
$ ./build/src/value_generators/value_generators_test -d yes --order lex -\# '[#DocGenYamlTestCaseRunner]'
```

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: yes but ... not in this ticket.

I have uncommented this test as there is already another failing test in here.

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: This could also be related to the suggestion to extend the test runner to support some other approximation checked (see the next comment).

But not in this ticket in any case.

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: Mongo BSON date, I've added a comment to generate the time:

```
$> date --utc -d@$((1606971090942 / 1000))  +"%Y-%m-%dT%H:%M:%S.%3N"
2020-12-03T04:51:30.000
```


Making these more 'friendly' would be part of another ticket (see the last 2 comments). I'm not completely convinced of the merit of this part though.

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: yes! max and min now have defaults.

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: I think I've covered this now, but let me know if I haven't.

jimoleary (commented):
I've generally provided comments inline, but I also have some top level comments.

1) It would be surprising to me if 'Now' did not return the current date / time,
2) Date is really RandomDate so I've renamed the operator to reflect this.
3) We are missing a new operator called Date which will generate date ranges from ```start``` (default to epoch) to ```stop``` (defaults to 2030-01-01) with a ```step``` (a timespec that defaults to 1ms). Think python slice. Rollover should probably reset back to start and step, minimum step value is also 1ms but we can discuss this. This new operator requires a new ticket and would address Ryans' concern about being able to generate the same dates for every test run.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -663,7 +663,7 @@ class NowGenerator : public Generator<bsoncxx::types::b_date> {
 // see https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
 // We strive to use smart pointers where possible. In this case this is not possible
 // but not a huge deal as these objects are statically allocated.
-const std::array formats = {
+const static auto formats = {
```
jimoleary, line 666: This ends up being a ```std::initializer_list```. I assume that is ok.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,92 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+    const std::string& datetime, const boost::posix_time::ptime& default_time = epoch) {
+    boost::posix_time::ptime pt;
+    auto millis = 0LL;
+
+    // Empty datetime is 0.
+    if (!datetime.empty()) {
+        for (const auto& format : formats) {
+            std::istringstream is{datetime};
+            is.imbue(format);
+            boost::local_time::local_date_time ldt{boost::local_time::not_a_date_time};
```
dalyd, line 685: is ldt and pt from date time examples? We try to avoid abbreviations, but I accept that things could get big here if we don't.

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: Could you add follow-up ticket to the comment?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -648,6 +650,92 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+
+            std::string msg = "^RandomDate: max('" + max + " / " + std::to_string(_max.count()) +
+                "') must be greater than min('" + min + " / " + std::to_string(_min.count()) + "')";
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{msg});
+        }
+    }
+
+    bsoncxx::types::b_date evaluate() override {
+        boost::random::uniform_int_distribution<long long> dist{_min.count(), _max.count() - 1};
+        return bsoncxx::types::b_date{std::chrono::milliseconds{dist(_rng)}};
```
rtimmons, line 730: this turned out really nice

rtimmons (approved):
I concur with David's comments, but the issues I originally brought up seem 💯. Thanks for withstanding the myriad of comments. This turned into very easy-to-read and idiomatic C++ (which is quite a feat).

src/value_generators/src/DocumentGenerator.cpp
jimoleary: yeah, they are from the examples, I'll remove the abbreviations.

jimoleary (commented):
New tickets are:

- PERF-2084 'Add ^Date value generator to Genny'
- PERF-2085 'Update ^RandomDate Generator inputs.'
- PERF-2086 'Add approximation checks support to genny ValGenTestCase'

jimoleary:
The evergreen tests pass, the comments have been addressed and new tickets have been created .. goiing to merge now,

Thanks for all the great feedback!

jimoleary:
evergreen merge

PR #399, opened 2020-12-18
TIG-2714 Move src/python into src/lamplib

rtimmons:
Only the first part of TIG-2714. I want to make most of the path and curator/results logic in python, and having it spread out was getting confusing. We have a lot less python now, and I want to consolidate so I can stop confusing myself over what lives where.

Also:

- clean up fixture files that are no longer needed now that csv parsing is no longer in genny
- move to pytest because it's nicer

Patch-build here:
https://spruce.mongodb.com/version/5fdd22b31e2d170329ee2498/tasks


src/lamplib/requirements.txt
```diff
@@ -1,7 +1,8 @@
-nose==1.3.7
 pymongo>=3.9
 PyYAML==5.1
 requests==2.21.0
 yamllint==1.15.0
 shrub.py==1.1.0
 black==19.10b0
+setuptools==51.0.0
```
guoyr, line 7: sigh, Python

guoyr (approved):
LGTM 💯

rtimmons:
evergreen merge

PR #400, opened 2020-12-21
WT-6968 Configure new workload to use fixed-rate scanning

tetsuo-cpp:
This PR is configuring the new LSWA workload to take advantage of the functionality added in #396.

### DIFF
--- a/src/phases/scale/LargeScalePhases.yml
+++ b/src/phases/scale/LargeScalePhases.yml
@@ -236,6 +236,15 @@ SnapshotScanner5GigabytesCmd:
   CollectionSortOrder: forward
   GlobalRate: 1 per 5 minutes
 
+# Leave out the ScanDuration since we'll be supplying the ScanRateMegaBytes option.
+SnapshotScanner5GigabytesFixedRateCmd:
+  Duration: *Duration
+  ScanSizeBytes: 500000000
+  ScanType: snapshot
+  CollectionSkip: 300
+  CollectionSortOrder: forward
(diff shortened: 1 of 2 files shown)


tetsuo-cpp:
@sulabhM 
I ran this workload locally and verified that the rate was being adhered to. I've pasted the diff + the testing output onto the associated JIRA ticket in case you're interested.

src/phases/scale/LargeScalePhases.yml
```diff
@@ -236,6 +236,15 @@ SnapshotScanner5GigabytesCmd:
...
   GlobalRate: 1 per 5 minutes
 
+# Leave out the ScanDuration since we'll be supplying the ScanRateMegaBytes option.
+SnapshotScanner5GigabytesFixedRateCmd:
+  Duration: *Duration
+  ScanSizeBytes: 500000000
+  ScanType: snapshot
+  CollectionSkip: 300
+  CollectionSortOrder: forward
+  GlobalRate: 1 per 5 minutes
```
sulabhM, line 246: What is the global rate here, how does it play along scan rate megabytes ?

tetsuo-cpp: The GlobalRate is how often the scanner is invoked during a phase. So if it finishes and the phase is still happening due to some other actors, it will start another one in 5 minutes.

sulabhM: ok, makes sense.

sulabhM (approved):
LGTM

PR #401, opened 2020-12-21
TIG-2586 Add support for parameterizing load (thread level, rate limit)

aggrand:
The existing system treated parameters as global variables, so I went ahead and added a dynamic scoping system. Hopefully that'll make this more robust as we add more preprocessing stuff (and defend against weird outcomes with external phases inside templates).

Patch running `mixed_workloads_genny`: https://evergreen.mongodb.com/version/5fe11e4c850e6137395dec3a


src/driver/test/workload_parsers_test.cpp
```diff
@@ -57,5 +102,59 @@ TEST_CASE("WorkloadParser can run generate smoke test configurations") {
...
+    const auto input = (R"(
+ActorTemplates:
+- TemplateName: TestTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "IncorrectDefault"}}
+    SomeKey: SomeValue
+    Phases:
+      OnlyIn:
+        Active: [{^Parameter: {Name: "Phase", Default: 1}}]
+        Max: 3
```
rtimmons, line 115: I don't get what `Max` is here - is that documented more explicitly somewhere? If it wants to be active in more than 1 Phase, can it be:

```yaml
ActiveInPhases: [{^Parameter: {Name: ActiveInPhases, Default: 3}]
# Or:
ActiveInPhases: All
# Or:
OnlyActiveInPhaseRange:
  From: {^Parameter: {Name: ActiveFromPhase, Default: 7}
  To: {^Parameter: {Name: ActiveToPhase, Default: 10}
# Or:
```

And then enforce that `ActiveInPhases` is not empty xor `ActiveInPhases==All` xor `ActiveInPhaseRange` is non-empty.

Would hopefully be a simple change to extract the logic:

```c++
std::pair<int, int> extractPhaseRange(YAML::Node phasesBlock); // => min/max inclusive
```

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -33,6 +33,70 @@ Clients:
...
+    Type: CrudActor
+    Database: *dbname
+    ClientName: Update
+    Threads: {^Parameter: {Name: "Threads", Default: 1}}
+    Phases:
+      OnlyIn:
+        Active: [{^Parameter: {Name: "Phase", Default: 1}}]
+        Max: 10
+        Config:
+          ExternalPhaseConfig:
```
rtimmons, line 49: hrm. Do we still really want/need `ExternalPhaseConfig` **and** templates? It's kinda neat that it works and that you can `genny evaluate` to undo all of the magic, but it seems really hard to reason about. I would almost be inclined to say you can't use `ExternalPhaseConfig` inside `ActorTemplate` but perhaps that's overly limiting?

src/driver/test/workload_parsers_test.cpp
```diff
@@ -19,6 +19,51 @@
...
+
+        auto retrievedOldInner = context->get("innerName1", Type::kParameter);
+        REQUIRE_FALSE(retrievedOldInner == inner);
+
+    }
+
+    auto retrievedOuter = context->get("outerName", Type::kParameter);
+    REQUIRE(retrievedOuter == outer);
+
+    REQUIRE_THROWS(context->get("outerName", Type::kActorTemplate));
```
rtimmons, line 63: I don't think this test would guard against regressions that could be introduced if we move away from `ContextGuard` to something else. Is there a way to test the scoping functionality using actor yaml directly?

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -81,278 +145,153 @@ Actors:
...
-  Phases:
-  - Phase: 0..1
-    Nop: true
-  - ExternalPhaseConfig:
-      Path: &PhasePath ../../phases/scale/MixPhases.yml
-      Key: UpdatePhase
-  - Phase: 3..10
-    Nop: true
+# Update Actors
+- ActorInstance:
```
rtimmons, line 149: How about `ActorFromTemplate`?

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -81,278 +145,153 @@ Actors:
...
+    Template: UpdateTemplate
+    Parameters:
```
rtimmons, line 151: How about `TemplateName` and `TemplateParameters` to make it clear how these blocks are treated?

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -81,278 +145,153 @@ Actors:
...
+      Name: Update_1
+      Threads: 1
+      Phase: 2
```
rtimmons, line 154: `OnlyActiveInPhase`?

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -15,16 +15,72 @@
...
+public:
+    using ContextMap = std::map<std::string, std::pair<YAML::Node, Type>>;
+    Context(Context* enclosing) : _enclosing{enclosing} {}
+
+    std::optional<YAML::Node> get(const std::string&, const Type& type);
+    void insert(const std::string&, const YAML::Node&, const Type& type);
+
+    // Insert all the values of a node, assuming they are of a type.
+    void insert(const YAML::Node&, const Type& type);
+    Context* enclosing() {return _enclosing;}
```
rtimmons, line 56: 💅🏽 plz ensure you run `clang-format` prior to merge

src/driver/src/workload_parsers.cpp
```diff
@@ -109,7 +157,54 @@ void WorkloadParser::convertExternal(std::string key, YAML::Node value, YAML::No
...
+    }
+    return out;
+}
+
+YAML::Node WorkloadParser::parseInstance(YAML::Node instance) {
+    YAML::Node actor;
+
+    {
+        ContextGuard guard(_context);
+        auto temp = _context->get(instance["Template"].as<std::string>(), Type::kActorTemplate);
```
rtimmons, line 192: had to read this a couple times to pasre `temp` as `template` rather than `temporary`. Same applies to another place or two - can you rename to `auto template`?

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -15,16 +15,72 @@
...
+
+/**
+ * Opens a context, closes it when exiting scope.
+ */
+class ContextGuard {
+public:
+    ContextGuard(Context*& contextPtr) : _contextPtr{contextPtr}, 
+        _managedContext{std::make_unique<Context>(_contextPtr)} {
+
+        _contextPtr = _managedContext.get();
```
rtimmons, line 72: It took me longer than I care to admit to understand what's going on here. I think that's mostly due to Monday-after-holiday, but I think there could be some slight tweaks to make this rather simple and clever idea easier to grok. Mostly only important if it needs to remain in the header.

It _looks_ weird to have `*&` types, especially since you have the `unique_ptr` manage the ptr-ref that is presumably managed by the caller. Reassigning a ref (especially an member variable) seems potentially quite confusing.

There are cases to use this--and perhaps this is one of those--but I think you could instead have this be something like a linked-list or stack of params (which are the "meat" of the contexts):

```c++
// hpp
class WorkloadParser {
  // lets you defer the impl to the cpp
  std::unique_ptr<typename Context> _context;
  // not a u_p<Ctx*> or u_p<Ctx*&>, just u_p<Ctx>
};

// cpp
struct Context {
  std::stack<std::unordered_map<Node,Type>> _params;
  struct ScopeGuard {
    ScopeGuard() {
      _params.push(...);
    }
    ~ScopeGuard() { _params.pop(); }
  };
  get(k,t) {
     auto top = _params.top();
     while(top) {
       if (k in top.keys()) { return top[k]; }
       top = _params.previous(top); // not sure of the stack api
     }
  }
  insert(k,v,t) {
    _params.top().insert(k,v,t);
  }
  ScopeGuard enter() { return ScopeGuard(); }
};
...
parseExternal() {
  _context->enter();
}
```

To me this is clearer at the expense of not being quite as clever with reassigning refs, but this may not all work out or it may end up less clear than what you started with here. 

ANYWYAY: It seems like this is all "helper" logic for the impl. If that's the case, see if you can put it in the CPP in which case these concerns are, imho, not really all that important as long as the code works (and it seems to according to your tests).

My design-spidey-senses perk up whenever I see things like this in headers which have a tendency to increase in prevalence over time.

rtimmons (changes requested):
Took me a while to figure out how this worked. Very clever idea and I like the new functionality a lot. I have some ideas about how to make it easier to grok for those who are a bit more dense (like myself), but like I said in the comments, I think all is fair in cpp; the actual design concerns only "matter" in headers.

So for me the only blocking things are:

1. change some of the Context logic to be a bit more explicit or move it into the cpp (perhaps both if you agree with the design change)
2. some minor naming/syntax things

I hate to blow this up, but syntax is one of the most important part of Genny and getting it wrong will sour new users on it, so I try to have strong opinions (loosely held) about it. Please push back wherever you disagree since multiple opinions are infinitely better than 1.

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -15,16 +15,72 @@
...
+};
+
+/**
+ * Manages scoped context for stored values.
+ *
+ * Only create using a ContextGuard.
+ */
+class Context {
+public:
+    using ContextMap = std::map<std::string, std::pair<YAML::Node, Type>>;
```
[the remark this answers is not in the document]
aggrand: I started doing this, but ended up having to do `std::make_pair` every single time `insert` was called, or do `returnVal->first` every single time `get` was called (the type is redundant for `get`, since you already assert the type by passing it in, so it'll literally always be `returnVal-first`.) That seemed a lot less ergonomic to me.

Instead I used `ContextValue` internally to `Context`, which hopefully at least makes it clearer. Please push back if you think using it in `get` or `insert`'s signature is also worth doing!

src/driver/test/workload_parsers_test.cpp
aggrand: `Max` lets the evaluator know the total number of phases that are going to run, so that inactive phases can be listed as `Nop`. I'd originally hoped that we could just not list phases in genny and they'll be assumed a Nop, but that causes a throw [here](https://github.com/mongodb/genny/blob/8f22265a48b6c9fd56ffb35c428db6e13a7a5cff/src/gennylib/include/gennylib/PhaseLoop.hpp#L491).

Another option would be to do a second preprocessing pass after all the templates and actors have been evaluated to determine how many phases will run, but I feel like it's hairy to have the result of evaluation rely on "non-local" information.

src/driver/src/workload_parsers.cpp
aggrand: Unfortunately `template` gets interpreted as a keyword by the compiler, so I changed it to `templateNode`.

src/workloads/scale/MixedWorkloadsGenny.yml
aggrand: Done.

src/workloads/scale/MixedWorkloadsGenny.yml
aggrand: Done.

src/workloads/scale/MixedWorkloadsGenny.yml
aggrand: Done.

src/driver/test/workload_parsers_test.cpp
aggrand: Added an actor yaml-based test. I think keeping this test has value too; this one is a lot more explicit than the yaml-based one.

src/workloads/scale/MixedWorkloadsGenny.yml
aggrand: My gut instinct is that more composability is usually a good thing, and as per the slack convo request to allow the `ActorTemplate`s to someday be a separate file that can be specified, I think that sort of composability to mix/match actor templates and phases may be desired by users over adding an artificial limit.

Right now all this preprocessing is just parameterized substitutions with a couple different convenience syntaxes, so I feel like if we keep to that principle we should hopefully be able to manage the complexity?

src/driver/include/driver/workload_parsers.hpp
aggrand: I followed your stack-based approach; you're right that's a lot clearer. Is there a way to move this into the cpp file but still allow us to unit-test it without the class declaration in the header? I'd prefer to keep the unit tests for `Context`; it's a lot more explicit than the actor-yaml one I added which relies on the parsing behavior of templates and external phases.

src/driver/test/workload_parsers_test.cpp
rtimmons: Oh, each Actor must know the total number of Phases across all other Actors. Gotcha. Hm. I'm not super keen on modifying `Context` just yet.

Maybe rename to `NopInPhasesUpTo`?

```yaml
Phases:
  OnlyActiveInPhases: [1]
  NopInPhasesUpTo: 2
  PhaseConfig: {foo: bar}
```

Equivalent to:

```yaml
Phases:
- {Phase: 0, Nop: true}
- {Phase: 1, foo: bar}
- {Phase: 2, Nop: true}
```
 Naming is hard.

src/workloads/scale/MixedWorkloadsGenny.yml
rtimmons: I'm afraid of somebody looking at a complicated example involving both templates and external phases and coming to the conclusion that genny is too hard to configure and use.

So ya such a limit would be artificial, but I view it more as "guardrails" encouraging somebody to simplify and to maybe come to us if they want the rails removed.

I'm really fine to leave it--up to you--but perhaps spit out a warning or make it clear in the docs that using templates plus external phases might mean writing a new Actor would be simpler. And let's also mention in the docs/examples of templates and external phases that `genny evaluate` can be used to show the yaml with all includes/parameters applied.

src/driver/include/driver/workload_parsers.hpp
rtimmons: Cool. And no if we wanna test these impl details specifically then it's gotta be in the header. I sympathize with wanting to keep the test. In that case, please ensure the helper stuff is in the `v1` namespace and/or that it's documented in comments that it's a helper detail and only in the header for ease of testing.

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -16,15 +16,68 @@
...
+
+    // Insert all the values of a node, assuming they are of a type.
+    void insert(const YAML::Node&, const Type& type);
+
+    /**
+     * Opens a scope, closes it when exiting scope.
+     */
+    struct ScopeGuard {
+        ScopeGuard(Context& context) : _context{context} {
+            _context._scopes.emplace_back();
```
rtimmons, line 62: I had to triple-read this to understand that we're creating a new `Context` here. It's easy to forget that `emplace` forwards to the ctor. Maybe add a comment along the lines of  `// Emplace (push) a new Context onto the stack`. Idk. If I had more C++14 paged into my memory maybe this would have been more obvious.

src/driver/include/driver/workload_parsers.hpp
```diff
@@ -16,15 +16,68 @@
...
+};
+
+/**
+ * Manages scoped context for stored values.
+ *
+ * Only create using a ContextGuard.
+ */
+class Context {
+public:
+    using ContextValue = std::pair<YAML::Node, Type>;
```
rtimmons, line 48: I don't think you need this in the header anymore 😄

rtimmons (approved):
This turned out quite nicely. A couple minor things to consider mostly around helping workload-authors not footgun.

src/driver/test/workload_parsers_test.cpp
aggrand: Good idea. Done!

aggrand:
evergreen merge

PR #403, opened 2021-01-04
WT-7041 Add MonotonicLoader actor

tetsuo-cpp:
@guoyr 
As discussed with @sulabhM, we need some way of generating monotonically increasing _ids.

I've voted for this feature on the ideas doc, but in the interim to unblock ourselves, I'm proposing to add a `MonotonicLoader` which basically does the same thing but with monotonically increasing _ids (the guts of this change is in e229083). When this feature gets implemented, we can remove this actor.

### DIFF
--- a/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
+++ b/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
@@ -1,0 +1,67 @@
+// Copyright 2021-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


rtimmons (approved):
LGTM. Nice workaround for the lack of support in value-generators.

tetsuo-cpp:
@sulabhM I'll hold off merging this until we figure out what's going on with the update rate.

sulabhM (approved):

tetsuo-cpp:
Ok @sulabhM and I decided to merge this now to keep things moving. We can follow up with another change if the WT-7055 investigation yields something interesting.

PR #407, opened 2021-01-06
WT-7057 Fix AutoRun indentation in LSWA workload

tetsuo-cpp:
Not sure if this is the culprit behind why this task sometimes doesn't get generated in patch builds. But I figure we may as well eliminate that possibility and fix the indentation.


rtimmons (approved):
LGTM. I haven't given your autorun issue a look but I'm not sure this would fix it. This indentation makes it look nicer so 👍🏽. But I think we may need to look at it together. Let me know if you could use a hand with it. It's pretty easy to run locally but it's not super obvious how to do so.

PR #409, opened 2021-01-13
PERF-2077 incrementing counter

apodelko:
Here is the first draft of the incrementing counter https://jira.mongodb.org/browse/PERF-2077 for discussion. I don't like it at all - not what it does, not code itself... Don't think that it is ready for a merge in any way. But after wasting a lot of time trying to figure out architecture / new C++ features / yaml / etc, it looks like I can utilize some help and I decided that bringing it out as a PR may be a good way to solicit some help. Here is the list of questions I have for now (and I probably didn't get to all questions yet].

1. It looks like the current DocumentGenerator architecture doesn't allow to have a shared counter between Actors - as its instances are per Actor.  [It appears that a shared counter may be in the workload context - but it appears that it will require much more changes, including architecture.]

2. If we have it per Actor, can we have multiple parallel threads per Actor? Do we need synchronization for the counter here? [I guess for document generation we probably don't need it]

3. So, as discussed in https://jira.mongodb.org/browse/PERF-2077, it appears that we basically have only multiplier and step as parameters. Not sure how we can use 'start' here - I guess we may hardcode 1 here, but it doesn't sound exciting.

4. It doesn't look like the test architecture used allows to verify if incrementing actually works (in DocumentGeneratorTestCases.yml). It appears that every iteration of the loop in DocGenYamlTestCaseRunner.cpp  
for (const auto&& nextValue : this->_thenReturns) {
resets the counter, so we can test only for the hardcoded initial counter value.

5. It appears that we have a problem with the assert in DocGenYamlTestCaseRunner.cpp - it does fail differing in a space (either you have it or not):
.../src/value_generators/test/DocGenYamlTestCaseRunner.cpp:88: FAILED:
  REQUIRE( toString(expected.view()) == toString(actual.view()) )
with expansion:
  "{ "int" : { "$numberLong" : "202" } }"
  ==
  "{ "int" : { "$numberLong" : "202"} }"
Actually it looks like it is not a new issue: https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocumentGeneratorTestCases.yml#L491-L492

 Any input / feedback would be much appreciated.

Thanks,
Alex

### DIFF
--- a/src/value_generators/include/value_generators/DocumentGenerator.hpp
+++ b/src/value_generators/include/value_generators/DocumentGenerator.hpp
@@ -39,6 +39,7 @@ public:
 struct GeneratorArgs {
     DefaultRandom& rng;
     ActorId actorId;
+    int64_t& counter;
 };
 
 
--- a/src/value_generators/include/value_generators/DocumentGenerator.hpp
+++ b/src/value_generators/include/value_generators/DocumentGenerator.hpp
@@ -39,7 +39,6 @@ public:
 struct GeneratorArgs {
     DefaultRandom& rng;
     ActorId actorId;
-    int64_t& counter;
 };
 
 
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -743,8 +743,10 @@ private:
 class IncGenerator : public Generator<int64_t> {
 public:
     IncGenerator(const Node& node, GeneratorArgs generatorArgs)
-        : _step{node["step"].maybe<int64_t>().value_or(1)} 
-        {_counter = node["start"].maybe<int64_t>().value_or(1) + generatorArgs.actorId * node["multiplier"].maybe<int64_t>().value_or(1);}
+        : _step{node["step"].maybe<int64_t>().value_or(1)} {
+        _counter = node["start"].maybe<int64_t>().value_or(1) +
+            generatorArgs.actorId * node["multiplier"].maybe<int64_t>().value_or(1);
+    }
 
     int64_t evaluate() override {
         auto inc_value = _counter;
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -230,6 +230,9 @@ static const char nstr[] = "n";
 static const char sstr[] = "s";
 static const char minstr[] = "min";
 static const char maxstr[] = "max";
+static const char stepstr[] = "step";
+static const char startstr[] = "start";
+static const char multiplierstr[] = "multiplier";
 static const char alphastr[] = "alpha";
 static const char betastr[] = "beta";
 static const char lambdastr[] = "lambda";
@@ -736,6 +739,33 @@ private:
     const std::chrono::milliseconds _max;
 };
 
+
+class IncGenerator : public Generator<int64_t> {
+public:
+    IncGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _actorId{generatorArgs.actorId},
+          _counter{generatorArgs.counter},
+          _step{extract(node, "step", "^Inc").to<int64_t>()},
+          _multiplier{extract(node, "multiplier", "^Inc").to<int64_t>()}
+          {}
+
+    int64_t evaluate() override {
+//        std::lock_guard<std::mutex> lg(mtx);
+        std::cout << "Counter before " << _counter << std::endl;
+        _counter=_actorId*_multiplier + _counter + _step;
+        std::cout << "Counter after " << _counter << std::endl;
+        return _counter;
+    }
+
+private:
+//    std::mutex mtx;
+    int64_t _actorId;
+    int64_t& _counter;
+    int64_t _step;
+    int64_t _multiplier;
+};
+
+
 /** `{a: [...]}` */
 class ArrayGenerator : public Generator<bsoncxx::array::value> {
 public:
@@ -889,6 +919,10 @@ const static std::map<std::string, Parser<UniqueAppendable>> allParsers{
     {"^Verbatim",
      [](const Node& node, GeneratorArgs generatorArgs) {
          return valueGenerator<true, UniqueAppendable>(node, generatorArgs, allParsers);
+     }},
+    {"^Inc",
+     [](const Node& node, GeneratorArgs generatorArgs) {
+         return std::make_unique<IncGenerator>(node, generatorArgs);
      }},
 };
 
@@ -1148,12 +1182,13 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
 }  // namespace
 
 // Kick the recursion into motion
+int64_t counter=0;
 DocumentGenerator::DocumentGenerator(const Node& node, GeneratorArgs generatorArgs)
     : _impl{documentGenerator<false>(node, generatorArgs)} {}
 DocumentGenerator::DocumentGenerator(const Node& node, PhaseContext& phaseContext, ActorId actorId)
-    : DocumentGenerator{node, GeneratorArgs{phaseContext.rng(actorId), actorId}} {}
+    : DocumentGenerator{node, GeneratorArgs{phaseContext.rng(actorId), actorId, counter}} {}
 DocumentGenerator::DocumentGenerator(const Node& node, ActorContext& actorContext, ActorId actorId)
-    : DocumentGenerator{node, GeneratorArgs{actorContext.rng(actorId), actorId}} {}
+    : DocumentGenerator{node, GeneratorArgs{actorContext.rng(actorId), actorId, counter}} {}
 
 
 DocumentGenerator::DocumentGenerator(DocumentGenerator&&) noexcept = default;
@@ -1168,5 +1203,6 @@ bsoncxx::document::value DocumentGenerator::operator()() {
 bsoncxx::document::value DocumentGenerator::evaluate() {
     return operator()();
 }
+
 
 }  // namespace genny
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -743,26 +743,18 @@ private:
 class IncGenerator : public Generator<int64_t> {
 public:
     IncGenerator(const Node& node, GeneratorArgs generatorArgs)
-        : _actorId{generatorArgs.actorId},
-          _counter{generatorArgs.counter},
-          _step{extract(node, "step", "^Inc").to<int64_t>()},
-          _multiplier{extract(node, "multiplier", "^Inc").to<int64_t>()}
-          {}
+        : _step{node["step"].maybe<int64_t>().value_or(1)} 
+        {_counter = node["start"].maybe<int64_t>().value_or(1) + generatorArgs.actorId * node["multiplier"].maybe<int64_t>().value_or(1);}
 
     int64_t evaluate() override {
-//        std::lock_guard<std::mutex> lg(mtx);
-        std::cout << "Counter before " << _counter << std::endl;
-        _counter=_actorId*_multiplier + _counter + _step;
-        std::cout << "Counter after " << _counter << std::endl;
-        return _counter;
+        auto inc_value = _counter;
+        _counter += _step;
+        return inc_value;
     }
 
 private:
-//    std::mutex mtx;
-    int64_t _actorId;
-    int64_t& _counter;
     int64_t _step;
-    int64_t _multiplier;
+    int64_t _counter;
 };
 
 
@@ -1182,13 +1174,12 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
 }  // namespace
 
 // Kick the recursion into motion
-int64_t counter=0;
 DocumentGenerator::DocumentGenerator(const Node& node, GeneratorArgs generatorArgs)
     : _impl{documentGenerator<false>(node, generatorArgs)} {}
 DocumentGenerator::DocumentGenerator(const Node& node, PhaseContext& phaseContext, ActorId actorId)
-    : DocumentGenerator{node, GeneratorArgs{phaseContext.rng(actorId), actorId, counter}} {}
+    : DocumentGenerator{node, GeneratorArgs{phaseContext.rng(actorId), actorId}} {}
 DocumentGenerator::DocumentGenerator(const Node& node, ActorContext& actorContext, ActorId actorId)
-    : DocumentGenerator{node, GeneratorArgs{actorContext.rng(actorId), actorId, counter}} {}
+    : DocumentGenerator{node, GeneratorArgs{actorContext.rng(actorId), actorId}} {}
 
 
 DocumentGenerator::DocumentGenerator(DocumentGenerator&&) noexcept = default;
@@ -1203,6 +1194,5 @@ bsoncxx::document::value DocumentGenerator::operator()() {
 bsoncxx::document::value DocumentGenerator::evaluate() {
     return operator()();
 }
-
 
 }  // namespace genny
--- a/src/value_generators/test/DocGenYamlTestCaseRunner.cpp
+++ b/src/value_generators/test/DocGenYamlTestCaseRunner.cpp
@@ -65,10 +65,12 @@ public:
 
     void run() const {
         DYNAMIC_SECTION("DocGenYamlTestCaseRunner " << name()) {
+            int64_t counter=1;
+            //int64_t * pcounter = &counter;
             if (_runMode == RunMode::kExpectException) {
                 try {
                     NodeSource ns = toNode(this->_givenTemplate);
-                    genny::DocumentGenerator(ns.root(), GeneratorArgs{rng, 1});
+                    genny::DocumentGenerator(ns.root(), GeneratorArgs{rng, 1, counter});
                     FAIL("Expected exception " << this->_expectedExceptionMessage.as<std::string>()
                                                << " but none occurred");
                 } catch (const std::exception& x) {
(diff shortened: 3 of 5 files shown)


rtimmons (commented):
I think this has the right "bones," but--as you point out--the state-management (a static-allocated int64) is less than ideal.

My back-of-the-envelope design was for there to be a static-allocated `unordered_map<string, atomic<int>>` and each call to the `^Inc` generator constructor would pass in a name:

```yaml
Document:
  key: {^Inc: {Name: MyDocument}}
```

And all generators that have the same `Name` would share the same atomic.

This is marginally better than what's here but only marginally.

I think your intuition about putting this shared state into the `WorkloadContext` is corret - that's basically the "global context" of the whole runtime environment. This is not unreasonable but would require the `value_generators` module to know about `gennylib` which I've tried pretty hard to avoid to keep the complexity sane. Perhaps that's not an insane thing to do, though...

The "best" solution there off-hand--without adding cmake modules or dependencies--is to make a `HasSharedState` class or something and `WorkloadContext` can be a subclass, and the value generators can take in a ptr. I suspect there would be gotchas with the templates involved, so I'm hesitant to go too far there without doing some compiler-wrestling first.

_ANYWAY_, in my opinion, the `static map<string, atomicint>` isn't all that terrible. In reality I'd have a struct that held the atomicint as well as the step value etc just to make the bookkeeping easier. This is where I'd start and see how it actually ends up. Moving it to be stored by the `WorkloadContext` would be primarily an academic exercise since we have no real use-case for there to be multiple contexts in existence.

I'm happy to chat further or give a more "formal" review if you'd like. Otherwise let me know if any of this makes sense or if it inspires a better idea.

rtimmons:
(You had some other questions in your PR description as well, and the answers may depend on the impl - we can work through those when you've found the impl you like, but happy to go into more detail if those qs are blocking you from digging deeper.)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -736,6 +739,33 @@ class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
...
+        : _actorId{generatorArgs.actorId},
+          _counter{generatorArgs.counter},
+          _step{extract(node, "step", "^Inc").to<int64_t>()},
+          _multiplier{extract(node, "multiplier", "^Inc").to<int64_t>()}
+          {}
+
+    int64_t evaluate() override {
+//        std::lock_guard<std::mutex> lg(mtx);
+        std::cout << "Counter before " << _counter << std::endl;
+        _counter=_actorId*_multiplier + _counter + _step;
```
dalyd, line 755: The _actorId*multiplier should be constant. You could make that a member of the generator.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -736,6 +739,33 @@ class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
     const std::chrono::milliseconds _max;
 };
 
+
+class IncGenerator : public Generator<int64_t> {
+public:
+    IncGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _actorId{generatorArgs.actorId},
+          _counter{generatorArgs.counter},
```
dalyd, line 747: Rather than passing in generatorArgs.counter as a new argument, I would be tempted to take this initializer out, and on line 750 (in the body) have
`_counter = _start + _multiplier*_actorId;`
If you find you don't need multiplier or start otherwise as members, you could move all of that up into the initializer as well. 

Then in evaluate you just need 
`_counter += _step;`

src/value_generators/include/value_generators/DocumentGenerator.hpp
```diff
@@ -39,6 +39,7 @@ class InvalidValueGeneratorSyntax : public std::invalid_argument {
 struct GeneratorArgs {
     DefaultRandom& rng;
     ActorId actorId;
+    int64_t& counter;
```
dalyd, line 42: Do we really need a shared counter?

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -592,3 +592,22 @@ Tests:
     ThenReturns:
     - { "date" : { "$date" : { "$numberLong" : "1369875424299" } } }
 
+  - Name: Inc0
+    GivenTemplate:
+      int: {^Inc: {multiplier: 100, step: 1}}
+    ThenReturns:
+    - "int": { "$numberLong" : "201"}
```
dalyd, line 599: Here you should have multiple entries under ThenRetuns like
`- "int": { "$numberLong" : "201"}`
`- "int": { "$numberLong" : "202"}`
`- "int": { "$numberLong" : "203"}`

dalyd (changes requested):
I'd go with a local member in the generator as opposed to having anything shared. I put some comments on. I'll respond to Alex's questions in a separate response.

dalyd:
> Here is the first draft of the incrementing counter https://jira.mongodb.org/browse/PERF-2077 for discussion. I don't like it at all - not what it does, not code itself... Don't think that it is ready for a merge in any way. But after wasting a lot of time trying to figure out architecture / new C++ features / yaml / etc, it looks like I can utilize some help and I decided that bringing it out as a PR may be a good way to solicit some help. Here is the list of questions I have for now (and I probably didn't get to all questions yet].
> 
> 1. It looks like the current DocumentGenerator architecture doesn't allow to have a shared counter between Actors - as its instances are per Actor.  [It appears that a shared counter may be in the workload context - but it appears that it will require much more changes, including architecture.]

Yes, no current support for shared counters right now as far as I know. 

> 2. If we have it per Actor, can we have multiple parallel threads per Actor? Do we need synchronization for the counter here? [I guess for document generation we probably don't need it]

Each Actor might be instantiated multiple times. Each instantiation will have its own copy of the generator and its own ActorID. 

> 3. So, as discussed in https://jira.mongodb.org/browse/PERF-2077, it appears that we basically have only multiplier and step as parameters. Not sure how we can use 'start' here - I guess we may hardcode 1 here, but it doesn't sound exciting.

Hm, I think I give a different answer each time on start. Sorry. I think each thread will want to iniatilize its counter to some start value, which should be a function of some generic start value (1 for default probably) plus an offset determined by which thread (ActorID) it is. 

> 4. It doesn't look like the test architecture used allows to verify if incrementing actually works (in DocumentGeneratorTestCases.yml). It appears that every iteration of the loop in DocGenYamlTestCaseRunner.cpp
>    for (const auto&& nextValue : this->_thenReturns) {
>    resets the counter, so we can test only for the hardcoded initial counter value.

I think that is a bug. This should work. 

> 5. It appears that we have a problem with the assert in DocGenYamlTestCaseRunner.cpp - it does fail differing in a space (either you have it or not):
>    .../src/value_generators/test/DocGenYamlTestCaseRunner.cpp:88: FAILED:
>    REQUIRE( toString(expected.view()) == toString(actual.view()) )
>    with expansion:
>    "{ "int" : { "$numberLong" : "202" } }"
>    ==
>    "{ "int" : { "$numberLong" : "202"} }"
>    Actually it looks like it is not a new issue: https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocumentGeneratorTestCases.yml#L491-L492
> 

I ran into a similar problem and gave up at some point. Hopefully this goes away for you. 

> Any input / feedback would be much appreciated.
> 
> Thanks,
> Alex

apodelko:
Here is a very simple implementation following David's suggestions [at least as I got them]. 

It includes a workaround as described in [HELP-21664](https://jira.mongodb.org/projects/HELP/queues/issue/HELP-21664) - probably it would make sense to update libbson and remove that workaround before merging.

Still want to get this simple implementation for review to see if it is acceptable.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -736,6 +739,31 @@ class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
...
+        _counter += _step;
+        return inc_value;
+    }
+
+private:
+    int64_t _actorId;
+    int64_t _counter;
+    int64_t _step;
+    int64_t _multiplier;
+    int64_t _start;
```
rtimmons, line 763: looks like you don't actually need to store `_actorId`, `_multiplier`, or `_start`, you just use those to compute the initial value of `_counter`.

Also you generally want to initialize things in the same order as they are declared in the struct/class (`: _actorId{...}, _counter{...}` etc). We're loose with this though.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -736,6 +739,31 @@ class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
...
 };
 
+
+class IncGenerator : public Generator<int64_t> {
+public:
+    IncGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _actorId{generatorArgs.actorId},
+          _step{extract(node, "step", "^Inc").to<int64_t>()},
+          _multiplier{extract(node, "multiplier", "^Inc").to<int64_t>()},
+          _start{extract(node, "start", "^Inc").to<int64_t>()}
```
rtimmons, line 749: consider making this optional and defaulting to 1:

```c++
_start{node["^Inc"].maybe<int64_t>().value_or(1)
```

src/value_generators/test/DocGenYamlTestCaseRunner.cpp
```diff
@@ -83,7 +83,8 @@ class ValGenTestCase {
             for (const auto&& nextValue : this->_thenReturns) {
                 auto expected = testing::toDocumentBson(nextValue);
                 auto actual = docGen();
-                REQUIRE(toString(expected.view()) == toString(actual.view()));
+                auto expectedFix = bsoncxx::from_json(toString(expected.view()));
+                REQUIRE(toString(expectedFix.view()) == toString(actual.view()));
```
rtimmons, line 87: thanks for making this fix and working with drivers to uncover the root of this issue. Can you add a TODO comment with a link to the TIG/PERF ticket to remove once the bson bug is fixed?

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -592,3 +592,18 @@ Tests:
...
+    - "int": { "$numberLong" : "202"}
+    - "int": { "$numberLong" : "203"}
+
+  - Name: IncrementWithStep
+    GivenTemplate:
+      int: {^Inc: {start: 100, multiplier: 200, step: 2}}
+    ThenReturns:
+    - "int" : { "$numberLong" :"500"}
+    - "int" : { "$numberLong" :"502"}
+    - "int" : { "$numberLong" :"504"}
```
rtimmons, line 609: I think there is one or two other test-cases that are commented out because of that bson bug and would hopefully work now - can you try uncommenting them?

(Unfortunately this yaml-based testing infrastructure relies on the RNG returning values in the same order, so adding test-cases in the middle of the file involves changing all the expected values that occur after those changes. If this is too tedious lemme know I can pull your branch and help with that part.)

src/value_generators/test/DocGenYamlTestCaseRunner.cpp
dalyd: Nice!

dalyd: Wait, does this work? I thought you would need the extra round trip on the "actual" value, not the expected value.

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: I was going to ask the same thing. 

Question for Alex and Ryan. Should we reseed the RNG between test cases? That would allow us to call it multiple times to get different values, but we wouldn't carry state from one test to the next. It would probably require updating all of the expected values though :(

dalyd (commented):
I like it! Nice work.

rtimmons:
> Should we reseed the RNG between test cases?

Strong yes. I tried for like 3 minutes to figure out how to do that when initially creating this yaml runner thing but shrugged and moved on. If someone can figure out how to make it re-seed between tests I have no problem updating all the expected value (I got kinda good at it when initially writing these - not something I'm super proud of tbh).

apodelko:
@rtimmons Any idea why [Ubuntu compilation fails](https://spruce.mongodb.com/version/6011e974d6d80a793f834d4e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=ubuntu1804)? It looks like it is failing for a while and it doesn't look like it is actually related to the PR changes.

dalyd:
> > Should we reseed the RNG between test cases?
> 
> Strong yes. I tried for like 3 minutes to figure out how to do that when initially creating this yaml runner thing but shrugged and moved on. If someone can figure out how to make it re-seed between tests I have no problem updating all the expected value (I got kinda good at it when initially writing these - not something I'm super proud of tbh).

I think we can just copy the rng before use and then copy back

> Note that all pseudo-random number generators described below are CopyConstructible and Assignable. Copying or assigning a generator will copy all its internal state, so the original and the copy will generate the identical sequence of random numbers. Often, such behavior is not wanted. In particular, beware of the algorithms from the standard library such as std::generate. They take a functor argument by value, thereby invoking the copy constructor when called.

https://www.boost.org/doc/libs/1_66_0/doc/html/boost_random/reference.html#boost_random.reference.generators

Hm, you've deleted copy and assignment from the wrapper though. 

Alternatively all the generators support a seed function. If you exposed that in the defaulRandom class we could reset the seed each time in the test class. Maybe reset on this line: https://github.com/mongodb/genny/blob/8d2e2c960a3f39cc2482bc4a7299eec3a120d1e0/src/value_generators/test/DocGenYamlTestCaseRunner.cpp#L82

Or we could just pass in a new generator each time here https://github.com/mongodb/genny/blob/8d2e2c960a3f39cc2482bc4a7299eec3a120d1e0/src/value_generators/test/DocGenYamlTestCaseRunner.cpp#L71

rtimmons:
> Or we could just pass in a new generator each time here 

Hey that's a good idea. I see no reason why that wouldn't. I'm surprised I didn't think of that (or if I did try it I certainly don't remember why I didn't go with it).

apodelko:
Submitted a new version with most suggestions implemented.
As re-seeding doesn't actually related to that ticket, split it into a separate task https://jira.mongodb.org/browse/PERF-2118

dalyd:
@apodelko I got a bit confused. I'm going to take a look, but try not to force push on PRs. You can just do a merge commit and push the merge commit and github can do stuff with that. It's also okay to have a lot of commits in a PR.

apodelko:
> @apodelko I got a bit confused. I'm going to take a look, but try not to force push on PRs. You can just do a merge commit and push the merge commit and github can do stuff with that. It's also okay to have a lot of commits in a PR.

As discussed, I did rebase and [probably] it didn't allow me to do regular push. Reading through some internal git guidelines, my understanding was that it is recommended to do rebase before submitting any changes. But it looks like it is not a good approach for updating a PR.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -736,6 +739,25 @@ class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
     const std::chrono::milliseconds _max;
 };
 
+
+class IncGenerator : public Generator<int64_t> {
+public:
+    IncGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _step{node["step"].maybe<int64_t>().value_or(1)} 
+        {_counter = node["start"].maybe<int64_t>().value_or(1) + generatorArgs.actorId * node["multiplier"].maybe<int64_t>().value_or(1);}
```
rtimmons, line 747: please ensure you run `clang-format` prior to merging

```sh
$ brew install clang-format
$ clang-format --version
#> clang-format version 11.0.0 # must be 11.0.0

clang-format -i ./src/**/*.cpp ./src/**/*.hpp
```

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -488,16 +488,18 @@ Tests:
...
       string: {^ActorIdString: {}}
     ThenReturns:
-    #    - { "int" : { "$numberLong" : "2000000000"}, "string" : "2000000000" }
     - "string": "2"
 
+  - Name: Actor ID
+    GivenTemplate:
+      int: {^ActorId: {}}
+    ThenReturns:
+    - { "int" : { "$numberLong" : "2000000000"}, "string" : "2000000000" }
```
rtimmons, line 501: I don't think this works - there's no `string` key in the `GivenTemplate`

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: > Should we reseed the RNG between test cases?

[PERF-2118](https://jira.mongodb.org/browse/PERF-2118) is to do this, but it's actually only a 2-line change - just move this line:

https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocGenYamlTestCaseRunner.cpp#L22

down to the top of the `DYNAMIC SECTION`:

https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocGenYamlTestCaseRunner.cpp#L68

This will of course break all the existing `ThenReturns` values though, as David noted.

rtimmons (approved):
LGTM assuming a successful CI run (excluding ubuntu18 or whatever it is that has a broken compile task right now)

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: I think the test case got split into two distinct test cases here, when the original had string and int.

dalyd (approved):
I think Ryan's comment on the ActorID test case is correct, but otherwise LGTM and nice work. It looks like the base ubuntu build is broken, which is unfortunate. I don't know if that changes how to merge the change in.

src/value_generators/src/DocumentGenerator.cpp
apodelko: Ran. Absolutely no output. Can I assume that everything is fine?

apodelko:
Yes, I split that ActorId case into two - leaving original string case and copying the commented out int case into a separate test case. Yep, probably should look at it more attentively - but as it worked, I didn't... Interesting why it worked - in the latest upload I just dropped the string part and it is working... Maybe it just somehow ignored the string part?

How am I supposed to merge it? 'evergreen merge' ? Will it merge it with Ununtu compile not working?

apodelko:
evergreen merge

PR #410, opened 2021-01-14
PERF-2087: Add workload for reshardCollection with reads and writes

visemet:
Also adds support for "ClientName" in RunCommand actor for running operations through a connection pool other than the "Default" one.

Sys-perf patch build: https://evergreen.mongodb.com/task/sys_perf_linux_shard_lite_reshard_collection_patch_4b6a89a429c39a4f0848e7723f7bda9d22576cf9_5fff91fbd6d80a156b0ca5fe_21_01_14_00_37_52/1

Sys-perf patch build <span>#</span>2: https://evergreen.mongodb.com/task/sys_perf_linux_shard_lite_reshard_collection_patch_cc8671bcf846a6aeda28a5cf81aea750c2fca4bd_60073a961e2d17558c27f16a_21_01_19_20_04_06/0

### DIFF
--- a/src/cast_core/src/RunCommand.cpp
+++ b/src/cast_core/src/RunCommand.cpp
@@ -257,7 +257,8 @@ void actor::RunCommand::run() {
 
 actor::RunCommand::RunCommand(ActorContext& context)
     : Actor(context),
-      _client{std::move(context.client())},
+      _client{std::move(
+          context.client(context.get("ClientName").maybe<std::string>().value_or("Default")))},
       _loop{context, context, _client, RunCommand::id()} {}
 
 namespace {
--- a/src/workloads/sharding/ReshardCollection.yml
+++ b/src/workloads/sharding/ReshardCollection.yml
@@ -1,0 +1,154 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+Description: |
+  Runs the reshardCollection command while read and write operations are active on the collection
+  being resharded.
+
+  The workload consists of 5 phases:
+    1. Creating an empty sharded collection distributed across all shards in the cluster.
+    2. Populating the sharded collection with data.
+    3. Running read and write operations on the collection before it is resharded.
+    4. Running read and write operations on the collection while it is being resharded.
+    5. Running read and write operations on the collection after it has been resharded.
+
+GlobalDefaults:
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+  # Collection0 is the default collection populated by the MonotonicLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentSize: &DocumentSize 10000  # = 10kB
+  DocumentSize80Pct: &DocumentSize80Pct 8000  # = 8kB
+  DocumentCount: &DocumentCount 100000  # for an approximate total of 1GB
+
+  ShardKeyValueMin: &ShardKeyValueMin 1
+  ShardKeyValueMax: &ShardKeyValueMax 100
+
+  ReadWriteOperations: &ReadWriteOperations
+  - OperationName: findOne
+    OperationCommand:
+      Filter: {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
+  - OperationName: updateOne
+    OperationCommand:
+      Filter: {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
+      Update: {$inc: {counter: 1}}
+      # Write operations can fail with a ShardInvalidatedForTargeting error response due to the
+      # routing information for the temporary resharding collection not being available on the
+      # donor shard. TODO: Investigate why mongos isn't automatically retrying on it.
+      ThrowOnFailure: false
+
+Clients:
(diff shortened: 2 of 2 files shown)


src/workloads/sharding/ReshardCollection.yml
```diff
@@ -0,0 +1,154 @@
...
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+  # Collection0 is the default collection populated by the MonotonicLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentSize: &DocumentSize 10000  # = 10kB
+  DocumentSize80Pct: &DocumentSize80Pct 8000  # = 8kB
+  DocumentCount: &DocumentCount 100000  # for an approximate total of 1GB
```
visemet, line 24: [note] 1GB of data has the resharding operation taking about 1 minute. I had been mainly interested in getting an initial version of a resharding workload running in Evergreen and then making adjustments to it as the feature continues to be developed.

Garaudy is interested in have 20GB per shard (4000000 = 2 shards * 20 * 100000) but I held off on using larger numbers for now. `key: {oldKey: hashed, newKey: 1}` isn't having data be redistributed (the data does get copied and writes are amplified) but I couldn't say the performance characteristics will match what happens after we'd switch to `key: {newKey: 1}`.

src/workloads/sharding/ReshardCollection.yml
```diff
@@ -0,0 +1,154 @@
...
+  Threads: 1
+  ClientName: ReshardCollection
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    # Until SERVER-53373 is implemented, donor shards aren't guaranteed to preserve enough history
```
visemet, line 106: [note] Assuming this seems is acceptable to the reviewers, I'll put up a PR for https://github.com/visemet/dsi/commit/d3b93efcd3a2b0f8a3f62dc0f46adb44f0ab622c too.

dalyd (approved):
Commit 1 LGTM. I have a few questions or suggestions, but overall very nice work. 

Top level question: do you want it running current in regression, or do you want to investigate and run by patch build until you get more final configurations? Either is fine by me. I wanted to ask as this is ongoing work and it sounds like you are still learning about the behavior.

src/workloads/sharding/ReshardCollection.yml
```diff
@@ -0,0 +1,154 @@
...
+        # the shards.
+        key: {oldKey: hashed}
+        numInitialChunks: 60
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: LoadInitialData
+  Type: MonotonicLoader
```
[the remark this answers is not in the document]
visemet: Yeah, PERF-2077 looks related. I believe the MonotonicLoader is meant to be a temporary thing.

https://github.com/mongodb/genny/blob/6e4359d0ab73c170523dbda8562fa1d389604059/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp#L38-L39

Maybe to give other folks more work (not me!) - removing MonotonicLoader seems like a good acceptance criterion for PERF-2077.

src/workloads/sharding/ReshardCollection.yml
visemet: > If it could be done at the beginning of the test we could put in a js check on version before enabling or disabling.

We set the WTPreserveSnapshotHistoryIndefinitely failpoint at startup in our correctness testing.

https://github.com/mongodb/mongo/blob/f3b3164a196fa930b17ac2f703db433d8fc79f36/jstests/sharding/libs/resharding_test_fixture.js#L77-L79

We could do something similar for our performance testing too. The reason I had chosen not to is because we'd be pinning the oldest\_timestamp from before the 1GB of data was inserted. If that 1GB of data becomes 40GB of data then I think we're going to see durable history stressed in a way that wouldn't be representative of the actual behavior during resharding.

> Can we version this somehow? Once we disable this we won't be able to run this test against versions before SERVER-53373 goes in.

To save myself some time, I had a goal in my initial cut of this workload to not write a new actor in C++. I'm willing to give up successfully running the ReshardCollection.yml workload on earlier mongodb/mongo commits.

src/cast_core/src/RunCommand.cpp
```diff
@@ -257,7 +257,8 @@ void actor::RunCommand::run() {
 
 actor::RunCommand::RunCommand(ActorContext& context)
     : Actor(context),
-      _client{std::move(context.client())},
+      _client{std::move(
+          context.client(context.get("ClientName").maybe<std::string>().value_or("Default")))},
```
guoyr, line 261: LGTM.

(Note for future reference: the CRUD and RunCommand actors are duplicating the default arg of "Default" here. It's  unlikely for other existing actors to require a configurable client name; if new use cases come up, we can switch the arg to `std::optional` to replace the `value_or()` part.

guoyr (approved):
Changes to RunCommand.cpp LGTM

visemet:
evergreen merge

visemet:
Going to manually merge due to Ubuntu compile failing from not having libmongocrypt installed.

PR #411, opened 2021-01-15
TIG-2826 Metrics are not being reported for CrudActor in ReshardCollection.yml workload

aggrand:
Basically stops extra streams from being created, details in ticket.


guoyr (approved):
Wow nice find! Glad to see the data is still being recorded 😅

aggrand:
This seems to be failing the Ubuntu compile, but I retried a couple of waterfall ubuntu compile tasks and they are failing now too, so something outside this repo is breaking that. I'll squash-merge this and file a separate ticket for the failure.

PR #412, opened 2021-01-19
Modernize genny python and paths

rtimmons:
This is bigger than I was hoping, but I think it's straightforward enough. Getting this out there for some early feedback.

Related:

- https://mongodbcr.appspot.com/749470020/
- https://github.com/10gen/dsi/pull/826 
- https://github.com/mongodb/genny/pull/412

A few things going on here:

- all genny invocations go through `run-genny` which is basically a straight lift of `run-dsi`.
- The existing `lamp` commands have been folded into the python that `run-genny` invokes. That now uses click for arg parsing.
- All of the shell scripts have been ported to python.
- **Note**: A lot of the python diff is just moving things around. The only new python is `cli.py`, `resmoke.py`, and `dry_run.py`, 
- The default `Metrics.Path` is now `build/CedarMetrics`

Simply changing the metrics path to `./build/CedarMetrics` turned rather unpleasant because we run genny either from the genny repo root or from pwd having repo root of `./src/genny`. With everything going through `run-genny` and being in python we can now easily compute paths and know exactly where things are.

A few minor things:

- Moved logging to structlog
- Moved testing to pytest
- The evergreen config no longer uploads to S3 (it was never used)
- The evergreen config now uses `run-genny` via `subprocess.exec`
- The mongo repo (used by resmoke tests) is now an evergreen module

Still remaining before merge:

- Ensure this works in a sane way for running from either the genny repo root or from the new evergreen conventional location which has `./src/genny` as the repo root.
- Update paths in DSI (mostly done as I did it in a branch in parallel with these changes)
- Update docs
- Add pytests for some of the nontrivial logic added (although there is very little real logic)
- Of course I'll ensure CI tests pass - going to be a bit annoying with broken compile but I'll manage

If this PR feels too big, I can split it into:

- kill s3 in evergreen
- mongo is a module
- run everything through run-genny
- moving to structlog and pytest
- moving shell to python

But that would be a fair amount of coordination with the DSI repo and doesn't seem worth it. Please push back if you disagree.


richardsamuels (approved):
I don't know if you have the patience to start mypy-ing this, or if it's even valuable. Your call here.

Please ticket the TODOs in cli.py.

Also where is the resmoke.py that you mentioned earlier?

run-genny
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
```
aggrand, line 1: Just nosing in with a thought, feel free to ignore, but one of the original goals behind the `./scripts/genny` script in the long-running workloads project was to make the fact that it's a shell script running python running a binary as invisible as possible, since most people just want to run "genny".

Would it make sense to preserve that illusion and rename this to `genny`, keep passing top-level commands through to the binary, and have some sort of `lamp` subcommand that exposes the rest of the script? Since the bread-and-butter of invoking this would be to run the genny binary anyway?

Totally optional opinion though, feel free to ignore.

richardsamuels:
When you're ready to sys-perf patch this, please include tpcc given the recent issue with expanded metrics and us moving directories around.

run-genny
rtimmons: Good idea. I'm going to revamp this a bit to be `./genny` that forwards args to `genny_core` except for subcommands that it knows about (or something like that).

PR #414, opened 2021-01-21
TIG-1201 Genny: Create Quiesce System Utility Method

aggrand:
Implements topology discovery, and the quiesce method which waits for the oplog to catch up, fsyncs, and waits for collection to drop.

Also took a stab at formalizing the topology traversal code. I feel like in e.g. resmoke we end up with every kitchen-sink method attached to the cluster representation, whereas the shell seems to have ad-hoc traversal everywhere, so I tried to get around both of those.

The actual quiescing followed the JS implementation closely, so I didn't add automated testing for that part as discussed in the ticket and templatizing on both `client` and `database` and `collection` seems tricky, but I did check locally that it works for different clusters. It could be possible to wrap some of those interactions in `DBService` methods to better test it though, if you think that'd be worth doing. Or, after TIG-2823, doing a resmoke integration test.


src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+            mongocxx::client client(mongocxx::uri(desc.mongodUri));
+            auto admin = client["admin"];
+            admin.run_command(make_document(kvp("fsync", 1)));
+        }
+    };
+    DoFsyncVisitor v;
+    topology.accept(v);
+}
+
+bool checkForDroppedCollections(mongocxx::database db) {
```
aggrand, line 97: This logic is different for 4.0 and earlier, but my understanding is that we use legacy genny on the legacy DSI branches, so I assumed adding a version check here would be unnecessary?

src/gennylib/include/gennylib/service.hpp
```diff
@@ -0,0 +1,71 @@
+// Copyright 2021-present MongoDB Inc.
```
aggrand, line 1: I added this file because it made unit testing `Topology` easier. Another option would be to templatize it but I wanted to avoid having all the details of discovery in the header file.

src/cast_core/src/QuiesceActor.cpp
```diff
@@ -0,0 +1,72 @@
...
+namespace genny::actor {
+
+struct QuiesceActor::PhaseConfig {
+    PhaseConfig(PhaseContext& context) {
+        auto actorDb = context.actor()["Database"].maybe<std::string>();
+        if (!actorDb) {
+            BOOST_THROW_EXCEPTION(
+                InvalidConfigurationException("Must give Database in Actor block."));
+        }
+        dbName = *actorDb;
```
rtimmons, line 42: nit but you don't need this error-checking - `Node` does it for you when you call `.to<>()`:

```
PhaseConfig(ctx)
: dbName{context.actor()["Database"].to<string>()} {}
```

will barf with a helpful error message if not present.

src/cast_core/src/QuiesceActor.cpp
```diff
@@ -0,0 +1,72 @@
...
+    std::string dbName;
+};
+
+
+void QuiesceActor::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto quiesceContext = _totalQuiesces.start();
+            BOOST_LOG_TRIVIAL(debug) << "QuiesceActor quiescing cluster.";
+            if (quiesce(_client, config->dbName)) {
```
rtimmons, line 54: note to circle back here - it seems odd to require a db name to quiesce teh whole cluster

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <boost/log/trivial.hpp>
+
+#include <gennylib/topology.hpp>
+
+namespace genny {
+
+const int DROPPED_COLLECTION_RETRIES = 1000;
+
+bool waitOplog(Topology& topology) {
```
rtimmons, line 36: 1. see if it makes sense to put most/all of these free functions into `topology.hpp`?
2. these aren't templated so you probably only need to decl them and then define them in the cpp (either topology or quiesce cpp). Not a huge deal but potentially speeds up some compile and link etc

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+            auto admin = client["admin"];
+
+            // Assert that all nodes are in reasonable states.
+            auto replSetRes = admin.run_command(make_document(kvp("replSetGetStatus", 1)));
+            auto members = replSetRes.view()["members"];
+            if (members && members.type() == bsoncxx::type::k_array) {
+                bsoncxx::array::view members_view = members.get_array();
+                for (auto member : members_view) {
+                    std::string state(member["stateStr"].get_utf8().value);
+                    if (state != "PRIMARY" && state != "SECONDARY" && state != "ARBITER") {
```
rtimmons, line 52: what other states could we be? standalone?

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            // Do flush
+            auto collection = admin["wait_oplog"];
+            mongocxx::write_concern wc;
+            wc.nodes(desc.nodes.size());
+            wc.journal(true);
+            mongocxx::options::insert opts;
+            opts.write_concern(wc);
+            auto insertRes = collection.insert_one(make_document(kvp("x", "flush")), opts);
```
rtimmons, line 69: I suspect you're just copying the js impl here and that there's a reason for the `{x:flush}` doc and whatnot, but this seems like voodoo. Can you either add a comment about what this is doing from a conceptual level or add a permalink to the js impl?

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+        
+            success_acc = success_acc && insertRes && insertRes->result().inserted_count() == 1;
+        }
+        bool success_acc = true;
```
rtimmons, line 73: 💅🏽 
1. `camelCase`
2. Please put this in `private` and with leading underscore
3. I don't grok why this is stateful or what this `success_acc` actually means
4. If we run this member-function a second time do we need to re-run the state-check loop above again? If so, do we want to reset `success_acc` to false? (Probably a dumb question given I don't quite grok what this var is doing)

This took me a while to parse as a member-variable 😸

src/gennylib/include/gennylib/quiesce.hpp
rtimmons: The `legacy` branch is going away so all branches will be on genny master. It's not unreasonable to say that workloads that use this can't be run on 4.0. If anyone balks we can perhaps do runtime version checks.

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+            using bsoncxx::builder::basic::kvp;
+            using bsoncxx::builder::basic::make_document;
+
+            mongocxx::client client(mongocxx::uri(desc.mongodUri));
+            auto admin = client["admin"];
+            admin.run_command(make_document(kvp("fsync", 1)));
+        }
+    };
+    DoFsyncVisitor v;
+    topology.accept(v);
```
rtimmons, line 94: note to self that `v` is stack-allocated: ensure pass-by-lvalue isn't allowed

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+    using bsoncxx::builder::basic::kvp;
+    using bsoncxx::builder::basic::make_document;
+
+    auto res = db.run_command(make_document(kvp("serverStatus", 1)));
+    auto idents = res.view()["storageEngine"]["dropPendingIdents"];
+
+    return idents.get_int64() != 0;
+}
+
+bool checkForDroppedCollectionsTestDB(mongocxx::pool::entry& client, std::string dbName) {
```
rtimmons, line 107: I'm not sure if you need any of these helpers to be "visible" once you move the impls to cpp - perhaps just need to decl the `quiesce` free-function and the rest can just live in the anonymous namespace in the cpp.

src/gennylib/include/gennylib/quiesce.hpp
```diff
@@ -0,0 +1,138 @@
...
+    auto db = client->database(dbName);
+    int retries = 0;
+    while (checkForDroppedCollections(db) && retries < DROPPED_COLLECTION_RETRIES) {
+        BOOST_LOG_TRIVIAL(debug) << "Sleeping 1 second while waiting for collection to finish dropping.";
+        retries++;
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
rtimmons, line 113: 1. I'm super skeptical of how this will play with `PhaseLoop`. I'm holding my breath for now but note to self to come back here.
2. separate thought: do we want to ensure there aren't concurrent invocations of this?

src/gennylib/include/gennylib/service.hpp
rtimmons: Note to self to come back here.

I think this is a good design. Genny doesn't use a lot of virtuals in its core code but virtuals are a great strategy to ease testing in utilities.

But `service` is a bit of a smurf term. Perhaps it makes sense to rename these and make them have a more concrete connection (both in name and code location) to `topology.hpp`.

src/gennylib/include/gennylib/service.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+    ServiceUri uri() { return _client.uri().to_string(); }
+
+    bsoncxx::document::value runAdminCommand(std::string command) { 
+        using bsoncxx::builder::basic::kvp;
+        using bsoncxx::builder::basic::make_document;
+
+        auto admin = _client.database("admin");
+        return admin.run_command(make_document(kvp(command, 1)));
+    }
```
rtimmons, line 59: may as well make everything virtual and make the impl in the cpp

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+#include <variant>
+#include <mutex>
+#include <sstream>
+
+#include <gennylib/v1/PoolFactory.hpp>
+#include <gennylib/service.hpp>
+
+namespace genny {
+
+class AbstractTopologyDescription;
```
rtimmons, line 29: super nit: Just call this `Topology` - the fact that it's abstract is its business, and "description" doesn't add much to the name. Push back if you disagree no biggie.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+    virtual void visitTopLevelPost(const AbstractTopologyDescription&) {}
+
+    virtual void visitMongodDescription(const MongodDescription&) {}
+    virtual void visitMongosDescription(const MongosDescription&) {}
+
+    virtual void visitReplSetDescriptionPre(const ReplSetDescription&) {}
+    virtual void visitReplSetDescriptionPost(const ReplSetDescription&) {}
+
+    virtual void visitShardedDescriptionPre(const ShardedDescription&) {}
+    virtual void visitShardedDescriptionPost(const ShardedDescription&) {}
```
rtimmons, line 56: Consider names

1. `on{Before,After}Topology`
2. `on{Before,After}{Mongo{d,s},Shard}`
3. `onBetween{Mongod,Shard}s`

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    // Called before/after/between visiting shards.
+    virtual void visitShardsPre(const ShardedDescription&) {}
+    virtual void visitShardsPost(const ShardedDescription&) {}
+    virtual void visitShardsBetween(const ShardedDescription&) {}
+
+    // Called before/after/between visiting mongoses.
+    virtual void visitMongosesPre(const ShardedDescription&) {}
+    virtual void visitMongosesPost(const ShardedDescription&) {}
+    virtual void visitMongosesBetween(const ShardedDescription&) {}
```
rtimmons, line 72: 1. Why not pure-virtual (`= 0`)?
2. If most visitors won't need them maybe just don't include them for now?

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    virtual ~TopologyVisitor() {}
```
rtimmons, line 74: let's make this `=0` to make this a pure virtual class.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+};
+
+// Be careful changing the traversal order of the cluster; visitors may depend on it.
+struct AbstractTopologyDescription {
+    virtual void accept(TopologyVisitor&) = 0;
+    virtual ~AbstractTopologyDescription() {};
```
rtimmons, line 80: sane here - `=0`.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+    void visitMongosesPre(const ShardedDescription&) { result << " mongoses: ["; }
+    void visitMongosesBetween(const ShardedDescription&) { result << ", "; }
+    void visitMongosesPost(const ShardedDescription&) { result << "]"; }
+
+    std::string str() { return result.str(); }
+
+private:
+    std::stringstream result;
+};
```
rtimmons, line 214: impls can be in .cpp (I realize it's annoying so my apologies for asking for banal code ceremony)

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+            v.visitTopLevelPre(*_topology);
+            _topology->accept(v); 
+            v.visitTopLevelPost(*_topology);
+        }
+    }
+
+    /**
+     * Update the Topology's view of the cluster.
+     */
+    void update(DBService& service);
```
rtimmons, line 168: likely wants to be private

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+
+private:
+    v1::PoolFactory _factory;
+    std::string nameToUri(const std::string& name);
+    void getDataMemberConnectionStrings(DBService& service);
+    void findConnectedNodesViaMongos(DBService& service);
+    std::unique_ptr<AbstractTopologyDescription> _topology = nullptr;
```
rtimmons, line 175: erm. seems odd for `class Topology` to have a `_topology` member-var.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,219 @@
...
+};
+
+class ToJsonVisitor : public TopologyVisitor {
```
rtimmons, line 178: hey this is neat

src/gennylib/src/topology.cpp
```diff
@@ -0,0 +1,135 @@
...
+    auto strippedName = name.substr(name.find('/') + 1, std::string::npos);
+    nameSet.insert(strippedName);
+    _factory.overrideHosts(nameSet);
+    return _factory.makeUri();
+}
+
+void Topology::getDataMemberConnectionStrings(DBService& service) {
+    auto res = service.runAdminCommand("isMaster");
+    if (!res.view()["setName"]) {
+        std::unique_ptr<MongodDescription> desc = std::make_unique<MongodDescription>();
```
rtimmons, line 37: maybe make the uri a ctor arg of the `MongoDescription`? (there are a few other things here that maybe want to be pushed into impl details (perhaps in ctors) of the `*Description` classes. Not sure of implications on testing but may actually make it easier so not a strong opinion here yet.

src/gennylib/src/topology.cpp
```diff
@@ -0,0 +1,135 @@
...
+    if (passives && passives.type() == bsoncxx::type::k_array) {
+        bsoncxx::array::view passives_view = passives.get_array();
+        for (auto member : passives_view) {
+            MongodDescription memberDesc;
+            memberDesc.mongodUri = std::string(member.get_utf8().value);
+            desc->nodes.push_back(memberDesc);
+        }
+    }
+
+    this->_topology.reset(desc.release());
```
rtimmons, line 70: I think this would be more testable if you had a `MongodDescription` ctor that took in a `view_or_value` (which you can trivially construct without live connections to a db).

src/gennylib/src/topology.cpp
```diff
@@ -0,0 +1,135 @@
...
+std::string Topology::nameToUri(const std::string& name) {
+    std::set<std::string> nameSet;
+    // Hostnames returned from some commands have the form configRepl/hostname:port
+    auto strippedName = name.substr(name.find('/') + 1, std::string::npos);
+    nameSet.insert(strippedName);
+    _factory.overrideHosts(nameSet);
+    return _factory.makeUri();
+}
+
+void Topology::getDataMemberConnectionStrings(DBService& service) {
```
rtimmons, line 34: odd to have a `void` `get` function - maybe this wants to be just a helper/free function or maybe `s/get/compute/`.

rtimmons (commented):
Nice starting point and clean code. I don't know the js impl well (er..even at all) so I'm going to trust your faithfulness to that impl.

I didn't quite finish yet - just started poking at the cpp files and tests - but putting what I have so far. Take a quick look for now and let me know if I've gone down the wrong path on what to look at. Otherwise I'll poke more at the impl this afternoon.

src/cast_core/src/QuiesceActor.cpp
aggrand: This is because my understanding is the collection drop checking is db-specific. At least in the 4.0-and-earlier implementation in the js it was. The new implementation uses `serverStatus` which [says](https://docs.mongodb.com/manual/reference/command/serverStatus/) it provides an overview of the db's state so I assume this added field is also db-specific.

src/gennylib/include/gennylib/quiesce.hpp
aggrand: There are a few [exotic](https://docs.mongodb.com/manual/reference/replica-states/) states if the replica set isn't in a stable position. This check was just copied from the js though we can delete if we don't think genny will ever talk to a repl set in those states.

src/gennylib/include/gennylib/quiesce.hpp
aggrand: Yeah will do! The js code isn't explained either but my understanding is to force the oplog to flush by doing a write with `w: all` and `j: true` so when that write is finished you know everything before it is replicated.

There's also [awaitReplication()](https://github.com/mongodb/mongo/blob/r4.1.4/src/mongo/shell/replsettest.js#L1354) in the shell which I think does something similar but its implementation is _way_ more complicated.

src/gennylib/include/gennylib/quiesce.hpp
aggrand: Will make those changes. It's stateful because in a sharded cluster `visitReplSetPre` will be called in the traversal on each shard, so we need to only return a success if the op succeeded on every shard.

src/gennylib/include/gennylib/topology.hpp
aggrand: 1. The visit methods can't be pure-virtual because they're always called in the topology description traversal, even as a nop. The idea is that subclass visitors only need to override the visit methods they care about, instead of overriding to add a bunch of nops.
2. The `JsonVisitor` needs them. That visitor isn't 100% necessary but it was useful for testing both the discovery and that traversal works correctly.

src/gennylib/include/gennylib/topology.hpp
aggrand: I thought pure virtual destructors were bad practice? Because the linker requires a definition for the destructor whether it's pure virtual or not, since every destructor call also calls the parent's destructor, its parent's, etc so a definition would need to be added outside the class declaration and having at least one function be pure virtual is all that's needed to make the class an abstract class?

src/gennylib/include/gennylib/quiesce.hpp
aggrand: I made those changes but as a more general question, do we care a lot about those standards (private, etc) for function-level classes like this? This object is sort of serving as a lambda

src/gennylib/include/gennylib/quiesce.hpp
aggrand: Yeah moved!

src/gennylib/include/gennylib/service.hpp
aggrand: I renamed it to `Connection`. Part of why I put it in a separate file is I thought it might be useful elsewhere in other places, the only other way we have to test driver-level things is with the resmoke integration and APM stuff, which might be more involved. Let me know though if we should just cross that bridge when we get there and I'll put it in the `topology.hpp` file.

src/gennylib/src/topology.cpp
aggrand: I tried this, but unfortunately a bson `element` doesn't seem to be a `view`, so couldn't be passed in, unless I'm overlooking something. I tried doing two different constructors, one for `element` and one for `std::string`, but it turns out that the document `element` and the `array::element` are different. They share a private base class that I can't seem to figure out how to convince the compiler to let me `dynamic_cast` to.

I figured having three different implementations of that constructor might not be an improvement over the existing situation, but let me know if that'd still be better!

src/cast_core/src/QuiesceActor.cpp
aggrand: An alternative might be to discover all the dbs and then wait for drops on them?

src/gennylib/include/gennylib/quiesce.hpp
aggrand: Added a locking mechanism to `quiesce.cpp` for the latter

rtimmons (approved):
Approved modulo some moving of things around and doing a bit of syntax 💅🏽 . Please push back wherever you disagree.

1. I looked at the actual quiesce business-logic and it seems reasonable, but please do put a link somewhere to the js impl for future maintainers. 
2. On a similar, note, consider asking Max H (`@visemet`) if he wouldn't mind taking a look at the quiesce cpp. He may also have good insights on the other C++isms, but he has a pretty good understanding of the quiesce logic iirc. Getting a quick 👍🏽 from him--if he has time--would be great.
3. Let's consider marking this as "beta" for now, even if just in a comment. I have very little doubt that this is correct (it's certainly well-designed and has reasonable testing), but I'd like us to know about uses of this so we can keep an eye on them. We're quite likely to be asked to field BFs if there is any suspicion of the quiesce logic making things go pear-shaped, so if we can proactively understand where it's being used we can double-check our impl and their usage is correct.

src/gennylib/src/quiesce.cpp
```diff
@@ -0,0 +1,147 @@
...
+    if (quiesceLock.try_lock()) {
+        Topology topology(*client);
+        bool waitOplogSuccess = waitOplog(topology);
+        doFsync(topology);
+        bool checkDroppedCollectionSuccess = checkForDroppedCollectionsTestDB(client, dbName);
+        success = waitOplogSuccess && checkDroppedCollectionSuccess;
+        quiesceLock.unlock();
+    } else {
+        quiesceLock.lock();
+        quiesceLock.unlock();
```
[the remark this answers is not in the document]
aggrand: I wasn't too sure what we wanted the concurrent behavior to be here. Really I think it only makes sense to ever have one, single-threaded actor do these quiesces. But if this function was ever incorporated into a larger actor or something, I think we want:

1. Only one thread does the quiesce.
2. The function only exits successfully when a quiesce has just occurred.
3. A thread won't do a redundant quiesce.

So this implementation lets one thread quiesce, and if other threads enter the function they fail to get the lock immediately, so block waiting for the lock. When they get it, they know the quiesce just happened so they can unlock and leave.

Something similar could be done with condition variables, but we'd have to deal with spurious wakeups and such so this seemed simpler and comparably efficient.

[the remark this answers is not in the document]
aggrand: I'll add a comment explaining this.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,211 @@
...
+        v.onAfterReplSet(*this); 
+    }
+};
+
+struct ShardedDescription : public TopologyDescription {
+    ReplSetDescription configsvr;
+    std::vector<ReplSetDescription> shards;
+    std::vector<MongosDescription> mongoses;
+
+    void accept(TopologyVisitor& v) {
```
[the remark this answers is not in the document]
aggrand: I was kind of thinking having these in the header would be nice for publicly documenting the traversal order of the cluster, but most of the time that's probably not a major concern so I'm good with moving it.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+    void accept(TopologyVisitor& v) { 
+        v.onBeforeSharded(*this); 
+        configsvr.accept(v);
+
+        v.onBeforeShards(*this); 
+        for (int i = 0; i < shards.size() - 1; i++) {
+            shards[i].accept(v);
+            v.onBetweenShards(*this);
+        }
```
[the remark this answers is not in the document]
aggrand: This isn't possible with the `onBetweenShards`, since we don't want it called after the last iteration. Maybe a bit of an unnecessary visit method, but `ToJsonVisitor` used it and it was nice for debugging.

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+
+    /** Misc hooks that most visitors won't need. **/
+
+    // Called between mongods in a replica set.
+    virtual void onBetweenMongods(const ReplSetDescription&) {}
+
+    // Called before/after/between visiting shards.
+    virtual void onBeforeShards(const ShardedDescription&) {}
+    virtual void onAfterShards(const ShardedDescription&) {}
```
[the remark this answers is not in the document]
aggrand: `onAfterSharded` refers to a sharded cluster, whereas `onAfterShards` is after going through each shard in the shard list. I'll rename `onAfterSharded` to `onAfterShardedCluster` to clarify.

src/gennylib/include/gennylib/topology.hpp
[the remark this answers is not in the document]
aggrand: I think this would preclude visitors that accumulate state as they traverse the cluster, like `ToJsonVisitor`

src/gennylib/include/gennylib/topology.hpp
```diff
@@ -0,0 +1,211 @@
...
+            v.onBeforeTopology(*_topologyDesc);
+            _topologyDesc->accept(v); 
+            v.onAfterTopology(*_topologyDesc);
+        }
+    }
+
+    /**
+     * Update the Topology's view of the cluster.
+     */
+    void update(DBConnection& connection);
```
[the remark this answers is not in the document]
aggrand: It should be safe to pass-by-reference since `update` doesn't store `DBConnection` anywhere and the enclosing scope outlives the `update` function? My concern with pass-by-value is that we don't want to create a new connection to the server, we want to reuse the existing one. And I think we don't want to have the `DBConnection` use `shard_ptr` to store the `mongocxx::client`?

src/gennylib/include/gennylib/Connection.hpp
```diff
@@ -0,0 +1,58 @@
...
+     */
+    virtual std::unique_ptr<DBConnection> makePeer(ConnectionUri uri) = 0;
+
+    virtual ~DBConnection() = 0;
+};
+
+class MongoConnection : public DBConnection {
+public:
+    MongoConnection(ConnectionUri uri);
+    ConnectionUri uri();
```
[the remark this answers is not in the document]
aggrand: Thanks this is amazing I added it everywhere.

src/gennylib/src/topology.cpp
```diff
@@ -0,0 +1,172 @@
...
+namespace genny {
+
+TopologyVisitor::~TopologyVisitor() {}
+
+TopologyDescription::~TopologyDescription() {}
+
+std::string Topology::nameToUri(const std::string& name) {
+    std::set<std::string> nameSet;
+    // Hostnames returned from some commands have the form configRepl/hostname:port
+    auto strippedName = name.substr(name.find('/') + 1, std::string::npos);
```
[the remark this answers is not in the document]
aggrand: This is interesting. Some of the names _do_ have the form `hostname:port` without the '/', but this code was still working fine because `npos + 1 == 0` on many implementations, so the substring was "correctly" including the entire string. 

Thanks for pointing this out, I'll update this to be less implementation-dependent.

src/gennylib/src/quiesce.cpp
```diff
@@ -0,0 +1,147 @@
...
+            // Assert that all nodes are in reasonable states.
+            auto replSetRes = admin.run_command(make_document(kvp("replSetGetStatus", 1)));
+            auto members = replSetRes.view()["members"];
+            if (members && members.type() == bsoncxx::type::k_array) {
+                bsoncxx::array::view members_view = members.get_array();
+                for (auto member : members_view) {
+                    std::string state(member["stateStr"].get_utf8().value);
+                    if (state != "PRIMARY" && state != "SECONDARY" && state != "ARBITER") {
+                        std::string name(member["name"].get_utf8().value);
+                        BOOST_LOG_TRIVIAL(error) << "Cannot wait oplog, replset member "
```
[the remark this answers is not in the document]
aggrand: Ended up making this a `warning` as per our offline discussion.

PR #415, opened 2021-02-02
WT-7130 Ensure that we pass our session into each operation when doing a snapshot scan

tetsuo-cpp:
@sulabhM 
This is why we were seeing rollbacks in the scanning part of our Genny workload. If we don't explicitly pass the session in, it's not executed as part of our transaction.

I'm not sure whether this is sufficient to get lookaside usage. I'm running some tests at the moment so we'll see.


sulabhM:
@tetsuo-cpp , interesting. Is it because a mongodb user session (not execution/wiredtiger session) needs to be established for transactions to work?

tetsuo-cpp:
> @tetsuo-cpp , interesting. Is it because a mongodb user session (not execution/wiredtiger session) needs to be established for transactions to work?

For snapshot reads, we already had established a MongoDB session and started a transaction. But we need to explicitly pass the session into each operation we do to make sure it uses the transaction.

sulabhM (approved):

PR #418, opened 2021-02-03
TIG-2823 Genny's actor integration tests don't error in evergreen

aggrand:
It's a little tricky to track down the errors through lamp-resmoke-cmake-actors. 

@guoyr and I discussed deleting the standalone tests, since that shouldn't be representative of actual use and it specifically fails the tests after the driver update.

I had to update the actors to specify csv output, since ftdc output doesn't let us store results for test purposes. I think either adding a test metrics format or adding a ftdc-parsing tool for our ctests will be necessary when we completely remove csv.


src/lamplib/tasks/__init__.py
```diff
@@ -33,6 +33,7 @@ def cmake(context, toolchain_dir, env, cmdline_cmake_args):
     cmake_cmd += [
         "-DCMAKE_PREFIX_PATH={}".format(cmake_prefix_path),
         "-DCMAKE_TOOLCHAIN_FILE={}".format(cmake_toolchain_file),
+        "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON",
```
richardsamuels, line 36: Does this cause compile_commands.json to be emitted? If so, can you add it to `.gitignore`?

src/lamplib/__main__.py
```diff
@@ -101,15 +102,19 @@ def main():
         elif args.subcommand == "benchmark-test":
             tasks.run_tests.benchmark_test(compile_env)
         elif args.subcommand == "resmoke-test":
-            tasks.run_tests.resmoke_test(
+            success = tasks.run_tests.resmoke_test(
```
richardsamuels, line 105: If an exception gets raised in this function, it should just percolate up and cause a non-zero return anyway, right?

richardsamuels (approved):
I don't have the best context on genny, so please take my approval with a grain of salt. Could you explain the Metrics: Format: csv change as well?

src/lamplib/tasks/__init__.py
aggrand: It gets dumped in `build` so is ignored anyway! I actually didn't mean to include that I just use it locally for pylinter; I can remove it if anyone prefers it not be there.

rtimmons: Hey this is a really nice change I wish we'd thought of it earlier. Huge 👍🏽

src/metrics/include/metrics/metrics.hpp
```diff
@@ -166,6 +166,7 @@ class RegistryT final {
                        bool assertMetricsBuffer = true)
         : _format{std::move(format)}, _pathPrefix{std::move(pathPrefix)} {
         if (_format.useGrpc()) {
+            boost::filesystem::remove_all(_pathPrefix);
```
rtimmons, line 169: Do we need this now that we've set the format to csv for test actors?

Alternatively maybe we can have a `CLEAR_POPLAR_METRICS` env var that this checks or something? I think most users will not want to delete prior metrics data. My WIP change jumps through some hoops to look for existing data and moves it to a timestamp'ed dir (does this on the genny python side) so it's really only tests that will need this remove functionality.

evergreen.yml
```diff
@@ -163,12 +163,6 @@ tasks:
   commands:
   - func: f_benchmark_test
 
-- name: t_integration_test_standalone
-  commands:
-  - func: f_resmoke_test
-    vars:
-      resmoke_test_args: "--suites src/resmokeconfig/genny_standalone.yml"
```
rtimmons, line 170: let's also remove the suite yaml file.

src/lamplib/__main__.py
aggrand: The resmoke test failure as a subprocess call doesn't actually throw an exception, just has its own return code. I'd be cool with testing and throwing inside if that's preferred

rtimmons (changes requested):
> [standalone] specifically fails the tests after the driver update.

Good catch! Let's add a comment _somewhere_ (maybe in the evergreen yaml right here) that it's not possible to add a standalone version due to the reliance on the apm callback (intuitively I don't understand why the driver change would break this but if there's some docs that lay it out then a link would suffice as an explanation why).

I'm okay with this except for the C++ removing existing metrics data. Please push back if you disagree anywhere.

rtimmons:
Maybe also remove the failing ubuntu1804 compile (or comment out the variant with a TODO link to the ticket we already have for it?). Fine if you wanna let it live to see another day for now.

aggrand:
@richardsamuels regarding the csv format, that metrics format is deprecated but most of our unittests rely on it since it stores the data locally. This test hadn't been updated to specify it explicitly like the unit tests because the error was hidden. 

When we remove csv as an option entirely we'll need some alternative for testing, but I think that'd be its own ticket

src/metrics/include/metrics/metrics.hpp
aggrand: Yeah it's not necessary; I had left it in since there's a relevant ticket but I like your approach. Removed.

aggrand:
Also a patch where I broke an actor test to check that it errors in evergreen: https://evergreen.mongodb.com/version/601b44ec7742ae5b85462c97

rtimmons (approved):
Sorry I forgot to check back on this sooner.

aggrand:
evergreen merge

PR #419, opened 2021-02-11
PERF-2089 Update ^RandomDate Generator to accept other generators as input

jimoleary:
**dateStringParserGenerator** calls **stringGenerator**. As a result, it supports all the generators (even if they don't make sense like ^IP).

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -153,6 +153,7 @@ using Parser = std::function<O(const Node&, GeneratorArgs)>;
 
 // Pre-declaring all at once
 // Documentation is at the implementations-site.
+const static int64_t parseStringToMillis(const std::string& datetime);
 const static boost::posix_time::ptime epoch(boost::gregorian::date(1970, 1, 1));
 const static boost::posix_time::ptime max_date(boost::gregorian::date(2150, 1, 1));
 
@@ -678,32 +679,81 @@ const static auto formats = {
                 new boost::local_time::local_time_input_facet("%Y-%m-%d %H:%M:%s%ZP")),
 };
 
+class DateToIntGenerator : public Generator<int64_t> {
+public:
+    explicit DateToIntGenerator(UniqueGenerator<bsoncxx::types::b_date>& generator) {
+        std::swap(_generator, generator);
+    }
+    int64_t evaluate() override {
+        auto date = _generator->evaluate();
+        auto millis = date.to_int64();
+        // BOOST_LOG_TRIVIAL(warning) << " DateToIntGenerator " << millis;
+        return millis;
+    }
+
+protected:
+    UniqueGenerator<bsoncxx::types::b_date> _generator;
+};
+
+class DoubleToIntGenerator : public Generator<int64_t> {
+public:
+    explicit DoubleToIntGenerator(UniqueGenerator<double>& generator) {
+        std::swap(_generator, generator);
+    }
+    int64_t evaluate() override {
+        auto millis = (long long)_generator->evaluate();
+        // BOOST_LOG_TRIVIAL(warning) << " DoubleToIntGenerator " << millis;
+        return millis;
+    }
+
+protected:
+    UniqueGenerator<double> _generator;
+};
+
+class DateStringParserGenerator : public Generator<int64_t> {
+public:
+    explicit DateStringParserGenerator(UniqueGenerator<std::string>& generator) {
+        std::swap(_generator, generator);
+    }
+    int64_t evaluate() override {
+        auto datetime = _generator->evaluate();
+        auto millis = parseStringToMillis(datetime);
+        // BOOST_LOG_TRIVIAL(warning) << " DateStringParserGenerator " << millis;
+        return millis;
+    }
+
+protected:
+    UniqueGenerator<std::string> _generator;
+};
+
 /** `{^RandomDate: {min: "2015-01-01", max: "2015-01-01T23:59:59.999Z"}}` */
 class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
 public:
     RandomDateGenerator(const Node& node, GeneratorArgs generatorArgs)
         : _rng{generatorArgs.rng},
-          _min{dateGenerator(node["min"], generatorArgs)->evaluate()},
-          _max{dateGenerator(node["max"], generatorArgs, max_date)->evaluate()} {
-        if (_max <= _min) {
-            auto max = node["max"].maybe<std::string>().value_or("");
-            auto min = node["min"].maybe<std::string>().value_or("");
-
-            std::string msg = "^RandomDate: max('" + max + " / " + std::to_string(_max) +
-                "') must be greater than min('" + min + " / " + std::to_string(_min) + "')";
-            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{msg});
-        }
-    }
+          _node{node},
+          _minGen{dateGenerator(node["min"], generatorArgs)},
+          _maxGen{dateGenerator(node["max"], generatorArgs, max_date)} {}
 
     bsoncxx::types::b_date evaluate() override {
-        boost::random::uniform_int_distribution<long long> dist{_min, _max - 1};
+        auto min = _minGen->evaluate();
+        auto max = _maxGen->evaluate();
+        if (max <= min) {
+            std::ostringstream msg;
+            msg << "^RandomDate: " << _node << ", max (" << max << ") must be greater than min ("
+                << min << ")";
+            BOOST_LOG_TRIVIAL(warning) << " RandomDateGenerator " << msg.str();
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{msg.str()});
+        }
+        boost::random::uniform_int_distribution<long long> dist{min, max - 1};
         return bsoncxx::types::b_date{std::chrono::milliseconds{dist(_rng)}};
     }
 
 private:
     DefaultRandom& _rng;
-    int64_t _min;
-    int64_t _max;
+    const Node& _node;
+    const UniqueGenerator<int64_t> _minGen;
+    const UniqueGenerator<int64_t> _maxGen;
 };
 
 
@@ -1168,7 +1218,7 @@ bool isUnknownDateFormat(const InvalidValueGeneratorSyntax& exception) {
  * @return
  *   The datetime as millis.
  */
-static auto parseStringToMillis(const std::string& datetime) {
+const static int64_t parseStringToMillis(const std::string& datetime) {
     boost::posix_time::ptime utc_time{boost::posix_time::not_a_date_time};
 
     auto millis = 0LL;
@@ -1204,11 +1254,21 @@ static auto parseStringToMillis(const std::string& datetime) {
  * @return
  *   a constant generator if given a constant/scalar.
  */
-UniqueGenerator<int64_t> dateStringParserGenerator(const Node& node, GeneratorArgs generatorArgs) {
+UniqueGenerator<int64_t> dateStringTimeGenerator(const Node& node, GeneratorArgs generatorArgs) {
     auto generator = stringGenerator(node, generatorArgs);
-    auto datetime = generator->evaluate();
-    auto millis = parseStringToMillis(datetime);
-    return std::make_unique<ConstantAppender<int64_t>>(millis);
+    return std::make_unique<DateStringParserGenerator>(generator);
+}
+
+/**
+ * @private
+ * @param node
+ *   a node double value i.e. either a scalar or a `^RandomDouble` value
+ * @return
+ *   a constant generator if given a constant/scalar.
+ */
+UniqueGenerator<int64_t> doubleTimeGenerator(const Node& node, GeneratorArgs generatorArgs) {
+    auto generator = doubleGenerator(node, generatorArgs);
+    return std::make_unique<DoubleToIntGenerator>(generator);
 }
 
 /**
@@ -1236,9 +1296,7 @@ UniqueGenerator<int64_t> dateTimeGenerator(const Node& node, GeneratorArgs gener
     if (auto parserPair = extractKnownParser(node, generatorArgs, dateParsers)) {
         // known parser type
         auto generator = parserPair->first(node[parserPair->second], generatorArgs);
-        auto datetime = generator->evaluate();
-        auto millis = datetime.to_int64();
-        return std::make_unique<ConstantAppender<int64_t>>(millis);
+        return std::make_unique<DateToIntGenerator>(generator);
     }
     // This is an internal private function, we have no sane way to interpret a scalar here,
     // so throw an exception.
@@ -1257,50 +1315,19 @@ UniqueGenerator<int64_t> dateTimeGenerator(const Node& node, GeneratorArgs gener
 UniqueGenerator<int64_t> dateGenerator(const Node& node,
                                        GeneratorArgs generatorArgs,
                                        const boost::posix_time::ptime& default_time) {
+    const static auto timeGenerators = {
+        dateTimeGenerator, dateStringTimeGenerator, intGenerator, doubleTimeGenerator};
     if (node) {
-        if (node.isMap()) {
+        for (auto&& generator : timeGenerators) {
             try {
-                auto generator = dateTimeGenerator(node, generatorArgs);
-                return std::make_unique<ConstantAppender<int64_t>>(
-                    (long long)generator->evaluate());
+                return generator(node, generatorArgs);
             } catch (const InvalidValueGeneratorSyntax& e) {
-                if (!isUnknownParser(e)) {
+                if (!isUnknownParser(e) && !isUnknownDateFormat(e)) {
                     throw e;
                 }
             } catch (const InvalidConversionException& e) {
             }
         }
-
-        try {
-            auto generator = dateStringParserGenerator(node, generatorArgs);
-            return std::make_unique<ConstantAppender<int64_t>>((long long)generator->evaluate());
-        } catch (const InvalidValueGeneratorSyntax& e) {
-            if (!isUnknownParser(e) && !isUnknownDateFormat(e)) {
-                throw e;
-            }
-        } catch (const InvalidConversionException& e) {
-        }
-
-        try {
-            auto generator = intGenerator(node, generatorArgs);
-            return generator;
-        } catch (const InvalidValueGeneratorSyntax& e) {
-            if (!isUnknownParser(e)) {
-                throw e;
-            }
-        } catch (const InvalidConversionException& e) {
-        }
-
-        try {
-            auto generator = doubleGenerator(node, generatorArgs);
-            return std::make_unique<ConstantAppender<int64_t>>((long long)generator->evaluate());
-        } catch (const InvalidValueGeneratorSyntax& e) {
-            if (!isUnknownParser(e)) {
-                throw e;
-            }
-        } catch (const InvalidConversionException& e) {
-        }
-
         std::stringstream msg;
         msg << "Unknown parser: not an expected type " << node;
         BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg.str()));
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -681,7 +681,7 @@ const static auto formats = {
 
 class DateToIntGenerator : public Generator<int64_t> {
 public:
-    explicit DateToIntGenerator(UniqueGenerator<bsoncxx::types::b_date>& generator)
+    explicit DateToIntGenerator(UniqueGenerator<bsoncxx::types::b_date>&& generator)
         : _generator{std::move(generator)} {}
     int64_t evaluate() override {
         auto date = _generator->evaluate();
@@ -694,7 +694,7 @@ private:
 
 class CastDoubleToInt64Generator : public Generator<int64_t> {
 public:
-    explicit CastDoubleToInt64Generator(UniqueGenerator<double>& generator)
+    explicit CastDoubleToInt64Generator(UniqueGenerator<double>&& generator)
         : _generator{std::move(generator)} {}
     int64_t evaluate() override {
         return (int64_t)_generator->evaluate();
@@ -706,7 +706,7 @@ private:
 
 class DateStringParserGenerator : public Generator<int64_t> {
 public:
-    explicit DateStringParserGenerator(UniqueGenerator<std::string>& generator)
+    explicit DateStringParserGenerator(UniqueGenerator<std::string>&& generator)
         : _generator{std::move(generator)} {}
     int64_t evaluate() override {
         auto datetime = _generator->evaluate();
@@ -1218,7 +1218,7 @@ int64_t parseStringToMillis(const std::string& datetime) {
  */
 UniqueGenerator<int64_t> dateStringTimeGenerator(const Node& node, GeneratorArgs generatorArgs) {
     auto generator = stringGenerator(node, generatorArgs);
-    return std::make_unique<DateStringParserGenerator>(generator);
+    return std::make_unique<DateStringParserGenerator>(std::move(generator));
 }
 
 /**
@@ -1230,7 +1230,7 @@ UniqueGenerator<int64_t> dateStringTimeGenerator(const Node& node, GeneratorArgs
  */
 UniqueGenerator<int64_t> doubleTimeGenerator(const Node& node, GeneratorArgs generatorArgs) {
     auto generator = doubleGenerator(node, generatorArgs);
-    return std::make_unique<CastDoubleToInt64Generator>(generator);
+    return std::make_unique<CastDoubleToInt64Generator>(std::move(generator));
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -153,6 +153,8 @@ using Parser = std::function<O(const Node&, GeneratorArgs)>;
 
 // Pre-declaring all at once
 // Documentation is at the implementations-site.
+const static boost::posix_time::ptime epoch(boost::gregorian::date(1970, 1, 1));
+const static boost::posix_time::ptime max_date(boost::gregorian::date(2030, 1, 1));
```
dalyd, line 157: I feel like max_date should be a lot higher. I know it's what we were already using, but it feels quite stark now.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +677,33 @@ const static auto formats = {
...
-}
 
 /** `{^RandomDate: {min: "2015-01-01", max: "2015-01-01T23:59:59.999Z"}}` */
 class RandomDateGenerator : public Generator<bsoncxx::types::b_date> {
 public:
     RandomDateGenerator(const Node& node, GeneratorArgs generatorArgs)
         : _rng{generatorArgs.rng},
-          _min{parseDateAsMillis(node["min"].maybe<std::string>().value_or(""))},
-          _max{parseDateAsMillis(node["max"].maybe<std::string>().value_or(""), max_date)} {
+          _minGen{dateGenerator(node["min"], generatorArgs)},
```
dalyd, line 686: Can I use {^Now: {}} as the min value? Does dateGenerator case to an int64?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +677,33 @@ const static auto formats = {
...
+            std::stringstream ex;
+            ex << "^RandomDate: max('" << maxValue << "') must be greater than min('" << minValue
+               << "')";
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{ex.str()});
         }
-    }
 
-    bsoncxx::types::b_date evaluate() override {
-        boost::random::uniform_int_distribution<long long> dist{_min.count(), _max.count() - 1};
+        boost::random::uniform_int_distribution<long long> dist{minValue, maxValue - 1};
```
dalyd, line 699: With the extra parameterization and generators, do you want to make kinds based on all the int distribution generators?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1140,150 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
     }
 }
 
+/**
+ * @private
+ * @param exception exception received from mongocxx::database::run_command()
+ * @return if the exception is a result of a connection error or timeout
+ */
+bool isUnknownParser(const InvalidValueGeneratorSyntax& exception) {
```
dalyd, line 1148: This looks nice.

src/value_generators/src/DocumentGenerator.cpp
dalyd: It looks like it can. ^Now is in dateGenerator. Does the system automatically do the conversion?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1140,150 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+             return std::make_unique<RandomDateGenerator>(node, generatorArgs);
+         }},
+    };
+
+    if (auto parserPair = extractKnownParser(node, generatorArgs, dateParsers)) {
+        // known parser type
+        auto generator = parserPair->first(node[parserPair->second], generatorArgs);
+        auto datetime = generator->evaluate();
+        auto millis = datetime.to_int64();
+        return std::make_unique<ConstantAppender<int64_t>>(millis);
```
dalyd, line 1229: Oh, we need to be careful here. So this will evaluate only once. Whatever string you get the first time is what you have the whole time. I'm going back and forth with myself over what makes the most sense. 

Could we make a UniqueGenerator<int64_t> that wraps a UniqueGenerator<string>?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1140,150 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+            auto generator = dateStringParserGenerator(node, generatorArgs);
+            return std::make_unique<ConstantAppender<int64_t>>((long long)generator->evaluate());
+        } catch (const InvalidValueGeneratorSyntax& e) {
+            if (!isUnknownParser(e)) {
+                throw e;
+            }
+        } catch (const InvalidConversionException& e) {
+        }
+
+        auto generator = dateTimeGenerator(node, generatorArgs);
```
dalyd, line 1278: Should we order these by expected frequency? Like dateStringParser is probably most likely, then dateTimeGenerator. I'm thinking about all the try catches. I wonder if there is a way around it. It's a valid way to do it.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -594,6 +594,55 @@ Tests:
     ThenReturns:
     - { "date" : { "$date" : { "$numberLong" : "1369875424299" } } }
 
+  # Date: Random Dates from RandomInt.
+  - Name: RandomWithInt
```
dalyd, line 598: This is really cool!

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -594,6 +594,55 @@ Tests:
...
+    GivenTemplate:
+      date: {^RandomDate: {min: {^RandomInt: {min: 1369875424299, max: 1369875424299}}, max: {^RandomInt: {min: 1369875424300, max: 1369875424300}}}}
```
dalyd, line 600: I missed it, do we check that max > min? With generators it will be easy to construct a case in which that is not true.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -594,6 +594,55 @@ Tests:
...
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "1369875424299" } } }
+
+  # Date: Random Dates From RandomDouble.
+  - Name: RandomFromRandomhDouble
```
dalyd, line 605: Typo in name

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -594,6 +594,55 @@ Tests:
...
+  - Name: RandomDateFromRandomString
+    GivenTemplate:
+      date: {^RandomDate: {min: {^RandomString: {length: 1, alphabet: "0"}}, max: {^RandomString: {length: 1, alphabet: "1"}}}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "0" } } }
+
+  # Date: Random Dates from Join.
+  - Name: RandomDateFromJoin
+    GivenTemplate:
+      date: {^RandomDate: {min: {^Join: {array: ["2013","05","30T07:27:04.299+06:30"], sep: "-"}}, max: {^Join: {array: ["2013","05","30 06:27:04.300+05:30"], sep: "-"}}}}
```
dalyd, line 628: I love this!

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -594,6 +594,55 @@ Tests:
...
+  - Name: RandomDateFromChoose
+    GivenTemplate:
+      date: {^RandomDate: {min: {^Choose: {from: ["2013-05-30T07:27:04.299+06:30"]}}, max: {^Choose: {from: ["2013-05-30 06:27:04.300+05:30"]}}}}
+    ThenReturns:
+    - { "date" : { "$date" : { "$numberLong" : "1369875424299" } } }
+
+  # Date: RandomDates with invalid generator.
+  # Test that the correct exception type is raised.
+  - Name: RandomDateException
+    GivenTemplate:
```
dalyd, line 642: I can get a random generator that is (2021-01-01, ^Now). Can I somehow specify (^Now - 30 days, ^Now). I would like that, but can manage with the first. Obviously not that syntax.

src/value_generators/src/DocumentGenerator.cpp
jimoleary: 2050?

dalyd: 2150?

src/value_generators/src/DocumentGenerator.cpp
dalyd: That's cool. Nice job.

src/value_generators/src/DocumentGenerator.cpp
dalyd: Yeah, keep the scope reasonable. I could imagine though having a set range (today to end of year), but not wanting that to be uniformly distributed.

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: Cool.  Sounds good.

src/value_generators/src/DocumentGenerator.cpp
dalyd: Could you make sure to include comments with the documentation for this? That may only be the test cases, but still.

dalyd (approved):
As I'm out next week, I'm going to approve now. I'll leave the rest with Ryan. 
One last request: Could we include some of these generators in a workload in the Docs directory? In Generators.yml?

src/value_generators/src/DocumentGenerator.cpp
jimoleary: Is the fact that  a ```{^Now: {}}``` evaluated once a surprise? 

If so then it may make sense to create a lazy generator for these two types.

dalyd: It is. I had proposed another generator of {^Once: {# Some other generator}} to allow explicitly doing that because there are use cases for it.

rtimmons (approved):
I'm OOO and can't look deeper until Tuesday. I took a quick look and I saw nothing would prevent merging, so if this needs to go in sooner I'm happy if David and the CI tests are happy.

jimoleary:
@rtimmons I'm happy to wait for your review. And it will give me a bit of time to look at lazy evaluating generators.

src/value_generators/src/DocumentGenerator.cpp
jimoleary: ok, I'll have a look at lazy evaluating generators. To be correct / consistent all the non-constant cases should also be lazy evaluated (i.e.^RandomInt etc.... ).

There wil be a side effect though that the min / max check could pass or fail depending on what values are generated.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,82 @@ const static auto formats = {
...
-const static boost::posix_time::ptime max_date(boost::gregorian::date(2030, 1, 1));
 
-static std::chrono::milliseconds parseDateAsMillis(
-    const std::string& datetime, const boost::posix_time::ptime& default_time = epoch) {
-    boost::posix_time::ptime utc_time{boost::posix_time::not_a_date_time};
-    auto millis = 0LL;
+class DateToIntGenerator : public Generator<int64_t> {
+public:
+    explicit DateToIntGenerator(UniqueGenerator<bsoncxx::types::b_date>& generator) {
+        std::swap(_generator, generator);
```
jimoleary, line 685: Is this the correct approach here?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,82 @@ const static auto formats = {
...
-            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{msg});
-        }
-    }
+          _node{node},
+          _minGen{dateGenerator(node["min"], generatorArgs)},
+          _maxGen{dateGenerator(node["max"], generatorArgs, max_date)} {}
 
     bsoncxx::types::b_date evaluate() override {
-        boost::random::uniform_int_distribution<long long> dist{_min.count(), _max.count() - 1};
+        auto min = _minGen->evaluate();
```
jimoleary, line 739: The error detection happens in evaluate so the test runner needed to be changed to take this into account.

src/value_generators/test/DocGenYamlTestCaseRunner.cpp
```diff
@@ -68,7 +68,8 @@ class ValGenTestCase {
             if (_runMode == RunMode::kExpectException) {
                 try {
                     NodeSource ns = toNode(this->_givenTemplate);
-                    genny::DocumentGenerator(ns.root(), GeneratorArgs{rng, 1});
+                    auto docGen = genny::DocumentGenerator(ns.root(), GeneratorArgs{rng, 1});
+                    docGen();
```
jimoleary, line 72: evaluate for ^RandomDate can throw the exception.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -527,8 +539,8 @@ Tests:
     ThenReturns:
     # The $date notation is extended json
     # (see https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson.Date).
-    - { "date" : { "$date" : { "$numberLong" : "1778468968160" } } }
-    - { "date" : { "$date" : { "$numberLong" : "1871664110292" } } }
+    - { "date" : { "$date" : { "$numberLong" : "4185670857742" } } }
+    - { "date" : { "$date" : { "$numberLong" : "5397029274960" } } }
```
jimoleary, line 543: max date changed in the last commit which has a knock on effect on the dates generated.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -153,6 +153,9 @@ using Parser = std::function<O(const Node&, GeneratorArgs)>;
 
 // Pre-declaring all at once
 // Documentation is at the implementations-site.
+const static int64_t parseStringToMillis(const std::string& datetime);
+const static boost::posix_time::ptime epoch(boost::gregorian::date(1970, 1, 1));
```
rtimmons, line 157: plz use `=` or curlies to make it clear these are static members not forward references.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -153,6 +153,9 @@ using Parser = std::function<O(const Node&, GeneratorArgs)>;
 
 // Pre-declaring all at once
 // Documentation is at the implementations-site.
+const static int64_t parseStringToMillis(const std::string& datetime);
```
rtimmons, line 156: c++-ism - the `static` isn't needed here

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -161,6 +164,9 @@ UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generato
 UniqueGenerator<double> doubleGeneratorBasedOnDistribution(const Node& node,
                                                            GeneratorArgs generatorArgs);
 UniqueGenerator<std::string> stringGenerator(const Node& node, GeneratorArgs generatorArgs);
+UniqueGenerator<int64_t> dateGenerator(const Node& node,
+                                       GeneratorArgs generatorArgs,
+                                       const boost::posix_time::ptime& default_time = epoch);
```
rtimmons, line 169: 💅🏽 plz use `camelCase` - `const ptime& defaultTime`

src/value_generators/src/DocumentGenerator.cpp
rtimmons: I'm not sure what the intent here is and why the `std::swap` does what you want? I don't usually see `swap` used with refs - usually it's with value types or values. This could be totally kosher but I for one could use a comment here about why this is the right thing.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,82 @@ const static auto formats = {
...
-    auto millis = 0LL;
+class DateToIntGenerator : public Generator<int64_t> {
+public:
+    explicit DateToIntGenerator(UniqueGenerator<bsoncxx::types::b_date>& generator) {
+        std::swap(_generator, generator);
+    }
+    int64_t evaluate() override {
+        auto date = _generator->evaluate();
+        auto millis = date.to_int64();
+        // BOOST_LOG_TRIVIAL(warning) << " DateToIntGenerator " << millis;
```
rtimmons, line 690: 💅🏽 plz remove the commented-out logs

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,82 @@ const static auto formats = {
...
-        if (utc_time == boost::posix_time::not_a_date_time) {
-            try {
-                // Last gasp try to interpret unsigned long long.
-                millis = std::stoull(datetime);
-            } catch (const std::invalid_argument& _) {
-                auto msg = "^RandomDate: Invalid Dateformat '" + datetime + "'";
-                BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax{msg});
-            }
-        }
+protected:
```
rtimmons, line 709: meh - `private` unless you intend these to be subclasses?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,82 @@ const static auto formats = {
...
-                break;
-            }
-        }
+class DoubleToIntGenerator : public Generator<int64_t> {
+public:
+    explicit DoubleToIntGenerator(UniqueGenerator<double>& generator) {
+        std::swap(_generator, generator);
+    }
+    int64_t evaluate() override {
+        auto millis = (long long)_generator->evaluate();
```
rtimmons, line 704: so `_gen->eval()` returns `double` which we cast to `long long` but the expected return type is `int64`? I think this is working by casting twice which is why you needed the separate `auto millis` declaration.

I think you can use a single cast:

```c++
int64_t evaluate() override {
  return (int64_t) _gen->eval();
}
```

Also consider naming this `CastDoubleToInt64Generator` or similar.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1190,154 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
 
+/**
+ * @private
+ * @param exception exception received from Generator factory (e.g. ::intGenerator()) invocation.
+ * @return if the exception is a result of the generator being unknown to the factory. For
+ * example, a ^RandomDate node being passed to the intGenerator().
+ */
+bool isUnknownParser(const InvalidValueGeneratorSyntax& exception) {
+    auto what = std::string{exception.what()};
+    return what.find("Unknown parser") != std::string::npos;
```
rtimmons, line 1201: we could make a separate `UnknownParserException` if that would help? Relying on exception messages for logic has bitten me a number of times in the past (trying to make exceptions easier to read has caused bugs). If that doesn't seem worthwhile or like it will work plz add a comment wherever it adds the "Unknown parser" string letting future maintainers know about `isUnknownParser`. (Obv same to the format one below)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1190,154 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+    if (!datetime.empty()) {
+        for (const auto& format : formats) {
+            std::istringstream date_stream{datetime};
+            date_stream.imbue(format);
+            boost::local_time::local_date_time local_date{boost::local_time::not_a_date_time};
+            if (date_stream >> local_date) {
+                utc_time = local_date.utc_time();
+
+                millis = (utc_time - epoch).total_milliseconds();
+                break;
```
rtimmons, line 1234: why not `return millis` here?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1190,154 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+
+/**
+ * @private
+ * @param datetime
+ *   a datetime string in a known format or unsigned long long string.
+ * @return
+ *   The datetime as millis.
+ */
+const static int64_t parseStringToMillis(const std::string& datetime) {
+    boost::posix_time::ptime utc_time{boost::posix_time::not_a_date_time};
```
rtimmons, line 1222: another `camelCase` reminder

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1173,6 +1190,154 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+            }
+        }
+        std::stringstream msg;
+        msg << "Unknown parser: not an expected type " << node;
+        BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg.str()));
+    }
+
+    // No value, get the appropriate default.
+    auto millis = (default_time - epoch).total_milliseconds();
+    return std::make_unique<ConstantAppender<int64_t>>(millis);
```
rtimmons, line 1338: Does a lot of this of this go away if we don't allow mixed types in the min/max fields? I.e. if you want a literal min/max rather than generated simply require the use of `^Verbatim`? What you have here is clever and well-written, but I wonder if it's complexity worth having.

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: > it's still easy to construct a case where it aribrarily fails depending on the inputs.

This seems a bit...fraught. Would it make sense for there to be a "min" and "max" member-function on the date generators that returns the min/max values that the generator could ever produce and then apply the min/max checking logic at construction time? This is obviously rather tricky with `^Choose` and perhaps others but maybe that's a good thing - meaning maybe it shouldn't be possible to internally use something other than `DateGen` (isA `Generator<boost::time_point>` or whatever the boost type is)?

rtimmons (approved):
Looks good - I like seeing the recursion machinery working out reasonably well for this.

My only hesitation is around the `std::swap` which may be correct but just not paging into my C++ memory banks on correct idioms.

Oh ya one more thing--
Adding a couple `ThenReturn` values for repeated executions may be helpful to ensure the runtime evaluation of certain things. E.g. I can imagine

```yaml
GivenTemplate:
  date: {^RandomDate: {min: {^Choose: {from: ["1500-01-01", "2020-02-16"]}, max: "2021-02-16"}
ThenReturns: [{date: foo}, {date: bar}] # ideally where bar couldn't have been from a choice of 1500 being the min
```

I'm not feeling clever enough to come up with a better example, but right now I think it's possible to introduce a regression that reverts back to instantiation-time evaluation and these test may not protect against that. If I'm wrong I'm happy to leave the tests as they are and not address this explicitly.

rtimmons:
Also this whole generators-of-generators thing is starting to feel like we're dangerously close to implementing a lisp. I'm not opposed to that per see but we could decide to lean into it more if we want to be more and more clever over time. Just some :philosoraptor: commentary.

jimoleary (commented):
> Also this whole generators-of-generators thing is starting to feel like we're dangerously close to implementing a lisp. I'm not opposed to that per see but we could decide to lean into it more if we want to be more and more clever over time. Just some :philosoraptor: commentary.

The 'killer' feature that this work allows is **_{^RandomDate: {min: {^Now:{}}}}_** where Now is the current time (rather than the first invocation time).

It seemed natural to me to make all the other generators evaluate at runtime also. It didn't add a huge amount of extra work.

In terms of min / max validation, your suggestion of adding a set of min / max methods to return the correct values to allow for validation has merit although unfortunately ^Now would require extra thought. Should I raise a new ticket to investigate?

There is an issue with boost date parsing that needs a bit of investigation, I've raised PERF-2153 to track this.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -672,71 +678,73 @@ const static auto formats = {
...
+private:
+    const UniqueGenerator<double> _generator;
+};
 
-    } else {
-        millis = (default_time - epoch).total_milliseconds();
+class DateStringParserGenerator : public Generator<int64_t> {
+public:
+    explicit DateStringParserGenerator(UniqueGenerator<std::string>& generator)
+        : _generator{std::move(generator)} {}
```
rtimmons, line 710: I think this can introduce use-after-move problems:

```c++
UniqueGenerator<std::string> og = ....;
DateStringParserGenerator other {og};
og->evaluate(); // 💥
```

I realize mostly academic for this code, but see if you can just pass in an rvalue and do the `std::move` at the callsite:

```c++
class ... {
     explicit DateStringParserGenerator(UniqueGenerator<std::string>&& generator)
     : _gen{generator} {}
}
...

UniqueGenerator<std::string> og = ....;
DateStringParserGenerator other {std::move(of)};
og->evaluate(); // still runtime 💥 but compiler can detect it
```

Ugh my goodness C++ makes it far too easy to footgun.

jimoleary: Thanks, I'll change it. I did think of doing the move at the call site, but thought it would be easier to do it in the receiver.

jimoleary: It looks like there is some oddness in my linux dev environment, the code above fails to compile on ubuntu (it works on MacOS (thanks @rtimmons). I will investigate my dev env a bit more and come back.

jimoleary:
evergreen merge

PR #420, opened 2021-02-11
TIG-2816 revert workaround for retryable writes

guoyr:
Patch: https://spruce.mongodb.com/version/602582841e2d17709ec6825c


brooookemiller:
Should we move forward with this, @guoyr or @aggrand? 

Based on the last comment on TIG-2816 from 10/18/2021, we mentioned that "There's a workaround to deal with the fact that we were getting write concern errors before we had upgraded the c++ driver, and now that we had upgrade it (awhile ago), we can remove this workaround."

PR #421, opened 2021-02-12
WT-7193 Cleanup Evergreen tasks for LSWA

tetsuo-cpp:
- Fix ScanWithLongLived to run for a shorter duration and use replica sets.
- Stop running LargeScaleModel since it generates BFs and doesn't do anything useful.

### DIFF
--- a/src/phases/scale/LargeScalePhases.yml
+++ b/src/phases/scale/LargeScalePhases.yml
@@ -14,7 +14,7 @@ GlobalDefaults:
   Random30String: &rand_30b_string {^RandomString: {length: 30}}
   Random130String: &rand_100b_string {^RandomString: {length: 100}}
 
-  Duration: &Duration 12 hours
+  Duration: &Duration 2 hours
 
   # We want about 200 bytes, 9 fields (they will all be indexed).
   RollingDocument: &RollingDocument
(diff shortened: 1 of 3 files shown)


src/phases/scale/LargeScalePhases.yml
```diff
@@ -14,7 +14,7 @@ GlobalDefaults:
   Random30String: &rand_30b_string {^RandomString: {length: 30}}
   Random130String: &rand_100b_string {^RandomString: {length: 100}}
 
-  Duration: &Duration 12 hours
+  Duration: &Duration 2 hours
```
tetsuo-cpp, line 17: Let's reduce to 2 hours and see if Cedar is ok with it.

src/workloads/scale/LargeScaleModel.yml
```diff
@@ -337,8 +337,3 @@ Actors:
   - ExternalPhaseConfig:
       Path: ../../phases/scale/LargeScalePhases.yml
       Key: OplogTrailerCmd
-
-AutoRun:
-  Requires:
-    mongodb_setup:
-    - standalone
```
tetsuo-cpp, line 344: This workload generates BFs sometimes. It's not doing anything useful so let's just disable it for now.

src/workloads/scale/ScanWithLongLived.yml
```diff
@@ -88,4 +88,5 @@ Actors:
 AutoRun:
   Requires:
     mongodb_setup:
-    - standalone
+    - single-replica
```
tetsuo-cpp, line 91: Our workload is now running with read concern "snapshot" which doesn't work for standalone.
Let's just run against `single-replica` and without flow control.

sulabhM (approved):

PR #423, opened 2021-02-21
PERF-2159 Adding options support for find and findOne in CrudActor

apodelko:
CrudActor ignores options for most operations including find. It does process some options only for four operations: BulkWrite, CountDocuments, EstimateDocumentCount, and StartTransaction. All options must be spell out in the code to be processed. It just ignores it otherwise without any error and warning.

So this PR adds support of a few options for find and findOne and adding a test that supposed to check it.

Ideally it would be good to get a warning when something in the yml file is ignored.

As I need it for PERF-2096 while it is not directly related to it - submitting this small PR separately.

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -799,7 +799,11 @@ struct FindOperation : public BaseOperation {
           _onSession{onSession},
           _collection{std::move(collection)},
           _operation{operation},
-          _filter{opNode["Filter"].to<DocumentGenerator>(context, id)} {}
+          _filter{opNode["Filter"].to<DocumentGenerator>(context, id)} {
+        if (opNode["Options"]) {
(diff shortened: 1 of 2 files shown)


rtimmons (approved):
Thanks for adding this. I understand how frustrating this probably was to track down.

> Ideally it would be good to get a warning when something in the yml file is ignored.

I've got a WIP PR https://github.com/mongodb/genny/pull/424 that I think would have helped you in this case. It checks that all yaml structures were accessed, and it does this during the "dry-run" operation that we do at CI/compile time. E.g. [here](https://spruce.mongodb.com/task/genny_amazon2_t_compile_patch_db500ad573b018aa7d3485ec3efb23d98eb27c87_6032eb4b7742ae62aaaae591_21_02_21_23_22_52/logs?execution=0) is the output. It's still WIP but thought it might make you feel somewhat better to know that it's coming.

apodelko:
evergreen merge

PR #424, opened 2021-02-21
TIG-2766 Warn of unused yaml structures

rtimmons:
80% of the solution needed to warn of potential mis-configuration. This is such a common annoyance that it seemed worthwhile to put out a "beta" solution:

1. Doesn't actually error on unused structures
2. Structures used via `.to<std::vector>` and other STL containers are reported as unused
3. The top-level root structure shows as unused (line noise). I didn't have the patience to fix that yet.

Before we look into handling these issues, I first want to see how useful this is / if it catches anything.

### DIFF
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -17,6 +17,7 @@
 #include <thread>
 #include <vector>
 
+#include <boost/algorithm/string/join.hpp>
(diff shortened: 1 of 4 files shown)


aggrand:
I'm gonna move back to draft and use this as a template for a new PR (so I don't have to push to ryan's fork and to more easily base off a newer commit)

rtimmons:
Re-opened as #713

PR #429, opened 2021-03-08
PERF-2175 load phase

jimoleary:
This [patch](https://evergreen.mongodb.com/task/sys_perf_linux_1_node_replSet_insert_llt_patch_fccad833c9081bf0cd61364afdb4ec01ceeb42fa_604152b13e8e8645bfb2ba56_21_03_04_21_36_21)  is for the first commit but there is not a huge change in functionality.

I want to create 2 workloads with difference values for InitialDocumentCount and SecondaryDocumentCount (an in-memory and not in-memory configuration).

I could do an in-memory test phase (the current workload) , drop the database and then run the test again with different values for the not in-memory but I think keeping them separate is simpler.

At the moment, it looks like I will need to make a copy og src/workloads/scale/InsertLLT.yml and update the values. This seems a very verbose approach. Is there a better way to do this or is there any way to move more of the boiler plate into src/phases/scale/LLTPhases.yml?

It would be nice if the ActorTemplates could also live in src/phases/scale/LLTPhases.yml.

### DIFF
--- a/src/phases/scale/LLTPhases.yml
+++ b/src/phases/scale/LLTPhases.yml
@@ -6,11 +6,12 @@ Description: |
 
 GlobalDefaults:
   dbname: &dbname llt
-  DocumentCount: &NumDocs 10000000
+  InitialDocumentCount: &InitialNumDocs 10000000
+  SecondaryDocumentCount: &SecondaryNumDocs 10000000
+
   Document: &Doc
     ts: {^Now: {}}
     caid: {^RandomInt: {min: 0, max: 1000}}
-    #  int0: &int {^RandomInt: {distribution: binomial, t: 200, p: 0.5}}
     cuid: {^RandomInt: {min: 0, max: 100000}}
     prod: {^RandomInt: {min: 0, max: 10000}}
     prid: {^RandomDouble: {min: 0.0, max: 1000.0}}
@@ -25,7 +26,7 @@ LoaderPhase:
     Repeat: 1
     BatchSize: 100
     Threads: 4
-    DocumentCount: { ^Parameter: { Name: "DocumentCount", Default: *NumDocs } }
+    DocumentCount: { ^Parameter: { Name: "DocumentCount", Default: *InitialNumDocs } }
     Database: *dbname
     CollectionCount: 4
     Document: *Doc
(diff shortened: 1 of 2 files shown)


rtimmons:
@aggrand do you have any insight into using `ActorTemplate` to solve the annoyance Jim points out here?

aggrand:
@rtimmons I filed TIG-2912 for the pain point around specifying actor templates externally. After doing that, moving more of the config into the actor template would help reduce copy-paste between workloads, but not eliminate it. (The actor would still need to be instantiated and such.) It sounds like a form of workload-level parameterization or "virtual workloads" generated from the same config file would be necessary to address that.

dalyd:
The code itself looks fine to me, but it's not clear to me that it works. Does that kind of parameterization work?

jimoleary:
> The code itself looks fine to me, but it's not clear to me that it works. Does that kind of parameterization work?

it created 4 collections with the correct number of documents in each. Is this not how it is intended to be used?

dalyd:
> > The code itself looks fine to me, but it's not clear to me that it works. Does that kind of parameterization work?
> 
> it created 4 collections with the correct number of documents in each. Is this not how it is intended to be used?

I just haven't seen the parameterization used that aggressively before :) LGTM.

dalyd (approved):

apodelko (approved):
LGTM - appear to closely follow the design. Only thought that came to my mind: as it appears that we will have multiple workload files, maybe we should name all files, say, starting from LLT to have it grouped together inside the folder?

jimoleary:
> @rtimmons I filed TIG-2912 for the pain point around specifying actor templates externally. After doing that, moving more of the config into the actor template would help reduce copy-paste between workloads, but not eliminate it. (The actor would still need to be instantiated and such.) It sounds like a form of workload-level parameterization or "virtual workloads" generated from the same config file would be necessary to address that.

Thanks, that sounds like it would really help cut down the boiler plate!

But what I'd (also) really like is something like TIG-2915.

I'll add the verbose version for the moment and we refactor it later.

> I just haven't seen the parameterization used that aggressively before :) LGTM.

@dalyd fair point! I'll commit this version for now and cut down on the parameters in the next version.

PR #430, opened 2021-03-09
PERF-2176 Create LLT Benchmark Insert Workload

jimoleary:
Renamed file as per AlexPs' suggestion.
On the way to working but open to suggestions on changes.

I find the GlobalRate a bit of an impediment given that it will be shared amongst the 4 threads and really we want the sum of the 4 phases as the metric (assuming this is how GlobalRate works but it seems to be the case).

It would be better IMO if there was something like a local thread rate. @rtimmons is something like this possible? 

The logRotate and profiling are just for debugging and will be removed in the final ticket (PERF-2177).


jimoleary:
I've scheduled a patch [here](https://spruce.mongodb.com/version/6047b8853627e043f4981c72/tasks)

jimoleary:
One feature I noticed while testing 'locally' is that the workload takes a considerable amount of time to start:

```bash
[ec2-user@ip-10-2-0-10 genny]$ mongo --host md0 llt --eval 'db.dropDatabase()' ; ./run-genny workload -- -v debug run --workload-file ./src/workloads/scale/LLTInsert.yml   --mongo-uri mongodb://md0:27017
MongoDB shell version v4.9.0-alpha5-207-gfccad83-patch-604152b13e8e8645bfb2ba56
connecting to: mongodb://md0:27017/llt?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("764deebe-4d6b-408c-915e-9c94c6ad0c00") }
MongoDB server version: 4.9.0-alpha5-207-gfccad83-patch-604152b13e8e8645bfb2ba56
{
	"info" : "database does not exist",
	"ok" : 1,
	"$clusterTime" : {
		"clusterTime" : Timestamp(1615306451, 1),
		"signature" : {
			"hash" : BinData(0,"AAAAAAAAAAAAAAAAAAAAAAAAAAA="),
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1615306451, 1)
}
[info ] [genny.curator       ] Moved existing metrics (presumably from a prior run). cwd=/media/ephemeral0/src/genny existing=build/CedarMetrics moved_to=build/CedarMetrics-2021-03-09T161420Z-8d5e90e8 timestamp=2021-03-09T16:14:20Z
[info ] [genny.curator       ] Starting poplar grpc in the background. command=['/media/ephemeral0/src/genny/build/curator/curator', 'poplar', 'grpc'] cwd=/media/ephemeral0/src/genny timestamp=2021-03-09T16:14:20Z
[curator] 2021/03/09 16:14:20 [p=info]: starting poplar gRPC service at 'localhost:2288'
[2021-03-09 16:23:08.616631] [0x00007f822791c8c0] [info]    Constructing pool with MongoURI 'mongodb://md0:27017/?appName=Genny&maxPoolSize=500'
[2021-03-09 16:23:08.667040] [0x00007f822791c8c0] [info]    Constructing pool with MongoURI 'mongodb://md0:27017/?appName=Genny'
[2021-03-09 16:23:09.161994] [0x00007f8003412700] [debug]   Beginning phase 0
[2021-03-09 16:23:09.163114] [0x00007f8005c17700] [debug]   Ended phase 0
```

@rtimmons would we expect this sort of workload to take ~10 minutes to start? It pegs 1 cpu on the workload client at 100% during this period.

src/workloads/scale/LLTInsert.yml
```diff
@@ -0,0 +1,248 @@
...
+
+  - ActorFromTemplate:
+      TemplateName: LoaderTemplate
+      TemplateParameters:
+        Name: InitialLoadWithIndexes
+        DocumentCount: *SecondaryNumDocs
+        OnlyActiveInPhase: 2
+        Indexes: {}
+
+  # <-----  4 threads @ 100 ops / second in batches of 50
```
apodelko, line 128: Where do we set 100 ops / second? We don't have global rate in InsertTemplate.100.50.

src/phases/scale/LLTPhases.yml
```diff
@@ -22,12 +23,75 @@ GlobalDefaults:
     - keys: {caid: 1, prid: 1, cuid: 1}
     - keys: {prid: 1, ts: 1, cuid: 1}
 
+  FiftyDocumentsList: &50DocumentList
```
apodelko, line 26: Is it the best we can do here? Appears that we are missing some kind of structure here- not sure what it could be here... [I guess it is rather an enhancement suggestion for genny then this code review]

jimoleary: I couldn't see a better way than this (and this is how one of the other workloads does it).

It does look like a genny feature request.

apodelko (approved):
LGTM assuming that it works and we can't make it more elegant at the moment. Numerous ActorFromTemplate look somewhat confusing for rather a straightforward task. As the outcome of the project we probably need some genny enhancements - but for now, I guess, we rather need it to make working.

jimoleary:
> LGTM assuming that it works and we can't make it more elegant at the moment. Numerous ActorFromTemplate look somewhat confusing for rather a straightforward task. As the outcome of the project we probably need some genny enhancements - but for now, I guess, we rather need it to make working.

Still attempting to validate that it works as expected, should be finished today (it is slow going).

Lack of parameterization does indeed make this verbose, confusing and error prone. I think that it will actually be worse in the next  ticket.

There will likely be a number of genny feature requests as a result.

jimoleary:
The average insert rate is correct (the variability is probably due to the way the sampling occurs):

image

Some more work is needed to investigate if this is the best we can do in terms of throughput:

image

PR #431, opened 2021-03-11
TIG-2909 Any test that runs replica should also run replica-all-featu…

dalyd:
…re-flags

### DIFF
--- a/src/workloads/docs/ParallelInsert.yml
+++ b/src/workloads/docs/ParallelInsert.yml
@@ -185,6 +185,7 @@ AutoRun:
     mongodb_setup:
     - replica
     - replica-noflowcontrol
+    - replica-all-feature-flags
   PrepareEnvironmentWith:
     mongodb_setup:
     - replica-delay-mixed
--- a/src/workloads/execution/CreateBigIndex.yml
+++ b/src/workloads/execution/CreateBigIndex.yml
@@ -94,3 +94,4 @@ AutoRun:
     - standalone
     - single-replica
     - replica
+    - replica-all-feature-flags
--- a/src/workloads/execution/PipelineUpdate.yml
+++ b/src/workloads/execution/PipelineUpdate.yml
@@ -240,3 +240,4 @@ AutoRun:
     mongodb_setup:
     - standalone
     - replica
+    - replica-all-feature-flags
--- a/src/workloads/execution/UnionWith.yml
+++ b/src/workloads/execution/UnionWith.yml
@@ -257,4 +257,5 @@ AutoRun:
     mongodb_setup:
     - standalone
     - replica
+    - replica-all-feature-flags
     - shard-lite
--- a/src/workloads/networking/ServiceArchitectureWorkloads.yml
+++ b/src/workloads/networking/ServiceArchitectureWorkloads.yml
@@ -74,3 +74,4 @@ AutoRun:
     mongodb_setup: 
     - replica
     - replica-noflowcontrol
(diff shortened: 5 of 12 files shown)


src/workloads/docs/ParallelInsert.yml
```diff
@@ -185,6 +185,7 @@ AutoRun:
     mongodb_setup:
     - replica
     - replica-noflowcontrol
+    - replica-all-feature-flags
```
dalyd, line 188: This one confused me wiht the PrepareEnvironmentWith. Do I need to do something more here?

rtimmons: Hm maybe. I think this is saying "run with mongodb_setup=replica-delay-mixed whenever the existing value of mongodb_stup is replica-all-feature-flags" so I'm not sure if this workload would ever run with mongodb_setup as anything _other_ than replica-delay-mixed. We now include the generated tasks in DSI artifacts so this would be pretty easy to patch-build (just select genny-auto-tasks on the variant and ensure the task is generated on the variant task-gen task)

I proposed a different syntax for this in [TIG-2841](https://jira.mongodb.org/browse/TIG-2841) that may be worth revisiting.

dalyd: I took some time to look through the code and this code is not doing what it is expected to do. I've removed my addition, and the noflowcontrol setup. 

What this was doing: For both the 3-node-replica and 3-node-replica-no-flowcontrol build variants, this would generate two tasks, using replica-delay-mixed and replica mongodb_setups respectively. That is, the no flow control and the default version were running *identical* tests. I think the PrepareEnvironmentWith.mongodbs_setup wants to be a list of lists, with a list per entry in Requires.mongodb_setup. 

I'm queuing up a patch build. Will link it in presently.

src/workloads/docs/ParallelInsert.yml
```diff
@@ -184,7 +184,6 @@ AutoRun:
   Requires:
     mongodb_setup:
     - replica
-    - replica-noflowcontrol
```
dalyd, line 187: See this comment for why this went away https://github.com/mongodb/genny/pull/431#discussion_r592692086

dalyd (commented):
Patch build https://evergreen.mongodb.com/version/6050cec132f4177273d509f1

src/workloads/docs/ParallelInsert.yml
rtimmons: Thanks for digging. We really didn't know what we wanted it to look like when we added the AutoRun section. I don't think the current syntax is worth augmenting. I'd much rather see something like the proposal in [TIG-2841](https://jira.mongodb.org/browse/TIG-2841) (it's not that much code) which would make what you're after here a lot more intuititive.

dalyd: That sounds fine to me. Thanks Ryan. The patch build ran by the way. I didn't check it closely, but I did see that it ran the genny tasks.  https://evergreen.mongodb.com/version/6050cec132f4177273d509f1

dalyd:
Ping @rtimmons -- did you have any changes for this one? The patch build looked good https://evergreen.mongodb.com/version/6050cec132f4177273d509f1

rtimmons (approved):

dalyd:
evergreen merge

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #437, opened 2021-03-24
TIG-2918 generate arrays

jimoleary:
I renamed arrayGenerator to literalArrayGenerator (mean it is a literal array of one or more generators) so that I could then implement an array generator.

I think that makes sense but I'm happy to take feedback to clarify.

### DIFF
--- a/src/value_generators/test/DocumentGeneratorTestCases.yml
+++ b/src/value_generators/test/DocumentGeneratorTestCases.yml
@@ -61,6 +61,86 @@ Tests:
     GivenTemplate: {a: []}
     ThenReturns: [{a: []}]
 
+  - Name: Array
+    GivenTemplate:
+      a: {^Array: {of: 1, number: 1}}
+    ThenReturns:
+      - {a: [1]}
+      - {a: [1]}
+      - {a: [1]}
+      - {a: [1]}
+      - {a: [1]}
+
+  - Name: Array10
+    GivenTemplate:
+      a: {^Array: {of: 1, number: 10}}
+    ThenReturns:
+      - {a: [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}
+
+  - Name: ArrayRandomLen
+    GivenTemplate:
+      a: {^Array: {of: 1, number: {^RandomInt: {min: 1, max: 5}}}}
+    ThenReturns:
+      - {a: [1]}
+      - {a: [1]}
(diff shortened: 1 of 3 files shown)


rtimmons (approved):
Nice!

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -61,6 +61,86 @@ Tests:
...
+    GivenTemplate:
+      a: {^Array: {of: {"first": {^RandomInt: {min: 10000000000, max: 10000000010}}}, number: 1}}
+    ThenReturns:
+      - {a: [{"first": 10000000001 }]}
+
+  - Name: ArrayInc
+    GivenTemplate:
+      a: {^Array: {of: {^Inc: {start: 10000000000, step: 1}}, number: 5}}
+    ThenReturns:
+      - {a: [10000000002,10000000003,10000000004,10000000005,10000000006]}
```
dalyd, line 107: Why does this start with 10000000002 rather than 10000000000?

dalyd (approved):
I love this! I

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: That is the way that ^Inc works, it uses the ActorId as a multiplier (personally I was surprised by this).

dalyd: Ah, the actorId part. Does it default multiplier to 1? Yes it does. Maybe that should be zero? Thoughts? Either way could you put an explicit "multiplier" in here of zero, or add a comment for future us?

dalyd: @apodelko do you have thoughts on the default value here? I'm currently thinking that it should default to 0 rather than 1, that is it should do nothing unless we explicitly choose to use it, but I could easily be missing something.

apodelko: @jimoleary @dalyd So does ActorId=2 here? Well, I guess it depends on what would be the main use here. If it mainly one thread operation (as appears here), multiplier=0 makes sense. For multi-threads operations multiplier is needed - but probably shouldn't be 1... So we may indeed change it to 0 if it makes more sense for everybody. One digit change...

PR #438, opened 2021-03-25
Fix test invocations in readme; delete outdated lamp readme

vrachev:
Fixing some things I encountered during onboarding


src/lamplib/README.md
```diff
@@ -1,110 +0,0 @@
...
-```
-
-Lint most of the YAML files in the Genny repository,
-including workloads, phases, and `evergreen.yml`.
-
-Make you run the script from the root of the Genny repository
-(so it can pick up on the `.yamllint` linter config file)
-
-
-Script: `genny-auto-tasks`
```
vrachev, line 52: This is the only section that isn't reflected in the general readme. Is this something we should port over? Not sure if anyone is actually running genny-auto-tasks locally.

rtimmons: We can kill it. Auto-tasks is a part of `run-genny`.

rtimmons (approved):

vrachev:
evergreen merge

vrachev:
evergreen merge

PR #439, opened 2021-03-25
PERF-2232 catch and count transaction operation exceptions

jimoleary:
Without this try / catch the new LLT workloads can eixt with something like:

```
[2021-03-25 20:37:53.612019] [0x00007f04e59d3700] [error]   Operation exception Transaction 76771 has been aborted.: generic server error
[2021-03-25 20:37:53.612088] [0x00007f04e59d3700] [debug]   
Dynamic exception type: mongocxx::v_noabi::operation_exception
std::exception::what: Transaction 76771 has been aborted.: generic server error
```

These sort of exceptions are transient and recoverable and shouldn't cause the workload to exit

### DIFF
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -41,6 +41,8 @@
 #include <value_generators/DocumentGenerator.hpp>
 
 namespace genny::actor {
+static const char* const transientTransactionLabel = "TransientTransactionError";
+
 enum class ScanType { kCount, kSnapshot, kStandard, kPointInTime };
 enum class SortOrderType { kSortNone, kSortForward, kSortReverse };
 // We don't need a metrics clock, as we're using this for measuring
@@ -423,10 +425,14 @@ void CollectionScanner::run() {
                     }
                     session.commit_transaction();
                 } catch (const mongocxx::operation_exception& e) {
-                    BOOST_LOG_TRIVIAL(debug) << "Snapshot Scanner operation exception: " << e.what();
-                    auto exceptionsCaught = config->exceptionsCaught.start();
-                    exceptionsCaught.addDocuments(1);
-                    exceptionsCaught.success();
+                    if(e.has_error_label(transientTransactionLabel) ) {
+                        BOOST_LOG_TRIVIAL(debug) << "Snapshot Scanner operation exception: " << e.what();
+                        auto exceptionsCaught = config->exceptionsCaught.start();
+                        exceptionsCaught.addDocuments(1);
+                        exceptionsCaught.success();
+                    } else {
+                        throw e;
+                    }
                 }
             } else if (config->scanType == ScanType::kPointInTime) {
                 pointInTimeScan(_client, config, readClusterTime);
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -398,29 +398,36 @@ void CollectionScanner::run() {
             if (config->scanType == ScanType::kCount) {
                 countScan(config, collections);
             } else if (config->scanType == ScanType::kSnapshot) {
-                mongocxx::client_session session = _client->start_session({});
-                session.start_transaction(*config->transactionOptions);
-                collectionScan(config, collections, _rateLimiter, session);
-                // If a scan duration was specified, we must make the scan
-                // last at least that long.  We'll do this within any
-                // running transaction, so we keep the "long running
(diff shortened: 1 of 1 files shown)


jimoleary:
BF-20586 has been raised for the waterfall failure.

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -398,29 +398,36 @@ void CollectionScanner::run() {
...
+                        if (stop > now) {
+                            auto sleepDuration = stop - now;
+                            auto secs = sleepDuration.count() / (1000 * 1000 * 1000);
+                            BOOST_LOG_TRIVIAL(info)
+                                << "Scanner id: " << this->_index << " sleeping " << secs;
+                            std::this_thread::sleep_for(sleepDuration);
+                        }
                     }
+                    session.commit_transaction();
+                } catch (const mongocxx::operation_exception& e) {
```
dalyd, line 425: Can we narrow the exception here at all? I looked at the docs, and am not sure how much can be done http://mongocxx.org/api/current/classmongocxx_1_1operation__exception.html
Maybe transient transaction error and retryable write error labels? See https://jira.mongodb.org/browse/HELP-14706 for a discussion on error labels.

dalyd (commented):
Should we loop Alex Cameron on to this review also? 

I suspect this fix should exist many places, but not suggesting we change it everywhere.

jimoleary:
@tetsuo-cpp: does this change seem reasonable?

src/cast_core/src/CollectionScanner.cpp
jimoleary: I didn't see a way to narrow it aside from checking the labels.

I had a first version that accepted a list of labels to ignore but that seemed overkill. I'll update it to just check for transient transactions and we can see how we feel about that.

The **collectionScan** function essentially has this try / catch wrapped around internally.

dalyd: Eh, do what seems right to you. I'm asking questions, but frankly don't understand the Mongo exceptions as well as I would like.

dalyd (approved):

src/cast_core/src/CollectionScanner.cpp
jimoleary: Scoping back to use the error label is the safest way forward for the moment. It means that the other workloads will execute exactly the same (given that they have never encountered this error anyway).

Plus it means that any new exceptions will result in failed waterfall tasks (rather than being silently eaten by the actors).

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -398,29 +400,40 @@ void CollectionScanner::run() {
...
-                        std::this_thread::sleep_for(sleepDuration);
+                    session.commit_transaction();
+                } catch (const mongocxx::operation_exception& e) {
+                    if(e.has_error_label(transientTransactionLabel) ) {
+                        BOOST_LOG_TRIVIAL(debug) << "Snapshot Scanner operation exception: " << e.what();
+                        auto exceptionsCaught = config->exceptionsCaught.start();
+                        exceptionsCaught.addDocuments(1);
+                        exceptionsCaught.success();
+                    } else {
+                        throw e;
```
rtimmons, line 434: just `throw;` will rethrow whatever underlying exception type it happens to be without potentially slicing it.

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -398,29 +400,40 @@ void CollectionScanner::run() {
...
-                        BOOST_LOG_TRIVIAL(info)
-                            << "Scanner id: " << this->_index << " sleeping " << secs;
-                        std::this_thread::sleep_for(sleepDuration);
+                    session.commit_transaction();
+                } catch (const mongocxx::operation_exception& e) {
+                    if(e.has_error_label(transientTransactionLabel) ) {
+                        BOOST_LOG_TRIVIAL(debug) << "Snapshot Scanner operation exception: " << e.what();
+                        auto exceptionsCaught = config->exceptionsCaught.start();
+                        exceptionsCaught.addDocuments(1);
+                        exceptionsCaught.success();
```
rtimmons, line 432: Consider adding a TransactionRetries metric. That would change the behavior of the Actor so up to you, but it seems like capturing the transaction retries as distinct would be a useful thing for analysis.

rtimmons (approved):
Minor thing on `throw`. Otherwise LGTM assuming the owners are also 👍🏽

tetsuo-cpp (approved):
LGTM.

PR #441, opened 2021-03-30
PERF-2183 update LLT 

jimoleary:
@tetsuo-cpp I added you as I have changed the CrudActor. This change is additive and backwards compatible. None of the old workloads will use the new collection name generation functionality.

### DIFF
--- a/src/cast_core/include/cast_core/actors/OptionsConversion.hpp
+++ b/src/cast_core/include/cast_core/actors/OptionsConversion.hpp
@@ -222,6 +222,32 @@ struct NodeConvert<mongocxx::options::transaction> {
         return rhs;
     }
 };
-}  // namespace YAML
 
+template <>
+struct NodeConvert<mongocxx::options::update> {
+    using type = mongocxx::options::update;
+
+    static type convert(const Node& node) {
+        type rhs{};
+
+        if (node["BypassDocumentValidation"]) {
+            rhs.bypass_document_validation(node["BypassDocumentValidation"].to<bool>());
+        }
+        if (node["Hint"]) {
+            auto h = node["Hint"].to<std::string>();
+            auto hint = mongocxx::hint(std::move(h));
+            rhs.hint(mongocxx::hint(hint));
+        }
+        if (node["Upsert"]) {
+            rhs.upsert(node["Upsert"].to<bool>());
+        }
+        if (node["WriteConcern"]) {
+            rhs.write_concern(node["WriteConcern"].to<mongocxx::write_concern>());
+        }
+
+        return rhs;
+    }
+};
+
+}
 #endif  // HEADER_1AFC7FF3_F491_452B_9805_18CAEDE4663D_INCLUDED
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -26,6 +26,7 @@
 #include <boost/throw_exception.hpp>
 
 #include <bsoncxx/json.hpp>
+#include <bsoncxx/string/to_string.hpp>
 
 #include <gennylib/Cast.hpp>
 #include <gennylib/MongoException.hpp>
@@ -34,6 +35,7 @@
 
 using BsonView = bsoncxx::document::view;
 using CrudActor = genny::actor::CrudActor;
+using bsoncxx::type;
 
 namespace {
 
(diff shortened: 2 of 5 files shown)


src/cast_core/include/cast_core/actors/OptionsConversion.hpp
```diff
@@ -222,6 +222,32 @@ struct NodeConvert<mongocxx::options::transaction> {
...
 
+template <>
+struct NodeConvert<mongocxx::options::update> {
+    using type = mongocxx::options::update;
+
+    static type convert(const Node& node) {
+        type rhs{};
+
+        if (node["BypassDocumentValidation"]) {
+            rhs.bypass_document_validation(node["BypassDocumentValidation"].to<bool>());
```
rtimmons, line 234: In the future you may wish to use the if-with-initializer C++17ism:

```c++
if (const auto& bypass = node["Bypass"]; bypass) {
  rhs.bypass(bypsas.to<bool>());
}
```

Existing code doesn't use this much so keeping with existing practice is also reasonable.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1044,15 +1048,24 @@ std::string getDbName(const PhaseContext& phaseContext) {
 namespace genny::actor {
 
 struct CrudActor::PhaseConfig {
-    mongocxx::collection collection;
     std::vector<std::unique_ptr<BaseOperation>> operations;
     metrics::Operation metrics;
+    std::optional<std::string> collectionName;
+    std::optional<int64_t> numCollections;
```
rtimmons, line 1054: consider `struct CollectionParams { string name; int num }` and then `std::optional<CollectionParams>`. Same exact binary layout but imho slightly clearer that one optional is present iff the other is.

Bonus you could have a `CollectionParams(const Node&)` ctor and do the throwing there (cheap thrills).

(You could potentially even make the `generateCollectionName` a member-function of CollectionParams. but potentially diminishing returns the further this goes.)

src/workloads/docs/CrudActor.yml
```diff
@@ -46,3 +46,33 @@ Actors:
...
           WriteConcern:
             Level: majority
+  - Repeat: 10
+    Threads: 8
+    # Generate names from Collection0 to Collection<CollectionCount -1>.
+    # The collection names are generated from the current actor id mod CollectionCount. So
+    # for at least one thread per collection the number of Threads must be greater than or equal to CollectionCount.
+    # If Threads < CollectionCount then the names generated will be based on the ActorIds so not all
+    # Collections will be used and it is difficult to predict which collection names will be generated. In this case,
+    # you may want to stick with Collection and explicitly pick the collection that you want.
```
rtimmons, line 56: This warning reads a bit more ominous than it is.

Would it instead make sense to get the number from a distribution rather than mod count? Functionaly not that different I don't think but perhaps a bit easier to reason about / explain / honest about hard-to-predict. - "Each instance gets CollectionX where X \in {1,n] for n=NumCollections".

src/workloads/scale/LLTInsert.yml
```diff
@@ -167,13 +191,12 @@ Actors:
...
         FindOptions:
-          BatchSzie: 1000
-          Hint: _id_ # Currently only support the index name.
+          BatchSize: 1000
+          Hint: _id_
           Comment: SnapshotScannerShort
 
-  # <-----  4 threads @ 25% in batches of 50
 - Name: InsertBenchmark.Low.50
-  Type: MultiCollectionInsert
```
rtimmons, line 176: Ooh so seems like maybe MultiCollectionInsert can be axed?

rtimmons (approved):
I didn't look too closely at the workload yaml and instead focused on the Actor C++. Nothing that would block merge, just a few passing thoughts. Nice change to this Actor.

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
jimoleary: c++17 is ok to use in genny now?

src/cast_core/src/CrudActor.cpp
jimoleary: yeah, that sounds cleaner.

src/workloads/docs/CrudActor.yml
jimoleary: my concern with a distribution was that it may not give you what you want for low numbers of threads (in one of my tests the prng produced 0,0,0, 1 ... which isn't what I wanted.

I'll lighten the text to make it informational rather than a dire warning.

src/workloads/scale/LLTInsert.yml
jimoleary: yes, I will do that now! Less is more. (TBH maybe all the MCX could be removed ).

jimoleary (commented):
thanks ryan!

tetsuo-cpp (approved):
LGTM. `CrudActor` part looks good.

jimoleary (commented):
thanks

dalyd (approved):
I only took a high level look, but I love that view. Very nice to see.

PR #442, opened 2021-03-31
Perf 2239 Set multiplier default value to 0 in ^Inc value generator

apodelko:
One symbol change to set multiplier default value to 0 in ^Inc value generator.

It appears that current multiplier default value 1 in ^Inc value generator is not helpful - if we generate data with a single thread, 0 makes sense. If we generate data with multiple threads, it should be a large value. So after a team discussion it was decided to change the default for multiplier to 0.


rtimmons (approved):
Thanks for fighting the tests to get this in.

jimoleary (approved):
lgtm

dalyd (approved):
LGTM. Some check fell on the evergreen merge. Can just try entering that again.

apodelko:
evergreen merge

apodelko:
evergreen merge

apodelko:
evergreen merge

PR #445, opened 2021-03-31
TIG-2695 add withTransaction to CrudActor and replace startTxn/commitTxn in workloads

vrachev:
I'll probably do some DRY cleanup in crud actor but this should be functional (at least the yaml test works). Running the workloads in a patch now.

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -109,7 +109,7 @@ using OpCallback = std::function<std::unique_ptr<BaseOperation>(const Node&,
                                                                 PhaseContext& context,
                                                                 ActorId id)>;
 
-std::unordered_map<std::string, OpCallback&> getOpConstructors(std::string);
+std::unordered_map<std::string, OpCallback&> getOpConstructors();
 
 struct WriteOperation : public BaseOperation {
     WriteOperation(PhaseContext& phaseContext, const Node& operation)
@@ -946,6 +946,7 @@ private:
  */
 
 struct WithTransactionOperation : public BaseOperation {
+
     WithTransactionOperation(const Node& opNode,
                              bool onSession,
                              mongocxx::collection collection,
@@ -959,27 +960,14 @@ struct WithTransactionOperation : public BaseOperation {
         auto& opsInTxn = opNode["OperationsInTransaction"];
         if (!opsInTxn.isSequence()) {
             BOOST_THROW_EXCEPTION(InvalidConfigurationException(
-                                          "'WithTransaction' requires an 'OperationsInTransaction' node of sequence type."));
+                 "'WithTransaction' requires an 'OperationsInTransaction' node of sequence type."));
         }
-        for (const auto& [k, txnOp] : opsInTxn) {
+        for (const auto&& [k, txnOp] : opsInTxn) {
             createTxnOps(txnOp, context, id);
         }
         if (opNode["Options"]) {
             _options = opNode["Options"].to<mongocxx::options::transaction>();
         }
-    }
-
-    void createTxnOps(const Node& txnOp, PhaseContext& context, ActorId id) {
-        auto opName = txnOp["OperationName"].to<std::string>();
-        auto inTxnOpConstructors = getOpConstructors("withTransaction");
-        auto opConstructor = inTxnOpConstructors.find(opName);
-        if (opConstructor == inTxnOpConstructors.end()) {
-            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
-                                          "Operation '" + opName + "' not supported inside a 'with_transaction' operation."));
-        }
-        auto createOp = opConstructor->second;
-        auto& yamlCommand = txnOp["OperationCommand"];
-        _txnOps.push_back(createOp(yamlCommand, _onSession, _collection, _operation, context, id));
     }
 
     void run(mongocxx::client_session& session) override {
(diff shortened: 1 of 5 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -928,6 +930,85 @@ struct CommitTransactionOperation : public BaseOperation {
...
+        }
+        for (const auto& [k, txnOp] : opsInTxn) {
+            createTxnOps(txnOp, context, id);
+        }
+        if (opNode["Options"]) {
+            _options = opNode["Options"].to<mongocxx::options::transaction>();
+        }
+    }
+
+    void createTxnOps(const Node& txnOp, PhaseContext& context, ActorId id) {
```
aggrand, line 972: Can this be in the private section?

src/cast_core/src/CrudActor.cpp
```diff
@@ -928,6 +930,85 @@ struct CommitTransactionOperation : public BaseOperation {
...
+            : BaseOperation(context, opNode),
+              _onSession{onSession},
+              _collection{std::move(collection)},
+              _operation{operation} {
+        auto& opsInTxn = opNode["OperationsInTransaction"];
+        if (!opsInTxn.isSequence()) {
+            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                                          "'WithTransaction' requires an 'OperationsInTransaction' node of sequence type."));
+        }
+        for (const auto& [k, txnOp] : opsInTxn) {
```
aggrand, line 964: (nit) `const auto&&`, I think forward-referencing in ranged for loops is idiomatic in genny.

aggrand (changes requested):
Great job tackling this! Mostly had a few nitpicks. Main question was regarding capturing the exception when the transaction fails.

Also, it might be a good idea to add a crud actor yaml test [here](https://github.com/mongodb/genny/blob/334f61ce6391ce464c0888180f474a5665095d7b/src/cast_core/test/CrudActorYamlTests.yml#L1). At least as a basic sanity check that the operations inside the transaction end up being performed.

If you want to locally end to end test the metrics, one approach is to use `curator ftdc` to export the metrics data and examine it. What I usually did for `InsertRemove` is do a single iteration, single threaded run then hand-check the data. Not sure if there'd be a good way to force the transaction to fail in order check for multiple "success" outputs for operations inside the transaction. Maybe temporarily adding a thread-sleep inside the callback, then in the mongo shell doing `rs.stepdown()`? (I presume that'll cancel the transaction?) Another approach is to compare the metrics data in patch builds from this PR with the same task on the current waterfall.

src/cast_core/src/CrudActor.cpp
vrachev: 👍

src/cast_core/src/CrudActor.cpp
vrachev: 👍

src/cast_core/src/CrudActor.cpp
```diff
@@ -1028,6 +1110,21 @@ std::unordered_map<std::string, OpCallback&> opConstructors = {
     {"updateMany", baseCallback<BaseOperation, OpCallback, UpdateManyOperation>},
     {"replaceOne", baseCallback<BaseOperation, OpCallback, ReplaceOneOperation>}};
 
+std::unordered_map<std::string, OpCallback&> getOpConstructors(std::string mode) {
```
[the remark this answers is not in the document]
vrachev: Good points. I looked into several options but I ended up just simplifying completely and writing the check for excluded ops in the `createTxnOps` function, and not putting any logic in the `getOpConstructors` function.

src/workloads/docs/CrudActorTransaction.yml
```diff
@@ -22,20 +22,20 @@ Actors:
...
-        - WriteCommand: updateOne
-          Filter: {a: 1}
-          Update: {$set: {a: 5}}
-        Options:
-          Ordered: true
-        # The OnSession: true flag tells the driver to use a logical session. The actor will use the
-        # same session for the duration of a phase.
-        OnSession: true
-    - OperationName: commitTransaction
+        OperationsInTransaction:
```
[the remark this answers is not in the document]
vrachev: I added 

```        
ThrowOnFailure: false  # Transactions are allowed to fail.
RecordFailure: true  # We still want to record the metrics even if the transaction is rolled back.
```

above

vrachev (commented):
Thanks for the review @aggrand! RFAL

> Also, it might be a good idea to add a crud actor yaml test here. At least as a basic sanity check that the operations inside the transaction end up being performed.

I did add a yaml test but you probably missed it in the midst of all the yaml files 🙂

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -572,3 +572,21 @@ Tests:
     OutcomeCounts:
       - Filter: {a: 1}
         Count: 1
+
```
aggrand, line 575: Oof you're right I totally overlooked this as a workload yaml last time, my bad. This looks great!

aggrand (approved):
This LGTM assuming the metrics look right! Great work and clean code :)

vrachev:
Thanks @aggrand, I checked the metrics with curator ftdc and they look correct

vrachev:
evergreen merge

vrachev:
evergreen merge

PR #446, opened 2021-04-01
PERF-2179 add LLT query workload

jimoleary:


src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  Workload to Benchmark the effect of LongLivedTransactions on an Update workload.
```
dalyd, line 4: s/Update/Query/ ?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+
+GlobalDefaults:
```
dalyd, line 6: I assume a lot of this is the same in the other workloads? It's a shame we can't pull out that commonality.

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+    - keys: {price: 1, ts: 1, cuid: 1}     # PDC
+    - keys: {price: 1, cuid: 1}            # Market
+    - keys: {caid: 1, price: 1, cuid: 1}   # Register
+
+  # Loader Config.
+  LoadThreads: &LoadThreads 4
+  LoadBatchSize: &LoadBatchSize 1000
+
+  # In-memory: Database size works out about 12GB.
+  InitialDocumentCount: &InitialNumDocs 100
```
dalyd, line 30: Does this need to be scaled up? Are 100 docs giving you 12 GB?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+  SecondaryDocumentCount: &SecondaryNumDocs 100
+
+  # WIP make a copy of the workload once these values are tuned correctly and
+  # the workload runs consistently.
+  # Not In-memory: Database size works out about 30GB.
+  # InitialDocumentCount: &InitialNumDocs 49000000
+  # SecondaryDocumentCount: &SecondaryNumDocs 1000000
+  CollectionCount: &CollectionCount 4
+
+  PdcQueryOperation: &PdcQueryOperation
```
dalyd, line 40: What does Pdc stand for?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+    OperationName: findOne
+    OperationCommand:
+      Filter: &PdcFilter {price: {'$gte' : {^RandomDouble: { min: 0.0, max: 500.0 }}}}
```
dalyd, line 43: Do you use this anchor anywhere?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+      Options: &PdcQueryOperationOptions
+        Hint: price_1_ts_1_cuid_1
+        Comment: PdcQueryOperation
+
+  MarketQueryOperation: &MarketQueryOperation
+    OperationName: findOne
+    OperationCommand:
+      Filter: &MarketFilter {'price': {'$gte' : {^RandomDouble: { min: 0.0, max: 500.0 }}}}
+      Options: &MarketQueryOperationOptions
+        Hint: price_1_cuid_1
```
dalyd, line 53: Is the hint the only difference between this and the PdcQueryOperation?

src/workloads/scale/LLTQuery.yml
dalyd: Could you add a comment of what these different queries are intending to do and how they stress the system differently?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Threads: *LoadThreads
+        CollectionCount: *CollectionCount
+        Database: *dbname
+        Repeat: 1
+        Document: *Doc
```
dalyd, line 102: Do we want all the docs loaded with ^Now or maybe some predefined range of dates?

src/workloads/scale/LLTQuery.yml
```diff
@@ -0,0 +1,421 @@
...
+  Threads: 1
+  Database: *dbname
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
+      NopInPhasesUpTo: 20
+      PhaseConfig:
+        Repeat: 1
+
+- Name: UpdateBaseline.Low.50
```
dalyd, line 135: I think you want to update the actor name here.

dalyd (changes requested):
I think you have some copy/pasta to fix, and would like more documentation/explanation so I can follow along better.

src/workloads/scale/LLTQuery.yml
jimoleary: I went for now so that the documents are incrementing. I think it is something we could change but I don't see the need at the moment.

src/workloads/scale/LLTQuery.yml
jimoleary: :+1:  more cut and paste, I thought I had fixed this already .

dalyd (approved):

PR #447, opened 2021-04-01
 PERF-2181 LLT remove workloads

jimoleary:
Continuing to add more workloads.

### DIFF
--- a/src/workloads/scale/LLTRemove.yml
+++ b/src/workloads/scale/LLTRemove.yml
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 Description: |
-  Workload to Benchmark the effect of LongLivedTransactions on an Update workload.
+  Workload to Benchmark the effect of LongLivedTransactions on a Remove workload.
 
 GlobalDefaults:
   # These values should match those are the top of LLTPhases.yml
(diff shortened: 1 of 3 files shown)


src/workloads/scale/LLTRemove.yml
```diff
@@ -0,0 +1,424 @@
...
+  Threads: 1
+  Database: *dbname
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0, 2, 4, 6, 8, 10, 12, 14, 16, 18, 20]
+      NopInPhasesUpTo: 20
+      PhaseConfig:
+        Repeat: 1
+
+- Name: UpdateBaseline.Low.50
```
dalyd, line 138: Actor name needs update

dalyd (changes requested):
The delete option part LGTM. Philosophical question of whether we would like changes like that in separate commit from the workload or not. No strong opinion from me. 

Same comments as query workload here  plus one more:
Do we need to insert more documents during this test? Will be run out of documents to remove during the run of this test? The number of documents should be decreasing across this test. How fast do they decrease?

src/workloads/scale/LLTRemove.yml
jimoleary: :+1:  plus the query names etc...

jimoleary (commented):
> The delete option part LGTM. Philosophical question of whether we would like changes like that in separate commit from the workload or not. No strong opinion from me.
> 
In this case, I think it is ok to have this change with the workload. In general, I would probably do it this way anyway (just to keep the number of reviews down). 

> Do we need to insert more documents during this test? Will be run out of documents to remove during the run of this test? The number of documents should be decreasing across this test. How fast do they decrease?

Good question. I think for this test, we will probably need to put another load phase somewhere or add more documents in the initial load (or both). I may leave that to the evaluate ticket. At this stage, I want to get the workloads in place,

jimoleary:
Note: I had gotten ahead of myself on the PR title (PERF-2181 is correct).

dalyd (approved):

jimoleary:
@dalyd FYI I have a workload run that is just about to finish and there are ~18,000,000 docs per collection.

PR #450, opened 2021-04-05
PERF-2102: Update resharding key pattern used in ReshardCollection.yml

visemet:


dalyd (approved):
This will change results, yes? Assuming so, could you put a note in #perf-build-baroning letting the build barons know?

visemet:
It may change the results, yes. There is still the same amount of data to be cloned but it is being redistributed differently than before. Will do about posting a note in the Slack channel.

guoyr (approved):

PR #451, opened 2021-04-05
PERF-2240: 'Create LLT Benchmark Mixed Workload' added LoggingActor

apodelko:
Added LoggingActor to the mixed workload - otherwise it fails.


jimoleary (approved):
Thanks Alex. I'll double check and make sure that this gets added to the other workloads too.

apodelko:
evergreen merge

PR #455, opened 2021-04-21
PERF-2264 Add examples of ^Inc to src/workloads/docs/generators.yml

apodelko:
This is to add information about ^Inc to src/workloads/docs/generators.yml
Would be that enough?

Actually everything below actorString (now, year2015, greaterThan2105, etc) is not in the document sample at the top. Does it still look consistent enough?

### DIFF
--- a/src/workloads/docs/Generators.yml
+++ b/src/workloads/docs/Generators.yml
@@ -16,7 +16,7 @@ Owner: "@mongodb/stm"
 # 	"string1" : "T336d",
 # 	"string2" : "B2A54",
 # 	"string3" : "LChkGtHKBI",
-#   "counter" : 1008,
+# 	"counter" : 1008,
 # 	"choose" : 12,
 # 	"join" : "This is a joined string",
 # 	"ip" : "77.220.130.163",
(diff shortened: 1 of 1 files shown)


src/workloads/docs/Generators.yml
```diff
@@ -16,6 +16,7 @@ Owner: "@mongodb/stm"
 # 	"string1" : "T336d",
 # 	"string2" : "B2A54",
 # 	"string3" : "LChkGtHKBI",
+#   "counter" : 1008,
```
jimoleary, line 19: [nit] indentation.

jimoleary (approved):
lgtm mod small not

src/workloads/docs/Generators.yml
apodelko: Fixed. It was a combination of spaces and tabs that showed right in VS Code - but not in github... Now made it same way as other lines.

src/workloads/docs/Generators.yml
```diff
@@ -90,6 +91,12 @@ Actors:
             # likely. It matches the string generational algorithm in YCSB.
             string3: {^FastRandomString: {length: 10}}
 
+            # increment generatot ^Inc with parameters start (default 1), multiplier (default 0}, and step (default 1)
```
dalyd, line 94: s/generatot/generator/

dalyd (approved):
1 nit

src/workloads/docs/Generators.yml
apodelko: thanks. fixed.

apodelko:
evergreen merge

PR #460, opened 2021-05-03
TIG-2976 Preprocess genny workloads in python

aggrand:
Still need to delete the .cpp code, but thought I'd go ahead and get eyes on the python code. It seems to succeed at fixing the performance issue in local runs.


src/lamplib/src/genny/tasks/genny_runner.py
```diff
@@ -26,6 +28,22 @@ def main_genny_runner(
             raise Exception(f"genny_core not found at {path}.")
         cmd = [path, *genny_args]
 
-        run_command(
-            cmd=cmd, capture=False, check=True, cwd=workspace_root,
-        )
+        # Intercept the workload given to the core binary.
```
aggrand, line 31: This is the part that I'm iffy about; it's pretty hacky. We could add more of this logic to click directly, but then the `workload` subcommand would have some commands in cpp and some in python, with different help messages in different locations, etc. 

I think a longer-term solution might be to just remove the cpp CLI parsing completely.

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -0,0 +1,334 @@
+import os
```
aggrand, line 1: This is mostly a line-by-line translation of workload_parsers.cpp in the genny driver code.

src/phases/scale/MixPhases.yml
```diff
@@ -19,7 +19,7 @@ UpdatePhase:
   Operations:
   - OperationName: updateOne
     OperationCommand:
-      Filter: &filter {id: {^RandomInt: {min: 0, max: *NumDocs}}}
+      Filter: *filter
```
aggrand, line 22: This was a duplicate of an existing anchor.

src/lamplib/src/genny/tasks/genny_runner.py
```diff
@@ -26,6 +28,22 @@ def main_genny_runner(
...
+        with tempfile.TemporaryDirectory() as tempdir:
```
jimoleary, line 32: It would be nice if it was also possible to generate and store the post-processed files. Plus it would make life easier if the post-processed files were in the dsi diagnostics archive.

Otherwise it will be difficult to debug / develop new workloads and actors.

aggrand: I added the `./run-genny evaluate` command which has the same behavior as the `evaluate` subcommand currently part of the core binary, and could be piped to a file. It's also deterministic so there's no additional data from evergreen runs that would be needed if people can generate the file locally. Would that work for your use case?

jimoleary: Ah nice. I was confused by the function name **create_new_actor**. That works for me (assuming it also gets added to the README) .

src/lamplib/src/genny/tasks/genny_runner.py
vrachev: I think this is ok. Not much of a difference in this logic living here vs click. 

As you said, the correct solution would be to move the cpp CLI parsing up to click, in which case we can revisit the logic here.

src/lamplib/src/genny/tasks/genny_runner.py
vrachev: Jim, for the purpose of debugging/developing new workloads, it is probably easier to use the new `./run-dsi evaluate` command that Raiden added. This way you don't need to run the actual workload to see what the processed yaml looks like. 

Nonetheless, it's not a bad idea to add the post-processed files to the artifacts, because the `evaluate` command only does minimal validation. If there is an error in the workload yaml, it would good to have the post-processed file to check exactly where that error is happening.

vrachev (changes requested):
A few small comments, but looks great overall! 

I mostly skimmed the preprocessing/test code since it's just translating the cpp parser/tests + I ran the `evaluate` command locally and the output looks good.

Can you also file a ticket to remove the cpp parsing code? And maybe another ticket to move the cpp cli code over to click just to track that we may want to do that at some point in the future.

rtimmons:
Glad to see this transition, but I think let's also kill the c++ version to prevent zombie code.

I'm happy to give a deeper look if you'd like, but my $0.02 is to do a few before/after diffs. Like run `genny evaluate` on a random collection of workload yamls (all?) before and after this change and ensure no diffs.

Also, consider adding a comment at the top of the generated file indicating that it's generated and by what code and process.

lmk if my input would be helpful otherwise I trust y'all to knock this out.

src/lamplib/src/genny/cli.py
```diff
@@ -116,6 +116,20 @@ def cmake_compile_install(
     )
 
 
+@cli.command(
+    "evaluate", help=("Print the evaluated YAML workload file with minimal validation."),
+)
+@click.argument("workload_path")
+@click.option(
+    "-s", "--smoke", is_flag=True, help="Smoke test, runs every actor phase once.",
```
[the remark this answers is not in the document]
aggrand: Done!

src/lamplib/src/genny/cli.py
```diff
@@ -116,6 +116,20 @@ def cmake_compile_install(
     )
 
 
+@cli.command(
+    "evaluate", help=("Print the evaluated YAML workload file with minimal validation."),
```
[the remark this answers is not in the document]
aggrand: I'm confused, there's the argument below this line for the `workload_path`, is there another filepath that should be specified too?

src/lamplib/src/genny/tasks/genny_runner.py
rtimmons: Lurking. Fwiw any files in build/WorkloadOutput get included. Could just put it in there in a place that makes sense.

aggrand: Ooh cool that's the fancy conventional locations work results. Will try that, thanks!

aggrand:
Removed the .cpp code. There were a decent number of cmake unit tests that depended on workload preprocessing, but they were basically testing workload preprocessing itself so I went ahead and made python tests for them. 

There's pretty good existing test coverage of preprocessing behavior (though the tests were translated). I also manually checked some more complicated workloads like MixedWorkloads and they seemed good! Running a [patch](https://spruce.mongodb.com/version/6095a9d39ccd4e190cb69e86/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) of Genny tasks on standalone and will compare the runtimes/outputs/general behavior with the current waterfall.

PR #461, opened 2021-05-04
PERF-2185 LLTMixed re-designed to run in different threads

apodelko:
PERF-2185 LLTMixed re-designed to run different types of operations in different threads.

[The 4.4 patch with the latest yml file](https://spruce.mongodb.com/task/sys_perf_linux_1_node_replSet_llt_mixed_patch_bba00a925c257e4ae8c5a6ec13b66890f8433302_609182ae2fbabe0ca85494c0_21_05_04_17_27_12/logs?execution=0)


apodelko:
Full results for the previous version of the LLTMixed.yml posted in [PERF-2185](https://jira.mongodb.org/browse/PERF-2185). A few changes made since to synchronize with latest Jim's changes (scan timing, no client name).

src/workloads/scale/LLTMixed.yml
```diff
@@ -241,56 +241,81 @@ Actors:
         LogEvery: 15 minutes
         Blocking: None
 
-- Name: MixedBaseline.Low
+- Name: Low.Short.Insert
   Type: CrudActor
   Threads: *LowThreads
```
dalyd, line 246: One of the two Threads (line 246 and 253) is unused. I think it's the one within the phase, but not sure. 
I think. 
They would have to be the same, regardless.

src/workloads/scale/LLTMixed.yml
```diff
@@ -241,56 +241,81 @@ Actors:
...
   Phases:
     OnlyActiveInPhases:
-      Active: [5]
+      Active: [5,7]
```
dalyd, line 249: Could you add a comment to the effect of "The test is run in phase 5 (without snapshot read) and phase 7 (with snapshot read)". Assuming that is correct and it hasn't been stated some place i can't see because github PRs are "interesting".

src/workloads/scale/LLTMixed.yml
```diff
@@ -337,8 +369,57 @@ Actors:
         Database: *dbname
         Duration: *SnapshotScannerMediumDuration
         Blocking: none
-        Operations: *Operations
+        Operations: *InsertOps
 
+- Name: Low.Medium.Query
```
dalyd, line 374: How come this is only in one phase but the previous was in 2 phases?

src/workloads/scale/LLTMixed.yml
```diff
@@ -241,56 +241,81 @@ Actors:
...
+        Operations: &InsertOps
           # Create
           - *InsertOperation
           - *InsertOperation
           - *InsertOperation
           - *InsertOperation
           - *InsertOperation
           - *InsertOperation
+
+- Name: Low.Short.Query
```
dalyd, line 267: Could you explain the naming convention for this somewhere in this file?

src/workloads/scale/LLTMixed.yml
apodelko: It looks like we have a bug here - whenever you place it, you get an error. Only when you have it in two places, it works. I didn't checked now, but it is what I faced before. 

And yes, they are supposed to be the same everywhere.

src/workloads/scale/LLTMixed.yml
apodelko: Baseline is only for the "short" duration for both low and high load. So all "shorts" are in two phases (without and with scanner). "Mediums" and "longs" are in one phase.

src/workloads/scale/LLTMixed.yml
```diff
@@ -241,56 +241,81 @@ Actors:
         LogEvery: 15 minutes
         Blocking: None
 
-- Name: MixedBaseline.Low
+- Name: Low.Short.Insert
```
jimoleary, line 244: these names are different to the other workloads.

That is Low.Short.Insert v InsertBenchmark.Low.Medium.

jimoleary (changes requested):
The naming and merging of phases in this workload is now different to the other 4. If we are going to rename them (and merge benchmarks / baselines), then also change the other 4 workloads to make them all consistent.

I think I prefer separate phases as the names will be "Low.Short.Insert" and Low.Baseline.Insert rather than Low.Short.Insert.5.5 and Low.Short.Insert.7.7.

Also note these names are opposite the other workloads (Low.Short.Insert v InsertBenchmark.High.Short).

Hopefully when TIG-2976 is finished we may be able to refactoring to remove a lot of duplication.

src/workloads/scale/LLTMixed.yml
```diff
@@ -148,11 +148,11 @@ GlobalDefaults:
           Level: majority
 
   # Low Baseline / Benchmark
-  LowGlobalRate: &LowGlobalRate 400 per 1 second
+  LowGlobalRate: &LowGlobalRate 100 per 1 second
```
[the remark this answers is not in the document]
apodelko: No, I tried match the load in the previous version of LLTMixed. As we have 4 actors now instead of one, I used a quarter of the original rate. We are pretty close to the max - user_time got to 7.5 cpus at its peak (and there was a peak there) and around 5 cpu at the beginning/end of high load (of 8 cpus we do use). I will run more - but I guess it would be more stress testing and probably should be separate from this PR.

apodelko:
All names are changed to match other tests. Baseline and Benchmark split. Some comments added. If I missed anything, please let me know.

apodelko:
Doubled the rate and some minor changes. It looks like it still handles it.

[The patch](https://spruce.mongodb.com/version/6092b70b7742ae3059c12cc9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

See some charts in [PERF-2185](https://jira.mongodb.org/browse/PERF-2185)

jimoleary (approved):
Thanks Alex, I think that this version makes most sense since the drop off is more prominent.

It will be interesting to see what the graphs look like once the continuous scan change is also committed.

src/workloads/scale/LLTMixed.yml
dalyd: Asking in slack. I'd point out that mixed_workloads_genny uses CrudActor and only sets it at the Actor level https://github.com/mongodb/genny/blob/ac2c06a1ec477c569ea27f818a549b38cad08e01/src/workloads/scale/MixedWorkloadsGenny.yml

dalyd (approved):

src/workloads/scale/LLTMixed.yml
dalyd: https://mongodb.slack.com/archives/C01VD0LQZED/p1620307395080700

dalyd: Raiden and Ryan agree that setting it within the phase doesn't do anything and TIG-2766 should catch such settings in the future. Can you pull out the threads from the phases, and just sanity check that it still runs?

apodelko:
@dalyd Let me merge this PR and I will investigate threads settings separately (as, anyway, it is across all scripts). It appears even more funny than I thought - but let me run a few experiments first.

apodelko:
evergreen merge

apodelko:
Oops.. Looks like we got conflicts. Looks like git got confused with so many changes - trying to figure out. Rebase even with manual resolution somehow doesn't work.

jimoleary:
@apodelko:  Still lgtm in case you were waiting on a new review.

apodelko:
evergreen merge

PR #464, opened 2021-05-06
PERF-2276 Performance workload for window functions

jlap199:


src/workloads/execution/SlidingWindows.yml
```diff
@@ -0,0 +1,138 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{$setWindowFields: {
+                partitionBy: "$part",
+                sortBy: {time: 1},
+                output: {avg: {$avg: {input: "$temp", documents: [-2, 2]}}}}}]
+          cursor: {batchSize: 10000}
+          allowDiskUse: true
```
nzolnierzmdb, line 54: Was this required/do we actually need to spill for this workload? I ended up filing a different ticket to test this specifically so it may be better to keep this one in-memory

src/workloads/execution/SlidingWindows.yml
```diff
@@ -0,0 +1,138 @@
...
+    - OperationMetricsName: MovingAvgPositionBased
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{$setWindowFields: {
+                partitionBy: "$part",
+                sortBy: {time: 1},
+                output: {avg: {$avg: {input: "$temp", documents: [-2, 2]}}}}}]
+          cursor: {batchSize: 10000}
```
nzolnierzmdb, line 53: Can you use the BatchSize parameter at the top here?

src/workloads/execution/SlidingWindows.yml
```diff
@@ -0,0 +1,138 @@
...
+    Operations:
+    - OperationMetricsName: MovingAvgPositionBased
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{$setWindowFields: {
+                partitionBy: "$part",
+                sortBy: {time: 1},
+                output: {avg: {$avg: {input: "$temp", documents: [-2, 2]}}}}}]
```
nzolnierzmdb, line 52: This isn't quite right, the 'documents' spec should be embedded in a 'window' object. IIRC, we've been bitten by this in the past, where genny appears to be running benchmarks just fine but in fact the commands are asserting. Can you double check what happens locally?

src/workloads/execution/SlidingWindows.yml
```diff
@@ -0,0 +1,138 @@
...
+                output: {rate: {$derivative: {input: "$temp"}, window: {documents: [-500, 0]}}}}}]
+          cursor: {batchSize: 10000}
+          allowDiskUse: true
+
+AutoRun:
+  Requires:
+    mongodb_setup:
+    - standalone
+    - replica
+    - replica-all-feature-flags
```
nzolnierzmdb, line 138: Can we add shard-lite here? Also I don't think we need the -all-feature-flags version given WF is on by default, but I guess it can't hurt?

nzolnierzmdb (changes requested):

src/workloads/execution/SlidingWindows.yml
jlap199: When I was testing I noticed some of the runs were failing without this. But that might have been with a very large dataset.

src/workloads/execution/SlidingWindows.yml
jlap199: Sure.

src/workloads/execution/SlidingWindows.yml
jlap199: It actually runs without error.

src/workloads/execution/SlidingWindows.yml
```diff
@@ -0,0 +1,138 @@
...
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 100000
+    BatchSize: 10000
```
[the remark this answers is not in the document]
jlap199: I did, but perhaps it's not necessary for a benchmark.

src/workloads/execution/SlidingWindows.yml
jlap199: I still get memory errors when running without spill to disk.

jlap199: I reduced the number of partitions in order to remove the need for spill.

src/workloads/execution/SlidingWindows.yml
jlap199: Updated the queries to the current valid syntax.

src/workloads/execution/SlidingWindows.yml
```diff
@@ -10,11 +10,11 @@ Actors:
     Database: &db test
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 100000
-    BatchSize: 10000
+    DocumentCount: 10000
```
nzolnierzmdb, line 13: Can you drop a comment here that this is chosen specifically to avoid spilling to disk?

src/workloads/execution/SlidingWindows.yml
```diff
@@ -10,11 +10,11 @@ Actors:
...
+    BatchSize: &batchSize 10000
     Document:
-      part: {^RandomInt: {min: 1, max: 10}}
-      time: {^RandomDate: {min: 1588626902000, max: 1620162902000}}
+      partitionKey: {^RandomInt: {min: 1, max: 5}}
```
nzolnierzmdb, line 16: I'm surprised you needed to change this, since having less partitions conceptually means that each partition has more documents and is more likely to spill. But if it works then keep it!

nzolnierzmdb (approved):
Looks good to me, thanks Joshua!

PR #467, opened 2021-05-13
TIG-2875 ValGenTestCase supports ThenExecuteAndIgnore

vrachev:


src/value_generators/test/DocGenYamlTestCaseRunner.cpp
```diff
@@ -32,21 +33,33 @@ class ValGenTestCase {
             msg << "Can't have ThenReturns and ThenThrows in '" << toString(node) << "'";
             throw std::invalid_argument(msg.str());
         }
+        if (_thenExecuteAndIgnore && _expectedExceptionMessage) {
```
aggrand, line 36: I think this is still leaving open the possibility of someone putting `ThenReturns` and also `ThenExecuteAndIgnore`, which doesn't seem to make sense. Should we combine this with the previous if-statement and check that only one of the three is set?

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -606,15 +606,10 @@ Tests:
...
 
-  # Now moves with time so would always fail until PERF-2086 is available.
-#  - Name: Now
-#    GivenTemplate:
-#      date: {^Now: {}}
-#    ThenReturns:
-#    # The $date notation is extended json
-#    # (see https://docs.mongodb.com/manual/reference/mongodb-extended-json/#bson.Date).
-#    # Check with linux command line `date --utc -d@$((1453689676330 / 1000.0))  +"%Y-%m-%dT%H:%M:%S.%3N"`
-#    - { "date" : { "$date" : { "$numberLong" : "1453689676330" } } }
```
aggrand, line 617: Should we still leave a comment referencing PERF-2086? Since that ticket is on the backlog product perf might want to update this again if they get the chance to do that ticket.

aggrand (approved):
Had a note about the guardrails and ticket cross-referencing, but otherwise LGTM!

src/value_generators/test/DocumentGeneratorTestCases.yml
jimoleary: _ThenExecuteAndIgnore_ fixes this issue for me. I would be happy to just close PERF-2086 out (it is a nice to have IMO but a very small edge case).

src/value_generators/test/DocGenYamlTestCaseRunner.cpp
vrachev: Yeah I thought about this too and couldn't figure out a nifty way of doing a 3-way XOR (Though https://stackoverflow.com/questions/6228252/xor-with-3-values has some fun ideas of assigning 1 to true values and 0 to false values and throwing if the sum != 1) 

I went ahead and added a 3rd if statement for `ThenReturns` and `ThenExecuteAndIgnore` which also does the job but is a bit clunky

vrachev:
evergreen merge

PR #468, opened 2021-05-13
PERF-2300: Add expressive queries to sys-perf workload

laplab:
[First Evergreen run](https://evergreen.mongodb.com/version/609d25fc2fbabe6ccd812b53)

[Evergreen for commit f613ed8](https://evergreen.mongodb.com/version/60a4d01fa4cf47685f54cf12)

### DIFF
--- a/src/workloads/execution/ExpressiveQueries.yml
+++ b/src/workloads/execution/ExpressiveQueries.yml
@@ -1,0 +1,207 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+
+Actors:
+
+# Phases:
+# 1. Insert data
+# 2. Fsync
+# 3. Plain query (without index or projection)
+# 4. Query with projection, but without index
+# 5. Create index
+# 6. Fsync
+# 7. Query with projection and with index
+# 8. Drop collection
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1 # Loader doesn't support multiple threads for a single collection
+  Phases:
+  - Repeat: 1
+    Database: &DB test
+    Threads: 1
+    CollectionCount: 1 # Collection name will be Collection0, this is not configurable.
+    DocumentCount: 1000000 # 1M
+    BatchSize: 1000
+    Document:
+      region: {^RandomInt: {min: 0, max: 1999}}
+      firstname: {^RandomString: {length: {^RandomInt: {min: 3, max: 10}}}}
+      lastname: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}}
+      cell: {^RandomInt: {min: 1111111111, max: 9999999999}}
+      email: {^RandomString: {length: {^RandomInt: {min: 10, max: 30}}}}
+      dob: {^RandomDate: {min: "1930-01-01", max: "2016-12-31"}}
+      gender: {^Choose: {from: ["Male", "Female", "Other"]}}
+      address: {
+        number: {^RandomInt: {min: 1, max: 9999}},
+        street: {^RandomString: {length: {^RandomInt: {min: 30, max: 60}}}},
+        city: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}},
+        state: {^Choose: {from: ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY"]}},
+        zip: {^RandomString: {length: 6}},
+        location: {position: {type: "Point", coordinates: [
+          {^RandomDouble: {min: -118.668469, max: -82.062023}},
+          {^RandomDouble: {min: 32.924436, max: 41.679986}}
+        ]}}
+      }
+      policies: {^Array: {number: {^RandomInt: {min: 3, max: 6}}, of: {^Choose: {from: [
+        {
+          policyType: "auto",
+          policyNum: {^RandomString: {length: 30}},
+          nextRenewalDt: {^RandomDate: {min: "2017-01-01", max: "2017-12-31"}},
+          year: {^RandomInt: {min: 1980, max: 2017}},
+          value: {^RandomInt: {min: 500, max: 500000}},
+          model: {^RandomString: {length: 30}},
+        },
+        {
+          policyType: "home",
+          policyNum: {^RandomString: {length: 30}},
+          nextRenewalDt: {^RandomDate: {min: "2017-01-01", max: "2017-12-31"}},
+          year: {^RandomInt: {min: 1970, max: 2017}},
+          value: {^RandomInt: {min: 500, max: 10000000}},
+          address: {
+            number: {^RandomInt: {min: 1, max: 9999}},
+            street: {^RandomString: {length: {^RandomInt: {min: 30, max: 60}}}},
+            city: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}},
+            state: {^Choose: {from: ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY"]}},
+            zip: {^RandomString: {length: 6}},
+            location: {position: {type: "Point", coordinates: [
+              {^RandomDouble: {min: -118.668469, max: -82.062023}},
+              {^RandomDouble: {min: 32.924436, max: 41.679986}}
+            ]}}
+          }
+        },
+        {
+          policyType: "life",
+          policyNum: {^RandomString: {length: 30}},
+          nextRenewalDt: {^RandomDate: {min: "2017-01-01", max: "2017-12-31"}},
+          insured_person: {
+            firstname: {^RandomString: {length: {^RandomInt: {min: 3, max: 10}}}},
+            lastname: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}},
+            yob: {^RandomDate: {min: "1930-01-01", max: "2016-12-31"}},
+            smoking: {^RandomInt: {min: 0, max: 1}}
+          }
+        }
+      ]}}}}
+  - &Nop {Nop: true}
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: ManageDatabase
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+  - *Nop
+  - Repeat: 5
+    # Duration: 60 seconds
+    Database: *DB
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        createIndexes: Collection0
+        indexes:
+        - key:
+            address.state: 1
+            policies.policyType: 1
+            policies.insured_person.smoking: 1
+            gender: 1
+            dob: 1
+          name: usefulIndex
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+  - Repeat: 1
+    Database: *DB
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        drop: Collection0
+
+- Name: RunQueries
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryNoProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
+          gender: "Female",
+          dob: {
+            $gte: "1990-01-01",
+            $lte: "1990-12-31"
+          },
+          address.state: "UT",
+          policies: {
+            $elemMatch: {
+              policyType: "life",
+              insured_person.smoking: 1
+            }
+          }
+        }
+  - Repeat: 10
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryWithProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
+          gender: "Female",
+          dob: {
+            $gte: "1990-01-01",
+            $lte: "1990-12-31"
+          },
+          address.state: "UT",
+          policies: {
+            $elemMatch: {
+              policyType: "life",
+              insured_person.smoking: 1
+            }
+          }
+        }
+        projection: {_id: 0, firstname: 1, lastname: 1, dob: 1}
+  - *Nop
+  - *Nop
+  - Repeat: 10
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryWithProjectionWithIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
+          gender: "Female",
+          dob: {
+            $gte: "1990-01-01",
+            $lte: "1990-12-31"
+          },
+          address.state: "UT",
+          policies: {
+            $elemMatch: {
+              policyType: "life",
+              insured_person.smoking: 1
+            }
+          }
+        }
+        projection: {_id: 0, firstname: 1, lastname: 1, dob: 1}
+  - *Nop
--- a/src/workloads/execution/ExpressiveQueries.yml
+++ b/src/workloads/execution/ExpressiveQueries.yml
@@ -177,3 +177,11 @@ Actors:
         find: Collection0
         filter: *filter
         projection: *projection
+
+AutoRun:
+  Requires:
+    mongodb_setup:
+    - atlas
+    - replica
+    - single-replica
+    - standalone
--- a/src/workloads/execution/ExpressiveQueries.yml
+++ b/src/workloads/execution/ExpressiveQueries.yml
@@ -14,7 +14,9 @@ Actors:
 
 - Name: InsertData
   Type: Loader
-  Threads: 1 # Loader doesn't support multiple threads for a single collection.
+  # Loader doesn't support multiple threads for a single collection.
+  # TODO: TIG-2938 Use multiple threads to speedup insertion.
+  Threads: 1
   Phases:
   - Repeat: 1
     Database: &DB test
@@ -122,6 +124,7 @@ Actors:
   - *Nop
   - *Nop
 
+# TODO: TIG-3050 Switch to CrudActor once projection argument is supported for find operation.
 - Name: RunQueries
   Type: RunCommand
   Threads: 1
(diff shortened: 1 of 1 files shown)


src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+  - *Nop
+  - Repeat: 5
+    # Duration: 60 seconds
```
laplab, line 106: This was copied from `CreateIndexPhase.yml`, I am not sure if we actually need the `Duration`. Also, it looks like `Repeat: 1` would be enough.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryNoProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
+          gender: "Female",
+          dob: {
+            $gte: "1990-01-01",
```
laplab, line 150: This is treated by parser as a string, not a date. Does yaml workload description support `ISODate("...")` format?

apodelko: @jimoleary any thoughts on how we need to specify date in the filter?

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+  - *Nop
+  - Repeat: 5
```
apodelko, line 105: Why do we need Repeat: 5 here? Wouldn't 1 be enough? [I guess CreateIndexPhase do it intentionally several times to get accurate timing for index creation  - my understanding that we don't care about that here]

apodelko (approved):
Overall LGTM (assuming that it works and create data as specified) if we just want to run a query several times (10). 

The setups are pretty powerful, so 10 executions may face some transient effects. Maybe create everything and do some warmup first to make sure that everything got into the cache?

I'd probably did many more executions - perhaps it is worth to check if it makes a difference.

src/workloads/execution/ExpressiveQueries.yml
jimoleary: Maybe '$gte:: { "$date" : { "$numberLong" : "631152000000" } }' or  '$gte: { "$date" : "1990-01-01" }'. See [bson date](https://docs.mongodb.com/manual/reference/mongodb-extended-json/#mongodb-bsontype-Date)

The following _{^RandomDate: {min: "1990-01-01", max: "1990-01-01T00:00:00.001"}}_ will work but it's not very obvious.

If the mongodb bsontype doesn't work we should raise a ticket for it.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
```
jimoleary, line 102: there is a [QuiesceActor](https://github.com/mongodb/genny/blob/ac2c06a1ec477c569ea27f818a549b38cad08e01/src/workloads/docs/QuiesceActor.yml) if that is what you are trying to do here.

src/workloads/execution/ExpressiveQueries.yml
jimoleary: [PERF-2084](https://jira.mongodb.org/browse/PERF-2084)  was created for a new generator. We should maybe consider scheduling it.

For now generating a random point date ( _{^RandomDate: {min: "1990-01-01", max: "1990-01-01T00:00:00.001"}}_ ) is the only way forward that I can see.

Supporting extended json would be a better fix but that is a large piece of work (if it is even possible at the moment).

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+# 3. Plain query (without index or projection)
+# 4. Query with projection, but without index
+# 5. Create index
+# 6. Fsync
+# 7. Query with projection and with index
+# 8. Drop collection
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1 # Loader doesn't support multiple threads for a single collection
```
dalyd, line 18: FYI: This has come up recently and led to https://jira.mongodb.org/browse/TIG-2938. Consider adding a comment and/or voting for the ticket. Every bit of evidence helps when we all try to schedule and prioritize these things.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  Phases:
+  - Repeat: 1
+    Database: &DB test
+    Threads: 1
+    CollectionCount: 1 # Collection name will be Collection0, this is not configurable.
+    DocumentCount: 1000000 # 1M
```
dalyd, line 24: I *think* Genny supports smart numbers, so you may be able to write 1M or 1e6. Looks like 1e6 should work https://github.com/mongodb/genny/blob/ac2c06a1ec477c569ea27f818a549b38cad08e01/src/workloads/selftests/GennyOverhead.yml#L32

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+    BatchSize: 1000
+    Document:
+      region: {^RandomInt: {min: 0, max: 1999}}
+      firstname: {^RandomString: {length: {^RandomInt: {min: 3, max: 10}}}}
```
dalyd, line 28: Cool -- love the nesting :) RandomString also supports an alphabet if you want to constrain it. Two other things may be useful for you:
1. ^join. This contatenates multiple strings. If you care about compressibility you can join a random string with a string of a repeated character. 
2. ^Choose has a string variant and can be nested with the join. 
3. There's a thing called ^FastRandomString -- it's semantically the same as ^RandomString, but it is faster at the cost of lower quality of randomness (not all letters will be equally likely). The quality difference often does not matter.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+      lastname: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}}
+      cell: {^RandomInt: {min: 1111111111, max: 9999999999}}
+      email: {^RandomString: {length: {^RandomInt: {min: 10, max: 30}}}}
+      dob: {^RandomDate: {min: "1930-01-01", max: "2016-12-31"}}
+      gender: {^Choose: {from: ["Male", "Female", "Other"]}}
+      address: {
+        number: {^RandomInt: {min: 1, max: 9999}},
+        street: {^RandomString: {length: {^RandomInt: {min: 30, max: 60}}}},
+        city: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}},
+        state: {^Choose: {from: ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY"]}},
```
dalyd, line 38: Cool! You don't need to, but if you wanted to you could define the states list as a yaml anchor elsewhere.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+        zip: {^RandomString: {length: 6}},
+        location: {position: {type: "Point", coordinates: [
+          {^RandomDouble: {min: -118.668469, max: -82.062023}},
+          {^RandomDouble: {min: 32.924436, max: 41.679986}}
+        ]}}
+      }
+      policies: {^Array: {number: {^RandomInt: {min: 3, max: 6}}, of: {^Choose: {from: [
```
dalyd, line 45: I love this!

src/workloads/execution/ExpressiveQueries.yml
dalyd: Ah, you do use it again later. You should at least define a Yaml anchor for it here, and reuse below. 
That would make it here:

> state: {^Choose: {from: &states ["AL"...]}}

and then future uses would be 
> state: {^Choose: {from: *states}}

You could go further and do
>state: &states {^Choose: {from: ["AL", ....]}}

Then future uses would be 
> state: *states

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+          policyNum: {^RandomString: {length: 30}},
+          nextRenewalDt: {^RandomDate: {min: "2017-01-01", max: "2017-12-31"}},
+          year: {^RandomInt: {min: 1970, max: 2017}},
+          value: {^RandomInt: {min: 500, max: 10000000}},
+          address: {
+            number: {^RandomInt: {min: 1, max: 9999}},
+            street: {^RandomString: {length: {^RandomInt: {min: 30, max: 60}}}},
+            city: {^RandomString: {length: {^RandomInt: {min: 5, max: 10}}}},
+            state: {^Choose: {from: ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY"]}},
+            zip: {^RandomString: {length: 6}},
```
dalyd, line 65: Consider {^RandomString: {length: 6, alphabet: "0123456789"}}. I think that would be closer to what you intend.

src/workloads/execution/ExpressiveQueries.yml
dalyd: +1 here -- you should use the QuiesceActor here instead.

src/workloads/execution/ExpressiveQueries.yml
dalyd: 1 should be fine. Also, the loader is perfectly happy to add an index for you. ~Do you need the index build in a different phase than the loader?~ Consider adding a comment repeating that you need to add the comment after running the queries without the index

src/workloads/execution/ExpressiveQueries.yml
dalyd: 1 should be fine. See discussion above.

src/workloads/execution/ExpressiveQueries.yml
dalyd: I believe Genny uses stod, which is documented here https://en.cppreference.com/w/cpp/string/basic_string/stof
The best documentation in Genny seems to be the test cases today. https://github.com/mongodb/genny/blob/e8ebcbee0d5b9e99a97cdfffa75e70ed4b25078e/src/gennylib/test/conventions_test.cpp

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  - *Nop
+  - Repeat: 1
+    Database: *DB
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        drop: Collection0
+
+- Name: RunQueries
+  Type: RunCommand
```
dalyd, line 135: Any reason you are using RunCommand actor instead of the Crud Actor? It might be more compact with the Crud Actor.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10
```
dalyd, line 140: Consider doing a duration here instead of a repeat, unless the operation is really slow.

src/workloads/execution/ExpressiveQueries.yml
dalyd: After reading more -- I think you have the runCommand actor so you can create the index and do the queries. I'd suggest splitting those parts into separate actors.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryNoProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
```
dalyd, line 147: I'd add a YAML achor here to be used below. 
filter: &filter {....

Then below you can have 
filter: *filter

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+          }
+        }
+  - Repeat: 10
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryWithProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: {
```
dalyd, line 168: If you define &filter anchor above, here you can replace lines 168-181 with
filter: *filter

It also means you can guarantee your queries use exactly the same filter.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+              insured_person.smoking: 1
+            }
+          }
+        }
+  - Repeat: 10
+    Database: *DB
+    Operations:
+    - OperationMetricsName: QueryWithProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
```
dalyd, line 166: Can define a YAML anchor here and use it for the last operation
OperationCommand: &OpQueryWithProject

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+            dob: 1
+          name: usefulIndex
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+  - Repeat: 1
```
dalyd, line 127: Not sure if this is needed -- in production we will tear down the cluster afterwards. Actually, it's better to not drop the collection as we run dbValidate, and there will be nothing to validate if you drop the collection.

dalyd (commented):
I enjoyed reviewing that a little too much :) I love the advanced use of the generators. 
I think this is almost all small stuff. Most of it is suggestions you can take or leave as you see fit.

dalyd:
FYI: https://jira.mongodb.org/browse/DAG-1199 and https://mongodb.slack.com/archives/C01VD0LQZED/p1621019179183300 around how to view your patch results.

src/workloads/execution/ExpressiveQueries.yml
dalyd: It looks like the query itself is over 1 second here, so the 10 takes about 14 seconds. Given all the overheads, there's no problem with running this for a few minutes.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,207 @@
...
+            $elemMatch: {
+              policyType: "life",
+              insured_person.smoking: 1
+            }
+          }
+        }
+        projection: {_id: 0, firstname: 1, lastname: 1, dob: 1}
+  - *Nop
+  - *Nop
+  - Repeat: 10
```
dalyd, line 185: This one is really fast, so running it 10 times only is very limited data.

src/workloads/execution/ExpressiveQueries.yml
laplab: Thanks for the suggestions!
1. Used it for making valid email addresses
2. I was thinking about using a pre-defined list of names, but it seems that random strings would suffice. The workload does not depend on the name values
3. Replaced all `RandomString` with `FastRandomString`

src/workloads/execution/ExpressiveQueries.yml
laplab: I have extracted whole address object into anchor.

src/workloads/execution/ExpressiveQueries.yml
laplab: `RunQueries` does not create index, `ManageDatabase` actor does that (renamed it to `CreateIndex`). I have tried using `CrudActor`, but it seems like it does not support projection argument for find operation. Keeping `RunCommand` for now.

src/workloads/execution/ExpressiveQueries.yml
laplab: If you do not feel strongly about it, I would prefer to leave filter and projection arguments in place. I have switched to the usage of anchors for filter and projection separately.

src/workloads/execution/ExpressiveQueries.yml
laplab: Switched to 1000 and added a duration.

src/workloads/execution/ExpressiveQueries.yml
dalyd: Could you open a TIG ticket for the projection argument support in the CrudActor, and link back to this PR?

src/workloads/execution/ExpressiveQueries.yml
dalyd: That's fine.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -117,38 +119,29 @@ Actors:
...
+  - *Nop
 
 - Name: RunQueries
   Type: RunCommand
   Threads: 1
   Phases:
   - *Nop
   - *Nop
-  - Repeat: 10
+  - Repeat: 100
```
dalyd, line 131: Note, according to the [docs ](https://github.com/mongodb/genny/blob/86e796946f3faea4be63b33b088d0fba93543890/src/workloads/docs/HelloWorld-MultiplePhases.yml) this will run for at least 100 iterations and at least 5 minutes, whichever takes longer. Is that your intent here?

dalyd (approved):
LGTM modulo a question.

src/workloads/execution/ExpressiveQueries.yml
laplab: https://jira.mongodb.org/browse/TIG-3050

src/workloads/execution/ExpressiveQueries.yml
laplab: Thank you for the note! Yes, this was my intention.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -0,0 +1,175 @@
...
+    - OperationMetricsName: QueryNoProjectionNoIndex
+      OperationName: RunCommand
+      OperationCommand:
+        find: Collection0
+        filter: &filter {
+          gender: "Female",
+          dob: {
+            # Currently, using generator is the only way to generate date value with the appropriate BSON type.
+            $gte: {^RandomDate: {min: "1990-01-01", max: "1990-01-01T00:00:00.001"}},
+            $lte: {^RandomDate: {min: "1990-12-31", max: "1990-12-31T00:00:00.001"}}
```
rtimmons, line 144: yikes well this is gross. Can you cut a TIG ticket to support a better syntax and add a TODO here with that ticket number?

Just `{^Date:}` would be fine.  I'm also curious if `{^Verbatim: {$date:}}` (using the bson notation) would work, but not sure if the driver supports that syntax versus native types.

rtimmons (approved):
Nice use of generators. Thanks for working around some of the syntax things. Assuming you've done patch-builds and whatnot this looks good from a genny perspective.

src/workloads/execution/ExpressiveQueries.yml
jimoleary: Yeah this super ugly. There's already a ticket for ^Date (see PERF-2084). The driver doesn't support Extended json yet (which is a shame).

PR #470, opened 2021-05-14
TIG-3041 Add aggregation support to collection scanner.

jimoleary:
This commit makes PERF-2317 easier (and any new workloads could be committed to the waterfalls).

### DIFF
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -184,6 +184,10 @@ struct CollectionScanner::PhaseConfig {
         } else {
             queryCollectionList = true;
         }
+        if (aggregate && filterExpr) {
+            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                                      "Only one of filterExpr and AggregatePipeline can be set."));
+        }
     }
 
     void collectionsFromNameList(const mongocxx::database& db,
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -28,7 +28,6 @@
 #include <bsoncxx/types.hpp>
 #include <mongocxx/client.hpp>
 #include <mongocxx/collection.hpp>
-#include <mongocxx/cursor.hpp>
 #include <mongocxx/database.hpp>
 
 #include <boost/algorithm/string.hpp>
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -238,7 +238,8 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
     bool scanFinished = false;
     auto statTracker = config->scanOperation.start();
     for (auto& collection : collections) {
-        auto docs = [&]() {
+        // Use a lambda to hide calling the correct operation.
+        auto cursor = [&]() {
           if (config->aggregate) {
               return collection.aggregate(session, evaluatePipeline(config), config->aggregateOptions);
           } else {
@@ -253,7 +254,8 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
          * You can still do a find but it'll throw an exception when we iterate.
          */
         try {
-            for (auto& doc : docs()) {
+            // Execute the lambda, iterate over all the docs in the cursor.
+            for (auto& doc : cursor()) {
                 docCount += 1;
                 scanSize += doc.length();
                 if (config->documents != 0 && config->documents == docCount) {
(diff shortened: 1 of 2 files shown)


src/workloads/docs/CollectionScanner.yml
```diff
@@ -70,6 +70,22 @@ Actors:
...
     ScanType: standard
     GlobalRate: 10 per 10 seconds
+  - Duration: 2 minutes
+    # Similar to the previous phase but with an aggregation.
+    Documents: 1000
+    ScanType: standard
+    GlobalRate: 10 per 10 seconds
+    # Count the frequency of each character in array field "a".
+    AggregatePipeline:
+        {array: [{"$addFields":{"a":{"$map":{"input":{"$range":[0,{"$strLenCP":"$a"}]},"in":{"$substrCP":["$a","$$this",1]}}}}},
```
jimoleary, line 80: "array" makes this tivial to implement (and would allow random stages to be generated).

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -219,15 +234,22 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
     bool scanFinished = false;
     auto statTracker = config->scanOperation.start();
     for (auto& collection : collections) {
-        auto filter = config->filterExpr ? config->filterExpr->evaluate()
-                                         : bsoncxx::document::view_or_value{};
-        auto docs = collection.find(session, filter, config->findOptions);
+        auto docs = [&]() {
+          if (config->aggregate) {
+              return collection.aggregate(session, evaluatePipeline(config), config->aggregateOptions);
+          } else {
```
dalyd, line 240: You do one or the other here. I think we should either support running both, or the constructor should throw an error if you set both.

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -219,15 +234,22 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
...
+                                               : bsoncxx::document::view_or_value{};
+              return collection.find(session, filter, config->findOptions);
+          }
+        };
+
         /*
          * Try-catch this as the collection may have been deleted.
          * You can still do a find but it'll throw an exception when we iterate.
          */
         try {
```
dalyd, line 251: I think this code is dead code now. Shouldn't this be in the for loop?

src/cast_core/src/CollectionScanner.cpp
```diff
@@ -219,15 +234,22 @@ void collectionScan(CollectionScanner::PhaseConfig* config,
     bool scanFinished = false;
     auto statTracker = config->scanOperation.start();
     for (auto& collection : collections) {
-        auto filter = config->filterExpr ? config->filterExpr->evaluate()
-                                         : bsoncxx::document::view_or_value{};
-        auto docs = collection.find(session, filter, config->findOptions);
+        auto docs = [&]() {
+          if (config->aggregate) {
+              return collection.aggregate(session, evaluatePipeline(config), config->aggregateOptions);
```
dalyd, line 239: Should this really be a return? Wont' that execute once and then end as oppose to repeating in the for loop?

dalyd (changes requested):

src/cast_core/src/CollectionScanner.cpp
jimoleary: I'm not sure I get what you mean. 
There is a lambda for each iteration of the outer _for (auto& collection : collections) {_  and each lambda returns a cursor so the inner _for (auto& doc : docs()) {_ works as expected.

src/cast_core/src/CollectionScanner.cpp
jimoleary: It should throw an exception. I'll fix that.

src/cast_core/src/CollectionScanner.cpp
jimoleary: I don't get you here, this for loop is required to iteration over the cursor returned by the lambda.

src/cast_core/src/CollectionScanner.cpp
dalyd: My eyes missed the lambda. Could you add a comment for me?

src/cast_core/src/CollectionScanner.cpp
dalyd: I got completely tripped up on the lambda and then somehow thought the previous context had been closed (it hadn't). My bad.

src/cast_core/src/CollectionScanner.cpp
jimoleary: sure, also renamed docs to cursor.

tetsuo-cpp (approved):
LGTM modulo existing comments. Nice work.

dalyd (approved):

PR #471, opened 2021-05-16
Use ensurepip and wheel for more robust installs

rtimmons:
I was trying to install genny on my phone using [iSH](https://ish.app/) which uses a pretty stripped-down version of Arch. I needed `pip` and `wheel` installed. Also the README saying it's ready to go was..a bit too optimistic.

(FWIW the install still bails due to no version of curator and the toolchain, but this at least uncovered a few new assumptions.)


aggrand (approved):
LGTM! Considering that some workloads kill my computer, I'd love to see your phone run it someday.

rtimmons:
evergreen merge

rtimmons:
evergreen merge

PR #472, opened 2021-05-17
TIG-2863 Record Genny's Phase and Workload start/end times

aggrand:
Adds the `GennyInternal.Phase` operation that ends when each phase ends, with a duration of the length of the phase. Also added the `GennyInternal.Workload` operation that ends when the workload ends, which has a duration encompassing the workload (so also implicitly has the start time).

Ended up creating an actor for this, since `Orchestrator` doesn't have access to the metrics subsystem and I figured it'd be better to avoid coupling them. I also considered that we might want to add more internal ops to `GennyInternal` someday. (Maybe even toggle-able for debugging?)

The outputted timestamp being slightly off from UTC time (mentioned in ticket) wasn't actually a problem anywhere; it was just correctly accounting for time zones and poplar outputs the `ts` field in `json`-formatted outputs as microseconds rather than ns.

I didn't add the `canary_` prefix to these timing ops since they're not actually canaries, but serve a functional use in postprocessing. @vrachev I saw you did TIG-2836; I was wondering if you had thoughts on whether adding `GennyInternal` to the exclusion regex would be possible? If not I can add canary to the name.

Will link the DSI companion PR too. Ended up doing all the postprocessing there to keep with the "Genny only outputs ftdc" pattern.


aggrand:
DSI companion: https://github.com/10gen/dsi/pull/894

src/driver/src/DefaultDriver.cpp
```diff
@@ -150,20 +153,20 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
...
 
-    reportMetrics(metrics, workloadName, true, startTime);
+    reportMetrics(metrics, workloadName, "Setup", true, startTime);
 
     // The "ActorStarted" and "ActorFinished" operations are genny internal operations. We want the
     // trend graph to be hidden by default to not confuse users, so we prefix them with "canary_"
     // to hit the CANARY_EXCLUSION_REGEX in https://git.io/Jtjdr
-    auto startedActors = metrics.operation("canary_" + workloadName, "ActorStarted", 0u);
-    auto finishedActors = metrics.operation("canary_" + workloadName, "ActorFinished", 0u);
+    auto startedActors = metrics.operation("canary_" + workloadName, "ActorStarted", 0u, std::nullopt, true);
```
vrachev, line 168: 💅  remove the `canary_` here and directly below so it won't get prefixed twice

aggrand:
Fwiw, considering that `Orchestrator` is outside the `v1` namespace and exposed through the `ActorContext`, and since PhaseLoop is described [here](https://github.com/mongodb/genny/blob/04924c80e19908cb9532760c41d4fa1caaed37fd/src/gennylib/include/gennylib/context.hpp#L393) as opt-in, I think it's arguable whether using it counts as exposing the internals (though the metrics collected here are defs for internal use).

vrachev:
@aggrand Fair point about this not actually exposing genny internals. Mostly it just seems a little clunky to me to have an actor in `cast_core` that's not meant to be used by users (aside from being automatically tacked onto workloads).

src/cast_core/include/cast_core/actors/GennyInternal.hpp
```diff
@@ -0,0 +1,64 @@
...
+    static std::string_view defaultName() {
+        return "GennyInternal";
+    }
+
+private:
+    genny::metrics::Operation _phaseOp;
+
+    /** @private */
+    Orchestrator& _orchestrator;
+    void reportPhase(metrics::clock::time_point finishTime, metrics::clock::time_point startTime);
```
rtimmons, line 58: nit: start before end and `Time` is implicit in the type, so `void report(time_point started, time_point finished)`

src/cast_core/src/GennyInternal.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <gennylib/Cast.hpp>
+#include <gennylib/MongoException.hpp>
+#include <gennylib/context.hpp>
+
+#include <value_generators/DocumentGenerator.hpp>
+
+namespace genny::actor {
+
+void GennyInternal::run() {
```
rtimmons, line 38: The name is not super obvious and I fear it eventually attracting more exotic and unrelated functionality and turning into a junk drawer... Maybe call this Actor `PhaseMetricsCapturer` or something?

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -220,6 +223,13 @@ def _parse_templates(self, templates):
                 template_node["TemplateName"], template_node["Config"], _ContextType.ActorTemplate
             )
 
+    def _parse_actors(self, actors):
+        actor_list = self._recursive_parse(actors)
```
rtimmons, line 227: So..perhaps a hot-take but I actually dislike adding Actors the user didn't explicitly add since I (as a workload author) can't intuit about what's going on with this hidden step. 

Consider requiring the workload yamls to include this new Actor but perhaps with an easier syntax that this preprocesses into regular yaml?

```yaml
Actors:
  - Type: CollectPhaseMetrics
     # Barf on other keys and implicitly add Name and Threads
```

No biggie if this seems weird. Not a huge deal given this will show in `evaluate`.

rtimmons (approved):
This is an interesting approach that I wouldn't have thought of. The actual impl is not as awkward as I would have thought.

I'm not convinced on a separate `internal` subdir within the metrics output directory, but that's an easy thing to change or ignore in the future.

I agree with @vrachev's sentiment about this being an outcast (ha!) in `cast_core`. As far as I reason, there's nothing stopping this from perhaps living in the `gennylib` module and maybe even being in the `genny::v1` c++ namespace (perhaps `genny::v1::actor` to be more consistent with `genny:actor`? idk).

Nice solution.

src/lamplib/src/genny/tasks/preprocess.py
rtimmons: I'd personally steer away from `defaults.yml` for now. Although it's clever idea, I fear having more files to look at to understand things will breed confusion.

Whatever you decide here is fine by me. Easy to change depending on user feedback or future needs.

aggrand:
Updated to separate out the GennyInternal actor and give it a more specific name. I couldn't put it in GennyLib because it's a static library and the producer isn't being initialized, and won't be unless it's referenced. (See more details [here](https://www.bfilipek.com/2018/02/static-vars-static-lib.html). I think we're technically relying on undefined behavior for all this, so we might want to reconsider this mechanism in the future.) I was able to get it to initialize by creating a `cast_internal` library that mirrors `cast_core`. Since it's a dynamic library it worked.

Will check in the slack channel whether anyone objects to renaming the canaries. Nobody is triaging on it so should be fine, but it'll break history. If not I might revert that part.

Also added an `export` option for use in DSI. Ideally after TIG-3059 I'd like to remove curator from DSI totally.

src/lamplib/src/genny/cli.py
```diff
@@ -145,6 +145,29 @@ def evaluate(ctx: click.Context, workload_path: str, output: str, smoke: bool):
     preprocess.evaluate(workload_path=workload_path, smoke=smoke, output=output)
 
 
+@cli.command(
+    "export", help=("Export the given FTDC file to CSV."),
```
vrachev, line 149: nice addition!!

vrachev (approved):
Nice work! LGTM

src/cast_internal/src/PhaseTimingRecorder.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <gennylib/context.hpp>
+
+#include <value_generators/DocumentGenerator.hpp>
+
+namespace genny::v1::actor {
+
+void PhaseTimingRecorder::run() {
+
+    std::optional<metrics::clock::time_point> startTime;
+    metrics::clock::time_point finishTime;
```
rtimmons, line 41: these look unused now, ya?

aggrand: Yup good catch! Will remove.

aggrand:
@vrachev  @rtimmons  The unittests are working on my local machine, but it looks like rhel70 is hitting [this](https://github.com/grpc/grpc/issues/21280) grpc issue. My understanding is that one is linker/compiler dependent, and I guess adding one more dynamic library for `genny_internal` causes the magical twists and turns inside the compiler to cause this to fail on rhel70 specifically. 

Since the `PhaseTimingRecorder` cannot be placed inside `gennylib` either for the reason explained in my above comment, how would y'all feel filing a ticket to update the grpc version to one that isn't bugged, and put `PhaseTimingRecorder` in `cast_core` for the moment with a giant TODO  at the top of the actor saying to move it in that ticket?

aggrand:
evergreen merge

aggrand:
evergreen merge

PR #473, opened 2021-05-18
PERF-2260 move llt workloads to transactions directory

jimoleary:
To deploy to production needs a couple of commits, this first one is a no (well there is a single typo fix) change move to a new directory.

The next set of changes will split the workloads and add baselines to each phase (doubling the number of files). I'll take the opportunity to rename the files and will disable the llt_insert temporarily.

The one after that will create the Greater than cache versions (doubling again).


src/workloads/transactions/LLTMixed.yml
```diff
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 Description: |
-  Workload to Benchmark the effect of LongLivedTransactions on a Remove workload.
+  Workload to Benchmark the effect of LongLivedTransactions on a Mixed workload.
```
jimoleary, line 4: This is the only change, the files were all moved to this new directory.

dalyd (approved):

PR #474, opened 2021-05-18
PERF-2260 low Ltc workloads with baseline per duration, rename phases to sort together

jimoleary:
This commit creates the Low rate / Less than Cache workloads and adds a baseline for each phase. The current one only has a baseline for 1 minute, this doesn't sufficiently allow for comparision between the baseline and benchmark phases.

Additionally the phase names have been rewritten to sort close to each other.
Finally the LLTInsertLowLtc.yml is being temporarily disabled.

The next PR will add templates for the pahses to reduce the amount of duplication.
A final PR will then add the LowGtc, HighLtc and HighGtc workloads.


apodelko (approved):
Code LGTM.

PR #475, opened 2021-05-18
Fix onSession handling for createIndex

rtimmons:
I saw this while looking at other things.
Thankfully no usages of `createIndex` with session opts in any workload yamls.
I checked other places where we ternary on `_onSession` and everywhere else is fine.


aggrand (approved):
Yikes good catch, LGTM!

rtimmons:
evergreen merge

aggrand:
@rtimmons  I think you're seeing TIG-2943 I just put up a workaround PR since the EVG fix isn't in yet.

PR #476, opened 2021-05-19
PERF-2260 use actor templates

jimoleary:
Look at mixed last. This PR cuts down on the repetition (and will hopefully make life easier in future).

[Patch](https://spruce.mongodb.com/version/60a4ed630ae6063e3d0f1d20/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Elinux-1-node-replSet%24) is currently progressing.

There will be one more PR after this one, but while it will be a large diif it will be mostly boiler plate cut and paste.

### DIFF
--- a/src/workloads/transactions/LLTInsertLowLtc.yml
+++ b/src/workloads/transactions/LLTInsertLowLtc.yml
@@ -57,7 +57,7 @@ ActorTemplates:
       Threads: *ThreadsValue
       Phases:
         OnlyActiveInPhases:
-          Active: {^Parameter: {Name: "OnlyActiveInPhase", Default: [5]}}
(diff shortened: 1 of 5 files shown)


apodelko (approved):
LGTM assuming that it work. I guess that it is the fist step in parameterization - ideally it would it good to parameterize it much further, but I guess we have some issues for now.

dalyd (approved):

jimoleary:
> LGTM assuming that it work. I guess that it is the fist step in parameterization - ideally it would it good to parameterize it much further, but I guess we have some issues for now.

unfortunately the next PR is going to make this all worse. Even if the parameterization is not done on the short term being able to load actor templates from external files would also really help (there is a ticket for that).

PR #480, opened 2021-05-20
PERF-2260 fix high rate

jimoleary:
smooth out the rate.


apodelko (approved):
LGTM. I guess we agreed on that as a workaround for the current algorithm. Makes it more deterministic and we should keep an eye on how it works (I wonder how it works when it can't handle that load).

jimoleary:
There is a very odd artifact in the 800 per 1 second workload where the getmores are affected as well. 

I've added a note to the report to come back to it.

PR #481, opened 2021-05-26
PERF-2299 Test unbounded window functions, with spilling

dpercy:
From the [patch build results](https://evergreen.mongodb.com/task/sys_perf_linux_1_node_replSet_set_window_fields_unbounded_patch_441bef2c968ab24fe52d910c26dc3ecac366d0eb_60ad6c221e2d175d3bf3b122_21_05_25_21_30_55#):
Test | 1 | 8 | Max
-- | -- | -- | --
SetWindowFields.BothSpill | 1,377,322,332 | &nbsp; | 1,377,322,332
SetWindowFields.NothingSpills | 100,871,471 | &nbsp; | 100,871,471
SetWindowFields.PartitionSpills | 226,255,058 | &nbsp; | 226,255,058
SetWindowFields.SortSpills | 1,436,059,760 | &nbsp; | 1,436,059,760

I'm surprised that "BothSpill" would be slightly faster than "SortSpills".  Maybe that's just noise?  It is under 5% difference.

### DIFF
--- a/src/workloads/execution/SetWindowFieldsUnbounded.yml
+++ b/src/workloads/execution/SetWindowFieldsUnbounded.yml
@@ -1,0 +1,184 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of $setWindowFields with an [unbounded, unbounded] window.
+  This case is interesting because it requires two passes over the documents (within each
+  partition): one pass to compute the window function, and one pass to write the new field to
+  each document.  If any partition is too large, it has to spill.  Also, internally we $sort by the
+  partition key, so that can also spill, depending on the collection size and available indexes.
+
+  We can divide this up into 4 cases:
+    1. Nothing spills.
+    2. Only the $sort spills.
+    3. Only the $_internalSetWindowFields spills.
+    4. Both the $sort and the $_internalSetWindowFields spill.
+
+  - To control whether the $sort spills, we can use an index, or not.
+  - To control whether the $_internalSetWindowFields spills, we can change the cardinality of
+    the partition key.
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    # Through trial and error, it seems 60k documents, when divided into 12 partitions, does not spill,
+    # but divided into fewer partitions (10 partitions), fits in the memory limit.
+    DocumentCount: 60000
+    BatchSize: &batchSize 1000
+    Document:
+      # Each document has a unique _id. In phase 2 we use this to create evenly distributed
+      # partition keys.
+      _id: {^Inc: {}}
+      # This int field is what we sum. Its value shouldn't matter for this test.
+      int: &int {^RandomInt: {min: 1, max: 10000}}
+      # This string pads the doc size out to around 10KB.
+      string: &string {^RandomString: {length: 10000}}
+  - Nop: true
+  - Nop: true
+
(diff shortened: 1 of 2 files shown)


src/workloads/execution/SetWindowFieldsUnbounded.yml
```diff
@@ -0,0 +1,184 @@
...
+            partitionBy: "$twelve_indexed",
+            output: {
+              total: {$sum: "$int"}
+            }
+          }},
+          {$project: {
+            partitionKey: 1,
+            percentage: {$divide: ["$int", "$total"]},
+            keepString: {$type: "$string"},
+          }}
```
jlap199, line 107: I noticed you follow each $setWindowFields with a $project in these benchmarks. Any particular reason?

src/workloads/execution/SetWindowFieldsUnbounded.yml
```diff
@@ -0,0 +1,184 @@
...
+        updates: [
+          {
+            q: {},
+            u: [
+              {$set: {
+                ten_indexed: {$mod: ["$_id", 10]},
+                ten_unindexed: {$mod: ["$_id", 10]},
+                twelve_indexed: {$mod: ["$_id", 12]},
+                twelve_unindexed: {$mod: ["$_id", 12]},
+              }},
```
jlap199, line 64: Is the purpose of this update to have partitions of exactly equal size? As opposed to defining these fields in the `InsertData` phase using `RandomInt`.

jlap199 (commented):
What are the units in the table above?

src/workloads/execution/SetWindowFieldsUnbounded.yml
dpercy: Yes, I was trying to make the partition sizes deterministic, because the partition size determines whether or not spilling happens.  I was thinking that if the values are random, then the partition sizes might vary enough to spill / not spill, inconsistently.

dpercy: Maybe it would be useful for the "Inc" value generator to support wrapping?

dpercy:
> What are the units in the table above?

I'm not sure, do you know where to find this?  Locally, the whole thing takes ~230 seconds, but that's different hardware and includes the insert phase.  I would guess it's nanoseconds.

src/workloads/execution/SetWindowFieldsUnbounded.yml
dpercy: I think the main reason is to avoid sending those 10KB strings in the query response, since the benchmark is trying to focus on the cost of the query rather than the network.  We can't do `{$project: {string: 0}}` though, because it would get pushed down too early in the query, so the documents would all be small, so nothing would spill.

I think the reason for the $divide was just because users would probably do that in queries like this.  I wouldn't expect that to affect the performance much; maybe it's clearer to remove it.

jlap199: Ah I forgot about the 10KB string. That makes sense. Would you mind adding another a comment to that effect?

src/workloads/execution/SetWindowFieldsUnbounded.yml
jlap199: That would be useful, but no need to add support for it. I just wanted to make sure my understanding was correct.

jlap199 (approved):

src/workloads/execution/SetWindowFieldsUnbounded.yml
dpercy: Done

dpercy:
Thanks!

PR #482, opened 2021-05-26
PERF-2277 Benchmarks for cumulative aggregation window functions

jlap199:
https://jira.mongodb.org/browse/PERF-2277

Test | 1 | 8 | Max
-- | -- | -- | --
CumulativeWindows.Avg | 49,321,199 | &nbsp; | 49,321,199
CumulativeWindows.CovPop | 62,363,598 | &nbsp; | 62,363,598
CumulativeWindows.CovSamp | 62,340,876 | &nbsp; | 62,340,876
CumulativeWindows.StdDevPop | 49,672,157 | &nbsp; | 49,672,157
CumulativeWindows.StdDevSamp | 49,887,726 | &nbsp; | 49,887,726
CumulativeWindows.Sum | 53,042,569 | &nbsp; | 53,042,569

### DIFF
--- a/src/workloads/execution/CumulativeWindows.yml
+++ b/src/workloads/execution/CumulativeWindows.yml
@@ -1,0 +1,155 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    # Small collection size to keep from spilling to disk during sorts.
+    DocumentCount: 10000
+    BatchSize: &batchSize 10000
+    Document:
+      t: {^RandomDate: {min: "2020-01-01", max: "2021-01-01"}}
+      x: {^RandomDouble: {distribution: normal, mean: 0, sigma: 3}}
+      y: {^RandomDouble: {distribution: normal, mean: 1, sigma: 3}}
+      z: {^RandomString: {length: 8}}
+  - Nop: true
+  - Nop: true
+
+- Name: Quiesce
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Nop: true
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - Nop: true
+
+- Name: CumulativeWindows
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Nop: true
+  - Nop: true
+  - Repeat: 10
+    Database: *db
+    Operations:
+    - OperationMetricsName: Sum
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{
+              $setWindowFields: {
+                sortBy: {t: 1},
+                output: {
+                  sum: {
+                    $sum: "$x",
+                    window: {documents: ["unbounded", "current"]}
+                  }
+                }
+              }
+            }]
+          cursor: {batchSize: *batchSize}
+    - OperationMetricsName: Avg
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{
+              $setWindowFields: {
+                sortBy: {t: 1},
+                output: {
+                  avg: {
+                    $avg: "$x",
+                    window: {documents: ["unbounded", "current"]}
+                  }
+                }
+              }
+            }]
+          cursor: {batchSize: *batchSize}
+    - OperationMetricsName: StdDevPop
+      OperationName: RunCommand
+      OperationCommand:
+          aggregate: Collection0
+          pipeline:
+            [{
+              $setWindowFields: {
+                sortBy: {t: 1},
+                output: {
+                  stdDevPop: {
+                    $stdDevPop: "$x",
+                    window: {documents: ["unbounded", "current"]}
+                  }
+                }
+              }
+            }]
+          cursor: {batchSize: *batchSize}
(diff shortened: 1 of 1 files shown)


src/workloads/execution/CumulativeWindows.yml
```diff
@@ -0,0 +1,155 @@
...
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    # Small collection size to keep from spilling to disk during sorts.
```
nzolnierzmdb, line 13: This is true, but also the window function stage itself can now spill. Given the way the memory tracking is currently working, we'll use approx double the size of each document as well as the state in the window (which for these type of windows should be minimal). So I think we're still good on not spilling?

src/workloads/execution/CumulativeWindows.yml
```diff
@@ -0,0 +1,155 @@
...
+    DocumentCount: 10000
+    BatchSize: &batchSize 10000
+    Document:
+      t: {^RandomDate: {min: "2020-01-01", max: "2021-01-01"}}
+      x: {^RandomDouble: {distribution: normal, mean: 0, sigma: 3}}
+      y: {^RandomDouble: {distribution: normal, mean: 1, sigma: 3}}
+      z: {^RandomString: {length: 8}}
```
nzolnierzmdb, line 20: This one looks unused, was it intentional? It would be fine to leave in with a comment, given its likely that real world use cases will have more fields than just the ones used by WFs.

nzolnierzmdb (approved):
Couple nits otherwise looks good to me

src/workloads/execution/CumulativeWindows.yml
jlap199: Yes that was the intention. I can leave a comment to that effect.

src/workloads/execution/CumulativeWindows.yml
jlap199: I can be more specific, but yes for these windows the sort is the only operation that should introduce the potential for spilling.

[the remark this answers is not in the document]
jlap199: I wanted to find a way to avoid sorting, but it looks like a `sortBy` is still required even if there's an index defined on the `sortBy` field. Is that just a syntactic requirement? Does a sort get optimized away if the sort field is indexed?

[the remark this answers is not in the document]
dpercy: For example, here's a query with $sort then $group.  Explain shows the $sort is gone: instead the initial $cursor stage has an IXSCAN.
```
> db.c.explain().aggregate([ {$sort: {_id: 1}}, {$group: {_id: null, docs: {$push: "$$ROOT"}}} ])
{
        "explainVersion" : "2",
        "stages" : [
                {
                        "$cursor" : {
                                "queryPlanner" : {
                                        "namespace" : "test.c",
                                        "indexFilterSet" : false,
                                        "parsedQuery" : {

                                        },
                                        "queryHash" : "F44244FE",
                                        "planCacheKey" : "E15E9946",
                                        "maxIndexedOrSolutionsReached" : false,
                                        "maxIndexedAndSolutionsReached" : false,
                                        "maxScansToExplodeReached" : false,
                                        "winningPlan" : {
                                                "queryPlan" : {
                                                        "stage" : "FETCH",
                                                        "planNodeId" : 2,
                                                        "inputStage" : {
                                                                "stage" : "IXSCAN",
                                                                "planNodeId" : 1,
                                                                "keyPattern" : {
                                                                        "_id" : 1
                                                                },
                                                                "indexName" : "_id_",
                                                                "isMultiKey" : false,
                                                                "multiKeyPaths" : {
                                                                        "_id" : [ ]
                                                                },
                                                                "isUnique" : true,
                                                                "isSparse" : false,
                                                                "isPartial" : false,
                                                                "indexVersion" : 2,
                                                                "direction" : "forward",
                                                                "indexBounds" : {
                                                                        "_id" : [
                                                                                "[MinKey, MaxKey]"
                                                                        ]
                                                                }
                                                        }
                                                },
                                                "slotBasedPlan" : {
                                                        ...
                                                }
                                        },
                                        "rejectedPlans" : [ ]
                                }
                        }
                },
                {
                        "$group" : {
                                "_id" : {
                                        "$const" : null
                                },
                                "docs" : {
                                        "$push" : "$$ROOT"
                                }
                        }
                }
        ],
    ...
```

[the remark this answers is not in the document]
jlap199: Thanks @dpercy, in that case I will create an index on `t` since we already have other benchmarks that exercise the sort.

src/workloads/execution/CumulativeWindows.yml
```diff
@@ -0,0 +1,157 @@
...
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    # Choose a collection large enough such that a sort will trigger a spill to disk.
```
dpercy, line 13: What do you mean here?  The queries don't specify `allowDiskUse: true`, so they would fail rather than spill.

I think we also don't expect anything to use much space: the sort should use the index, and the [unbounded, current] window should only need a small, constant amount of space.

jlap199: Yes that's right. This comment says that without the index we would spill, which would result in failure for these tests.

dpercy (approved):

src/workloads/execution/CumulativeWindows.yml
dpercy: Oh, got it, thanks.

PR #484, opened 2021-06-01
TIG-2841 Revamp AutoRun syntax to be more powerful and expressive

vrachev:
Patches coming shortly. Existing `AutoRun` sections have been updated to the new syntax, and I've added info in the README (though ideally it would live in the TBD syntax primer). 

This ticket allows workloads to exclude running on certain branches, so I'm gonna file a new ticket to enable auto_tasks on the 4.0 and 4.2 branches. There will also be another genny ticket to add the branch exclusions for existing workloads that aren't compatible on those branches. I left that out here as the PR is already pretty big and I'd have to do some testing on the older branches to see which workloads can and can't run. Also will talk to product perf to see if they even want any of these workloads to be enabled in those branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,35 +327,67 @@ cd genny
 evergreen set-module -m genny -i <ID> # Use the build ID from the previous step.
 ```
 
-In the browser window, select either `genny_patch_tasks` or `genny_auto_tasks`.
-`genny_patch_tasks` will run any workloads that you have added or modified locally
+In the browser window, select either `schedule_patch_auto_tasks` or `schedule_variant_auto_tasks`.
+`schedule_patch_auto_tasks` will run any workloads that you have added or modified locally
 (based on your git history). This is useful if you want to test only the workload(s)
-you've been working on. 
+you've been working on. Note that `schedule_patch_auto_tasks` will only schedule a task for
+a variant if the workload is compatible (see AutoRun below).
 
-`genny_auto_tasks` automatically runs workloads based on the evergreen environment
+`schedule_variant_auto_tasks` automatically runs workloads based on the evergreen environment
 (variables from `bootstrap.yml` and `runtime.yml` in DSI) and an optional AutoRun
-section in any workload, doing simple key-value matching between them. For example,
+section in any workload. The AutoRun section is a list of <When/ThenRun> blocks,
+where if the When condition is met, tasks are scheduled with additional bootstrap
+values from ThenRun. For example,
 suppose we have a `test_workload.yml` file in a `workloads/*/` subdirectory,
 containing the following AutoRun section:
 
 ```yaml
 AutoRun:
-  Requires:
-    bootstrap:
-      mongodb_setup: 
-        - replica
-        - single-replica
+  - When:
+      mongodb_setup:
+        $eq:
+          - replica
+          - replica-noflowcontrol
+      branch_name:
+        $neq:
+          - v4.0
+          - v4.2
+    ThenRun:
+      - infrastructure_provisioning: foo
+      - infrastructure_provisioning: bar
+      - arbitrary_bootstrap: baz
 ```
 
-In this case, `test_workload` would be run whenever `bootstrap.yml`'s `mongodb_setup`
-variable has a value of `replica` or `single-replica`. In practice, the workload
-AutoRun sections are setup so that you can use `genny_auto_tasks` to run all relevant
-workloads on a specific buildvariant.
+In this case, it looks in the `bootstrap.yml` of `test_workload`, checks if `mongodb_setup`
+is either `replica` or `replica-noflowcontrol`, and also if `branch_name` is neither v4.0 or v4.2.
+If both conditions are true, then we schedule several tasks. Let's say the workload name is
+`DemoWorkload`, 3 tasks are scheduled - `demo_workload_foo`, `demo_workload_bar`, and `demo_workload_baz`.
+The first task is passed in the bootstrap value `infrastructure_provisioning: foo`, the second
+is passed in `infrastructure_provisioning: bar` and the third `arbitrary_bootstrap: baz`.
 
-Both `genny_patch_tasks` and `genny_auto_tasks` will compile mongodb and then run
-the relevant workloads.
+This is a more complex example of AutoRun. Here's a more simple one representing usual usecases:
 
-NB:
+```yaml
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq: standalone
+```
+
+Let's say this is `DemoWorkload` again. In this case, if `mongodb_setup` is `standalone`
+we schedule `demo_workload` with no params.
+
+A few notes on the syntax:
+- supports multiple When/ThenRun blocks per AutoRun. Each are evaluated independently.
+- When blocks can evaluate multiple conditions. All conditions must be true in this case.
+- When supports $eq and $neq. Both can accept either a scalar or list of values.
+- For a list of values, $eq evaluates to true if it is equal to at least one.
+- For a list of values, $neq evaluates to true if it is equal to none of the values.
+- AutoRun will fail if the bootstrap value used in When does not exist in expansions.yml.
+- ThenRun blocks are optional.
+- Each item in the ThenRun list can only support one {bootstrap_key: bootstrap_value} pair.
+
+NB on patch-testing:
 
 1.  After the task runs you can call `set-module` again with more local changes from Genny or DSI.
     This lets you skip the 20 minute wait to recompile the server.
--- a/README.md
+++ b/README.md
@@ -355,17 +355,17 @@ AutoRun:
     ThenRun:
       - infrastructure_provisioning: foo
       - infrastructure_provisioning: bar
-      - arbitrary_bootstrap: baz
+      - arbitrary_key: baz
 ```
 
 In this case, it looks in the `bootstrap.yml` of `test_workload`, checks if `mongodb_setup`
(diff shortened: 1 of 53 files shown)


src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -140,17 +140,23 @@ def has(self, key: str, acceptable_values: List[str]) -> bool:
...
-            raise Exception(f"Unknown key {key}. Know about {self.conts.keys()}")
+            return False
         actual = self.conts[key]
         return any(actual == acceptable_value for acceptable_value in acceptable_values)
 
 
 class GeneratedTask(NamedTuple):
     name: str
-    mongodb_setup: Optional[str]
+    bootstrap_key: Optional[str]
```
vrachev, line 150: Let's us set an arbitrary bootstrap value

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -281,17 +358,19 @@ def all_tasks(self) -> List[GeneratedTask]:
...
         return [task for workload in self.all_workloads() for task in workload.variant_tasks(build)]
 
-    def patch_tasks(self) -> List[GeneratedTask]:
+    def patch_tasks(self, build: CurrentBuildInfo) -> List[GeneratedTask]:
         """
         :return: Tasks for modified workloads current variant (runtime)
         """
-        return [task for workload in self.modified_workloads() for task in workload.all_tasks()]
+        return [
+            task for workload in self.modified_workloads() for task in workload.variant_tasks(build)
```
vrachev, line 372: patch_tasks now only schedules modified workloads if they are compatible with the variant. This is a change from before where patch_tasks would schedule all modified workloads for all variants selected. This meant if you modified a workload requiring a replica set, and selected patch_auto_tasks on the standalone variant, the task would be scheduled and proceed to fail.

vrachev (commented):
Patch with [variant_tasks](https://spruce.mongodb.com/version/60b64f2457e85a6f75de2ed4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), patch with [patch_tasks](https://spruce.mongodb.com/version/60b64eed0305b9589f4f48bb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), patch with [no changes (control)](https://spruce.mongodb.com/version/60b64a5e2a60ed6d6fd9e96e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

I cancelled the builds after the tasks were scheduled, the important thing is that the same tasks got scheduled with the new auto tasks.

Here's a [patch](https://spruce.mongodb.com/version/60b79c1b9ccd4e3ae650def9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) to test a workload with ThenRun

vrachev:
So even though I added the `ThenRun` functionality described in the ticket, we might want to consider whether we should actually add it to genny. Even though it's just an extension of the existing `PreparesEnvironmentWith` functionality, there's an argument that this functionality encourages Genny/DSI coupling by allowing users to modify DSI params in Genny (ThenRun is optional so it's not forced, but still allowed). The better behavior might be to have users create new sys-perf variants. The counter argument is that ThenRun means users can add new workloads without modifying sys-perf, added flexibility during testing, faster iteration and extending existing functionality. 

I've already added ThenRun so there wouldn't be further work if we decide to keep it, but perhaps we explicitly want to not include it. Thoughts? @aggrand @rtimmons @ancostas

rtimmons:
Glad to see this made it in!

My $0.02 is to leave the `ThenRun`. I want users to feel empowered to add new workloads without having to create SERVER tickets and CRs and try to synchronize commits to two repos. If they shoe-horn a workload into a variant that doesn't make the most intuitive sense that's fine, the variants are just string names :)

I didn't look through the code but the test-cases look 👍 so I'm fine if y'all are.

README.md
```diff
@@ -327,35 +327,66 @@ cd genny
...
+
+A few notes on the syntax:
+- Supports multiple When/ThenRun blocks per AutoRun. Each are evaluated independently.
+- When blocks can evaluate multiple conditions. All conditions must be true to schedule the task.
+- When supports $eq and $neq. Both can accept either a scalar or list of values.
+- For a list of values, $eq evaluates to true if it is equal to at least one.
+- For a list of values, $neq evaluates to true if it is equal to none of the values.
+- ThenRun blocks are optional.
+- Each item in the ThenRun list can only support one {bootstrap_key: bootstrap_value} pair.
+
```
aggrand, line 388: Following up on the concerns about `ThenRun` coupling too much with DSI, here might be a good spot to warn users to be judicious in its use.

vrachev: Good idea

aggrand (approved):
I like this change a ton. It makes AutoRun so much more powerful, and this code is super clean. I mostly just had some small cleanup points / nitpicks / questions, nothing to block on.

I agree that coupling Genny and DSI is potentially problematic, especially since `ThenRun` can modify _any_ arbitrary bootstrap var. That said, having to put up PRs for so many repos seems even worse in a lot of ways. I'm comfortable for the moment just policing its use / warning users in docs for now.

LGTM assuming no tasks missing in the patches above vs runs with the current master.

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -178,61 +181,135 @@ def __init__(self, workspace_root: str, file_path: str, is_modified: bool, reade
...
+                        f"Got key ${key} with condition ${condition}."
+                    )
+                if "$eq" in condition:
+                    acceptable_values = condition["$eq"]
+                    if not isinstance(acceptable_values, list):
+                        acceptable_values = [acceptable_values]
+                    if not build.has(key, acceptable_values):
+                        okay = False
+                elif "$neq" in condition:
+                    acceptable_values = condition["$neq"]
```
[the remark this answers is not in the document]
vrachev: Aha yeah that would make a lot more sense 😄

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -178,61 +181,135 @@ def __init__(self, workspace_root: str, file_path: str, is_modified: bool, reade
...
+                    )
+
+            if okay:
+                tasks += self.generate_requested_tasks(then_run)
+
+        return self._dedup_task(tasks)
 
-        return [task for task in self.all_tasks() if meets_criteria()]
+    @staticmethod
+    def _dedup_task(tasks: List[GeneratedTask]) -> List[GeneratedTask]:
```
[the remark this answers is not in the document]
vrachev: The autorun impl may schedule the same task twice (for example, if we have two When blocks that are both true). It was easier to just dedupe the list at the end vs keeping track of if we've added a task and only conditionally add a new task. 

I added a comment to make this clearer

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -178,61 +181,135 @@ def __init__(self, workspace_root: str, file_path: str, is_modified: bool, reade
...
+        """
+        tasks = []
+        if len(then_run) == 0:
+            tasks += [GeneratedTask(self.snake_case_base_name, None, None, self)]
+
+        for then_run_block in then_run:
+            # Just a sanity check; we check this in _validate_auto_run
+            assert len(then_run_block) == 1
+            [(bootstrap_key, bootstrap_value)] = then_run_block.items()
+            task_name = f"{self.snake_case_base_name}_{self._to_snake_case(bootstrap_value)}"
```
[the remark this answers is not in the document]
vrachev: I used just the `bootstrap_value` for history purposes. I added a comment though to articulate this. I also added that we may change this after PM-2310. 

This means a task like `workload_name_false` is possible lol, but I think we should just keep an eye on that in PRs.

vrachev:
Added the branch conditionals here (instead of https://jira.mongodb.org/browse/TIG-3105) to keep it in one PR. Patches from [4.0](https://spruce.mongodb.com/version/60c1067a2a60ed0d148a07a6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), [4.2](https://spruce.mongodb.com/version/60c106380ae60647007bb784/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [master](https://spruce.mongodb.com/version/60c105d3d6d80a690340e8a9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) are all green with the right tasks being created for each branch. 

@dalyd this is gonna add new tasks to 4.0 and 4.2. Is that ok from your end?

dalyd:
> Added the branch conditionals here (instead of https://jira.mongodb.org/browse/TIG-3105) to keep it in one PR. Patches from [4.0](https://spruce.mongodb.com/version/60c1067a2a60ed0d148a07a6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), [4.2](https://spruce.mongodb.com/version/60c106380ae60647007bb784/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [master](https://spruce.mongodb.com/version/60c105d3d6d80a690340e8a9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) are all green with the right tasks being created for each branch.
> 
> @dalyd this is gonna add new tasks to 4.0 and 4.2. Is that ok from your end?

👍 provided the new tasks aren't red or purple.

vrachev:
evergreen merge

PR #486, opened 2021-06-04
TIG-2819 Update toolchain download URLs to point to new S3 bucket

vrachev:
See https://github.com/10gen/vcpkg/pull/11 for more details. This makes genny download the toolchain from S3. Because the git rev is not constant amongst all the binaries, I added a workaround where we touch a file in the gennytoolchain repo with the githash as the name and check that it's the same githash as is requested. (The git hash is the hash of the final vcpkg change, but the macos gennytoolchain artifact was created before the commit was merged, and therefore its git hash is different).

### DIFF
--- a/src/lamplib/src/genny/toolchain.py
+++ b/src/lamplib/src/genny/toolchain.py
@@ -157,8 +157,8 @@ class ToolchainDownloader(Downloader):
     # These build IDs are from the genny-toolchain Evergreen task.
     # https://evergreen.mongodb.com/waterfall/genny-toolchain
 
-    TOOLCHAIN_BUILD_ID = "0db5d1544746c1570371f51109a0a312a7215b65_20_10_01_06_38_39"
-    TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split("_")[0]
+    AWS_BUCKET = "https://dsi-donot-remove.s3-us-west-2.amazonaws.com/genny-toolchain"
+    TOOLCHAIN_GIT_HASH = "815d1e868681f2a712f5518431d5bba44d40d014"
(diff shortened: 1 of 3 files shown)


vrachev:
Tested this locally on macos as well as a ubuntu1804 workstation

src/lamplib/src/genny/toolchain.py
```diff
@@ -195,7 +195,7 @@ def _can_ignore(self):
         )
 
     def _check_toolchain_githash(self):
-        res = "".join(
-            run_command(cmd=["git", "rev-parse", "HEAD"], cwd=self.result_dir, check=True).stdout
-        )
-        return res.strip() == ToolchainDownloader.TOOLCHAIN_GIT_HASH
+        if os.path.exists(self.toolchain_git_hash_file):
```
aggrand, line 198: Won't this only check whether there is a git hash file in the first place, without checking that it's the right hash?

aggrand (changes requested):
Great job getting this to work and deciphering all the vcpkg stuff!

Just a question about the aws bucket mechanism. Was the problem that macos is flaky but still works sometimes? If so, would be able to just re-execute the task until we get a green macos build, then leave it be? That way we could keep downloading from the task; I feel like if we want to use AWS buckets we'd want to have a dedicated "push" task like curator does. Manual aws bucketing seems a bit inscrutable. Also continuing to download from the task would mean it seamlessly starts working again after we update the toolchain.

vrachev:
@aggrand Yeah I think we can try just committing to the waterfall until we get a green build. Mind taking a look at https://github.com/10gen/vcpkg/pull/11 for that?

vrachev:
No longer needed with #488

PR #490, opened 2021-06-10
PERF-2317 Extend LLT measurements for Analytics workloads

apodelko:
Not sure if it makes sense to merge as is - but I'd like to get some feedback earlier as we are pressed to finish the project. I plan to play with data size and try mixed workload (copied from LLT). So at least these two things probably would be changed. 

The main part I want your feedback is data structure and aggregation pipeline. And, of course, everything else to improve it.

### DIFF
--- a/src/workloads/transactions/LLTAnalytics.yml
+++ b/src/workloads/transactions/LLTAnalytics.yml
@@ -163,22 +163,6 @@ ActorTemplates:
               Comment: {^Parameter: {Name: "Comment", Default: "Aggregate.Snapshot"}}
               AllowDiskUse: true # Naming convention follows c++ driver rather than mongo shell.  
 
-
-    #Documents: 1000
-    #ScanType: standard
-    #GlobalRate: 10 per 10 seconds
-    # Count the frequency of each character in array field "a".
-    #AggregatePipeline:
-    #    {array: [{"$addFields":{"a":{"$map":{"input":{"$range":[0,{"$strLenCP":"$a"}]},"in":{"$substrCP":["$a","$$this",1]}}}}},
-    #             {"$unwind": "$a"},
-    #             {"$group": {"_id": "$a", "count": {"$sum":1}}},
-    #             {"$sort": {"_id":1}},
-    #    ]}
-    #AggregateOptions:
-    #  BatchSize: 1000
-    #  Comment: Aggregation to count char frequency
-    #  AllowDiskUse: true # Naming convention follows c++ driver rather than mongo shell.             
-
 Clients:
   Default:
     QueryOptions:
@@ -279,14 +263,14 @@ Actors:
     TemplateName: UpdateTemplate
     TemplateParameters:
       Name: Long.Update.Benchmark
-      Active: [5]
+      Active: [7]
       Blocking: none
       Duration: *SnapshotScannerLongDuration
 
 - ActorFromTemplate:
     TemplateName: ScanTemplate
     TemplateParameters:
-      Name: Long.Aggregate.Snapshot
+      Name: Long.Aggregate.Benchmark
       Active: [7]
       Comment: SnapshotAggregateLong
       Duration: *SnapshotScannerLongDuration
--- a/src/workloads/transactions/LLTAnalytics.yml
+++ b/src/workloads/transactions/LLTAnalytics.yml
@@ -1,0 +1,319 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  Workload to Benchmark the effect of LongLivedTransactions on an Update workload.
+
+GlobalDefaults:
+  # Start: The following block gets updated for the other configurations.
+  # Not In-memory: Database size works out about 30GB.
+  InitialDocumentCount: &InitialNumDocs 49000000
+  SecondaryDocumentCount: &SecondaryNumDocs 10000000
+  GlobalRateValue: &GlobalRateValue 1 per 1250 microsecond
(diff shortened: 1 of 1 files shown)


src/workloads/transactions/LLTAnalytics.yml
```diff
@@ -0,0 +1,284 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  Workload to Benchmark the effect of LongLivedTransactions on an Update workload.
+
+GlobalDefaults:
+  # Start: The following block gets updated for the other configurations.
+  # Not In-memory: Database size works out about 30GB.
+  InitialDocumentCount: &InitialNumDocs 4900000 #0
```
jimoleary, line 9: is there also a LTC version?

src/workloads/transactions/LLTAnalytics.yml
```diff
@@ -0,0 +1,284 @@
...
+          Active: {^Parameter: {Name: "Active", Default: [7]}}# [7]
+          NopInPhasesUpTo: *MaxPhases
+          PhaseConfig:
+            Duration: {^Parameter: {Name: "Duration", Default: *SnapshotScannerShortDuration}}
+            #ScanDuration: {^Parameter: {Name: "ScanDuration", Default: *SnapshotScannerShortDuration}}
+            ScanType: snapshot
+            #ScanContinuous: true
+            # GenerateCollectionNames: true
+            AggregatePipeline:
+              {array: [{"$match":{"cuid":{"$gte":500}}},
```
jimoleary, line 156: The base scanner does a collscan, this is doing something different. Is there a reason not to do a full scan?

src/workloads/transactions/LLTAnalytics.yml
```diff
@@ -0,0 +1,284 @@
...
+                {"$unwind": "$orders"},
+                {"$group": {"_id": "$ts", "total": {"$sum":{ $multiply: [ "$orders.unitPrice", "$orders.quantity" ] }}}},
```
jimoleary, line 158: grouping on ts is going to give you some odd and large groups (I suspect). Would it make more sense to group by something like cuid?

What does the output of the aggregation actually look like?

src/workloads/transactions/LLTAnalytics.yml
```diff
@@ -0,0 +1,284 @@
...
+                {"$sort": {"_id":1}}]
+              }
+            AggregateOptions:
+              BatchSize: 1000
+              Comment: {^Parameter: {Name: "Comment", Default: "Aggregate.Snapshot"}}
+              AllowDiskUse: true # Naming convention follows c++ driver rather than mongo shell.
```
jimoleary, line 164: is this needed, I suspect it is based on the group by ts and the size of the collections.

src/workloads/transactions/LLTAnalytics.yml
```diff
@@ -0,0 +1,284 @@
...
+      GenerateCollectionNames: true
+      Phases:
+        OnlyActiveInPhases:
+          Active: {^Parameter: {Name: "Active", Default: [7]}}# [7]
+          NopInPhasesUpTo: *MaxPhases
+          PhaseConfig:
+            Duration: {^Parameter: {Name: "Duration", Default: *SnapshotScannerShortDuration}}
+            #ScanDuration: {^Parameter: {Name: "ScanDuration", Default: *SnapshotScannerShortDuration}}
+            ScanType: snapshot
+            #ScanContinuous: true
```
jimoleary, line 153: I think you may want this set to true

src/workloads/transactions/LLTAnalytics.yml
apodelko: The comment remained from your script - will need to be adjusted. The current size is 21.9GB for /data/dbs after the test. It appears to be LTC - both xvde and xvdf have zero reads (which is somewhat strange). Marketing wants drastically larger set of data - 80M documents according to the ticket - and we have 5.9M here.

dalyd (commented):
I"m going to defer to Jim for now.

src/workloads/transactions/LLTAnalytics.yml
apodelko: The output of aggregation looks like:
{ "_id" : ISODate("2021-06-11T20:36:16.838Z"), "total" : 1231829.2169177427 }
{ "_id" : ISODate("2021-06-11T20:36:16.839Z"), "total" : 1616938.9089608237 }
{ "_id" : ISODate("2021-06-11T20:36:16.840Z"), "total" : 1205669.5251776665 }
{ "_id" : ISODate("2021-06-11T20:36:16.841Z"), "total" : 1792501.6233907759 }
{ "_id" : ISODate("2021-06-11T20:36:16.842Z"), "total" : 2011877.5808415299 }
{ "_id" : ISODate("2021-06-11T20:36:16.843Z"), "total" : 1916505.504180574 }
Not too nice. Ideally it should be per document - say id, data, etc and total - but I didn't have time to figure it out yet.

apodelko:
As we stopped working on the project, I suggest to merge the file 'as is' to have it handy in case if we will need to continue to work on it in the future or will need something similar (which is quite possible). So I added some comments in the beginning explaining that. No point to polish it further - but I don't want to lose it either. @dalyd @jimoleary Any other thoughts on preserving the file?

jimoleary (approved):
I think it's a good idea to keep it. It won't run at the moment anyway, so lgtm from me.

dalyd (approved):

apodelko:
evergreen merge

PR #491, opened 2021-06-14
PERF-2360: Add new workload for untargeted sharded $lookup

HanaPearlman:
This new workload currently contains a phase which runs untargeted $lookups (not eligible for shard targeting) on a sharded collection into another sharded collection. We anticipate adding new phases to this workload to test other kinds of sharded $lookups, e.g. targeted $lookups and $lookups on unsharded collections into sharded collections. We will either add new phases to this workload or create a new, corresponding workload for sharded $graphLookup.

Here is the [workload proposal document](https://docs.google.com/document/d/1JLTTS7jJDiMxrhT2yMNSW7YA07KtVgoUbB9jXEKpujU/). It includes the following table showing some example output and robustness results:

<meta charset="utf-8"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:10pt;"><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">UntargetedLookupShardedToSharded data over 5 trials:</span></p><div dir="ltr" style="margin-left:0pt;" align="left">
Trial | Avg Latency | 95% Latency
-- | -- | -- 
1 | 5622326682 | 5737675489
2 | 5734970962 | 5849303387
3 | 5623881720 | 5748353800
4 | 5691124866 | 5771233556
5 | 5644313673 | 5724696416
(max-min)/med | 0.01995712617 | 0.02167698359

</div>

Here is an example [patch build result](https://spruce.mongodb.com/task/sys_perf_linux_shard_lite_all_feature_flags_sharded_lookup_patch_a11aba2bf9f18a8e73a23099519ad45a2939d238_60c36de630661579ddf38b03_21_06_11_14_06_39/logs?execution=1).

Last patch before merge: https://spruce.mongodb.com/version/60e5a7da3627e06703ed9d05/tasks

### DIFF
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -50,9 +50,9 @@ Actors:
   Phases:
   - *Nop
   - Repeat: 1
-    BatchSize: &BatchSize 1000
+    BatchSize: 1000
     Threads: 1
-    DocumentCount: 10000
+    DocumentCount: &NumDocs 3000
     Database: *Database
     CollectionCount: 2    # Loader will populate 'Collection0' then 'Collection1'.
     Document:
@@ -108,7 +108,9 @@ Actors:
               as: matches
             }
           }]
-        cursor: {batchSize: *BatchSize}
+        # To get meaningful results, the entire result set should fit in a single batch. This should
+        # be possible since both collections are small.
+        cursor: {batchSize: *NumDocs}
 
 AutoRun:
   - When:
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -17,6 +17,8 @@ Actors:
   - Repeat: 1
     Database: admin
     Operations:
+    # Shard Collection0 and Collection1 using hashed sharding to ensure that the chunks are evenly
+    # distributed across the shards.
     - OperationMetricsName: EnableSharding
       OperationName: AdminCommand
       OperationCommand:
@@ -25,7 +27,6 @@ Actors:
       OperationName: AdminCommand
       OperationCommand:
         shardCollection: test.Collection0
-        # Hashed sharding will ensure that the chunks are evenly distributed across the shards.
         key: {shardKey: hashed}
         numInitialChunks: &NumChunks 6
     - OperationMetricsName: ShardForeignCollection
@@ -34,6 +35,11 @@ Actors:
         shardCollection: test.Collection1
         key: {shardKey: hashed}
         numInitialChunks: *NumChunks
+    # Disable the balancer so that it can't skew results while the $lookups are running.
+    - OperationMetricsName: DisableBalancer
+      OperationName: AdminCommand
+      OperationCommand:
+        balancerStop: 1
   - &Nop {Nop: true}
   - *Nop
   - *Nop
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -1,0 +1,116 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+
+GlobalDefaults:
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+
+  LocalCollection: &LocalCollection Collection0
+  LocalNamespace: &LocalNamespace test.Collection0
+
+  ForeignCollection: &ForeignCollection Collection1
+  ForeignNamespace: &ForeignNamespace test.Collection1
+
+  DocumentCount: &DocumentCount 5000
+  BatchSize: &BatchSize 1000
+  NumChunks: &NumChunks 5
+
+Actors:
+- Name: CreateShardedCollections
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationMetricsName: EnableSharding
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: *Database
+    - OperationMetricsName: ShardLocalCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: *LocalNamespace
+        key: {shardKey: hashed}
+        numInitialChunks: *NumChunks
+    - OperationMetricsName: ShardForeignCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: *ForeignNamespace
+        key: {shardKey: hashed}
+        numInitialChunks: *NumChunks
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: LoadInitialData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    BatchSize: *BatchSize
+    Threads: 1
+    DocumentCount: *DocumentCount
+    Database: *Database
+    CollectionCount: 2    # Loader will populate 'Collection0' then 'Collection1'.
+    Document:
+      shardKey: {^RandomInt: {min: 1, max: 100}}
+      date: &Date {^RandomDate: {min: "2020-01-01", max: "2021-01-01"}}
+      int: {^RandomInt: {min: 1, max: 100}}
+  - *Nop
+  - *Nop
+
+- Name: Quiesce
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+
+- Name: RunLookups
+  Type: RunCommand
+  Database: *Database
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - Repeat: 10    # Untargeted $lookup from sharded collection to sharded collection
+    Database: *Database
+    Operations:
+    - OperationMetricsName: UntargetedLookupShardedToSharded
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *LocalCollection
+        pipeline:
+          [{
+            $lookup: {
+              from: *ForeignCollection,
+              let: {localInt: "$int"},
+              pipeline: [{
+                $match: {
+                  $expr: {
+                    $and: [
+                      {$eq: ["$int", "$$localInt"]},
+                      {$lte: ["$date", *Date]}
+                    ]
+                  }
+                }
+              }],
+              as: matches
+            }
+          }]
+        cursor: {batchSize: *BatchSize}
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq: shard-lite-all-feature-flags
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -7,7 +7,7 @@ Description: |
     1. Creating empty sharded collections distributed across all shards in the cluster.
     2. Populating collections with data.
     3. Fsync.
-    4. Running targeted $lookup from a sharded collection.
+    4. Running untargeted $lookup from a sharded collection.
 
 Actors:
 - Name: CreateShardedCollections
@@ -25,8 +25,9 @@ Actors:
       OperationName: AdminCommand
       OperationCommand:
         shardCollection: test.Collection0
+        # Hashed sharding will ensure that the chunks are evenly distributed across the shards.
         key: {shardKey: hashed}
-        numInitialChunks: &NumChunks 5
+        numInitialChunks: &NumChunks 6
     - OperationMetricsName: ShardForeignCollection
       OperationName: AdminCommand
       OperationCommand:
@@ -45,7 +46,7 @@ Actors:
   - Repeat: 1
     BatchSize: &BatchSize 1000
     Threads: 1
-    DocumentCount: 5000
+    DocumentCount: 10000
     Database: *Database
     CollectionCount: 2    # Loader will populate 'Collection0' then 'Collection1'.
     Document:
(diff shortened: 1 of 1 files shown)


src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,116 @@
...
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: *ForeignNamespace
+        key: {shardKey: hashed}
+        numInitialChunks: *NumChunks
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: LoadInitialData
```
kaywux, line 47: For my own knowledge, what's the difference between the LoadInitialData actor you're using here vs the InsertData actor?

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,116 @@
...
+
+  LocalCollection: &LocalCollection Collection0
+  LocalNamespace: &LocalNamespace test.Collection0
+
+  ForeignCollection: &ForeignCollection Collection1
+  ForeignNamespace: &ForeignNamespace test.Collection1
+
+  DocumentCount: &DocumentCount 5000
+  BatchSize: &BatchSize 1000
+  NumChunks: &NumChunks 5
```
kaywux, line 17: Do you need all of these to be defined as global defaults? I'm not sure what the preferred YAML style is here, but as a general rule I try not to use global when it's not necessary to do so.

Another way to define these would be right when you use it for the first time. For example, instead of defining Database here, you could have on line 30: `enableSharding: &db test` and then reuse it with `*db` in other phases, even within other actors.

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,116 @@
...
+          [{
+            $lookup: {
+              from: *ForeignCollection,
+              let: {localInt: "$int"},
+              pipeline: [{
+                $match: {
+                  $expr: {
+                    $and: [
+                      {$eq: ["$int", "$$localInt"]},
+                      {$lte: ["$date", *Date]}
```
kaywux, line 103: Is this `Date` always going to be the last date that is set on line 60? If so, I can see there being a lot of variability in this workload depending on what the last `Date` was. Is there a reason we aren't using something like `$$localDate` instead, similar to `$$localInt`?

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,116 @@
...
+              }],
+              as: matches
+            }
+          }]
+        cursor: {batchSize: *BatchSize}
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq: shard-lite-all-feature-flags
```
kaywux, line 116: I haven't seen this syntax before for AutoRun, is this required to run the test command feature flag? Usually I see something like:
```
AutoRun:
  Requires:
    mongodb_setup:
    - shard-lite-all-feature-flags
```

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,116 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+
```
kaywux, line 3: Could you add a Description field here to give a brief overview about what this workload will test?

kaywux (commented):
This looks like a great start, thanks Hana! I've left a few questions and suggestions in the comments.

src/workloads/execution/ShardedLookup.yml
HanaPearlman: Done!

src/workloads/execution/ShardedLookup.yml
HanaPearlman: As I understand it, the Loader actor allows us to insert into multiple collections in a single phase, while the Insert actor inserts into a single collection.

src/workloads/execution/ShardedLookup.yml
HanaPearlman: No, `*Date` just represents the yaml value of `Date`, so it will generate a new random date for each $lookup we run. We can see this in the logs, which show the pipelines that are generated; each pipeline has a different date value. 

I just wanted to avoid running identical queries on the foreign collection. If we went with something like `$$localDate`, then the subpipelines (and the top-level $lookup) would be identical on each run.

src/workloads/execution/ShardedLookup.yml
HanaPearlman: This is new as of a couple of days ago! Existing workloads have been updated to have this new syntax. See this ticket: [TIG-2841](https://jira.mongodb.org/browse/TIG-2841), especially the last comment.

src/workloads/execution/ShardedLookup.yml
HanaPearlman: The global default style is not required, but I went with it because I thought it was more clear, particularly for the local/foreign collection/namespace values, to define up-front. But, I see we don't do this elsewhere in the query workload code, so I've removed them in the newest commit.

HanaPearlman (commented):
Thanks Katherine! Went ahead with the changes you suggested and responded to your other questions.

src/workloads/execution/ShardedLookup.yml
kaywux: Ah that makes sense about `*Date` generating a new date every time - thanks for the clarification! I don't think it's necessarily a bad thing for the repeated runs to generate the same $lookup subpipeline for each document if we use `$$localDate`, since we'd then have smaller variance in the repeat performance and thus a stronger indication if there were ever a code change that caused perf regressions (although maybe more variance between runs since the performance for each run is more contingent on the generated documents). In saying that, I'm also fine with the `*Date` method you're using, and the variance between runs is looking good; feel free to keep this as is.

src/workloads/execution/ShardedLookup.yml
kaywux: Oh great, new syntax! Thanks for the context :)

kaywux (approved):
This LGTM mod the small description edit and changes required for the feature flag, as well as a look from the perf team as discussed. Thanks Hana!

jimoleary (commented):
I've left a few comments / suggestions. 

My main observation is that this is a very small workload (the 10 repetitions appear to run in about 90 seconds) and the collections are about 325KB each. 

Have you run any longer / larger tests to see if it worth scaling the workload up further?

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,109 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of $lookup against a sharded foreign collection.
+
+  The workload consists of the following phases:
+    1. Creating empty sharded collections distributed across all shards in the cluster.
```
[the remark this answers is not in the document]
HanaPearlman: Done! Added a comment about this lower down.

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,109 @@
...
+    Database: *Database
+    CollectionCount: 2    # Loader will populate 'Collection0' then 'Collection1'.
+    Document:
+      shardKey: {^RandomInt: {min: 1, max: 100}}
+      date: &Date {^RandomDate: {min: "2020-01-01", max: "2021-01-01"}}
+      int: {^RandomInt: {min: 1, max: 100}}
+  - *Nop
+  - *Nop
+
+- Name: Quiesce
```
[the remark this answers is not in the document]
HanaPearlman: Thanks for pointing this out! It's good to know about this actor. Unfortunately, it seems to me like we can't use it for this workload. Part of the [quiesce code](https://github.com/mongodb/genny/blob/ac2c06a1ec477c569ea27f818a549b38cad08e01/src/gennylib/src/quiesce.cpp#L117) examines the output of serverStatus. When serverStatus is run against a mongos, the output shape is different, so it fails at the linked line. Think we should stick with fsync for this specific use case, at least for now.

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,109 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of $lookup against a sharded foreign collection.
+
+  The workload consists of the following phases:
+    1. Creating empty sharded collections distributed across all shards in the cluster.
+    2. Populating collections with data.
```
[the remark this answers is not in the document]
HanaPearlman: Hmm, I'm not sure I understand why we might want to do that. As I understand, it's better to allow the balancer to rebalance the chunks in the case that the generated data isn't distributed very evenly.

src/workloads/execution/ShardedLookup.yml
HanaPearlman: Thanks for the explanation! I tried running with `$$localDate` instead and got nearly identical results. I'm going to keep it as-is for now, but may reconsider as we add new phases to this workload.

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,109 @@
...
+    - OperationMetricsName: EnableSharding
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: &Database test
+    - OperationMetricsName: ShardLocalCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: test.Collection0
+        key: {shardKey: hashed}
+        numInitialChunks: &NumChunks 5
```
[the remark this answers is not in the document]
HanaPearlman: Yes, this is a good point. Changed to 6.

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,109 @@
...
+
+- Name: LoadInitialData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    BatchSize: &BatchSize 1000
+    Threads: 1
+    DocumentCount: 5000
```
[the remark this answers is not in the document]
HanaPearlman: Yes, this is a good point. When we try increasing the document count, we get a linear increase in query time. We've gone with 10k documents in the next commit, which brings the query time to 10 seconds, and the total test time to around 2 minutes.

It's worth noting that we anticipate adding about 10 more, similar phases to this workload (or potentially split between two workloads) which will test other kinds of $lookups/$graphLookups. Some of these phases with be faster and some will be slower. If we end up with a 10 second query average, then we expect a test that takes closer to 15 minutes, or about half that if we create a second workload.

HanaPearlman (commented):
Thanks for the review!

src/workloads/execution/ShardedLookup.yml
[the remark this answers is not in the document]
jimoleary: This is a hashed shard key with randomly generated values so the collections **should** end up being evenly distributed at the end of the load phase (without the need for the balancer).

My concern would be that we have the balancer running during the early test iterations and that skews the results.

src/workloads/execution/ShardedLookup.yml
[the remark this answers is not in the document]
jimoleary: Good point, I wasn't aware of that!

I've raised [TIG-3152](https://jira.mongodb.org/browse/TIG-3152) to track this.

jimoleary (commented):
I still have a final question about the balancer but aside from that, the workloads looks good.

src/workloads/execution/ShardedLookup.yml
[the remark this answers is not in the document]
HanaPearlman: Thanks!

src/workloads/execution/ShardedLookup.yml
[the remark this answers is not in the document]
HanaPearlman: Yes, this makes sense to me. Updated in the next commit.

jimoleary (approved):
lgtm!

src/workloads/execution/ShardedLookup.yml
```diff
@@ -0,0 +1,118 @@
...
+                    ]
+                  }
+                }
+              }],
+              as: matches
+            }
+          }]
+        # To get meaningful results, the entire result set should fit in a single batch. This should
+        # be possible since both collections are small.
+        cursor: {batchSize: *NumDocs}
```
HanaPearlman, line 113: We need to make sure that the entire result set is calculated and returned in a single batch to get meaningful timing measurements. Whereas before we were seeing a linear increase in query time when increasing both collection sizes, now we see something that looks more quadratic (as expected for $lookup).

As a result, the queries are slower. Accordingly, I reduced the collection size and now each query takes about 11 seconds to run.

HanaPearlman (commented):
This commit includes an important change to increase the batch size of the $lookup results so we make sure we're getting the entire result set!

src/workloads/execution/ShardedLookup.yml
kaywux: Great, glad we were able to get to the bottom of those confounding scaling results!

As a back-of-the-envelope calculation, if each doc is about 0.7KB and each doc on the left side of the $lookup will match around 30 results (probably less) on the right, the total result set for each $lookup should never be more than 10MB, which is well below the 16MB limit. One thing to keep in mind when adding similar $graphLookup workloads is that the result set for a similar setup might be much larger than that.

kaywux (approved):
Still LGTM, thanks Hana!

src/workloads/execution/ShardedLookup.yml
jimoleary: That makes sense, so this is lgtm.

jimoleary (approved):

HanaPearlman:
evergreen merge

PR #492, opened 2021-06-16
PERF-2400: Add MonotonicSingleLoader actor and use it in resharding

visemet:


src/workloads/docs/MonotonicSingleLoader.yml
```diff
@@ -0,0 +1,16 @@
...
+  Type: MonotonicSingleLoader
+  Threads: 100
+  # The MonotonicSingleLoader must only ever be active in a single phase of the workload.
+  Phases:
+  - Repeat: 1
+    Database: test
+    # The collection name is optional and defaults to "Collection0" if omitted.
+    Collection: mycoll
+    BatchSize: 1000
+    DocumentCount: 100_000
```
samanca: We removed the underscores used to format numbers in an earlier comment: "Update number syntax". Is there a reason we are using an underscore here?

src/cast_core/include/cast_core/actors/MonotonicSingleLoader.hpp
```diff
@@ -0,0 +1,88 @@
...
+ * Bulk inserts data into a single collection using multiple threads. Similar to
+ * the MonotonicLoader actor, the generated documents have a monotonically
+ * increasing _id, starting from {_id: 0}.
+ *
+ * However, the MonotonicSingleLoader actor differs from the Loader and
+ * MonotonicLoader actors in a few notable ways:
+ *
+ *  - The collection name is optional and defaults to "Collection0" if omitted.
+ *
+ *  - The MonotonicSingleLoader actor must only ever be active in one phase of the
```
samanca, line 40: Should we also mention that "MonotonicSingleLoader" may only load data into a single collection?

src/workloads/docs/MonotonicSingleLoader.yml
visemet: I prefer the underscores for readability of large numbers. [It is allowed by the YAML spec](http://yaml.org/type/int.html) but breaks genny's IntegerSpec parsing.

samanca (approved):
LGTM, thank you!

src/workloads/docs/MonotonicSingleLoader.yml
aggrand: Worth noting that this should work for IntegerSpec now. After [TIG-2976](https://jira.mongodb.org/browse/TIG-2976) we produce an intermediate yaml file where `100_000` has been converted to `100000`, but please let me know if you hit any issues with that.

We'll make sure to document this in [TIG-3045](https://jira.mongodb.org/browse/TIG-3045).

dalyd (commented):
A couple of thoughts, but no blockers from me. 

1. Would you like to use ^Inc with multiplier instead of hard wiring the id? It feels a little strange to me that you specify a whole document, but then _id get's slipped in underneath. I'm not sure what happens if _id is in the template also.  https://github.com/mongodb/genny/blob/master/src/workloads/docs/Generators.yml#L94-L98
2. Are you okay with leaving a path for us to merge all of this into Loader at some point (but not by you?). Ideally the loader would be smart and handle the collections < threads case and split up the collections in that case. We have talked about replacing MonotonicLoader uses with Loader now that we have ^Inc. It would be nice to merge this one in sometime also. I worry that the proliferation of Loaders will lead to confusion for others trying to add workloads.

src/workloads/docs/MonotonicSingleLoader.yml
aggrand: (For workload yamls specifically. For actor integration tests probably not since they don't go through the preprocessor.)

dalyd: @aggrand is that little bit of niceness documents somewhere or in our examples? I didn't know it was a thing, but I like it and will encourage its use.

visemet: It didn't work for me though @aggrand?

```
[2021/06/16 01:51:48.402]     <testcase classname="cast_core_test.MongoTestFixture" name="MonotonicSingleLoader/Inserts documents into the collection" time="0.774172">
[2021/06/16 01:51:48.402]       <failure type="FAIL">
[2021/06/16 01:51:48.402] Dynamic exception type: genny::InvalidConfigurationException
[2021/06/16 01:51:48.402] std::exception::what: Invalid value for genny::IntegerSpec field: 10_000 from config: 10_000
[2021/06/16 01:51:48.403] CAUGHT Dynamic exception type: genny::InvalidConfigurationException
[2021/06/16 01:51:48.403] std::exception::what: Invalid value for genny::IntegerSpec field: 10_000 from config: 10_000
[2021/06/16 01:51:48.403] at ../src/cast_core/test/MonotonicSingleLoader_test.cpp:67
[2021/06/16 01:51:48.403]       </failure>
[2021/06/16 01:51:48.403]     </testcase>
```

https://evergreen.mongodb.com/lobster/evergreen/task/genny_amazon2_t_integration_test_single_node_replset_patch_7ed579c9ab4af496a502b2632866ce0ab3a9f4d5_60c9518661837d0e7948c33f_21_06_16_01_19_03/0/task#bookmarks=0%2C4868&l=1&shareLine=3889

dalyd: Doh -- nvm -- I see the documentation ticket now.

aggrand: Not at the moment; it was actually just a side effect of loading the yaml and writing it back out. We'll definitely be giving the genny documentation a lot of love very soon in [TIG-3045](https://jira.mongodb.org/browse/TIG-3045) though; I'll comment this there so it's not lost.

visemet: Oh nevermind, I see your comment about how the integration tests don't go through preprocessing. I'll be doing more work on ReshardCollection.yml momentarily so I can switch the larger numbers to use underscores in a separate pull request then.

visemet:
evergreen merge

PR #494, opened 2021-06-17
PERF-2405: Add another workload for reshardCollection with more docs

visemet:


dalyd (approved):

src/workloads/sharding/ReshardCollectionMixed.yml
```diff
@@ -0,0 +1,186 @@
...
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 201
+  # The reshardCollection command is expected to take a long to complete so we give the actor
+  # running it a much higher socket timeout.
+  ReshardCollection:
+    QueryOptions:
+      maxPoolSize: 1
+      socketTimeoutMS: 36_000_000  # = 10 hours
```
samanca, line 60: Don't we have a 6 hour timeout on DSI instance? Just checking if this value is actually honored, or we should use a smaller value.

samanca (approved):

src/workloads/sharding/ReshardCollectionMixed.yml
visemet: I used 10 hours as a round "hand" number than doing the math for 6 hours. The sys-perf task would fail if it took longer than 6 hours still. The resharding operation after d1eb8af5673d1a6130805e435791c24ed8ce9580 and 2044fe6532104a0eb63f94c3afb2393108884f02 finished in 26 minutes. I don't actually need to raise the socket timeout this much anymore but feel like keeping it as-is.

visemet:
evergreen merge

PR #495, opened 2021-06-17
Perf 2260 deploy workloads

jimoleary:
Patches:
* [master patch](https://spruce.mongodb.com/version/60cb513dd1fe0715e5c4daba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
* [v4.4 patch](https://spruce.mongodb.com/version/60cb52ee61837d23b6fe7e37/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
* [v4.2 patch](https://spruce.mongodb.com/version/60cb532f30661555ad16b46e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC): no actual workloads should run
* [v4.0 patch](https://spruce.mongodb.com/version/60cb529d7742ae161d8dca20/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC): no actual workloads should run


src/workloads/transactions/LLTMixedHigh.yml
```diff
@@ -124,7 +124,7 @@ GlobalDefaults:
   PtcRemoveOperation: &PtcRemoveOperation
     OperationName: deleteOne
     OperationCommand:
-      Filter: {price: {'$gte' : {^RandomDouble: { min: 0.0, max: 500.0 }}}}
+      Filter: {'price': {'$gte' : {^RandomDouble: { min: 0.0, max: 1000.0 }}}}
```
jimoleary, line 127: These changes were suggested from the delete workload ticket.

src/workloads/transactions/templates/LLTMixedLowLtc.tmpl
```diff
@@ -5,8 +5,8 @@ Description: |
 
 GlobalDefaults:
   # Start: The following block gets updated for the other configurations.
-  # Not In-memory: Database size works out about 30GB.
-  InitialDocumentCount: &InitialNumDocs 49000000
+  # In-memory: Database size works out about 12GB.
```
jimoleary, line 8: This is odd but I think it is just git is just making assumptions about what files were moved (the values are actually correct).

Also we could delete these files, but I wanted to keep them around in case we ever want to deploy other workloads.

apodelko:
What workloads do we want to operationalize? Won't MixedHigh be enough to catch any change - and we may keep other workloads to run if needed to investigate it further.

jimoleary:
> What workloads do we want to operationalize? Won't MixedHigh be enough to catch any change - and we may keep other workloads to run if needed to investigate it further.

Good call, I changed to just have mixed high.

The other workload can be regenerated from the templates directory using the script in the readme file.

jimoleary:
v4.2 and v4.0 did not run any llt workload (as I had expected / hoped).

I'll deploy this PR tomorrow (assuming everything is ok with the other 2 patches).

src/workloads/transactions/LLTMixed.yml
```diff
@@ -561,11 +561,15 @@ Actors:
...
-#        $eq:
-#          - atlas
-#          - replica
-#          - replica-all-feature-flags
-#          - single-replica
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - atlas
```
dalyd, line 568: Do you want atlas, atlas-like-replica, or both? Any of those are fine, just double checking intention.

src/workloads/transactions/templates/LLTMixedLowLtc.tmpl
```diff
@@ -124,7 +124,7 @@ GlobalDefaults:
   PtcRemoveOperation: &PtcRemoveOperation
     OperationName: deleteOne
     OperationCommand:
-      Filter: {price: {'$gte' : {^RandomDouble: { min: 0.0, max: 500.0 }}}}
+      Filter: {'price': {'$gte' : {^RandomDouble: { min: 0.0, max: 1000.0 }}}}
```
dalyd, line 127: Are quotes actually needed or desired here? I think I saw Brad taking them out of his PR at request from Ryan, but could be wrong.

dalyd: Oh, it's because it's templated, isn't it?

src/workloads/transactions/templates/README.md
```diff
@@ -3,19 +3,26 @@ Long Lived Transactions
...
 
 ```bash
-for file in src/workloads/transactions/LLT*LowLtc.yml; do
+out_dir=src/workloads/transactions/templates/out
+mkdir -pv $out_dir
+for file in src/workloads/transactions/templates/LLT*LowLtc.tmpl; do
+    cp $file $out_dir/$(basename ${file/tmpl/yml})
+done
+for file in $out_dir/LLT*LowLtc.yml; do
     sed -e 's/^  GlobalRateValue: .*$/  GlobalRateValue: \&GlobalRateValue 1 per 1250 microsecond/' \
```
dalyd, line 17: Ah, my eyes! I thought we got rid of all the sed in our codebase. 
I'm not sure I want to ask for a change. Maybe I'll pretend I didn't see this.

dalyd (approved):
Quick look from me and some comments. Approving to get out of the way tomorrow. I looked mainly at the autorun and template stuff.

src/workloads/transactions/LLTMixed.yml
jimoleary: good point, I should probably add  atlas-like-replica too.

src/workloads/transactions/templates/LLTMixedLowLtc.tmpl
jimoleary: no it was more to make them consistent (but it looks like I picked the wrong thing to standardize on).

I will leave it for the moment though.

src/workloads/transactions/templates/README.md
jimoleary: This is really only for convenience. These files are essentially mothballed and should not be used unless there is a compelling need.

jimoleary:
> What workloads do we want to operationalize? Won't MixedHigh be enough to catch any change - and we may keep other workloads to run if needed to investigate it further.

All the old workloads are kept in the templates directory (along with a script to recreate them).

I had picked both the high and the low rate to keep. But actually just one of them makes more sense. I think I would be inclined to keep the low rate version.

apodelko:
> All the old workloads are kept in the templates directory (along with a script to recreate them).
> 
> I had picked both the high and the low rate to keep. But actually just one of them makes more sense. I think I would be inclined to keep the low rate version.

Makes sense if high load is too close to stress testing - we probably want steady workload here.

Do we actually need AutoRun section in templates? Wouldn't it enough to have AutoRun for the workload we actually want to run?

apodelko (approved):
LGTM assuming that we actually run one mixed workload in mainline

jimoleary:
> Makes sense if high load is too close to stress testing - we probably want steady workload here.
> 
> Do we actually need AutoRun section in templates? Wouldn't it enough to have AutoRun for the workload we actually want to run?

if we list what we want it to run on then it will need to be updated for every release. If we list waht we don't want it to run on then it should run on every new version without the need for a change.

> LGTM assuming that we actually run one mixed workload in mainline

It's working as expected.

I had a look at the high and low profile. I'm going to select the low version as it should be (marginally) more stable. We can always change it later if we need to.

PR #496, opened 2021-06-18
PERF-2405: Add 95% read / 5% write reshardCollection workload

visemet:


src/workloads/sharding/ReshardCollectionReadHeavy.yml
```diff
@@ -0,0 +1,186 @@
...
+    - OperationMetricsName: ReshardCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        reshardCollection: *Namespace
+        key: {newKey: 1}
+  - *Nop
+
+- Name: ReadCollectionBeingResharded
+  Type: CrudActor
+  Threads: 95
```
samanca, line 125: I don't think this will give you 95% read and 5% write. Reader threads progress faster than writer threads, and that'd result in having more than 95% of total operations being reads. I think we'd want every thread to perform a write after every 19 read operations to emulate the desired mixture of CRUD operations.

samanca (changes requested):

src/workloads/sharding/ReshardCollectionReadHeavy.yml
visemet: Hmm, I thought over Slack you had suggested splitting up the reader clients from the writer clients? Should I change ReshardCollectionMixed.yml because it is using 50 threads for findOne and 50 threads for updateOne.

samanca: I think it's okay to split the threads as long as we use the rate-limiter to enforce the ratio. Otherwise, I'm not sure if this will give us the ratio we want. How do you feel about using the rate-limiter here?

visemet: It isn't possible to specify `GlobalRate` with `Blocking: None`. The DocumentThroughput values in my patch build suggest that we're closer to 2-4% write. I don't feel attached to the particular ratio other than wanting "almost entirely reads with some writes".

https://spruce.mongodb.com/task/sys_perf_linux_shard_lite_all_feature_flags_reshard_collection_read_heavy_patch_1d8a437c4d4f6ebf4c51546514bf8c4da99046fb_60cbf2fa2fbabe7e4152d83e_21_06_18_01_13_36/trend-charts?execution=2

samanca (approved):

visemet:
evergreen merge

PR #499, opened 2021-06-25
TIG-3028 Allow GlobalRate to be specified for background actors

aggrand:
The main actual change is to wake up all threads when the phase ends, so that the GlobalRate sleeps don't carry over to the next phase. Ended up being smaller than I expected since we could piggyback on the Orchestrator's CV.

### DIFF
--- a/src/gennylib/include/gennylib/Orchestrator.hpp
+++ b/src/gennylib/include/gennylib/Orchestrator.hpp
@@ -21,12 +21,11 @@
 #include <shared_mutex>
 #include <vector>
 
+#include <gennylib/conventions.hpp>
+
 namespace genny {
 
 class Orchestrator;
-
-// May eventually want a proper type for Phase, but for now just a typedef is sufficient.
-using PhaseNumber = unsigned int;
 
 using OrchestratorCB = std::function<void(const Orchestrator*)>;
 
@@ -104,6 +103,10 @@ public:
      */
     bool continueRunning() const;
 
+    /**
+     * Sleep until the given phase ends or time out.
+     */
+    void sleepToPhaseEnd(const PhaseNumber pn, Duration timeout);
 
 private:
     mutable std::shared_mutex _mutex;
--- a/src/gennylib/include/gennylib/Orchestrator.hpp
+++ b/src/gennylib/include/gennylib/Orchestrator.hpp
@@ -104,9 +104,10 @@ public:
     bool continueRunning() const;
 
     /**
-     * Sleep until the given phase ends or time out.
+     * Sleep for the duration, also waking up if the phase changes before the
+     * duration passes.
      */
-    void sleepToPhaseEnd(const PhaseNumber pn, Duration timeout);
+    void sleepToPhaseEnd(Duration timeout, const PhaseNumber pn);
 
 private:
     mutable std::shared_mutex _mutex;
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -148,7 +148,7 @@ public:
                     const auto rate = _rateLimiter->getRate() > 1e9 ? 1e9 : _rateLimiter->getRate();
 
                     // Add ±5% jitter to avoid threads waking up at once.
-                    _sleeper->sleep_for(orchestrator, inPhase, std::chrono::nanoseconds(
+                    _sleeper->sleepFor(orchestrator, inPhase, std::chrono::nanoseconds(
                         int64_t(rate * (0.95 + 0.1 * (double(rand()) / RAND_MAX)))), !_doesBlock);
                     continue;
                 }
--- a/src/gennylib/include/gennylib/PhaseLoop.hpp
+++ b/src/gennylib/include/gennylib/PhaseLoop.hpp
@@ -125,12 +125,6 @@ public:
             const auto rateLimiterName =
                 phaseContext["RateLimiterName"].maybe<std::string>().value_or(defaultRLName.str());
 
-            if (!_doesBlock) {
-                throw InvalidConfigurationException(
-                    "GlobalRate must be specified alongside either Duration or Repeat, otherwise "
-                    "there's no guarantee the rate limited operation will run in the correct "
-                    "phase");
-            }
             _rateLimiter =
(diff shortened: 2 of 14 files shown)


src/gennylib/include/gennylib/Orchestrator.hpp
```diff
@@ -21,13 +21,12 @@
...
 #include <vector>
 
+#include <gennylib/conventions.hpp>
+
 namespace genny {
 
 class Orchestrator;
 
-// May eventually want a proper type for Phase, but for now just a typedef is sufficient.
-using PhaseNumber = unsigned int;
```
aggrand, line 29: Had to move to conventions.hpp to avoid circular inclusions.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -125,38 +125,31 @@ class IterationChecker final {
...
     }
 
     constexpr void limitRate(const SteadyClock::time_point referenceStartingPoint,
                              const int64_t currentIteration,
+                             Orchestrator& orchestrator,
                              const PhaseNumber inPhase) {
-        // This function is called after each iteration, so we never rate limit the
-        // first iteration. This means the number of completed operations is always
-        // `n * GlobalRateLimiter::_burstSize + m` instead of an exact multiple of
-        // _burstSize. `m` here is the number of threads using the rate limiter.
```
aggrand, line 145: Removed this comment because it's wrong, TIG-1825 moved to rate limiter to the == operator that runs before each iteration.

src/gennylib/include/gennylib/v1/Sleeper.hpp
```diff
@@ -49,6 +50,18 @@ class Sleeper {
     Sleeper(Sleeper&& other) = delete;
     Sleeper& operator=(Sleeper&& other) = delete;
 
+    constexpr void sleep_for(Orchestrator& o, const PhaseNumber pn,
```
rtimmons, line 53: 💅🏽 on the spacing (run clang-format) and `camelCase` conventions
And don't hate me, but `o` and `pn` were hard to parse in the admittedly simple impl. Although I just realized you're copying the existing names. Shrug. Your call on the param names.

src/gennylib/include/gennylib/v1/Sleeper.hpp
```diff
@@ -49,6 +50,18 @@ class Sleeper {
...
+            const Duration period, bool phaseChangeWakeup = false) const {
+
+        if (phaseChangeWakeup) {
+            // Using locks / condition variables is less efficient/safe, so we
+            // only use this mechanism if the caller explicitly asked for it.
+            o.sleepToPhaseEnd(pn, period);
+        } else if (period.count() && o.currentPhase() == pn) {
```
rtimmons, line 60: `period.count() > 0`? (Just realized this is the same as elsewhere in this file but it's weird if it's doing the convert-sizes-to-bools thing. Maybe just me.)

src/gennylib/src/Orchestrator.cpp
```diff
@@ -160,4 +160,12 @@ void Orchestrator::abort() {
     _phaseChange.notify_all();
 }
 
+void Orchestrator::sleepToPhaseEnd(const PhaseNumber pn, Duration timeout) {
+    reader lock{_mutex};
+    if (this->_current != pn || state == State::PhaseEnded) {
+        return;
+    }
+    _phaseChange.wait_for(lock, timeout);
```
rtimmons, line 168: spurious wakeup could cause this to return early. I think you need to wrap this in a `while` with the inverted `if` predicate above being your loop predicate

```c++
while(this->_current == pn && state != PhaseEnded .....) { _phaseChange.wait_for(...) }
```

src/gennylib/test/GlobalRateLimiter_test.cpp
```diff
@@ -177,25 +177,43 @@ auto incProducer = std::make_shared<DefaultActorProducer<IncActor>>("IncActor");
...
-        };
-        REQUIRE_THROWS_WITH(fun(), Matches(R"(.*alongside either Duration or Repeat.*)"));
+        genny::ActorHelper ah{config, num_threads, {{"IncActor", incProducer}}};
+        auto runInBg = [&ah]() { ah.run(); };
+        std::thread t(runInBg);
+        std::this_thread::sleep_for(110ms);
+        t.join();
+
+        const auto state = getCurState();
+        REQUIRE(state == 72);
```
rtimmons, line 216: Noice

src/gennylib/include/gennylib/Orchestrator.hpp
```diff
@@ -104,6 +103,10 @@ class Orchestrator {
      */
     bool continueRunning() const;
 
+    /**
+     * Sleep until the given phase ends or time out.
```
rtimmons, line 107: What if the timeout occurs? What is the expected behavior for the rate-limiter? I don't know what the right answer is, but the possibility for timeouts seems real esp if we have very long phases and a very tight rate limits.

rtimmons (approved):
Change looks reasonable. Modulo some nitpicks and one potential concurrency thing to double-check me on. The timeout behavior leaves me a bit uneasy, but I could be too out of touch with this rn to have a real opinion on that.

src/gennylib/include/gennylib/Orchestrator.hpp
aggrand: Timeouts here are actually expected! The rate-limiter wraps its sleeps in a loop that checks whether it can consume a token, sleeping for the rate if not. The doc comment was bad it should really say "sleep for the duration, also waking up if the phase changes before the duration passes". I'll update to clarify.

src/gennylib/src/Orchestrator.cpp
aggrand: That's a good point. I originally ignored spurious wakeups because we do this check outside of the orchestrator, but you made me realize that if we don't account for the wakeups then we might not hit the correct rate (since we always sleep for the rate, if every thread woke up right before it was time to burst, then was put back to sleep for the rate again, then we'd go under the rate).

Added your suggestion and some timing-based extra details.

aggrand:
evergreen merge

PR #500, opened 2021-07-12
PERF-2330 & PERF-2255 miscellaneous changes

ghartnett:
Miscellaneous changes for PERF-2330 & PERF-2255.

The new loader test is used for PERF-2330

Other changes fix minor bugs in existing workloads or make it easier to change number of collections.

Commit 52b79ac can be useful during testing. If you're sometimes changing the number of documents being loaded and sometimes forget. This gives you a chance to spot the error when the test starts.


src/workloads/scale/LoadTest.yml
```diff
@@ -0,0 +1,119 @@
...
+  MaxPhases: &MaxPhases 20
+
+  # The Sample Document Shape.
+  Document: &Doc
+    ts: {^Now: {}}
+    caid: {^RandomInt: {min: 0, max: 1000}}
+    cuid: {^RandomInt: {min: 0, max: 100000}}
+    prod: {^RandomInt: {min: 0, max: 10000}}
+    price: {^RandomDouble: {min: 0.0, max: 1000.0}}
+    data: {^Join: {array: ["aaaaaaaaaa", {^FastRandomString: {length: {^RandomInt: {min: 0, max: 10}}}}]}}
```
dalyd, line 17: Why is one a FastRandomString and the next one a RandomString? Seems we should match them up.

src/workloads/scale/LoadTest.yml
```diff
@@ -0,0 +1,119 @@
...
+  FinalDocumentCount: &FinalDocumentCount 1000000
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
+      maxPoolSize: 400
+
+Actors:
+- Name: InitialLoad
```
dalyd, line 43: Maybe a more descriptive name, such as Load with no secondary indexes. It can be pithier than that.

src/workloads/scale/LoadTest.yml
```diff
@@ -0,0 +1,119 @@
...
+        Threads: *LoadThreads
+        CollectionCount: *CollectionCount
+        Database: *dbname
+        Repeat: 1
+        Document: *Doc
+        DocumentCount: *InitialNumDocs
+        Indexes: *FirstIndex
+        BatchSize: *LoadBatchSize
+
+- Name: SecondLoad
```
dalyd, line 60: Maybe Load with Three Indexes?

src/workloads/scale/LoadTest.yml
```diff
@@ -0,0 +1,119 @@
...
+      Active: [1, 3, 5]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        LogEvery: 10 seconds
+        Blocking: None
+
+AutoRun:
+  Requires:
+    mongodb_setup:
+      - replica
```
dalyd, line 119: Any other setups to include? Should we run this on standalone and single node also?

dalyd (approved):
Some optional suggestions. I would also suggest splitting this into two PRs so it goes as two commits into the repo. Is there any overlap between the two tickets?

src/workloads/scale/LoadTest.yml
ghartnett: Added shard-lite and single-replica

src/workloads/scale/LoadTest.yml
ghartnett: LoadWithTwoMoreIndexes

src/workloads/scale/LoadTest.yml
ghartnett: LoadWithOneIndex

src/workloads/scale/LoadTest.yml
ghartnett: I changed both to FastRandomString

ghartnett:
evergreen merge

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #502, opened 2021-07-14
TIG-3201 Update genny loader to support multiple threads per collections

jimoleary:
[patch](https://spruce.mongodb.com/version/60e82a780ae6060b4eaef626/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/cast_core/include/cast_core/actors/Loader.hpp
+++ b/src/cast_core/include/cast_core/actors/Loader.hpp
@@ -37,7 +37,7 @@ namespace genny::actor {
 class Loader : public Actor {
 
 public:
-    explicit Loader(ActorContext& context, uint thread, uint totalThreads);
+    explicit Loader(ActorContext& context, uint thread, size_t totalThreads);
     ~Loader() override = default;
 
     static std::string_view defaultName() {
--- a/src/cast_core/include/cast_core/actors/Loader.hpp
+++ b/src/cast_core/include/cast_core/actors/Loader.hpp
@@ -37,7 +37,7 @@ namespace genny::actor {
 class Loader : public Actor {
 
 public:
-    explicit Loader(ActorContext& context, uint thread);
+    explicit Loader(ActorContext& context, uint thread, uint totalThreads);
     ~Loader() override = default;
 
     static std::string_view defaultName() {
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -38,27 +38,66 @@ using index_type = std::pair<DocumentGenerator, std::optional<DocumentGenerator>
 
 /** @private */
 struct Loader::PhaseConfig {
-    PhaseConfig(PhaseContext& context, mongocxx::pool::entry& client, uint thread, ActorId id)
+    PhaseConfig(PhaseContext& context,
+                mongocxx::pool::entry& client,
+                uint thread,
+                uint totalThreads,
+                ActorId id)
         : database{(*client)[context["Database"].to<std::string>()]},
-          // The next line uses integer division. The Remainder is accounted for below.
-          numCollections{context["CollectionCount"].to<IntegerSpec>() /
-                         context["Threads"].to<IntegerSpec>()},
+          multipleThreadsPerCollection{
+              context["MultipleThreadsPerCollection"].maybe<bool>().value_or(false)},
+          // The next line uses integer division for non multithreaded configurations.
+          // The Remainder is accounted for below.
+          numCollections{multipleThreadsPerCollection
+                             ? 1
+                             : context["CollectionCount"].to<IntegerSpec>() /
+                                 context["Threads"].to<IntegerSpec>()},
           numDocuments{context["DocumentCount"].to<IntegerSpec>()},
           batchSize{context["BatchSize"].to<IntegerSpec>()},
           documentExpr{context["Document"].to<DocumentGenerator>(context, id)},
-          collectionOffset{numCollections * thread} {
-        auto& indexNodes = context["Indexes"];
-        for (auto [k, indexNode] : indexNodes) {
-            indexes.emplace_back(indexNode["keys"].to<DocumentGenerator>(context, id),
-                                 indexNode["options"].maybe<DocumentGenerator>(context, id));
-        }
-        if (thread == context["Threads"].to<int>() - 1) {
-            // Pick up any extra collections left over by the division
-            numCollections += context["CollectionCount"].to<uint>() % context["Threads"].to<uint>();
+          collectionOffset{multipleThreadsPerCollection
+                               ? thread % context["CollectionCount"].to<IntegerSpec>()
+                               : numCollections * thread} {
+        auto createIndexes = [&]() {
+            auto& indexNodes = context["Indexes"];
+            for (auto [k, indexNode] : indexNodes) {
+                indexes.emplace_back(indexNode["keys"].to<DocumentGenerator>(context, id),
+                                     indexNode["options"].maybe<DocumentGenerator>(context, id));
+            }
+        };
+
+        if (multipleThreadsPerCollection) {
+            if (context["Threads"]) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                    "Phase Config 'Threads' parameter is not supported if "
+                    "'MultipleThreadsPerCollection' is true."));
+            }
+            uint collectionCount = context["CollectionCount"].to<IntegerSpec>();
+            if (totalThreads % collectionCount != 0) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                    "'CollectionCount' must be an even divisor of 'Threads'."));
+            }
+            uint numThreads = totalThreads / collectionCount;
+            numDocuments = context["DocumentCount"].to<int64_t>() / numThreads;
+            if (thread / collectionCount == 0) {
+                // The first thread for each collection:
+                //   1. Picks up any extra documents left over by the division.
+                //   2. Is responsible for creating the indexes.
+                numDocuments += context["DocumentCount"].to<uint>() % numThreads;
+                createIndexes();
+            }
+        } else {
+            createIndexes();
+            if (thread == context["Threads"].to<int>() - 1) {
+                // Pick up any extra collections left over by the division
+                numCollections +=
+                    context["CollectionCount"].to<uint>() % context["Threads"].to<uint>();
+            }
         }
     }
 
     mongocxx::database database;
+    bool multipleThreadsPerCollection;
     int64_t numCollections;
     int64_t numDocuments;
     int64_t batchSize;
@@ -123,13 +162,13 @@ void genny::actor::Loader::run() {
     }
 }
 
-Loader::Loader(genny::ActorContext& context, uint thread)
+Loader::Loader(genny::ActorContext& context, uint thread, uint totalThreads)
     : Actor(context),
       _totalBulkLoad{context.operation("TotalBulkInsert", Loader::id())},
       _individualBulkLoad{context.operation("IndividualBulkInsert", Loader::id())},
       _indexBuild{context.operation("IndexBuild", Loader::id())},
       _client{std::move(context.client())},
-      _loop{context, _client, thread, Loader::id()} {}
+      _loop{context, _client, thread, totalThreads, Loader::id()} {}
 
 class LoaderProducer : public genny::ActorProducer {
 public:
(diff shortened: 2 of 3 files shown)


src/cast_core/src/Loader.cpp
```diff
@@ -38,27 +38,66 @@ using index_type = std::pair<DocumentGenerator, std::optional<DocumentGenerator>
...
+        if (multipleThreadsPerCollection) {
+            if (context["Threads"]) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                    "Phase Config 'Threads' parameter is not supported if "
+                    "'MultipleThreadsPerCollection' is true."));
+            }
+            uint collectionCount = context["CollectionCount"].to<IntegerSpec>();
+            if (totalThreads % collectionCount != 0) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                    "'CollectionCount' must be an even divisor of 'Threads'."));
```
rtimmons, line 78: consider including the configured `CollectionCount` and/or other yaml params for an easier time debugging.

rtimmons (approved):
LGTM from a genny perspective. Couple nits that are up to you. I didn't double-check all the division/mod logic but nothing obvious stood out to me.

Please ensure you do a `clang-format`, and do a cursory patch-build of workloads that use the Loader prior to merging.

src/cast_core/include/cast_core/actors/Loader.hpp
```diff
@@ -37,7 +37,7 @@ namespace genny::actor {
 class Loader : public Actor {
 
 public:
-    explicit Loader(ActorContext& context, uint thread);
+    explicit Loader(ActorContext& context, uint thread, uint totalThreads);
```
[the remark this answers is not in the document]
jimoleary: :+1:

src/cast_core/src/Loader.cpp
jimoleary: good idea.

jimoleary (commented):
Already did a ```clang-format-10 -i src/cast_core/src/Loader.cpp```.

I did a full [patch](https://spruce.mongodb.com/version/60e82a780ae6060b4eaef626/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) already. I forgot to add it to the description. 

I'm keeping the mod / div logic simple deliberately (hence the invalid configuration exception).

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,79 @@
...
+    Phases:
+      - Repeat: 1
+        Database: cold
+        CollectionCount: 6
+        Threads: 1
+        DocumentCount: 200000
+        BatchSize: 100000
+        Document:
+          a: {^RandomString: { length: 100 }}
+      - Repeat: 1
```
dalyd, line 23: I don't understand the point of his example outside of the last loader phase and maybe the first. What are the hot phases doing?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,79 @@
...
+        Database: hot
+        CollectionCount: 6
+        Threads: 1
+        DocumentCount: 200000
+        BatchSize: 100000
+        Document:
+          a: {^RandomString: { length: 100 }}
+        FindOptions:
```
dalyd, line 31: Is this doing anything? I don't see an index supported here.

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,79 @@
...
+          Hint: a_1 # Currently only support the index name.
+          Comment: "Phase 1 loader"
```
dalyd, line 44: This is Phase 2 now, isn't it?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,79 @@
...
+        CollectionCount: 2
+        MultipleThreadsPerCollection: true
+        DocumentCount: 200000
+        BatchSize: 100000
+        Document:
+          a: {^RandomString: { length: 100 }}
+        FindOptions:
+          Hint: a_1 # Currently only support the index name.
+          Comment: "Phase 1 loader"
+  - Name: MultipleLoadThreadsPerCollection
```
dalyd, line 57: Didn't we just use 5 threads per collection in the last phase of the previous actor?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,79 @@
...
+    Phases:
+      - Repeat: 1
+        Database: multithreaded
+        # CollectionCount must be an even divisor of Threads.
+        # Phase config cannot contain Threads if MultipleThreadsPerCollection is true.
+        # Create 10 collections and populate with 10 threads per collection.
+        MultipleThreadsPerCollection: true
+        CollectionCount: 10
+        # DocumentCount is the total document count. Each thread populate an equal fraction of the count.
+        DocumentCount: 200000
```
dalyd, line 69: Shouldn't one of document count and batchsize be adjusted here? I think there are 20k docs per collection total, and, 10 threads per collection would give 2k docs per thread, but the batch size is 100k. I guess that means that it's one insert per thread, but still a bit confusing.

dalyd (commented):
Thanks for adding an example, although I got confused while reading through it.

jimoleary (commented):
There was a bit of copy / pasta going on in the example.

I've tried to simplify and document it.

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,83 @@
...
+  Default:
+    QueryOptions:
+      maxPoolSize: 200
+
+Actors:
+  - Name: MultipleCollectionsPerLoaderThread
+    Type: Loader
+    Threads: 10
+    Phases:
+      # If CollectionCount is less than the PhaseConfig Threads, then unexpected outcomes will occur.
```
jimoleary, line 15: I added a comment here, if you specify a CollectionCount less than phase config Threads then only 1 thread will populate CollectionCount numbers of collections.

src/workloads/docs/Loader.yml
jimoleary: updated to include an index example.

src/workloads/docs/Loader.yml
jimoleary: yes but I also wanted to show a larger example.

src/cast_core/src/Loader.cpp
```diff
@@ -38,27 +38,68 @@ using index_type = std::pair<DocumentGenerator, std::optional<DocumentGenerator>
...
+                ActorId id)
         : database{(*client)[context["Database"].to<std::string>()]},
-          // The next line uses integer division. The Remainder is accounted for below.
-          numCollections{context["CollectionCount"].to<IntegerSpec>() /
-                         context["Threads"].to<IntegerSpec>()},
+          multipleThreadsPerCollection{
+              context["MultipleThreadsPerCollection"].maybe<bool>().value_or(false)},
+          // The next line uses integer division for non multithreaded configurations.
+          // The Remainder is accounted for below.
+          numCollections{multipleThreadsPerCollection
```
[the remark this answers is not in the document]
dalyd: Ack. Could you add a comment saying we are using an explicit option to avoid breaking existing functionality?

src/workloads/docs/Loader.yml
dalyd: Thanks. I think I could still use some more comments here or updates. The comments on line 17-19 -- are they out of date, or are they the confusing behavior?

dalyd: Maybe also add a comment that if you are using collections > threads, always make sure that the two thread entries match?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,83 @@
...
+        Document:
+          a: {^RandomString: { length: 100 }}
+        FindOptions:
+          Hint: a_index # Currently only support the index name.
+          Comment: "Phase 1 loader"
+        # Each thread creates the indexes for each collection it manages.
+        Indexes:
+          - keys: {a: 1}
+            options: {name: "a_index"}
+      - Repeat: 1
```
dalyd, line 35: This is another example of unintended and bad behavior, yes?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,83 @@
...
+      maxPoolSize: 200
+
+Actors:
+  - Name: MultipleCollectionsPerLoaderThread
+    Type: Loader
+    Threads: 10
+    Phases:
+      # If CollectionCount is less than the PhaseConfig Threads, then unexpected outcomes will occur.
+      - Repeat: 1
+        # create 6 collection (CollectionCount / PhaseConfig Threads) per thread over 10 threads.
```
dalyd, line 17: Oh, is this because threads below is 1, each (of the 10) thread creates 6 collections? Wow.

src/workloads/docs/Loader.yml
dalyd: Assuming these are examples of "bad" or surprising behavior, could we start with one that operates as expected, and explicitly call out that that should be the common case and what you should do?

src/workloads/docs/Loader.yml
```diff
@@ -0,0 +1,83 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+
+# More threads requires a larger connection pool.
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 200
+
```
dalyd, line 9: Could you add a comment up here explaining that there are two ways to use this actor. The first instantiation shows the default, and the second shows MultipleThreadsPerCollection: true. Use the second if you have more loader threads than collections to load.

src/cast_core/src/Loader.cpp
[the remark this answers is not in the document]
jimoleary: will do, I think that we should also open another new ticket to try to merge the functionality as you are suggesting. But I suspect that it may not be straight forward (mainly because there are existing tests).

src/workloads/docs/Loader.yml
jimoleary: Sure thing, more documentation is good.

src/workloads/docs/Loader.yml
jimoleary: This comment on line 15 is an explaining that there could be  bad behavior if the CollectionCount is ever greater than the phase thread count.

All of the examples here should be good behavior.

I will  move the line 15 comments up to a new top level intro comment (I think it will make more sense there).

src/workloads/docs/Loader.yml
jimoleary: Yes, hence the name **MultipleCollectionsPerLoaderThread**.

(Also all the indexes will be created once all the loading is complete).

src/workloads/docs/Loader.yml
jimoleary: I don't think so, this actually is internally consistent. I assumed this is how it was supposed to work?

jimoleary (commented):
I'll update the into block comment tomorrow.

It looks like we also need a new ticket to define what behavior is expected and validate that all the workloads are sane (and add some invariant error checking).

Then it may be possible to merge  the functionality without an explicit multithreaded flag. I suspect this is a non trivial amount of work (for little pay off).

src/cast_core/src/Loader.cpp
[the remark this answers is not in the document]
dalyd: Only open the ticket if there seems to be worth pushing on at some point. Otherwise, just document and move on.

src/workloads/docs/Loader.yml
dalyd: I thought the good case would be Threads: N in both the actor and phase config. Don't they differ in each of these examples? I find it surprising that the first example creates 10 collections (and I wrote the original thing). If the parameter was "CollectionCountPerThreadsAsSpecificInPhaseConfig" then it would make sense :)

src/workloads/docs/Loader.yml
dalyd: I expected Threads: 10, and ending up with exactly 6 collections would be the expected case, like in BigUpdate. My expectation is CollectionCount should be the total number of collections added. I think it did that when I first wrote it.  

Before this change https://github.com/mongodb/genny/pull/240 threads was only specified in one place.

dalyd (commented):
More on the conversation. Despite my questions, don't over think it :) Also, I'm out Thursday, Friday, so don't block on my approval either.

src/workloads/docs/Loader.yml
jimoleary: I'll add a 10:10:1 (threads / threads / collections) example first and keep the current ones.

I was trying to document the current behavior (as it isn't documented anywhere else).

src/workloads/docs/Loader.yml
jimoleary: I was also surprised that CollectionCount is not the number collections created but to be fair I was also surprised that threads must be specified twice (especially if it is the same value). 

I don't want to change the current behavior as LargeScaleModel.yml , LatrgeScaleSerial.yml, OutOfCacheScanner.yml and OutOfCacheSnapShotReads.yml use odd combinations of values.

jimoleary (commented):
I will likely update the comments and merge tomorrow.

src/workloads/docs/Loader.yml
jimoleary: done

PR #503, opened 2021-07-14
PERF-2448 Benchmark an unsharded lookup in a sharded environment

jlap199:
Ran a $lookup against two unsharded collections. One was run in a sharded environment with the sharded lookup feature flag turned on, the other was run with the feature flag turned off. The query was ~33% faster with the feature flag disabled.

Metric | shard-lite | shard-lite-all-feature-flags
-- | -- | --
AverageLatency | 981,263,356 | 1,467,256,868

[Evergreen patch](https://spruce.mongodb.com/version/60f21666d6d80a245ce1599a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/execution/UnshardedLookup.yml
+++ b/src/workloads/execution/UnshardedLookup.yml
@@ -1,0 +1,70 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Benchmark a $lookup from one unsharded collection to another in a sharded environment.
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    DocumentCount: 1000
+    BatchSize: &batchSize 1000
+    CollectionCount: 2
+    Document:
+      shardKey: {^RandomInt: {min: 1, max: 1000}}
+      a: {^RandomInt: {min: 1, max: 1000}}
+      b: &string {^RandomString: {length: 10000}}
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: Quiesce
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+
+- Name: Lookup
+  Type: RunCommand
+  Database: *db
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10
+    Database: *db
+    Operations:
+    - OperationMetricsName: Lookup
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline:
+          [{
+            $lookup: {
+              from: "Collection1",
+              let: {a0: "$a"},
+              pipeline: [{
+                $match: {
+                  $expr: {$eq: ["$a", "$$a0"]},
+                }
+              }],
+              as: "joined"
+            }
+          }]
+        cursor: {batchSize: *batchSize}
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - shard-lite
+          - shard-lite-all-feature-flags
--- a/src/workloads/execution/UnshardedLookup.yml
+++ b/src/workloads/execution/UnshardedLookup.yml
@@ -11,11 +11,11 @@ Actors:
   - Repeat: 1
     Database: &db test
     Threads: 1
-    DocumentCount: 1000
-    BatchSize: &batchSize 1000
+    DocumentCount: 500
+    BatchSize: &batchSize 500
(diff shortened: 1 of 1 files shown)


src/workloads/execution/UnshardedLookup.yml
```diff
@@ -0,0 +1,70 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    DocumentCount: 1000
+    BatchSize: &batchSize 1000
+    CollectionCount: 2
+    Document:
+      shardKey: {^RandomInt: {min: 1, max: 1000}}
```
kaywux, line 18: Since these are unsharded collections, do we need this 'shardKey' field?

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -0,0 +1,70 @@
...
+      OperationCommand:
+        aggregate: Collection0
+        pipeline:
+          [{
+            $lookup: {
+              from: "Collection1",
+              let: {a0: "$a"},
+              pipeline: [{
+                $match: {
+                  $expr: {$eq: ["$a", "$$a0"]},
```
kaywux, line 57: If 'a' is a random int from 1-1000, and we only have 1000 documents (is this for each collection or in total?) with an even distribution there won't be many "hits" right? Aka each doc flowing into the $lookup stage will on average only match one result. Can we add another phase to measure the performance for when we'd have to send more results over the network?

kaywux (changes requested):
This is looking great, thanks Joshua! I left a few comments below. The 25% regression is interesting and slightly unexpected - as talked about offline, I'm interested in seeing the difference that the local read optimization will make to this.

src/workloads/execution/UnshardedLookup.yml
jlap199: No we don't. I was originally creating a sharded collection with a single chunk, but I later realized I could just run with the feature flag on/off. This can be removed.

src/workloads/execution/UnshardedLookup.yml
jlap199: Yes that's correct, on average we'd expect 1 match per document. Note each document has a 10KB string field in order to make sure we're sending a non-trivial payload through the network interface. But I can certainly add another benchmark with a greater payload if you think it would be useful.

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -0,0 +1,88 @@
...
+              let: {b0: "$b"},
+              pipeline: [{
+                $match: {
+                  $expr: {$eq: ["$$b0", "$b"]},
+                }
+              }],
+              as: "joined"
+            }
+          }]
+        cursor: {batchSize: *batchSize}
```
HanaPearlman, line 81: I did want to raise the question of result set size, which was discussed on my PR [in this thread](https://github.com/mongodb/genny/pull/491#discussion_r660010303). Here are my quick estimations for this workload.

For the second query: each document size is around 10KB and we expect 10 matches per document on the left. This would give us an expected result set size of at least 100MB. 

For the first query: we expect 1 match per document on the left. The amount of data we expect to see in the result set is 1000 documents on the left plus 1000 documents on the right, each at 10KB = 20MB.

What do you think?

HanaPearlman:
This is looking pretty good to me! Just the one question about result set size.

src/workloads/execution/UnshardedLookup.yml
kaywux: Ah got it, I didn't see that the 'c' field was so big - thanks for that clarification!

src/workloads/execution/UnshardedLookup.yml
kaywux: Thanks for pointing that out Hana! Personally I'd like to avoid adding many new stages to filter the returned results unless it's really necessary, since they could add significant bulk to the latency (especially $count, which desugars to a group) and if anything changes with the implementation or optimization of those stages, these workload results will also reflect those non-lookup related changes.

Instead, what do you think of decreasing the size of the 'c' field for each document, and slightly increasing the chance for more "hits" - so that each payload going over the network has more variation instead of being either one large doc or nothing?

jlap199: Yes thank you @HanaPearlman, this peculiarity invalidated the original benchmark numbers.

@kaywux I see your point. Although I'm not sure the size of the data set is large enough for the extra aggregation phase to have a meaningful impact on the performance differences. The lookup is still the dominate operation in terms of execution time. But I'm happy to remove the aggregation if you think it's minimizing the difference between the two executions.

kaywux: Yep that would be great, thanks in advance for the changes!

kaywux (approved):
This LGTM, thanks for the changes Joshua!

HanaPearlman (approved):
LGTM! Thanks

PR #505, opened 2021-07-19
TIG-3200 Fix ColdScanner's CollectionCount

rtimmons:
[Not ready for merge yet just here as reference point.]

Each of the 6 instances wants to iterate over `Collection{0..9}` rather than instance `$i` scanning `Collection$i{0..9}`.


luke-pearson:
For now I think this is still wrong.

luke-pearson:
Closing in place of #595

PR #506, opened 2021-07-20
PERF-2363, PERF-2364, PERF-2365, PERF-2366: Targeted and untargeted g…

militsasotirova:
…raphLookup from sharded and unsharded local collections

This PR is for the $graphLookup workloads for [PERF-2363](https://jira.mongodb.org/browse/PERF-2363), [PERF-2364](https://jira.mongodb.org/browse/PERF-2364), [PERF-2365](https://jira.mongodb.org/browse/PERF-2365), [PERF-2366](https://jira.mongodb.org/browse/PERF-2366) (targeted and untargeted, with sharded and unsharded local collections). Here are the initial results (I'll run more trials on evergreen once the workload is approved/the questions below are resolved):
Screen Shot 2021-07-20 at 9 42 15 AM

I modified MonotonicLoader so that the actor can also generate a field that increases with the '_id' field plus an offset.

A couple open questions: 

1. The collection size is currently 20k so that we could get as many results that could fit into one batch. Should we reduce the collection size so that the untargeted queries take less time?
2. I had two queries for each ticket: one where the recursion of $graphLookup was not allowed and one where I allowed it to go to a `maxDepth` of 13, thinking that the timing of the queries would differ. However, it seems like they take about the same amount of time - this could be because of the way cacheing works in $graphLookup and I'm wondering if this is an issue/something we need to write a different kind of test to address.

### DIFF
--- a/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
+++ b/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
@@ -34,9 +34,9 @@ namespace genny::actor {
  * number of collections, and list of indexes are all adjustable from the yaml configuration.
  *
  * This actor is identical to `Loader` except it loads with monotonically increasing _ids and it has
- * an additional pair of arguments ('FieldIncreasingByOffsetFromID' and 'OffsetFromID') to generate
- * a field in each document whose value is 'OffsetFromID' greater than the '_id' of the document.
- * For example, if the field name is 'a' and the offset is 1, the documents will look like
+ * an optional additional pair of arguments ('FieldIncreasingByOffsetFromID' and 'OffsetFromID') to 
+ * generate a field in each document whose value is 'OffsetFromID' greater than the '_id' of the 
+ * document. For example, if the field name is 'a' and the offset is 1, the documents will look like
  *      {
  *          "_id": 1,
  *          "a": 2,
--- a/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
+++ b/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
@@ -33,20 +33,7 @@ namespace genny::actor {
  * indexes. Each collection is identically configured. The document shape, number of documents,
  * number of collections, and list of indexes are all adjustable from the yaml configuration.
  *
- * This actor is identical to `Loader` except it loads with monotonically increasing _ids and it has
- * an optional additional pair of arguments ('FieldIncreasingByOffsetFromID' and 'OffsetFromID') to 
- * generate a field in each document whose value is 'OffsetFromID' greater than the '_id' of the 
- * document. For example, if the field name is 'a' and the offset is 1, the documents will look like
- *      {
- *          "_id": 1,
- *          "a": 2,
- *          ...
- *      },
- *      {
- *          "_id": 2,
- *          "a": 3,
- *          ...
- *      }, ...
+ * This actor is identical to `Loader` except it loads with monotonically increasing _ids.
  *
  * TODO: This specialized actor should be removed once the current value generator supports
  * monotonically increasing values.
--- a/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
+++ b/src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
@@ -33,7 +33,20 @@ namespace genny::actor {
  * indexes. Each collection is identically configured. The document shape, number of documents,
  * number of collections, and list of indexes are all adjustable from the yaml configuration.
  *
- * This actor is identical to `Loader` except it loads with monotonically increasing _ids.
+ * This actor is identical to `Loader` except it loads with monotonically increasing _ids and it has
+ * an additional pair of arguments ('FieldIncreasingByOffsetFromID' and 'OffsetFromID') to generate
+ * a field in each document whose value is 'OffsetFromID' greater than the '_id' of the document.
+ * For example, if the field name is 'a' and the offset is 1, the documents will look like
+ *      {
+ *          "_id": 1,
+ *          "a": 2,
+ *          ...
+ *      },
+ *      {
+ *          "_id": 2,
+ *          "a": 3,
+ *          ...
+ *      }, ...
  *
  * TODO: This specialized actor should be removed once the current value generator supports
  * monotonically increasing values.
--- a/src/cast_core/src/MonotonicLoader.cpp
+++ b/src/cast_core/src/MonotonicLoader.cpp
@@ -14,6 +14,7 @@
 
 #include <cast_core/actors/MonotonicLoader.hpp>
 
+#include <iostream>
 #include <memory>
 
 #include <yaml-cpp/yaml.h>
@@ -50,6 +51,8 @@ struct MonotonicLoader::PhaseConfig {
                          context["Threads"].to<IntegerSpec>()},
           numDocuments{context["DocumentCount"].to<IntegerSpec>()},
           batchSize{context["BatchSize"].to<IntegerSpec>()},
+          fieldIncreasingByOffsetFromID{context["FieldIncreasingByOffsetFromID"].to<std::string>()},
+          offsetFromID{context["OffsetFromID"].to<IntegerSpec>()},
           documentExpr{context["Document"].to<DocumentGenerator>(context, id)},
           collectionOffset{numCollections * thread} {
         auto& indexNodes = context["Indexes"];
@@ -67,6 +70,8 @@ struct MonotonicLoader::PhaseConfig {
     int64_t numCollections;
     int64_t numDocuments;
     int64_t batchSize;
+    std::string fieldIncreasingByOffsetFromID;
+    int64_t offsetFromID;
     DocumentGenerator documentExpr;
     std::vector<index_type> indexes;
     int64_t collectionOffset;
@@ -95,6 +100,8 @@ void genny::actor::MonotonicLoader::run() {
                             auto tmpDoc = config->documentExpr();
                             auto builder = bsoncxx::builder::stream::document();
                             builder << "_id" << ++id_num;
+                            builder << config->fieldIncreasingByOffsetFromID 
+                                << (id_num + config->offsetFromID);
                             builder << bsoncxx::builder::concatenate(tmpDoc.view());
                             bsoncxx::document::value newDoc = builder
                                 << bsoncxx::builder::stream::finalize;
--- a/src/cast_core/src/MonotonicLoader.cpp
+++ b/src/cast_core/src/MonotonicLoader.cpp
@@ -14,7 +14,6 @@
 
 #include <cast_core/actors/MonotonicLoader.hpp>
 
-#include <iostream>
 #include <memory>
 
 #include <yaml-cpp/yaml.h>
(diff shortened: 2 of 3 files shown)


src/cast_core/src/MonotonicLoader.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <cast_core/actors/MonotonicLoader.hpp>
 
+#include <iostream>
```
HanaPearlman, line 17: I think we can remove this dependency, right?

src/workloads/execution/ShardedGraphLookup.yml
```diff
@@ -0,0 +1,268 @@
...
+    # all documents whose '_id' is greater than or equal to its own (limited by maxDepth).
+
+    # Similarly, for the recursive untargeted queries below: for each document X in Collection0
+    # we find the document Y in Collection0 where X's 'a' value (startWith) is equal to Y's 'a' 
+    # value (connectToField). Then, we look at Y's '_id' (connectFromField) and find the document Z
+    # in Collection0 that has the same value for its 'a' value (connectToField), and so on. 
+    # Because the collection is sharded by '_id' and 'a' is the connectToField, the $graphLookup
+    # is not targeted to any specific shard. Since the connectToField and connectFromField are 
+    # swapped from the targeted queries, each document will have in its matches all documents whose
+    # '_id' is less than or equal to its own (limited by maxDepth).
```
HanaPearlman, line 101: I think the comments in the file are super helpful-- thank you!

src/workloads/execution/ShardedGraphLookup.yml
```diff
@@ -0,0 +1,268 @@
...
+        aggregate: Collection0
+        pipeline:
+          [{
+            $graphLookup: {
+              from: "Collection0", 
+              startWith: "$_id", 
+              connectFromField: "a", 
+              connectToField: "_id", 
+              as: "matches",
+              maxDepth: 13
```
HanaPearlman, line 138: Note to other reviewers, here was the thinking about the maxDepth value: The size of the result set will look approximately like `numDocs * sizeOfDoc * (maxDepth + 1) + numDocs * sizeOfDoc`, where the first number represents all of the matches from the right side and the second represents the documents on the left side. With the current doc size of approx 35 bytes, and a collection size of 20k, we'll get around 10MB of output data for a depth of 13.

The untargeted queries take a long time (~2 minutes). This could be caused by the collection size, which is pretty large compared to the collection sizes of recent $lookup workloads. An alternative, while still maintaining a similar result set size, could be to increase the size of each document and decrease the collection size.

src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
```diff
@@ -33,7 +33,20 @@ namespace genny::actor {
...
+ *      {
+ *          "_id": 1,
+ *          "a": 2,
+ *          ...
+ *      },
+ *      {
+ *          "_id": 2,
+ *          "a": 3,
+ *          ...
+ *      }, ...
```
HanaPearlman, line 49: The idea behind this change was to allow easy calculation of result set size for the $graphLookup queries with `maxDepth > 0`, see my comment below for more details. But I _think_ this has had the unfortunate side effect of reducing the difference in performance for the $graphLookups with `maxDepth = 0` vs `maxDepth > 0`.

I have not confirmed this, but if the cache maintained by DocumentSourceGraphLookup looks like a mapping from `valueOfConnectToField` to `documentInForeignCollection`, then after just 1 iteration, all of the documents in the foreign collection should be in the cache and we can do iterations without having to do any additional querying. If this is true, then there seems to be a trade off between meaningfully testing `maxDepth > 0` and correctly bounding our result set size.

src/cast_core/src/MonotonicLoader.cpp
```diff
@@ -95,6 +100,8 @@ void genny::actor::MonotonicLoader::run() {
                             auto tmpDoc = config->documentExpr();
                             auto builder = bsoncxx::builder::stream::document();
                             builder << "_id" << ++id_num;
+                            builder << config->fieldIncreasingByOffsetFromID 
+                                << (id_num + config->offsetFromID);
```
HanaPearlman, line 104: Note that the changes in this file are causing failures for other workloads. These fields should be optional, since other workloads do not require this functionality.

HanaPearlman (commented):
Two quick points of feedback and a couple of notes with more details for other reviewers!

src/cast_core/include/cast_core/actors/MonotonicLoader.hpp
kaywux: Hmm yeah that would explain the benchmark results that we're seeing for recursive vs non-recursive $graphLookup. Perhaps instead of using the MonotonicLoader we could instead construct the collection such that the first iteration of the search only surfaces a portion of the total documents so that all of them aren't added to the cache at depth 0 - for example, if we construct documents with `{a: {^RandomInt: {min: 1, max: 100}}, b: {^RandomInt: {min: 1, max: 100}}, c: {^RandomInt: {min: 1, max: 10}}}` and we join on `a` and `b` but start with `c`. With this approach it might not be possible to completely avoid filling the cache such that later iterations wouldn't mostly consult the cache instead of doing additional querying, but I'm okay with that. This would at least give us a predictable measure of how the result set would grow so that we can continue bounding the result size of the recursive $graphLookup workloads using the `maxDepth` field.

Another option would be to add a new Loader Actor that would give us some control over what fraction of the generated documents have certain value ranges for each field. This way we could artificially ensure that new documents are introduced to the frontier for each iteration of the search. I haven't thought through all the details but this approach would obviously require more changes than the one above. Personally I'm not sure if this is worth the effort, but of either of you feel differently I could be convinced!

kaywux (changes requested):
Thanks for the first pass and for the explanatory comments! I have a few general comments about how to measure recursive $graphLookup workloads, interested to hear your thoughts on that.

militsasotirova:
I modified the workload so that the documents are randomized to avoid every document being in the cache after the first iteration. I also added more queries for different cases of the `maxDepth` field. When `maxDepth` wasn't specified, I used a smaller foreign collection because in the worst case scenario all of the documents in the local collection will match all of the documents in the foreign, and we wanted to stay under 16MB for the result set size. Here are the timing results for the updated workload:
Screen Shot 2021-07-22 at 4 30 58 PM

One possible reason for the lack of difference between targeted and untargeted for the maxDepth = 0 (No Recursion) and maxDepth = 2 (Depth Limited) cases is because the maxDepth is low relative to the situation where it is unspecified (Depth Unlimited).

src/workloads/execution/ShardedGraphLookup.yml
```diff
@@ -231,36 +254,117 @@ Actors:
...
               as: "matches",
-              maxDepth: 13
+              maxDepth: 2
+            }
+          }]
+        # To get meaningful results, the entire result set should fit in a single batch. This should
+        # be possible since both collections are small.
+        cursor: {batchSize: *num_docs_total}
+
+    # In the following queries the 'maxDepth' field is not specfied, so the $graphLookup runs
```
kaywux, line 288: nit: "specified"!

src/workloads/execution/ShardedGraphLookup.yml
```diff
@@ -0,0 +1,372 @@
...
+  - *Nop
+  - Repeat: 10
+    Database: *Database
+    Operations:
+    # The recursive targeted queries below work as follows: for each document X in Collection0 
+    # we find any document Y in Collection0 where X's 'c' field (startWith) is equal to Y's 'b'
+    # (connectToField). Then, we look at Y's 'a' (connectFromField) value and find any document Z in 
+    # Collection0 that has the same value for its 'b' (connectToField), and so on. 
+    # Because the collection is sharded by 'b' the $graphLookup is targeted towards 
+    # specific shard(s).
```
kaywux, line 116: Above when you're sharding collections on lines 35 and 41, you're sharding by `shardKey` - which isn't used again. Did you mean to shard by `b`? Otherwise these "targeted" queries will actually be untargeted since they'd also have to do a scatter-gather to all the shards.

kaywux (changes requested):
Thanks for making these changes Militsa! I noticed one thing that would explain the similarities between the targeted and untargeted benchmark results - it looks like the "targeted" queries might not actually be on the field that the collections are sharded on. Once that's changed, I'd be curious to see if the results are more in line with what we'd expect!

src/workloads/execution/ShardedGraphLookup.yml
militsasotirova: Yes I did mean to shard by `b`! Thanks for catching that.

militsasotirova:
After making the change to shard by `b`, I got the following results:
Screen Shot 2021-07-27 at 12 53 21 PM

There still isn’t much of a difference between targeted and untargeted for maxDepth = 0 and maxDepth = 2, but there is a much larger difference between targeted and untargeted for unspecified maxDepth (before, targeted unspecified maxDepth graphLookups took ~4 seconds but now its at ~2)

kaywux:
Thanks for running this again Militsa! Have you run these since Hana pushed her top-level parallel $graphLookup patch? At first I was confounded by the similarities between the benchmark results for sharded vs. unsharded local collection for the non-depth unlimited cases (I'd expect the sharded workload to be faster since there would be parallelism), would love to see the results when we take that into account!

militsasotirova:
With Hana's changes, we get:
Screen Shot 2021-07-30 at 9 23 15 AM
[Here's](https://spruce.mongodb.com/version/610318aa57e85a7fb12f2089/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) the link to the evergreen run.
There still doesn't seem to be difference between targeted and untargeted in the non-depth unlimited cases - maybe its the collection size?

HanaPearlman (commented):
Thanks for the back and forth so far Militsa!!

The depth unlimited case is turning out to be useful for showing the speedup both from untargeted to targeted and from unsharded local collection to sharded local collection.  These results looks great.

Unfortunately, it seems like the depth limited cases are not really reflecting these. My guess for this is around the way that we are doing the querying. Could it be that in attempting to limit the result set size (by creating field 'c' and using it as our start field) we are accidentally creating a scenario where after completing the search for the first 10ish documents from the local collection, we have _all_ relevant documents in our cache? This could 1) reduce the number of recursive queries we need to do, and therefore reduce the difference in targeted vs untargeted and 2) potentially give the unsharded case a small boost, since it has a shared cache whereas the sharded case does not have a (singular) shared cache but does benefit from parallelism.

If so, we're a bit stuck. Ideally we could write queries without starting with the 'c' field, but then our collections will need to be tiny, and we won't be able to get very meaningful results.

Editing to say: sorry for this late, critical comment! I thought this was ready to approve, but then I had a longer think about why we're seeing such similar numbers across the depth limited cases.

kaywux (approved):
Thanks for the work on this Militsa, I know there's been a lot of flip flopping and investigation!

After thinking about this, I'm okay with punting the depth-limited case (at least, figuring out a meaningful way of testing it) in favor of getting these other benchmarks in. The main goal of this ticket was to have a baseline benchmark for targeted/untargeted $graphLookup performance in both non-recursive and recursive cases, and in the latter case we have good information from the `DepthUnlimited` cases. It's up to you and Hana whether you want to rework or remove the `DepthLimited` benchmarks - I personally don't see the harm in leaving them in; if we do we might want to drop a comment explaining the cache behavior that leads to the surprising perf results.

LGTM, great job!

HanaPearlman (approved):
LGTM!

militsasotirova:
evergreen merge

PR #507, opened 2021-07-27
PERF-2358 Targeted lookup from an unsharded collection to a sharded one

jlap199:
PERF-2359 Targeted lookup from a sharded collection to a sharded collection
PERF-2362 Untargeted lookup from an unsharded collection to a sharded one

Benchmark | AverageLatency
-- | --
UntargetedLookupUnshardedIntoSharded | 6,460,528,788
UntargetedLookupShardedIntoSharded | 3,505,457,432
TargetedLookupUnshardedIntoSharded | 2,106,877,279
TargetedLookupShardedIntoSharded | 1,087,654,546

A few things to note are that targeted tests are faster than untargeted since certain shards can be left unread. Also an untargeted lookup where the left side is sharded is ~2X faster than when the left side is unsharded. This because the former lookup can be parallelized across shards.

### DIFF
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -7,7 +7,7 @@ Description: |
     1. Creating empty sharded collections distributed across all shards in the cluster.
     2. Populating collections with data.
     3. Fsync.
-    4. Running untargeted $lookup from a sharded collection.
+    4. Running targeted and untargeted $lookups against a sharded foreign collection.
 
 Actors:
 - Name: CreateShardedCollections
@@ -50,15 +50,15 @@ Actors:
   Phases:
   - *Nop
   - Repeat: 1
-    BatchSize: 1000
+    BatchSize: 3000
     Threads: 1
     DocumentCount: &NumDocs 3000
     Database: *Database
     CollectionCount: 3    # Loader will populate 'Collection0', 'Collection1' and 'Collection2'.
     Document:
       shardKey: {^RandomInt: {min: 1, max: 100}}
-      date: &Date {^RandomDate: {min: "2020-01-01", max: "2021-01-01"}}
       int: {^RandomInt: {min: 1, max: 100}}
+      str: {^RandomString: {length: 100}}
   - *Nop
   - *Nop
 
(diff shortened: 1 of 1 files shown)


src/workloads/execution/ShardedLookup.yml
```diff
@@ -111,6 +111,62 @@ Actors:
...
+          [
+            {$lookup: {
+              from: "Collection1",
+              localField: "shardKey",
+              foreignField: "shardKey",
+              as: "joined"
+            }}
+          ]
+        cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: UntargetedLookupShardedIntoSharded # PERF-2359
```
kaywux, line 128: Copy-paste error? The untargeted sharded to sharded workload already exists (line 89), PERF-2359 refers to the targeted version you have above on line 114.

kaywux (commented):
This is looking great, thanks Joshua! One general comment - for the sake of consistency since we already have the sharded to sharded untargeted operation in the same workload, could we keep the four benchmarks similar so we can draw comparisons between them without as many confounding variables? For example, the former benchmark has a $lookup subpipeline that's slightly more involved and varying (discussed briefly [here](https://github.com/mongodb/genny/pull/491/files#r651946985)) - I'm ambivalent about whether we want to change the existing one or your newly added ones, but do think that either way all four should have roughly the same structure (except for differences in whether or not the query is targeted or not). Happy to discuss this!

Adding Hana to the review as she added that first $lookup benchmark.

HanaPearlman:
Cool results! I agree with Katherine about consistency with the existing benchmark in this file. I'm not particularly attached to the way I wrote it, and for simplicity of code it could be nice to have everything use the localField/foreignField syntax instead of the pipeline syntax. I would be fine with it if you changed the existing pipeline to
`[
    {$lookup: {
        from: "Collection1",
        localField: "int",
        foreignField: "int",
        as: "joined"
    }}
]`

Otherwise, this LGTM!

src/workloads/execution/ShardedLookup.yml
jlap199: No this was purposeful. I wanted to get an apples to apples comparison to the targeted sharded to sharded workload, without modifying Hana's existing workload. But you're right, technically this is duplicated work.

HanaPearlman (approved):
LGTM! Thanks

PR #508, opened 2021-07-27
PERF-2390 Lookup from sharded collection to unsharded collection

jlap199:
With the feature flag turned on:

Benchmark | AverageLatency
-- | --
MatchSingleShardLookupUnsharded | 14,243,024
MatchAllShardsLookupUnsharded | 14,884,660

With the feature flag turned off:

Benchmark | AverageLatency
-- | --
MatchSingleShardLookupUnsharded | 9,508,384
MatchAllShardsLookupUnsharded | 12,764,823

When the feature flag is enabled, there is a bit of a regression in the performance of $lookup from a small sharded collection to an unsharded collection. Also surprisingly there doesn't seem to be much of a difference between the two benchmarks when the feature flag is enabled, even though one must scan all shards before performing the lookup whereas the other only needs to read from a single shard.

### DIFF
--- a/src/workloads/execution/UnshardedLookup.yml
+++ b/src/workloads/execution/UnshardedLookup.yml
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  Benchmark a $lookup where the foreign collection (right side) is unsharded.
+  Benchmark a $lookup in a sharded environment where the foreign collection is unsharded.
 
 Actors:
 - Name: CreateShardedCollections
@@ -14,13 +14,17 @@ Actors:
     - OperationMetricsName: EnableSharding
       OperationName: AdminCommand
       OperationCommand:
-        enableSharding: &Database test
+        enableSharding: &db test
     - OperationMetricsName: ShardSourceCollection
       OperationName: AdminCommand
       OperationCommand:
         shardCollection: test.Collection2
         key: {a: hashed}
-        numInitialChunks: &NumChunks 8
+        numInitialChunks: 8
+    - OperationMetricsName: DisableBalancer
+      OperationName: AdminCommand
+      OperationCommand:
+        balancerStop: 1
   - &Nop {Nop: true}
   - *Nop
   - *Nop
@@ -31,15 +35,17 @@ Actors:
   Phases:
   - *Nop
   - Repeat: 1
-    Database: &db test
+    Database: *db
     Threads: 1
-    DocumentCount: 1000
-    BatchSize: &batchSize 1000
+    DocumentCount: 2000
+    BatchSize: &batchSize 2000
     CollectionCount: 3
     Document:
       a: {^RandomInt: {min: 1, max: 100}}
       b: {^RandomInt: {min: 1, max: 10}}
-      c: {^RandomInt: {min: 0, max: 999}}
+      c: {^RandomInt: {min: 1, max: 10}}
+      d: {^RandomInt: {min: 1, max: 1000}}
+      s: {^RandomString: {length: 1000}}
   - *Nop
   - *Nop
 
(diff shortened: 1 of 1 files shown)


src/workloads/execution/UnshardedLookup.yml
```diff
@@ -71,6 +95,38 @@ Actors:
             }}
           ]
         cursor: {batchSize: *batchSize}
+    - OperationMetricsName: ShardedSourceLookup1
```
jlap199, line 98: I should give these better names. This benchmark needs to scan all shards of the left(input) collection. The other test however only needs to scan one since it matches on the shard key.

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -1,31 +1,54 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  Benchmark a $lookup from one unsharded collection to another in a sharded environment.
+  Benchmark a $lookup where the foreign collection (right side) is unsharded.
```
kaywux, line 4: nit: Could you change this to "Benchmark a $lookup in a sharded environment where ..."? It would help with potential confusion if a reader was wondering why we we're running even unsharded-unsharded workloads with sharding enabled.

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -71,6 +95,38 @@ Actors:
...
           ]
         cursor: {batchSize: *batchSize}
+    - OperationMetricsName: MatchAllShardsLookupUnsharded
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection2
+        pipeline:
+          [
+            {$match: {
+              c: {$in: [84, 112, 253, 396, 495, 515, 638, 754, 872, 998]}
```
kaywux, line 105: How do you feel about changing this match stage to something like `{$match: {c: {$lt: 5}}}`? Since the collection is sharded on 'a', we don't need to be so specific about the query in order to target all the shards - unless there was another reason for you to choose this match, in which case let me know!

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -71,6 +95,38 @@ Actors:
...
+            }}
+          ]
+        cursor: {batchSize: *batchSize}
+    - OperationMetricsName: MatchSingleShardLookupUnsharded
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection2
+        pipeline:
+          [
+            {$match: {a: 23}},
```
kaywux, line 121: At first glance I was worried that there might be cases where we wouldn't get any matches at all, but looks like that will only happen `0.99^1000 = 0.004%` of the time. Nice!

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -71,6 +95,38 @@ Actors:
             }}
           ]
         cursor: {batchSize: *batchSize}
+    - OperationMetricsName: MatchAllShardsLookupUnsharded
```
kaywux, line 98: Now that we have both sharded and unsharded to unsharded lookups, could you change this name to something like "ShardedToUnshardedLookupMatchAllShards", and same for the one below?

Oh and since I can't comment on unedited lines, similarly for line 70, "UnshardedToUnshardedLookupOneToFew"?

kaywux (dismissed):
> Also surprisingly there doesn't seem to be much of a difference between the two benchmarks when the feature flag is enabled, even though one must scan all shards before performing the lookup whereas the other only needs to read from a single shard.

That's because of the parallelism! $lookup can now run in parallel on the shards, meaning that the MatchAllShards benchmark has a $lookup running on each of the eight shards (but still has some regression compared to before because it needs to send results over the network). The MatchSingleShard on the other hand doesn't benefit from any parallelism, so it makes sense that it has a larger regression compared to with the feature flag off.

This is mostly ready, I left a couple small comments! Just to double check we're in the clear with the batch size, how big is the result set for each of the workloads?

src/workloads/execution/UnshardedLookup.yml
jlap199: Yes that would be better. I was originally filtering on `a`, and so this was used to target all the shards. I forgot to update the filter when I added `c`.

src/workloads/execution/UnshardedLookup.yml
jlap199: Yeah, although it would be nice if we had something like `{^SequentialInt: {min: 1, max: 100}}` for this purpose.

jlap199:
@kaywux you're right. How could I forget? Nice to see they align with expectations. As far as the batch size is concerned, we're good to go. The result set is ~2MB. As long as we're under 16MB, a single fetch will return the entire data set.

src/workloads/execution/UnshardedLookup.yml
kaywux: Agreed! Militsa edited the `MonotonicLoader` actor to accomplish something similar for the first iteration of her [$graphLookup benchmarks](https://github.com/mongodb/genny/pull/506), but as far as I know there's currently no built-in way in Genny to do this. Sounds like a feature request!

kaywux: Agreed! Militsa edited the MonotonicLoader actor to accomplish something similar for the first iteration of her [$graphLookup benchmarks](https://github.com/mongodb/genny/pull/506), but as far as I know there's currently no built-in way in Genny to do this. Sounds like a feature request!

kaywux:
> The result set is ~2MB. As long as we're under 16MB, a single fetch will return the entire data set.

This seems a little small, especially since (if I'm correct in assuming that the latency is measured in nanoseconds?) each phase finishes in only 0.014 seconds. Could you modify the $lookup subpipeline in these two benchmarks to try and hit more of the foreign collection, so we can get a better idea of what the regression would be when there's more to be sent over the network? This could be something as simple as `{$gt: ["$$b0, "$b"]}` instead of `$eq`, although I haven't done the math for what the expected result size would be for that.

kaywux:
Adding @HanaPearlman as a reviewer since I'll be out next week!

jlap199:
The updated benchmark results are:

With the feature flag turned on:

Benchmark | AverageLatency
-- | --
ShardedToUnshardedLookupMatchSomeShards | 189,236,846
ShardedToUnshardedLookupMatchAllShards | 190,400,885

With the feature flag turned off:

Benchmark | AverageLatency
-- | --
ShardedToUnshardedLookupMatchSomeShards | 163,162,257
ShardedToUnshardedLookupMatchSomeShards | 229,770,900

HanaPearlman (commented):
This is basically LGTM! Two small nits and then one question: could/should we be increasing the result set size even more? Do you have an estimate for how large it is currently?

src/workloads/execution/UnshardedLookup.yml
```diff
@@ -1,31 +1,56 @@
...
+    - OperationMetricsName: EnableSharding
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: &db test
+    - OperationMetricsName: ShardSourceCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: test.Collection2
+        key: {a: hashed}
+        numInitialChunks: 8
```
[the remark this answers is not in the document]
jlap199: It was kind of arbitrary. I picked it so that there wouldn't be too many or too few documents per chunk. In the case of two shards though we will get an even distribution.

jlap199:
@HanaPearlman in these new benchmarks the $lookup is processing ~16MB, however the final result set is much smaller since we project out the 1KB string field in each document after the join. This makes it so that most of the network latency we're measuring comes only from $lookup which is what we're concerned with. I would say the current data size is probably sufficient since the results are in line with expectations and processing time is on the order of seconds.

HanaPearlman (approved):
LGTM

src/workloads/execution/UnshardedLookup.yml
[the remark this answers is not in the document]
HanaPearlman: Sorry, one final comment here. Would you mind adding an operation here to disable the balancer? Had another brief conversation about it in my original PR. Wouldn't hurt to have it here too!

PR #512, opened 2021-08-05
PERF-2426: Add tests for caching behavior of unsharded $lookup

HanaPearlman:
From [this patch build](https://spruce.mongodb.com/version/610d3c9d1e2d173508a875db/tasks), we get the following results.

With feature flag disabled, in an unsharded environment:
|---|---|
| LookupWithCachedPrefixUnshardedToUnsharded | .1554993457 |

This is by far the fastest result. 

With feature flag disabled:
|---|---|
| LookupWithCachedPrefixUnshardedToUnsharded | 7.6843960346 |
| LookupWithCachedPrefixShardedToUnsharded | 7.6966880165|

With feature flag enabled:
|---|---|
| LookupWithCachedPrefixUnshardedToUnsharded | 9.1613125435 |
| LookupWithCachedPrefixShardedToUnsharded | 5.5058569847 |

When the feature flag is turned on, we see a slow down in the unsharded local case because we end up targetting shards instead of reading locally when we execute the $lookup subpipeline. Also, we see a speed up in the sharded local case because the $lookup can be parallelized.

We hoped that this testing might have revealed the bug where the $lookup cache is lost in a sharded environment, related to [SERVER-57168](https://jira.mongodb.org/browse/SERVER-57168). It turns out it's a bit hard to do that because there are other confounding factors that can lead to huge performance boosts in an unsharded environment.

### DIFF
--- a/src/workloads/execution/LookupInUnshardedEnvironment.yml
+++ b/src/workloads/execution/LookupInUnshardedEnvironment.yml
@@ -1,0 +1,73 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises $lookup in an unsharded environment.
+
+  The workload consists of the following phases:
+    1. Populating collections with data.
+    2. Fsync.
+    3. Running $lookup's with cacheable subpipeline prefixes.
+
+Actors:
+- Name: LoadInitialData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    BatchSize: 1000
+    Threads: 1
+    DocumentCount: &NumDocs 3000
+    Database: &Database test
+    CollectionCount: 2    # Loader will populate 'Collection0' and 'Collection1'.
+    Document:
+      key: {^RandomInt: {min: 1, max: 100}}
+      int: {^RandomInt: {min: 1, max: 100}}
+      str: {^RandomString: {length: 100}}
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: Quiesce
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        fsync: 1
+  - *Nop
+
+- Name: RunLookups
+  Type: RunCommand
+  Database: *Database
+  Phases:
+  - *Nop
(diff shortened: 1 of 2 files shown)


kaywux (approved):
Thanks Hana! These benchmark results are really interesting, a few thoughts overall:
- With the feature flag disabled, I'm surprised that the `LookupWithCachedPrefixShardedToUnsharded ` is basically equal with `LookupWithCachedPrefixUnshardedToUnsharded`! I expected to see a bit of a slowdown in the sharded case due to having to open cursors on the shards that contain the local collection and network latency from sending documents to the primary shard. Not sure why this is the case, do you have any ideas?
- I believe that the cache is being blown away in all cases that are run in a sharded environment, both with the flag on and off, since [the code that causes this](https://github.com/mongodb/mongo/blob/82f713b946460738ad15469c083f173ae0d1f80b/src/mongo/db/pipeline/sharded_agg_helpers.cpp#L1312) is run regardless. That might explain the order of magnitude difference between unsharded to unsharded lookups in a standalone vs sharded environment. Could you confirm if the $lookup is still executing the subpipeline on every input document in the latter cases (even with the flag off)? If so, then these tests would have (hopefully) served their purpose in highlighting the bug that SERVER-57168 will address.

Overall the actual workload LGTM, I'm interested in hearing your thoughts on the points above!

HanaPearlman:
Thanks Katherine!! Some responses:

- Yes, I found that weird too! Unfortunately, I don't have any guesses for why this might be; my reasoning was the same as yours.
- And again I agree, the cache should be blown away in any case in the sharded environment, which is why I added the unsharded environment test. While the presence of the cache might explain the order of magnitude improvement, it's hard to isolate just that effect. For example, another factor is the amount of data transfer over the network. Still, the queries definitely execute the cache vs no cache case (even if they may not have led us to find the bug immediately).

I have verified locally that the routing is working correctly for the cases in a sharded environment both with feature flag enabled and disabled. When I run the workload locally, I get one result per local document, and I can see that the top-level queries and all of the subpipelines are executing as expected (in terms of quantity and location) in the logs. Because of the $group, the result set size is not much larger than a single collection, so it definitely fits in a batch. 

I'm going to go forward with merging, given this verification, but I'm happy to continue the conversation!

kaywux:
Sounds good, thanks Hana!

HanaPearlman:
evergreen merge

PR #513, opened 2021-08-05
PERF-2378: Add tests for caching behavior of sharded $lookup

HanaPearlman:
From [this patch build](https://spruce.mongodb.com/version/610d3c9d1e2d173508a875db/tasks), we get the following results.

|---|---|
| LookupWithCachedPrefixUnshardedToSharded | 17.1584018563 |
| LookupWithCachedPrefixShardedToSharded | 9.4955312718 |

The queries with sharded local collections are faster because the $lookup can be parallelized.

It's interesting to see how these results relate to #512. Compared to the UnshardedToUnsharded (~9.2 sec) and ShardedToUnsharded (~5.5 sec) feature-flag-enabled benchmarks, we see a slow down when the foreign collection becomes sharded. A slow down is expected, since when the foreign collection is sharded we have to scatter-gather results instead of targeting just the primary. However, it's much slower than I expected-- the results for ShardedToSharded are even slower than ShardedToUnsharded  with the feature flag turned off. That is, it seems that the effect of having a sharded foreign collection is worse than the benefits of parallel execution, at least in the case of these queries.

### DIFF
--- a/src/workloads/execution/ShardedLookup.yml
+++ b/src/workloads/execution/ShardedLookup.yml
@@ -142,6 +142,40 @@ Actors:
             }}
           ]
         cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: LookupWithCachedPrefixShardedToSharded
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline:
+          [{
+            $lookup: {
+              from: Collection1,
+              let: {localInt: "$int"},
+              pipeline: [
+                {$group: {_id: {$sum: ["$int", 1]}}},
+                {$match: {$expr: {$eq: ["$_id", "$$localInt"]}}}
+              ],
+              as: matches
+            }
+          }]
+        cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: LookupWithCachedPrefixUnshardedToSharded
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection2
+        pipeline:
+          [{
+            $lookup: {
+              from: Collection1,
+              let: {localInt: "$int"},
+              pipeline: [
+                {$group: {_id: {$sum: ["$int", 1]}}},
+                {$match: {$expr: {$eq: ["$_id", "$$localInt"]}}}
+              ],
+              as: matches
+            }
+          }]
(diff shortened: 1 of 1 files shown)


src/workloads/execution/ShardedLookup.yml
```diff
@@ -142,6 +142,40 @@ Actors:
...
+        aggregate: Collection0
+        pipeline:
+          [{
+            $lookup: {
+              from: Collection1,
+              let: {localInt: "$int"},
+              pipeline: [
+                {$group: {_id: {$sum: ["$int", 1]}}},
+                {$match: {$expr: {$eq: ["$_id", "$$localInt"]}}}
+              ],
```
kaywux, line 157: Hmm I don't think this pipeline will be able to use the cache even after [SERVER-57168](https://jira.mongodb.org/browse/SERVER-57168), since the cache stage would be abandoned if sent to the shards: https://github.com/mongodb/mongo/blob/82f713b946460738ad15469c083f173ae0d1f80b/src/mongo/db/pipeline/sharded_agg_helpers.cpp#L628.

Could you add a sharded to sharded $lookup query that can cache? The pipeline will need to have a stage that forces a pipeline split (like a $group or $limit) before the $lookup. This would have the $lookup execute on a merger rather than in parallel and would (I believe) have similar performance to a similar unsharded to sharded scenario.

kaywux (commented):
Thanks Hana! I have one major comment, which is that (even after SERVER-57168 is resolved) the sharded to sharded $lookup query will not be able to use the cache. These results are really interesting when compared to your other PR! Scatter-gather queries are definitely not the ideal use-case for sharded $lookup, hopefully targeted queries would fare better.

src/workloads/execution/ShardedLookup.yml
kaywux: Edit: Brain outage moment! All of the above doesn't apply since the cache stage is added to the foreign pipeline, _not_ the original pipeline. Thanks for bringing that up Hana!

kaywux (approved):
LGTM, push away!

HanaPearlman:
evergreen merge

HanaPearlman:
evergreen merge

PR #514, opened 2021-08-09
STAR-2112 Expose curator cli command for running t2 translation in run-genny

edzh:


aggrand (approved):
You'll want to update the curator commit here: https://github.com/mongodb/genny/blob/e1abdddb149abbf97b72882531937582be112430/src/lamplib/src/genny/curator.py#L153

Otherwise it LGTM assuming it works with the updated commit!

PR #518, opened 2021-08-17
PERF-2501 Add a genny workload that runs large batch write commands with operations that fail with WouldChangeOwningShard errors

jz1242:


src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -0,0 +1,91 @@
...
+  - *Nop
+  - Repeat: 1
+    SleepBefore: *BalancerWait
+    BatchSize: 1000
+    DocumentCount: *DocumentCount
+    Database: *Database
+    Document: {x: 0}
+  - *Nop
+
+- Name: UpdateShardKey
```
jz1242, line 63: From https://github.com/mongodb/mongo/blob/master/src/mongo/s/commands/cluster_write_cmd.cpp#L152-L159 If we send a retryable write command and it fails with WouldChangeOwningShard, we implicitly retry in a transaction.

cheahuychou: Does genny use sessions by default? According to this, it looks like you need to specify `OnSession: true`. `retryableWrite` defaults to true so I think the operation will run as a retryable write by default after you specify `OnSession: true`.

https://github.com/mongodb/genny/blob/446f8ebdc10fb1d8d7d0a940887ef1b7a4203f71/src/workloads/docs/CrudActorTransaction.yml#L38-L40 

Also, could you add a phrase that runs the update inside a transaction? I think you can use `withTransaction` to do that. Here is an example: https://github.com/mongodb/genny/blob/446f8ebdc10fb1d8d7d0a940887ef1b7a4203f71/src/workloads/docs/CrudActorTransaction.yml#L12-L40

According to the source code for `WithTransactionOperation`, it corresponds to `with_transaction` in the driver API http://mongocxx.org/api/current/classmongocxx_1_1client__session.html#a2ca899e8c0018af269f2542b3538587f.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -0,0 +1,91 @@
...
+  - *Nop
+  - *Nop
+
+- Name: LoadInitialData
+  Type: MonotonicSingleLoader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    SleepBefore: *BalancerWait
```
cheahuychou, line 56: I opened the DSI artifacts for the EVG task you sent me and could not find any log lines for chunk migration (I did find log lines for shardCollection and splitChunk). So I was not sure if the balancer was on. I think you might need to explicitly enable it by running a balancerStart command.  https://github.com/mongodb/mongo/blob/66708080ab5db302558f7bd33293167d181f3858/src/mongo/shell/utils_sh.js#L192-L193

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -0,0 +1,91 @@
...
+  Threads: 1 # We want to use 1 thread to avoid updates throwing WriteConflict errors.
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: *DocumentCount
+    Database: *Database
+    Collection: *Collection
+    Operations:
+    - OperationMetricsName: UpdateShardKey
+      OperationName: bulkWrite
```
cheahuychou, line 74: I think we should specify `OperationName: updateOne` to be more explicit? https://github.com/mongodb/genny/blob/e8cc83e21b32a2713201207f8d8fe7624e0dd005/src/cast_core/src/CrudActor.cpp#L436

cheahuychou (changes requested):

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -0,0 +1,91 @@
...
+
+GlobalDefaults:
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+  # Collection0 is the default collection populated by the MonotonicSingleLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentCount: &DocumentCount 1000  # Number of documents to insert and modify.
```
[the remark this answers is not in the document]
jz1242: I can change it to 10000 or 100000

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -0,0 +1,93 @@
...
+Owner: "@mongodb/sharding"
+Description: |
+  Creates a sharded collection with 2 chunks on 2 different shards using ranged sharding and updates
+  the shard key to trigger WouldChangeOwningShard errors.
+
+  The workload consists of 3 phases:
+    1. Shard an empty collection (using ranged sharding) into two chunks and wait for the balancer 
+    to distribute each chunk to its own shard.
+    2. Populate the sharded collection with data.
+    3. Update the shard key value to trigger WouldChangeOwningShard errors.
```
aggrand, line 11: So the goal of this workload is to trigger these errors? If so you might want to configure the throw settings for CrudActor like [here](https://github.com/mongodb/genny/blob/5fdb13cc6d9b6153fd2567e9d536f509289a5725/src/workloads/docs/CrudActor.yml#L39). Otherwise the workload may crash out.

jz1242: The error itself is implicitly retried on so the client doesn't actually see that we throw WouldChangeOwningShard.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
jz1242: Done.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
jz1242: I added the balancerStart command just as a failsafe. I think that sharded clusters automatically start with the balancer enabled.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
jz1242: As discussed offline, upon getting WouldChangeOwningShard for a retryable write, we automatically retry it in a transaction. So adding a new phase for running it in a transaction is a little redundant. I think retryableWrites are enabled by default in drivers: https://docs.mongodb.com/v4.0/core/retryable-writes/ .

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -53,7 +53,7 @@ Actors:
 
 - Name: LoadInitialData
   Type: MonotonicSingleLoader
-  Threads: 50
+  Threads: 1
```
jz1242, line 56: I think having just a single thread inserting 100000 documents is probably enough.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
aggrand: Ah gotcha, thanks for clarifying!

aggrand (approved):
Regarding the genny syntax sanity-check I can provide, this LGTM!

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
cheahuychou: It looks like resharding workloads use 100 threads to insert 100000 documents. https://github.com/mongodb/genny/blob/bfaff11cabfa9e9f362458f597943259e754f1f1/src/workloads/sharding/ReshardCollection.yml#L87 So I think if we want to go with 100000 documents, we should not use just one thread. Alternatively, we can decrease the number of documents to 10000.

cheahuychou (approved):
LGTM mod one comment. Also, since this workload only runs batch write commands as retryable writes, I think we should rephrase the ticket title (i.e. commit message) to not say "in all session option configurations". Could you also add a comment in the JIRA ticket to explain the reason? Thank you.

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
jz1242: From my understanding using 100 threads with a doc count of 100000 inserts 100000 * 100 docs. I reduced the number of documents to 10000 just so the test doesnt take as long to run and should still give meaningful perf data.

jz1242:
evergreen merge

PR #524, opened 2021-08-20
PERF-2429 Add Stress Tests to Production (mixed stress and Load Test)

dalyd:
Patch build: https://evergreen.mongodb.com/build/sys_perf_linux_3_node_replSet_patch_01fb3fac6dbf5a6566e0ce6cf5218d4274df19f1_611d2d8b1e2d17098199f36a_21_08_18_15_56_52

The patch build was successful for MixedWorkloadsGennyStress and LoadTest, so I've included those two here. I need to figure out how to get BigUpdate10k to not timeout during validate. I'll split that problem off of this work. 

Note, this PR effectively re-enables PERF-2330 & PERF-2255 miscellaneous changes (#500) (#523) change that was previously reverted. It should also have a working auto run for it.

### DIFF
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -16,7 +16,6 @@
 
 #include <algorithm>
 #include <memory>
-#include <locale>
 
 #include <bsoncxx/json.hpp>
 
@@ -118,9 +117,9 @@ void genny::actor::Loader::run() {
                 auto collectionName = "Collection" + std::to_string(i);
                 auto collection = config->database[collectionName];
                 // Insert the documents
-                BOOST_LOG_TRIVIAL(info) << "Starting to insert: "
-                                        << config->numDocuments << " docs "
-                                        << "into " << collectionName;
+                BOOST_LOG_TRIVIAL(info)
+                    << "Starting to insert: " << config->numDocuments << " docs "
+                    << "into " << collectionName;
                 uint remainingInserts = config->numDocuments;
                 {
                     auto totalOpCtx = _totalBulkLoad.start();
@@ -162,8 +161,8 @@ void genny::actor::Loader::run() {
                         indexOpCtx.success();
                     }
                 }
-                BOOST_LOG_TRIVIAL(info) << "Done with load phase. All " <<
-                  config->numDocuments << " documents loaded into " << collectionName;
+                BOOST_LOG_TRIVIAL(info) << "Done with load phase. All " << config->numDocuments
+                                        << " documents loaded into " << collectionName;
             }
         }
     }
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <locale>
 
 #include <bsoncxx/json.hpp>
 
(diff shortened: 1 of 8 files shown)


src/cast_core/src/Loader.cpp
```diff
@@ -117,6 +118,9 @@ void genny::actor::Loader::run() {
                 auto collectionName = "Collection" + std::to_string(i);
                 auto collection = config->database[collectionName];
                 // Insert the documents
+                BOOST_LOG_TRIVIAL(info) << "Starting to insert: "
```
dalyd, line 121: This was from Ger's work earlier. It doesn't seem unreasonable to have the start and end logged, but I can pull it if it's distracting.

src/workloads/scale/BigUpdate10k.yml
```diff
@@ -0,0 +1,137 @@
...
+  Type: LoggingActor
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0,1]
+      NopInPhasesUpTo: 1
+      PhaseConfig:
+        LogEvery: 10 seconds
+        Blocking: None
+# AutoRun:
```
dalyd, line 132: Autorun commented out for now while I figure out the db validation timeout.

src/workloads/scale/LoadTest.yml
```diff
@@ -0,0 +1,121 @@
...
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1, 3, 5]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        LogEvery: 10 seconds
+        Blocking: None
+
+AutoRun:
```
dalyd, line 116: This should be functional autorun, unlike the original.

src/cast_core/src/Loader.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <locale>
```
brad-devlugt, line 19: Do we need this extra include?

src/workloads/scale/MixedWorkloadsGennyStress.yml
```diff
@@ -0,0 +1,220 @@
...
+      PhaseConfig:
+        LogEvery: 10 seconds
+        Blocking: None
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - atlas
+          - replica
```
brad-devlugt, line 220: What are your thoughts on adding the flow-control disabled variant here?

src/cast_core/src/Loader.cpp
brad-devlugt: Sounds good, it sounds like a reasonable log to have

brad-devlugt (approved):

src/cast_core/src/Loader.cpp
dalyd: This was from Ger's change. Let me try removing it and see if it builds.

src/workloads/scale/MixedWorkloadsGennyStress.yml
dalyd: I'm thinking. If you felt strongly I could definitely be talked into it. I wonder if it would flag because of the lag? Although we may still have that disabled.

brad-devlugt: I think it's disabled https://github.com/10gen/dsi/blob/master/configurations/analysis/analysis.common.yml#L25

It could be interesting to have a test where lag grows indefinitely - I wonder what happens with the longer 45min runtime.  If there's too many unknowns though I think it'd be fine as a separate ticket

src/cast_core/src/Loader.cpp
dalyd: Nope -- not needed.

aggrand (approved):
LGTM! Thanks for fixing the autorun.

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #526, opened 2021-08-23
PERF-2475: Add tests for $graphLookup to explore trade off of paralle…

HanaPearlman:
…lization and shared cache

Timing results here: https://spruce.mongodb.com/version/6123a66ba4cf4718ca4703a6/tasks

Feature flag off:
|---|---|
| RunGraphLookups.GraphLookupUnshardedToUnsharded | 6.5536703637 |
| RunGraphLookups.GraphLookupShardedToUnsharded | 6.4973098341 |

Feature flag on:
|---|---|
| RunGraphLookups.GraphLookupUnshardedToUnsharded | 6.8557203083 |
| RunGraphLookups.GraphLookupShardedToUnsharded | 4.4047303946 |

When the feature flag is turned on, we see a slight slowdown in the unsharded case (<5%) possibly because we end up targetting shards instead of reading locally. Also, we see a speed up in the sharded local case (~32%) even though the cache is not shared between the nodes, since the $graphLookup execution can be parallelized.

### DIFF
--- a/src/workloads/execution/UnshardedGraphLookup.yml
+++ b/src/workloads/execution/UnshardedGraphLookup.yml
@@ -1,0 +1,123 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of $graphLookup against an unsharded foreign collection.
+  The workload consists of the following phases:
+    1. Creating an empty sharded collection distributed across all shards in the cluster.
+    2. Populating collections with data.
+    3. Fsync.
+    4. Running $graphLookup's.
+
+Actors:
+- Name: CreateShardedCollections
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    Operations:
+    # Shard Collection0 using hashed sharding to ensure even distribution of chunks across shards.
+    - OperationMetricsName: EnableSharding
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: &Database test
+    - OperationMetricsName: ShardLocalCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: test.Collection0
+        key: {c: hashed}
+        numInitialChunks: &NumChunks 6
+    # Disable the balancer so that it can't skew results while the $lookups are running.
+    - OperationMetricsName: DisableBalancer
+      OperationName: AdminCommand
+      OperationCommand:
+        balancerStop: 1
+  - &Nop {Nop: true}
+  - *Nop
+  - *Nop
+
+- Name: LoadGraphLookupData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    BatchSize: 1000
+    Threads: 1
+    DocumentCount: &NumDocs 3000
+    Database: *Database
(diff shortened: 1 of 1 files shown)


src/workloads/execution/UnshardedGraphLookup.yml
```diff
@@ -0,0 +1,123 @@
...
+              from: "Collection1",
+              startWith: "$a",
+              connectFromField: "a",
+              connectToField: "b",
+              as: "matches",
+              maxDepth: 1
+            }},
+            {$project: {s: 0, "matches.s": 0}}
+          ]
+        cursor: {batchSize: *NumDocs}
```
HanaPearlman, line 98: I tried to come up with a query that would rely on the cache without filling it up too quickly, to avoid the issue I believe to be affecting Militsa's last PR #506 (comment thread [here](https://github.com/mongodb/genny/pull/506#pullrequestreview-725301223)) where the cache is immediately filled with the relevant documents. For this query, each document from the local collection will match 2 documents in the foreign collection in expectation, so there will be about 6000 queries for specific values of `b` during the $graphLookup stage. My very loose thinking is that each value of `b` (between 1 and 3000) will be queried ~2 times. So, if there's a single shared cache, it's more likely that we can take advantage of this, whereas if there are two or more caches due to parallel execution, we will likely not be able to rely on the cache as much. 

I would definitely welcome thoughts on this. Is this example too convoluted? Is there an obvious alternative? Ultimately to me it feels like the benefit of the parallel execution should outweigh the downside of not sharing a single cache almost always -- every query that would have been satisfied by looking up in a single shared cache could just be run in parallel on all relevant nodes, and we would likely end up with similar timing results.

A note on size: each collection is around 9MB, but the $project at the end of the pipeline and the low maxDepth ensures the result set fits in a single batch. After the project, the documents are around 50 bytes in size, and the result set ends up being less than 1 MB.

kaywux: Thanks for this comment! Just to make sure I understand what's happening: for a document like `{a: 1, b: 2, ...}`, the 0th $graphLookup iteration will query for a match to `b: 1`, which let's say returns doc `{a: 5, b: 1, ...}` and then at depth 1 queries again for `b: 5`. Am I understanding that right?

I think you're right, it seems like it would be hard (and more than slightly contrived) to construct a case where the shared cache benefits outweigh the parallelism - for the case you have here we already have a low level of parallelism (with shard-lite being only 2 shards) but could possibly do something with the documents and the way we shard them to ensure that each of the two queries for a specific `b` comes from a different shard. With the way genny workloads create documents with randomness I'm not sure if there's currently a good way to control for this.

kaywux (commented):
Thanks Hana, these results are enlightening - seems that in nearly all cases the parallelization would be more beneficial. I have one thought about a case that could show more shared cache benefits, but would probably be hard to construct, in which case it might not be worth adding - let me know your thoughts!

kaywux (approved):
As discussed online, there are a few approaches we could take to try and measure the edge case discussed above, but all would involve cpp changes to the Loader and necessitate a maxDepth limit to get "meaningful" results. Since this case is convoluted and something that we doubt a user would be running with $graphLookup, I'm good with the perf cases you have here. Thanks Hana!

HanaPearlman:
evergreen merge

PR #530, opened 2021-09-01
PERF-2531 Add docker support

jimoleary:
This is useful for running multiple genny clients on a single  workload client in 'local' DSI development mode.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+build
+# Except this file
+!build/.gitignore
+dist
+genny_venv
+docker
+mci
+.git
+.github
+.idea
+.tmp
+containers
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ genny.egg-info
 
 # tools
 curator
+
+# docker
+containers
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,1 @@
+Dockerfile.amazon2
--- a/docker/Dockerfile.1804
+++ b/docker/Dockerfile.1804
@@ -1,0 +1,66 @@
+#
+# Create an amazon2 docker container.
+#   $> docker build -t  genny/ubuntu:1804 ./ -f docker/Dockerfile.18004 --no-cache --rm=true
+#   $> docker run -it genny/amazon:2
+#     OR with ssl support
+#   $> docker run  -v /etc/pki:/etc/pki -it genny/ubuntu:1804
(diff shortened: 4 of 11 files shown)


aggrand:
Just a note, but if having multiple curator poplar instances is the blocker to multiple genny runs as described in the ticket, we could also instead tweak the python launcher to just not launch poplar if there's already one running. (And also tweak the output paths so they don't conflict.) Multiple genny processes should be fine using the same poplar service.

dalyd (commented):
@jimoleary I don't feel competent to review this one. I don't know any of the docker stuff.

rtimmons:
My docker-fu is weak, so I don't feel qualified to review in that regard. I poked and didn't see anything obvious though FWIW.

But..uhm..actually...what's this enabling?

Couple questions:

1. What does docker actually get us here? The toolchain already (mostly) isolates the build & runtime system. If curator/genny isolation is a thing, we could scheme on paths like Raiden suggested.
2. What does multiple genny instances give us? Splitting up N threads over M processes seems like more overhead than N threads on a single process?
3. How do you use the data that is produced? The UX for Phases and timeseries in general seems like it'd be confusing if you have results from multiple instances to analyze in parallel.

> I am running these processes manually, I suspect there would be a reasonable about of effort to be able to do this in evergreen and no definite need or benefit at the moment.

Correct. Making this "real" seems like a fair amount of toil, and I think most useful would be to support multiple genny machines (workload clients) which would likely obviate docker.

So...my fear is that adding this code sets the precedent that it's something we explicitly support or are working toward supporting.

That said, I don't want to reject useful things. If this is useful, let's add it and indicate that it's not "official".

There is already `src/workloads/contrib` that may seem right to you. With some slight "😱 EXPERIMENTAL 😱" warnings in the README plus some scenarios where it should/shouldn't be used would let others benefit from this code without the added maintenance/design burden of ensuring it fits with what's here today and what's planned in the future.

Please push back where you disagree.

jimoleary:
Hi @rtimmons, this was from the early phase of stress testing where I was hitting issues with limits, etc. 

I think that there is a use case of multiple genny instances at some stage but orchestrating the instances to run in step and handling the output data would be a  project in themselves. Raidans' suggested approach may be the right way forward, docker was just a simple way for me to avoid issues with multiple curator processes.

I think I am happy to close this and the ticket out as won't do. The PR will be available and the ticket can be found in JIRA if anyone ever needs to review it in future.

PR #531, opened 2021-09-01
PERF-2532 Add PerInstance Client flag to create a pool per actor inst…

jimoleary:
…ance

### DIFF
--- a/src/cast_core/src/CollectionScanner.cpp
+++ b/src/cast_core/src/CollectionScanner.cpp
@@ -516,7 +516,7 @@ void CollectionScanner::run() {
 CollectionScanner::CollectionScanner(genny::ActorContext& context)
     : Actor{context},
       _totalInserts{context.operation("Insert", CollectionScanner::id())},
-      _client{context.client()},
+      _client{context.client(v1::DEFAULT_CLIENT_NAME, CollectionScanner::id() )},
       _index{WorkloadContext::getActorSharedState<CollectionScanner, ActorCounter>().fetch_add(1)},
       _runningActorCounter{
           WorkloadContext::getActorSharedState<CollectionScanner, RunningActorCounter>()},
--- a/src/gennylib/src/PoolManager.cpp
+++ b/src/gennylib/src/PoolManager.cpp
@@ -60,7 +60,11 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
     std::unique_lock<std::mutex> lock{lap.first};
 
     Pools& pools = lap.second;
-
+    auto perInstance =
+        context["Clients"][name]["PerInstance"].maybe<bool>().value_or(false);
+    if (!perInstance) {
+        instance = 0;
+    }
     auto& pool = pools[instance];
     if (pool == nullptr) {
         pool = createPool(this->_mongoUri, name, this->_apmCallback, context);
--- a/src/cast_core/src/CommitLatency.cpp
+++ b/src/cast_core/src/CommitLatency.cpp
@@ -178,7 +178,7 @@ void CommitLatency::run() {
 CommitLatency::CommitLatency(genny::ActorContext& context)
     : Actor(context),
       _rng{context.workload().getRNGForThread(CommitLatency::id())},
-      _client{std::move(context.client())},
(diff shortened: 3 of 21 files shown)


src/cast_core/src/CollectionScanner.cpp
```diff
@@ -516,7 +516,7 @@ void CollectionScanner::run() {
 CollectionScanner::CollectionScanner(genny::ActorContext& context)
     : Actor{context},
       _totalInserts{context.operation("Insert", CollectionScanner::id())},
-      _client{context.client()},
+      _client{context.client(v1::DEFAULT_CLIENT_NAME, CollectionScanner::id() )},
```
dalyd, line 519: At the risk of introducing chaos, should we consider switching the order of the arguments? Then we could use the default value for the client name in all these files.

src/gennylib/src/PoolManager.cpp
```diff
@@ -60,7 +60,11 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
     std::unique_lock<std::mutex> lock{lap.first};
 
     Pools& pools = lap.second;
-
+    auto perInstance =
```
dalyd, line 63: Is there something missing here? It doesn't look like you use perInstance anywhere.

src/gennylib/src/PoolManager.cpp
```diff
@@ -60,7 +60,11 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
...
+        context["Clients"][name]["PerInstance"].maybe<bool>().value_or(false);
+    if (!perInstance) {
+        instance = 0;
```
dalyd, line 66: Isn't this changing the existing behavior as it will overwrite the instance value passed into the function? Did you maybe mean the opposite?

dalyd (changes requested):

src/cast_core/src/CollectionScanner.cpp
jimoleary: I did consider that but decided to put it the change out for review first. It might make the scale of the changes marginally smaller.

There is also another option to consider, should the configuration be on the Client or the Actor? The scale of changes for an Actor configuration would be about the same (and I don't think there would be any changes required to PoolManager ). However, I *think* the configuration being on the client pool itself makes more sense and is less verbose in the yml files.

src/gennylib/src/PoolManager.cpp
jimoleary: I assume you mean the yml config (as the variable is used on the next line).

The configuration is demonstrated [here](https://github.com/mongodb/genny/pull/531/files#diff-dc702b545e53500c48ea324dd47ba077c92d81a33320be2bebdbdeb8cbc2a394R48).

src/gennylib/src/PoolManager.cpp
jimoleary: No, I may have named it badly but if _perInstance_ is false (the default) then _instance_ will always be 0 so the pool will be shared. If _perinstance_ is true then the _instance_ value will be unqiue and the pool won't be shared.

Maybe a client property called _Shared_  with a default value of true would be clearer?

Also the current code base only ever uses a zero value for this parameter.

dalyd: Ah, the current code base only every using zero was what I was missing.

dalyd: Maybe a comment would help.

src/cast_core/src/CollectionScanner.cpp
dalyd: Alright, I'm good. I'll leave this thread to sit for a bit in case Ryan has anything to add.

src/gennylib/src/PoolManager.cpp
dalyd: What I was missing was that support for multiple instances has existed for 3 years in the code base and never used  :) I just spent some time in git blame and it makes more sense to me now.

dalyd (approved):

dalyd (approved):

jimoleary:
It's not worth going forward with this PR. The change is too niche to be reasonable (beyond some extreme stress testing that I am doing).

PR #532, opened 2021-09-03
PERF-2523: Add tests for $graphLookup into unsharded collection in sh…

HanaPearlman:
…arded cluster

Results can be found [here](https://spruce.mongodb.com/version/61313b3656234307b166d3b0/tasks).

Feature flag off
|---|---|
| RunGraphLookups.GraphLookupUnshardedToUnshardedOneToMany | 1.6552927594 |
| RunGraphLookups.GraphLookupShardedToUnshardedOneToMany  | 1.6444522419 |
| RunGraphLookups.GraphLookupShardedToUnshardedMatchSomeShards | .1211813272 |
| RunGraphLookups.GraphLookupShardedToUnshardedMatchAllShards | .0906489153 |

Feature flag on
|---|---|
| RunGraphLookups.GraphLookupUnshardedToUnshardedOneToMany | 1.8329503544 |
| RunGraphLookups.GraphLookupShardedToUnshardedOneToMany | 1.3836131205 |
| RunGraphLookups.GraphLookupShardedToUnshardedMatchSomeShards | .1091266271 |
| RunGraphLookups.GraphLookupShardedToUnshardedMatchAllShards | .0727160483 |

In every case except UnshardedToUnsharded, turning the feature flag on speeds up the queries because of parallelization. Good news!

Another thing to note is that the MatchAllShards case is faster than MatchSomeShards with the feature flag on because we can parallelize across more shards even though we need to get data from every shard. I'm not sure why we see the same thing when the feature flag is off, since I would expect the opposite (MatchSomeShards could be faster because it can use the existing index). I have verified that the execution is as expected locally, so I'm unsure here.

### DIFF
--- a/src/workloads/execution/UnshardedGraphLookup.yml
+++ b/src/workloads/execution/UnshardedGraphLookup.yml
@@ -27,7 +27,7 @@ Actors:
         shardCollection: test.Collection0
         key: {c: hashed}
         numInitialChunks: &NumChunks 6
-    # Disable the balancer so that it can't skew results while the $lookups are running.
+    # Disable the balancer so that it can't skew results while the $graphLookups are running.
     - OperationMetricsName: DisableBalancer
       OperationName: AdminCommand
       OperationCommand:
@@ -42,7 +42,7 @@ Actors:
   Phases:
   - *Nop
   - Repeat: 1
-    BatchSize: 1000
+    BatchSize: 3000
     Threads: 1
     DocumentCount: &NumDocs 3000
     Database: *Database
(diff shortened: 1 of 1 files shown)


kaywux (approved):
LGTM, glad to see that we won't need to special-case unsharded collections for $graphLookup! I'm similarly surprised about why the MatchSome case is slower than MatchAll even with the flag turned off - I definitely would have expected the first predicate to have taken advantage of the shard key / index on `c`. I did a quick test in the mongo shell and confirmed that the predicate works as expected on a hashed index on a standalone, so maybe it's not targeting shards efficiently? Either way - this doesn't seem like a $graphLookup issue so shouldn't hold this ticket up.

HanaPearlman:
evergreen merge

PR #534, opened 2021-09-09
PERF-2078 Example workload for proposed State Machine Functionality for Crud Actor.

dalyd:
This is just the example document. When we are happy with it I will start on implementing the beast. Please pay particular attention to the syntax and ergonomics. If there are small things we can do to make it more similar to other FSM we use, let me know.

### DIFF
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -43,8 +43,6 @@ Actors:
       OperationCommand:
         WriteOperations:
         - WriteCommand: insertOne
-          Document: {phoneID: 0, state: "Off"}
-        - WriteCommand: insertOne
           Document: {phoneID: 1, state: "Off"}
         - WriteCommand: insertOne
           Document: {phoneID: 2, state: "Off"}
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 
-description: |
+Description: |
   TODO: This functionality does not exist yet. It will be implemented in PERF-2516.  The Crud Actor
   will support FSM definition, with operations associated with each states. Choices are declared
   with weights that are converted into probabilities.
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -2,29 +2,27 @@ SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 
 description: |
-  TODO: This Actor does not exist yet. It will be implemented in PERF-2516.
-  The State Machine Crud Actor implements a FSM with crud operations. Choices are declared with
-  weights, which are converted into probabilities.
+  TODO: This functionality does not exist yet. It will be implemented in PERF-2516.  The Curd Actor
+  will support FSM definition, with operations associated with each states. Choices are declared
+  with weights that are converted into probabilities.
 
   There is an initial state vector, with weight. Each actor instance will probabilistically and
   independently choose it's initial state based on the weights.
 
   Each state specifies a list of transitions. The next transition is picked probabilistically, based
-  on the transition weights. Each transition specifies a delay, thens executes a (possibly empty) set
-  of operations, before transitioning to the next state. The state itself may execute its own set of
-  operations.
+  on the transition weights. Transitions may experience delay (sleepBefore), before changing the
+  state. The state has associated operations which execute while the FSM remains in that
+  state. (Question: what happens during the transition delay? I suggest nothing)
 
-  By default the system will measure the latency to execute each set of operations (per transition
-  or per state), and each operation. The latencies associated with each transition and state are
-  tracked separately.
+  By default the system will measure the latency to execute each set of operations associated with a
+  state, and each individual operation.
 
   This example workload is a toy example modeling the state of 5 smart phones. The model has four
   states for each phone: On, Off, Sleep, and Error, and transitions between those states. In this
   example, there is monitoring for the smart phones, and the database gets updated based on the
-  state of the phones and state changes. The workload (monitoring app) tracks the current phone state in
-  one document (update operations). Note that this is separate than the state machinery tracking of
-  state. The workload also counts the number of transitions from state On to each of
-  the other three states (also update operations).
+  state of the phones and state changes. The workload (monitoring app) tracks the current phone
+  state in one document per phone (update operations). Note that this is separate than the state
+  machinery tracking of state.
 
 Clients:
   Default:
@@ -54,12 +52,6 @@ Actors:
           Document: {phoneID: 3, state: "Off"}
         - WriteCommand: insertOne
           Document: {phoneID: 4, state: "Off"}
-        - WriteCommand: insertOne
-          Document: {stat: "OnOff", count: 0}
-        - WriteCommand: insertOne
-          Document: {stat: "OnSleep", count: 0}
-        - WriteCommand: insertOne
-          Document: {stat: "OnError", count: 0}
     - &nop {Nop: true}
     - *nop
 
@@ -74,9 +66,9 @@ Actors:
     States:
     - Name: "On"  # The device is on
 
-      # The states support the keywords Repeat, SleepBefore, SleepAfter, and Duration. These
-      # keywords behave the same as in Phase Control, but operate per state (i.e., per actor
-      # instance as it goes through the state machine.
+      # The states support the keywords Repeat, SleepBefore, SleepAfter, (local) Rate, and
+      # Duration. These keywords behave the same as in Phase Control, but operate per state (i.e.,
+      # per actor instance as it goes through the state machine.
       Repeat: 1
       Operations:
       # A state can have (optional) a set of operations that execute upon entering the state. This
@@ -86,6 +78,10 @@ Actors:
       # always non-blocking. When the phase ends, the state machine will finish any open operations,
       # and then exit. It if is waiting for a delay, it will exit immediately. This is required as
       # there may be states with long delays.
+      # Question: Should we support having a single fire operation with repeating operations? Here I
+      # could invent a case where the following operation is repeat 1, but there is another
+      # operation that repeats once a second while in this state. I don't know how to represent
+      # that.  
       - OperationName: updateOne
         OperationCommand:
           Filter: &filter {phoneID: {^ActorID: {}}}
@@ -107,35 +103,6 @@ Actors:
         # accepts any valid int or double generator or constant for value.
         SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
 
-        # TODO: REmove the transition operations
-        # In this advanced case, operations may also be associated with transitions.
-        Operations:
-        - OperationName: updateOne
-          OperationCommand:
-            Filter: {stat: "OnOff"}
-            Update:
-              $inc: {count: 1}
-      - To: "Sleep"
-        Weight: 2
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
-        Operations:
-        - OperationName: updateOne
-          OperationCommand:
-            Filter: {stat: "OnSleep"}
-            Update:
-              $inc: {count: 1}
-      - To: "Error"
-        Weight: 0.2
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
-        Operations:
-        - OperationName: updateOne
-          OperationCommand:
-            Filter: {stat: "OnError"}
-            Update:
-              $inc: {count: 1}
-
     - Name: "Off"  # The device is Off
       Operations:
       - OperationName: updateOne
@@ -146,8 +113,7 @@ Actors:
       Transitions:  # Can only go to On from Off
       - To: "On"
         Weight: 1
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
 
     - Name: Sleep
       Operations:
@@ -159,12 +125,11 @@ Actors:
       Transitions:  # Can only go to On or Off from sleep
       - To: "On"
         Weight: 1
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
       - To: "Off"
         Weight: 1
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+
     - Name: Error
       Operations:
       - OperationName: updateOne
@@ -175,8 +140,7 @@ Actors:
       Transitions:  # Turn off on error
       - To: "Off"
         Weight: 1
-        Delay:
-          Delay: {^RandomDouble: {min: 0, max: 15}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
 
     # In this example we always start in the off State
     InitialStates:
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -46,15 +46,15 @@ Actors:
       OperationCommand:
         WriteOperations:
         - WriteCommand: insertOne
-          Document: {phoneID: 1, currentPhoneState: "PhoneIsOff"}
+          Document: {phoneID: 1, currentPhoneState: PhoneIsOff}
         - WriteCommand: insertOne
-          Document: {phoneID: 2, currentPhoneState: "PhoneIsOff"}
+          Document: {phoneID: 2, currentPhoneState: PhoneIsOff}
         - WriteCommand: insertOne
-          Document: {phoneID: 3, currentPhoneState: "PhoneIsOff"}
+          Document: {phoneID: 3, currentPhoneState: PhoneIsOff}
         - WriteCommand: insertOne
-          Document: {phoneID: 4, currentPhoneState: "PhoneIsOff"}
+          Document: {phoneID: 4, currentPhoneState: PhoneIsOff}
         - WriteCommand: insertOne
-          Document: {phoneID: 5, currentPhoneState: "PhoneIsOff"}
+          Document: {phoneID: 5, currentPhoneState: PhoneIsOff}
     - &nop {Nop: true}
     - *nop
 
@@ -90,7 +90,7 @@ Actors:
         OperationCommand:
           Filter: &filter {phoneID: {^ActorId: {}}}
           Update:
-            $set: {currentPhoneState: "PhoneIsOn"}
+            $set: {currentPhoneState: PhoneIsOn}
 
       # Transitions specifies next states, delays, and operations. The On state can transition to
       # each of the three other states
@@ -102,15 +102,15 @@ Actors:
         # 0.
         Weight: 1
 
-        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.  For
-        # sleepBefore, the system is considered to remain in it's current state. If there is a
-        # repeating action, it will continue repeating until the transition fires.  For repeat, the
-        # operation is repeated immediately.  sleepAfter will not be supported for transitions with
-        # no operations.
+        # Transitions also support the phase keyword sleepBefore.  For sleepBefore, the system is
+        # considered to remain in it's current state. If there is a repeating action in the state,
+        # it will continue repeating until the transition fires. The transition always respects the
+        # blocking behavior of the phase. If the phase ends during a sleepBefore, the transition is
+        # aborted.
 
         # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
         # accepts any valid int or double generator or constant for value.
-        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: seconds}}
 
     - Name: "Off"  # The device is Off
       Operations:
@@ -118,11 +118,11 @@ Actors:
         OperationCommand:
           Filter: *filter
           Update:
-            $set: {currentPhoneState: "PhoneIsOff"}
+            $set: {currentPhoneState: PhoneIsOff}
       Transitions:  # Can only go to On from Off
       - To: "On"
         Weight: 1
-        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: seconds}}
 
     - Name: Sleep
       Operations:
@@ -130,14 +130,14 @@ Actors:
         OperationCommand:
           Filter: *filter
           Update:
-            $set: {currentPhoneState: "PhoneIsSleeping"}
+            $set: {currentPhoneState: PhoneIsSleeping}
       Transitions:  # Can only go to On or Off from sleep
       - To: "On"
         Weight: 1
-        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: seconds}}
       - To: "Off"
         Weight: 1
-        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: seconds}}
 
     - Name: Error
       Operations:
@@ -145,11 +145,11 @@ Actors:
         OperationCommand:
           Filter: *filter
           Update:
-            $set: {currentPhoneState: "PhoneIsInError"}
+            $set: {currentPhoneState: PhoneIsInError}
       Transitions:  # Turn off on error
       - To: "Off"
         Weight: 1
-        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: seconds}}
 
     # In this example we always start in the off State
     InitialStates:
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -6,6 +6,9 @@ description: |
   will support FSM definition, with operations associated with each states. Choices are declared
   with weights that are converted into probabilities.
 
+  TODO: Remove this file from skipped dry run list when implemented
+  (src/lamplib/src/genny/tasks/dry_run.py)
+
   There is an initial state vector, with weight. Each actor instance will probabilistically and
   independently choose it's initial state based on the weights.
 
--- a/src/workloads/docs/CrudActorFSM.yml
+++ b/src/workloads/docs/CrudActorFSM.yml
@@ -80,7 +80,7 @@ Actors:
 
       # For example, this phase has a 30 second duration, making it blocking. When that 30 seconds
       # end all the actors will end their phases. If an actor is in a sleepBefore or sleepAfter
-      # delay, it will exit the phase. If the actor is exucting a state's operations, it will
+      # delay, it will exit the phase. If the actor is executing a state's operations, it will
       # complete those operations, and then exit the phase.
       Operations:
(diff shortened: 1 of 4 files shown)


dalyd:
I've overshot with the reviewers I think, tagging TIPS and STM. It would be nice to have a Genny expert's opinion on this one.

richardsamuels:
If you're looking for genny expertise, your best options are @rtimmons (on vacation until 9/14) or perhaps @aggrand. I don't believe any current STM team members have significant expertise on genny.

ancostas:
I've added vlad and raiden here. One of them should be able to review this.

vrachev:
I have some bandwidth to take a look today

src/workloads/docs/StateMachineCrudActor.yml
```diff
@@ -0,0 +1,159 @@
...
+      OperationCommand:
+        WriteOperations:
+        - WriteCommand: insertOne
+          Document: {doc: 1, state: "Off"}
+        - WriteCommand: insertOne
+          Document: {stat: "OnOff", count: 0}
+        - WriteCommand: insertOne
+          Document: {stat: "OnSleep", count: 0}
+        - WriteCommand: insertOne
+          Document: {stat: "OnError", count: 0}
```
vrachev, line 51: We want to more clearly define the responsibilities of the actor and the yaml.

In the yaml we want to define the states, the transitions, and operations to perform during each state.

But the actual FSM itself should be contained in the actor. Meaning, the initialization of the states/transitions (represented by documents), the tracking of transitions (i.e. the updateOne to the transition doc) and the actual transition itself (changing the "currentState" doc to the next state) should all be done by the actor itself. So now this section becomes:

```
Operations:
- OperationName: initFSM
   OperationCommand:
     states: ["On", "Off", "Sleep", "Error"]
```

And the state definitions look like this:
```
  - Name: StateDemo
    Type: CrudState
    Threads: 1
    Database: StateTest
    Phases:
      - *nop
      - Duration: 30 seconds
        Collection: CrudDemo
        States:
          - Name: "On"  # The device is on
            Operations:
              # A state can have (optional) a set of operations that execute upon entering the state.
              # No longer need this, the actor handles changing the "currentState" doc to "On".
              # - OperationName: updateOne
              #  OperationCommand:
              #    Filter: &filter {doc: 1}
              #    Update:
              #      $set: {state: "On"}

            # Transitions specifies next states, delays, and operations. The On state can transition to
            # each of the three other states
            Transitions:
              - To: "Off"  # Transition to this state
                Weight: 1
                # The delay is the time before firing the operations. It can be any generator. The time is
                # in seconds. In this example it will take a uniform amount of time between 0 and 15 seconds
                # to execute. The time should be a floating point number.
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
                  # You can specify the time units. Seconds is the default value and may be ommitted. Valid
                  # choices are microseconds, milliseconds, seconds, minutes, hours
                  Units: seconds
                # The actor handles creating the transition doc {start: "On", end: "Off"} and incrementing it.
                # No other ops are allowed here. 
                # Operations:
                #  - OperationName: updateOne
                #    OperationCommand:
                #      Filter: {stat: "OnOff"}
                #      Update:
                #        $inc: {count: 1}
              - To: "Sleep"
                Weight: 2
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
                # Handled by actor.
                # Operations:
                #  - OperationName: updateOne
                #    OperationCommand:
                #      Filter: {stat: "OnSleep"}
                #      Update:
                #        $inc: {count: 1}
              - To: "Error"
                Weight: 0.2
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
                # Handled by actor.
                # Operations:
                #  - OperationName: updateOne
                #    OperationCommand:
                #      Filter: {stat: "OnError"}
                #      Update:
                #        $inc: {count: 1}
```

Here's what it looks like without the comments:
```
  - Name: StateDemo
    Type: CrudState
    Threads: 1
    Database: StateTest
    Phases:
      - *nop
      - Duration: 30 seconds
        Collection: CrudDemo
        States:
          - Name: "On"  # The device is on
            Operations:
               - OperationName: RunCommand
                 # User defined command here.
            Transitions:
              - To: "Off"  # Transition to this state
                Weight: 1
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
                  Units: seconds
              - To: "Sleep"
                Weight: 2
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
              - To: "Error"
                Weight: 0.2
                Delay:
                  Delay: {^RandomDouble: {min: 0, max: 15}}
```

vrachev (commented):
This is a great starting place, and having an FSM in Genny is a fun idea. We've done a similar thing in the correctness fuzzer for testing transactions. 

The idea here of using documents to model states and transitions is good, but I do have a few ideas for how to improve it.

src/workloads/docs/StateMachineCrudActor.yml
vrachev: The idea here is to not re-define a FSM for every workload.

src/workloads/docs/StateMachineCrudActor.yml
```diff
@@ -0,0 +1,159 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+
+description: |
+  TODO: This Actor does not exist yet. It will be implemented in PERF-2516.
```
rtimmons, line 5: This seems like a well-reasoned idea and a good starting point for the syntax.

I don't think this needs to extract/refactor Crud Actor code to be used in multiple Actors. Rather, I think something that encapsulates the "tell me which set of Operations to run now" would be a nice refactor seam.

```cpp
template<typename Op>
struct OpChooser {
    // Get list of operations to perform in this iteration of the Phase.
    std::vector<std::unique_ptr<Op>> ops(PhaseNumber currentPhase) const;
};

template<typename Op, typename...Args>
OpChooser<Op> make(
    const PhaseContext& ctx,
    std::initializer_list<std::pair<std::string, std::unique_ptr<Op>>> ops;
    ) {}

struct Operation {
    virtual void run() = 0;
    virtual ~Operation() = default;
};

struct OperationA : public Operation {
    void run() override {}
};
struct OperationB : public Operation {
    void run() override {}
};

struct PhaseConfig {
    OpChooser<Operation> chooser;
    PhaseConfig(const PhaseContext& ctx)
    : chooser{make<Operation>(
        ctx, {
            {"OperationA", std::make_unique<OperationA>()},
            {"OperationB", std::make_unique<OperationB>()}
        }
    )} {}
};

struct MyActor {
    void run() {
        for (auto&& config : _loop) {
            for (const auto&& op : config->chooser->ops()) {
                op->run()
            }
    }
    PhaseLoop<PhaseConfig> _loop;
};
```

Then the `OpsChooser` class knows how to translate the yaml structures of `Choice`/`Weight` and is constructed with the binding from operation names to impls. There is `Operations` xor `States` in the `Phase` block, so when it sees just `Operations`, the chooser is a static choice similar to what CrudActor currently does.

I'm dubious if this wants to be available to other Actors or if it wants to be isolated to CrudActor.{cpp,hpp} until its better proven out. I think it can easily start in just CrudActor and later brought out (or perhaps incorporated into PhaseLoop) if/when other Actors want similar functionality.

Also as mentioned in slack, there is reasonable fear/objection to changing CrudActor given its size. I suggest we move Crud Actor to its own cmake module as a peer with `cast_core` which will let us cleanly carve the cpp into multiple translation-units.

src/workloads/docs/StateMachineCrudActor.yml
dalyd: @vrachev The thought with this example was that the updateOne was part of the workload, not part of the state machine itself. It was an example of something you might want to do when you enter a state, or take a specific transition, but the actor would track its state independent of those operations. Does that make sense? 

As drawn above, what work on the cluster does the example do? I definitely would not want to hard code that work into the actor.

dalyd: Put another way -- the documents (updates  and all) are not modeling the states for the state machine. They are side-effects of the execution of the state machine. A real workload would have more interesting crud operations in their place.

src/workloads/docs/StateMachineCrudActor.yml
dalyd: @rtimmons would the only change here then be to make this "Crud" rather than "CrudState", or is there other syntax changes for it? I would be game for trying that. The code samples above scare me a bit, but I can deal and ask for help when I break the template stuff.

rtimmons: @dalyd 

> would the only change here then be to make this "Crud" rather than "CrudState"

Yes that's the biggest thing. I have some 💅 things to think on for syntax keywords, but the structure seems good. I'm glad the transition table is enumerated per-state rather than global or path/context-dependent.

But actually two questions upon further thought:

1. When is the state transition logic evaluated? After every iteration of the Operations loop, or at some interval? If after every iteration, the number of transitions taken could be rather large, which is maybe what you want. If at some time or iteration interval, do we want syntax to make that explicit?
2. I'm confused about the `Delay` functionality in `Transitions`. According to your comment, "The delay is the time before firing the operations," but what's the distinction between this and the  "Sleep" state?

The "sleep"/"delay" things may be slightly trickier to implement without impacting performance or digging kinda deep into `Orchestrator` or `PhaseLoop` internals. We already have [SleepBefore](https://github.com/mongodb/genny/blob/5d20179b3dafd7ae2d63e35f26e5d7782490e16e/src/workloads/docs/HelloWorld.yml#L11-L14) which we may want to unify with this concept.

dalyd: Reverse order on responses (kindof): 

1. The sleep state was meant to be part of the model. I had a cell phone or laptop in mind that could go into a sleep state. It's completely separate from the FSM machinery. If it's confusing I can completely put in a different state there, or I could expand the doc at the top to explain it better. 
2. Delay: My model is in each state, we pick a transition. After picking the transition, we wait for whatever the delay is before transitioning to the next state. Repeat. There can be a lot of transitions (or not), and that's my goal. 
3. The order of delays and choices, and if delays are associated with the state or the transition are big ones. I went with what made sense to me, but there are other options.  
4. The delay can be seen as related to the rate limiting, but it feels different to me. It's definitely within a phase though. 

Also, please throw on the comments about the syntax keywords.

aggrand: Related to the delay question, this workload looks like it's performing a single defined set of operations, then sleeping before transitioning and performing the next defined set of operations. Would there be cases instead where instead we'd have a set of operations repeated, with either a `Duration` or `Repeat` keyword, before transitioning? Or maybe a number of repeats or durations with sleeps before/after?

Asking because that would seem to be replicating more `PhaseLoop` logic.

src/workloads/docs/StateMachineCrudActor.yml
vrachev: As written the operations are doing things that the state machine should be doing, so it appeared that the state machine was being implemented *in* the yaml. I just wanted to make sure we wouldn't be doing things like tracking the current state or keeping metrics on what states we've visited already. The actor should be responsible for telling us what state we're in, or how many times we've made X transition or been in Y state. Perhaps it's fine as an example, but it just confused me a bit to see that being done manually in the yaml.

src/workloads/docs/StateMachineCrudActor.yml
rtimmons: > After picking the transition, we wait for whatever the delay is before transitioning

It sounds like `Duration` is the existing concept we could leverage.

And since we're getting `Duration` in States, I wonder if the rest of the Phase-level syntax keywords want to work here as well. I.e., `Duration`, `Repeat`, `SleepBefore`, rate-limiting, etc. That would certainly unify the syntax between Phases and States.

But it requires some nontrivial (though not huge) surgery to make it happen. I *think* this is more or less equivalent to baking the FSM machinery into PhaseLoop.

It gets hard to reason about rate-limiting and `Blocking` at the `Phase` level and at the `State` level, so we'd probably want to have some guidance and clear thinking on how to design with those.

Making the `Duration` value-generator-aware would be cool. We can clone `RandomDouble` or similar to make it a `ValueGenerator<TimeSpec>` to get your idea.

Here's my slight tweak on what you propose. Main differences (1) put Phase-like semantics into each `State`, and (2) use `WithProbability` as a percentage in favor of `Weights`. Last part is subjective, and first part has some pros/cons that are worth discussing.

```yaml
Actors:
- Name: Store
  Type: Crud
  Phases:
  - Duration: 10 minutes
    States:
    - Name: Find
      Duration:
        ^RandomDuration: {min: 1, max: 10, unit: seconds}
      GlobalRate: 1000 per second
      Operations: [{findOne}]
      Transitions:
      - To: Update
        WithProbability: 1%
      - To: {^Choice: [Update, Find]}
        WithProbability: 99%
    - Name: Update
      Repeat: 1000
      Operations: [{updateOne}]
      Transitions:
      - To: Find
        WithProbability: 10%
```

We could start with just `Duration` and reject other keywords for now. This would let us delay surgery to `PhaseLoop` at the expense of duplicating some [`TimeSpec` and time-check logic](https://github.com/mongodb/genny/blob/e8cc83e21b32a2713201207f8d8fe7624e0dd005/src/gennylib/include/gennylib/PhaseLoop.hpp#L170-L175).

dalyd: > Related to the delay question, this workload looks like it's performing a single defined set of operations, then sleeping before transitioning and performing the next defined set of operations. Would there be cases instead where instead we'd have a set of operations repeated, with either a `Duration` or `Repeat` keyword, before transitioning? Or maybe a number of repeats or durations with sleeps before/after?
> 
> Asking because that would seem to be replicating more `PhaseLoop` logic.

That is a great question. Not for my current case, but I could imagine cases in which it would be true. I'm thinking a lot about whether I want this or not.

dalyd: @rtimmons with your example, global rate doesn't make sense. I imagine it's a state per actor instance, as opposed to all the instantiations transitioning at the same time.

dalyd:
Small update on the example. I changed the type to Crud, and I updated the text on the example to hopefully clarify what is state machine machinery, and what is the workload behavior. I think that was causing some confusion.

dalyd:
Okay, updates are up from this morning's discussion. Please take a look.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+      Repeat: 1
+      Operations:
+      # A state can have (optional) a set of operations that execute upon entering the state. This
+      # state updates a specific document to indicate it is in state On.
+
+      # Note: Following the analogy to phases -- phases may be blocking or non-blocking. States are
+      # always non-blocking. When the phase ends, the state machine will finish any open operations,
+      # and then exit. It if is waiting for a delay, it will exit immediately. This is required as
+      # there may be states with long delays.
+      # Question: Should we support having a single fire operation with repeating operations? Here I
```
dalyd, line 81: I embedded this question in the comments so I wouldn't forget. 

> Should we support having a single fire operation with repeating operations? Here I
> could invent a case where the following operation is repeat 1, but there is another operation that repeats once a second while in this state. I don't know how to represent that.  

Thoughts?

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+
+        # Transitions are probabilistically chosen based on weight. A transition with weight 2 is
+        # twice as likely to be picked as a transition with weight 1. Weights must be greater than
+        # 0.
+        Weight: 1
+
+        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.
+        # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
aggrand, line 104: On one hand, I feel like `SleepBefore` should be in the state not the transition. That would fit with adding other `PhaseLoop`-like syntax such as `Duration` or `Repeat` (which would only make sense in the state). On the other hand, `SleepBefore` in states wouldn't let you differentiate the time before transitioning for different transitions without adding intermediate states. Assuming we go the route of only having ops and such in the state.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+      Operations:
+      - OperationName: updateOne
+        OperationCommand:
+          Filter: *filter
+          Update:
+            $set: {state: "Off"}
+      Transitions:  # Can only go to On from Off
+      - To: "On"
+        Weight: 1
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
aggrand, line 116: Should these generators be using RandomInt instead? We have a TimeSpec constructor from `int64_t` which might be easier to repurpose.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+      # always non-blocking. When the phase ends, the state machine will finish any open operations,
+      # and then exit. It if is waiting for a delay, it will exit immediately. This is required as
+      # there may be states with long delays.
+      # Question: Should we support having a single fire operation with repeating operations? Here I
+      # could invent a case where the following operation is repeat 1, but there is another
+      # operation that repeats once a second while in this state. I don't know how to represent
+      # that.  
+      - OperationName: updateOne
+        OperationCommand:
+          Filter: &filter {phoneID: {^ActorID: {}}}
```
jimoleary, line 87: typo: It should be ActorId.

Also will the actor ids be in the correct range (from 0 to 4)?

src/workloads/docs/CrudActorFSMAdvanced.yml
```diff
@@ -0,0 +1,186 @@
...
+      Operations:
+      - OperationName: updateOne
+        OperationCommand:
+          Filter: *filter
+          Update:
+            $set: {state: "Off"}
+      Transitions:  # Can only go to On from Off
+      - To: "On"
+        Weight: 1
+        Delay:
```
jimoleary, line 151: sleepBefore?

src/workloads/docs/CrudActorFSM.yml
dalyd: I *think* ActorIDs start at 0. If someone could confirm, that would be great. If not, I'll adjust at implementation time when this can actually run.

src/workloads/docs/CrudActorFSM.yml
dalyd: Exactly. That's hy I put the SleepBefore in the transition.

src/workloads/docs/CrudActorFSM.yml
dalyd: I'm open to pushing here, but since this includes units, shouldn't it be a float? I feel like 3.43 seconds should be a valid timespec.

src/workloads/docs/CrudActorFSMAdvanced.yml
dalyd: Yes. Thank you.

src/workloads/docs/CrudActorFSM.yml
aggrand: They start at 1, actually because of poplar weirdness: https://github.com/mongodb/genny/blob/2f2dff07f4b1908cb5a5707b90226bcd42f827c4/src/gennylib/include/gennylib/context.hpp#L284

dalyd: Thanks. Updated so setup inserts 1-5 rather than 0-4.

dalyd:
All, how are we looking on this one? I think I've addressed all the open questions.

src/workloads/docs/CrudActorFSM.yml
rtimmons: > a single fire operation with repeating operations

I don't understand that question tbh.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+
+    - Name: "Off"  # The device is Off
+      Operations:
+      - OperationName: updateOne
+        OperationCommand:
+          Filter: *filter
+          Update:
+            $set: {state: "Off"}
```
rtimmons, line 112: I grok the `state` here versus the `State.Name`, but naming both the same may muddy the waters for new people. Consider `$set: {currentPhoneState:PhoneIsOff}` to make it clear that `Name:Off` and `state:Off` are technically orthogonal from Genny's perspective.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,148 @@
...
+
+        # Transitions are probabilistically chosen based on weight. A transition with weight 2 is
+        # twice as likely to be picked as a transition with weight 1. Weights must be greater than
+        # 0.
+        Weight: 1
+
+        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.
+        # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
rtimmons, line 104: Does `SleepBefore` in the `Transition` indicate it will sleep before doing the Transition's Operations? During that `SleepBefore` time-period, is the Actor nop'ing or continuing to do its current State's Operations?

src/workloads/docs/CrudActorFSMAdvanced.yml
```diff
@@ -0,0 +1,180 @@
...
+        # twice as likely to be picked as a transition with weight 1. Weights must be greater than
+        # 0.
+        Weight: 1
+
+        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.
+        # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
+
+        # TODO: REmove the transition operations
```
rtimmons, line 112: Is this TODO still valid?

src/workloads/docs/CrudActorFSMAdvanced.yml
```diff
@@ -0,0 +1,180 @@
...
+        # In this advanced case, operations may also be associated with transitions.
+        Operations:
+        - OperationName: updateOne
+          OperationCommand:
+            Filter: {stat: "OnOff"}
```
rtimmons, line 117: `state` (couple others too)

rtimmons (commented):
This is getting better!

Maybe it's me being dense, but tldr for my review here is that I have a hard time reasoning about the intended behavior for

1. `Sleep{Before,After}` and `Transitions` wrt whe Actor is doing during that time-period
2. How sleep/repeat/duration interact with`Blocking` 

I'm also fine to sit back and see how this pans out for just the Crud Actor. As long as it's backward-compatible and limited to "just" the Crud Actor, the details can be ironed out after a bit of POC/implementation.

Over-specifying exploratory new features can be a buzz-kill, so please don't feel compelled to spend more time deliberating than actually coding in this case if you don't want to.

src/workloads/docs/CrudActorFSM.yml
dalyd: I was imagining having something that happens once when you enter the state, combined with something that happens with some rate while you are in that state. Does that make sense? Maybe we just don't do anything like that for now.

src/workloads/docs/CrudActorFSM.yml
dalyd: I read it that SleepBefore is before the transition, so the current state keeps doing its thing. I tried to capture that in updated comments.

src/workloads/docs/CrudActorFSM.yml
dalyd: Done

src/workloads/docs/CrudActorFSMAdvanced.yml
dalyd: No. thanks.

src/workloads/docs/CrudActorFSMAdvanced.yml
dalyd: This actually is a "stat". I'll update to statistic.

src/workloads/docs/CrudActorFSM.yml
dalyd: For now I've removed the question. We can continue the discussion though.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -68,23 +70,24 @@ Actors:
...
-      # always non-blocking. When the phase ends, the state machine will finish any open operations,
-      # and then exit. It if is waiting for a delay, it will exit immediately. This is required as
-      # there may be states with long delays.
-      # Question: Should we support having a single fire operation with repeating operations? Here I
-      # could invent a case where the following operation is repeat 1, but there is another
-      # operation that repeats once a second while in this state. I don't know how to represent
-      # that.
+      # always non-blocking, in the sense that any sleepBefore, SleepAfter delays will be
+      # interrupted to close the phase. However, any active operations will be finished.
+
```
dalyd, line 80: @rtimmons does this added text make sense to your question? Both above and below this comment.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,157 @@
...
+        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.  For
+        # sleepBefore, the system is considered to remain in it's current state. If there is a
+        # repeating action, it will continue repeating until the transition fires.  For repeat, the
+        # operation is repeated immediately.  sleepAfter will not be supported for transitions with
+        # no operations.
+
+        # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
rtimmons, line 113: Same comment / suggested example text above - the `Transition`s always respect the blocking behavior of the Phase they're in, ya?

src/workloads/docs/CrudActorFSM.yml
rtimmons: Consider only allowing integer values but allowing for more precision with the units `{^TimeSpec: {seconds: {^Int}}`. I have little strong opinion here fwiw beyond trying at least a bit to avoid decimals if we can.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -0,0 +1,157 @@
...
+      Operations:
+      - OperationName: updateOne
+        OperationCommand:
+          Filter: *filter
+          Update:
+            $set: {currentPhoneState: "PhoneIsInError"}
+      Transitions:  # Turn off on error
+      - To: "Off"
+        Weight: 1
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
rtimmons, line 152: Not gonna block merge or anything, but fwiw most of the quotes in this example aren't necessary and (imho) are a bit distracting. We don't (yet) lint quotes, but that would be the first rule I'd add 😸

src/workloads/docs/CrudActorFSMAdvanced.yml
```diff
@@ -0,0 +1,182 @@
...
+
+        # Transitions are probabilistically chosen based on weight. A transition with weight 2 is
+        # twice as likely to be picked as a transition with weight 1. Weights must be greater than
+        # 0.
+        Weight: 1
+
+        # Transitions also support the phase keywords of sleepBefore, sleepAfter, and repeat.
+        # ^TimeSpec is a proposed generator to randomly generate acceptable TimeSpecs with units. It
+        # accepts any valid int or double generator or constant for value.
+        SleepBefore: {^TimeSpec: {value: {^RandomDouble: {min: 0, max: 15}}, unit: "seconds"}}
```
rtimmons, line 113: I think I agree with this syntax and functionliaty, but it's hard for me to keep all the different Sleep/Duration/Repeat/Blocking combinations when they can be present in 3 different places. I think it'll be (slightly) clearer after the basic States functionality is a thing.

I I'm inclined to say let's wait to see how the syntax and functionality pans out without Operations in Transitions before adding them. So maybe keep this here but add a note to the top that it's more speculative than the other example and not (yet) prescriptive of what the featureset will be.

src/workloads/docs/CrudActorFSM.yml
dalyd: FWIW, yamllint now passes on this file. It really wants On and Off quoted, otherwise it complains about truthy values should be true or false.

src/workloads/docs/CrudActorFSMAdvanced.yml
dalyd: Yeah, can do. That was the intention anyways.

src/workloads/docs/CrudActorFSM.yml
dalyd: Added a comment in the comment block above.

dalyd:
I just had a talk with @rtimmons on this one. I'm going to merge now and continue working on the implementation. The aim will be the simplest working form, which may constrain some of the stated functionality. I will update the examples as appropriate as I go. Thanks.

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #538, opened 2021-09-14
TIG-3347 Disable LoadTest on 4.0

vrachev:


aggrand (approved):
The change itself LGTM, but should we reach out to product perf to make sure they're okay with this not running on 4.0? If they do want it running, we could do this to squash the BF and file a separate ticket to update quiesce (which we own) to work in 4.0 also and re-enable.

vrachev:
@dalyd is product perf ok with removing LoadTest from 4.0? If we absolutely need the test we could work on making the Quiesce actor compatible with 4.0 as Raiden mentioned. Thanks

dalyd:
> @dalyd is product perf ok with removing LoadTest from 4.0? If we absolutely need the test we could work on making the Quiesce actor compatible with 4.0 as Raiden mentioned. Thanks

I'm going to say yes. 
FYI @mongodb/product-perf can be used to signal the whole. Team. 
@ghartnett you created the test. Any objections? It sounds reasonable to me.

dalyd (approved):

ghartnett:
LGTM. No objections from me.

vrachev:
evergreen merge

PR #539, opened 2021-09-16
TIG-2835 Enforce Documentation for Genny workloads

aggrand:
Re-opened due to the revert. Works now with the auto-run test.


vrachev (approved):
Thanks for fixing!

aggrand:
evergreen merge

PR #540, opened 2021-09-17
Use BOOST_THROW_EXCEPTION more pervasively

rtimmons:
This ensures we always have location and diagnostic information in exception messages.

Idk how feasible it is to lint this, but that's something worth considering in the future.


aggrand (approved):
Good catch, thanks for adding these!

Yeah not sure how lintable it'd be without doing a lot of custom stuff with clang's api, but might be a fun spike.

PR #541, opened 2021-09-23
PERF-2571: Add genny tests for $graphLookup with maxDepth 0

HanaPearlman:
From [this](https://spruce.mongodb.com/version/6148dfd20305b97273016779/tasks) evergreen patch, we see the following results.

Feature flag off:

|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroOneToOne  | 5.0487785019 |
| GraphLookupShardedToUnshardedMaxDepthZeroOneToFive | 1.5662687845 |
| GraphLookupShardedToUnshardedMaxDepthZeroOneToOneHundred | 1.3148955413 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchOneToTenTargeted | .8854111305 |

Feature flag on:
|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroOneToOne | 3.660342830 |
| GraphLookupShardedToUnshardedMaxDepthZeroOneToFive | 1.6935450762 |
| GraphLookupShardedToUnshardedMaxDepthZeroOneToOneHundred | 1.178930092 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchOneToTenTargeted | .5505447558 |

In most cases except OneToFive, we're seeing improvements of 25%+. For OneToFive, there is a regression of about 6%.
These results are interesting! `$graphLookup` with `maxDepth: 0` is much more similar to `$lookup`, yet we still see better performance when parallelizing in the `ShardedToUnsharded` case than we do with `$lookup`! It must be that the cache is somehow helping us out here. Would welcome any other ideas!

### DIFF
--- a/src/workloads/execution/UnshardedGraphLookup.yml
+++ b/src/workloads/execution/UnshardedGraphLookup.yml
@@ -225,6 +225,7 @@ Actors:
         aggregate: Collection0
         pipeline:
           [
+            {$match: {a: {$lte: 30}}},
             {$graphLookup: {
               from: "Collection1",
               startWith: "$c",
--- a/src/workloads/execution/UnshardedGraphLookup.yml
+++ b/src/workloads/execution/UnshardedGraphLookup.yml
@@ -54,6 +54,7 @@ Actors:
       d: {^RandomInt: {min: 1, max: 600}}
       e: {^RandomInt: {min: 1, max: 600}}
       f: {^RandomInt: {min: 1, max: 30}}
+      g: {^RandomInt: {min: 1, max: 1500}}
       s: {^RandomString: {length: 3000}}
   - *Nop
   - *Nop
@@ -167,6 +168,23 @@ Actors:
             {$project: {s: 0, "matches.s": 0}}
           ]
         cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: GraphLookupShardedToUnshardedMaxDepthZeroOneToTwo
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline:
+          [
+            {$graphLookup: {
+              from: "Collection1",
+              startWith: "$g",
+              connectFromField: "g",
+              connectToField: "g",
+              as: "matches",
+              maxDepth: 0
+            }},
+            {$project: {s: 0, "matches.s": 0}}
+          ]
+        cursor: {batchSize: *NumDocs}
     - OperationMetricsName: GraphLookupShardedToUnshardedMaxDepthZeroOneToFive
       OperationName: RunCommand
       OperationCommand:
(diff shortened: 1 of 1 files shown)


src/workloads/execution/UnshardedGraphLookup.yml
```diff
@@ -149,6 +150,74 @@ Actors:
...
+              startWith: "$f",
+              connectFromField: "f",
+              connectToField: "f",
+              as: "matches",
+              maxDepth: 0
+            }},
+            {$project: {s: 0, "matches.s": 0}}
+          ]
+        cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: GraphLookupShardedToUnshardedMaxDepthZeroMatchOneToTenTargeted
```
kaywux, line 204: Could you explain what you mean by "targeted" here? My understanding is that the $graphLookup stage itself will be sent to execute on all shards in parallel, and then generate queries on "Collection 1" that will target the primary shard as it will with the other added tests. If the goal of this is to have $graphLookup execute on only a few shards I think you'll need a predicate on "c" before the $graphLookup (like there is in the MatchSomeShards test). Let me know if I'm missing something!

kaywux (changes requested):
Thanks for this Hana! And yes, I agree that the cache is really helping us out here. If I am understanding correctly - there is such a latency drop-off from 1:1 and 1:5 (with the flag both on and off) because in the latter case there are 5 documents with the same "d", so the $graphLookup only needs to match with "e" on the first time it sees that "d", and can use the cache to answer this the subsequent times. Since in the previous implementation the cache is shared we can use the cache a total of four times for each "d", but once we start executing in parallel each shard has its own cache so we won't be able to use the cache as often - which would explain the slight perf regression here. Once we get to 1:10 and 1:100 this difference starts being negligible with how often each cache is being accessed, and parallelization becomes more beneficial.

Could you also add a test for 1:2, where each `connectFromField` value is only in two documents? I'd expect that this would be a larger regression than 1:5, but would like to see by how much.

src/workloads/execution/UnshardedGraphLookup.yml
HanaPearlman: Yes, this was a mistake and you're right, I really meant to be testing MatchAllShards vs MatchSomeShards. Updated in the next commit!.

HanaPearlman:
I've added a couple new tests, and here are the new results

Feature flag off:

|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroOneToTwo  | 3.9163753362 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchSomeShards | .0119693981 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchAllShards | .9161633825 |

Feature flag on:
|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroOneToTwo | 3.234654659 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchSomeShards | .0095686224 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchAllShards | .5757880516 |

The results are found in the same patch [here](https://spruce.mongodb.com/version/6148dfd20305b97273016779/tasks).

Again, in these cases, parallelized $graphLookup matches or out-performs $graphLookup run on the primary only.

I did not expect to see a regression in the OneToTwo case. If there are two queries for the same foreign key, then it is essentially just as fast to do the queries in parallel as it is to do them serially -- one query for the key and then the second query can look up the result in the cache. As expected, the MatchSomeShards case is much closer in timing results!

src/workloads/execution/UnshardedGraphLookup.yml
```diff
@@ -149,6 +151,109 @@ Actors:
...
+            {$project: {s: 0, "matches.s": 0}}
+          ]
+        cursor: {batchSize: *NumDocs}
+    - OperationMetricsName: GraphLookupShardedToUnshardedMaxDepthZeroMatchAllShards
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline:
+          [
+            {$graphLookup: {
```
kaywux, line 228: Could you add a $match here to limit the docs flowing in to $graphLookup to around 30? This would make it easier to compare with the `MatchSomeShards` case

kaywux (approved):
This LGTM with one more minor change, thanks for the changes!

HanaPearlman:
Newest results for changed tests!

Feature flag off:

|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroMatchSomeShards | .0112497851 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchAllShards | .0837548874 |

Feature flag on:
|---|---|
| GraphLookupShardedToUnshardedMaxDepthZeroMatchSomeShards | .009571348.4 |
| GraphLookupShardedToUnshardedMaxDepthZeroMatchAllShards | .0554753451 |

The results are found in the same patch [here](https://spruce.mongodb.com/version/6148dfd20305b97273016779/tasks).

src/workloads/execution/UnshardedGraphLookup.yml
HanaPearlman: Done! Moving on to merging.

HanaPearlman:
evergreen merge

PR #542, opened 2021-09-29
PERF-2380 Add $densify performance tests

davish:


davish:
A note that DSI also need to be updated to include `featureFlagDensify` in order for the perf test to run properly. Link to the evergreen patch with the change: https://spruce.mongodb.com/task/sys_perf_linux_3_node_replSet_disabled_feature_flags_densify_patch_78ef9784e77a2411e8aa2b582bc80033f0f36228_615362831e2d1730c9d3469e_21_09_28_18_46_10/tests?execution=1&sortBy=STATUS&sortDir=ASC

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
+SchemaVersion: 2018-07-01
```
TedTuckman, line 1: Worth finding out if there is any kind of best practice to number of measurements per file.

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10000
+    BatchSize: &batchSize 1000
+    Document:
+      partitionKey: {^RandomInt: {min: 1, max: 100}}
```
TedTuckman, line 19: Do we think it would be interesting to test differences in number of partitions? I don't think it should have an impact, except maybe in cases where we need to finish densifying after the input.

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+      # Test a subset of units that have different characteristics.
+      number: {^RandomInt: {min: 0, max: 20000}} # Seperate codepath for numeric DensifyValues.
+      milliseconds: {^RandomDate: {min: "2021-01-01T00:00:00.000Z", max: "2021-01-01T00:00:20.000Z"}} # Most basic date unit.
+      hours: {^RandomDate: {min: "2021-01-01", max: "2023-01-01"}} # Date unit with constant number of milliseconds.
+      months: {^RandomDate: {min: "2021-01-01", max: "4000-01-01"}} # Date unit with non-constant number of milliseconds.
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: CreateIndex
```
TedTuckman, line 28: Do we use this index?

TedTuckman: Oh, we probably use it for sorting. Maybe a comment?

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+
+- Name: Densify
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10
+    Database: *db
+    Operations:
```
TedTuckman, line 58: Is there anything we can do with spacing or comments to make this section more easily digestible?

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+    # Numeric metrics. Each unit has a metric for each possible codepath (Partition/no partition, explicit/full/partitioned bounds.)
+    - OperationMetricsName: NumericFullDensify
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "number", range: {bounds: "full", step: {^RandomInt: {min: 1, max: 11}}}}}]
```
TedTuckman, line 64: Will random step (and other values) be consistent across repetitions of the test inside the same run? If not, should they be or do we think it doesn't matter?

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "number", partitionByFields: ["partitionKey"], range: {bounds: "partition", step: {^RandomInt: {min: 1, max: 11}}}}}]
+        cursor: {batchSize: *batchSize}
+    - OperationMetricsName: NumericExplicitRangeDensify
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "number", range: {step: {^RandomInt: {min: 1, max: 11}}, bounds: [10000, 30000]}}}]
```
TedTuckman, line 82: Not random bounds?

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "milliseconds", partitionByFields: ["partitionKey"], range: {bounds: "partition", unit: "millisecond", step: {^RandomInt: {min: 1, max: 11}}}}}]
+        cursor: {batchSize: *batchSize}
+    - OperationMetricsName: MillisecondExplicitRangeDensify
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "milliseconds", range: {step: {^RandomInt: {min: 1, max: 11}}, unit: "millisecond", bounds: [{^RandomDate: {min: "2021-01-01T00:00:05.000Z", max: "2021-01-01T00:00:05.001Z"}}, {^RandomDate: {min: "2021-01-01T00:00:15.000Z", max: "2021-01-01T00:00:15.001Z"}}]}}}]
```
TedTuckman, line 113: The random min/max here are very small -- with this range may as well hardcode, or should come up with a better random bounds that guarantees max is above min.

TedTuckman (commented):
Some comments, I think we're testing the important cases. Delegating approval to Katherine, feel free to merge without me while I'm on vacation if she approves after addressing my comments.

src/workloads/execution/Densify.yml
davish: I checked on the performance channel and apparently genny uses the same random seed every time, so across evergreen builds the tests will be consistent.

src/workloads/execution/Densify.yml
davish: this range is hardcoded right now -- the only way to add a date type in a query is to use $^RandomDate with a min and max bound. I'll add a  comment explaining that, there's an open ticket to add in a better way to do this. I'll probably move the ranges so that they are actually random to add some variation, though.

src/workloads/execution/Densify.yml
davish: done.

src/workloads/execution/Densify.yml
davish: I'll add in some blank lines to give some visual space between each unit.

src/workloads/execution/Densify.yml
davish: done.

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,194 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "number", range: {bounds: "full", step: {^RandomInt: {min: 1, max: 11}}}}}]
+        cursor: {batchSize: *batchSize}
+    - OperationMetricsName: NumericFullDensifyByPartition
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{$densify: {field: "number", partitionByFields: ["partitionKey"], range: {bounds: "full", step: {^RandomInt: {min: 1, max: 11}}}}}]
```
kaywux, line 70: Could you format these pipelines to be easier to read? Not sure if genny has a linter/formatter, but could be something as simple as 
 ```
[{
    $densify: {
        field: "number",
        partitionByFields: ["partitionKey"],
        range: {bounds: "full", step: {^RandomInt: {min: 1, max: 11}}}
    }
}]
```

File: src/workloads/execution/Densify.yml (hunk 1 above)
Comment (davish) [reply src/workloads/execution/Densify.yml:1]: Asked on slack, but looking at the other workloads it seems like the general idea is to have the workload as the logical unit of whatever you're testing.

File: src/workloads/execution/Densify.yml [hunk 9]
```diff
@@ -0,0 +1,194 @@
...
+        indexes: 
+        - key: {number: 1}
+          name: number
+        - key: {milliseconds: 1}
+          name: milliseconds
+        - key: {hours: 1}
+          name: hours
+        - key: {months: 1}
+          name: months
+  - *Nop
```
kaywux, line 48: Could you add a Quiesce phase before the densify commands? This will help with workload noise, especially on the first of the repetitions.

kaywux (commented):
Thanks Davis! I left a few general comments and will wait until you address Ted's requests to take a deeper look.

src/workloads/execution/Densify.yml
davish: Done

src/workloads/execution/Densify.yml
davish: Done

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,421 @@
...
+    CollectionCount: 1
+    DocumentCount: 10000
+    BatchSize: &batchSize 1000
+    Document:
+      partitionKey: {^RandomInt: {min: 1, max: 100}}
+      # Test a subset of units that have different characteristics.
+      number: {^RandomInt: {min: 0, max: 20000}} # Seperate codepath for numeric DensifyValues.
+      milliseconds: {^RandomDate: {min: "2021-01-01T00:00:00.000Z", max: "2021-01-01T00:00:20.000Z"}} # Most basic date unit.
+      hours: {^RandomDate: {min: "2021-01-01", max: "2023-01-01"}} # Date unit with constant number of milliseconds.
+      months: {^RandomDate: {min: "2021-01-01", max: "4000-01-01"}} # Date unit with non-constant number of milliseconds.
```
kaywux, line 24: Question: what is the difference between `hours` and `months` in terms of constant/non-constant number of milliseconds? What I'm seeing here is that the `months` values will just be spread across a wider domain.

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,421 @@
...
+    - OperationMetricsName: NumericFullDensify
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{
+          $densify: {
+            field: "number", 
+            range: {
+              bounds: "full", 
+              step: {^RandomInt: {min: 1, max: 11}}
```
kaywux, line 88: If I'm understanding $densify / this workload correctly - each document will have a `number` field that is a value from 0 - 20000. Since there are 10000 documents in the collection, there will be many "gaps" in between the `number` field for these documents, but there is a very small probability that there would be a "gap" that is as big as 10 documents (ie. we have a doc with `number: 0` but the second in the sorted order is `number: 11`).

For $densify does this mean that, if the step size is a certain number (let's say 10) we would only generate new documents if there is a "gap" in the data that is that big? And if so, have you done the [noise/robustness testing](https://docs.google.com/document/d/1336S0iRRqNBDc_xgnmqOFoIx6G4_8FJiSZg0vpQMRBE/edit#heading=h.bwieg7qd7dzx) for these workloads? I can imagine that if the first question is true (and let me know if I'm missing something!), then we would have a lot of variance in these results depending on the step size that is chosen on each repeat - a better approach to this kind of perf testing might be to have separate operations with different step sizes (for example NumericFullDensifySmallStepSize and NumericFullDensifyLargeStepSize).

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,421 @@
...
+            }
+          }
+        }]
+        cursor: {batchSize: *batchSize}
```
kaywux, line 92: Unfortunately with the way that Genny workloads currently work, the results from this operation as it's written now will only measure how much work it would take to get the first batch of results, and not the performance of the whole operation. And because the batchSize here is significantly smaller than the number of docs in the collection (and even smaller than the amount of documents we would eventually want to output) this could give us a false understanding of performance of the $densify stage as a whole.

There are several ways to get the performance results of the whole operation (like wrapping this command in an explain and using `executionStats`) but the one that's most consistent and least likely to couple the performance of these workloads with non-$densify code is to try and fit all the results into the first batch size. Could you make a back-of-the-envelope calculation of the size of each document, as well the expected number of documents we would generate and return, and adjust the size of the docs and the batch size accordingly? Happy to talk about other methods as well if you want to explore another avenue for how to do this!

src/workloads/execution/Densify.yml
kaywux: Depending on the answer to one of my questions later down in the workload, we might have to separate these existing operations out into even more test cases. In that case, it seems like the numeric testing and the date-type testing is fairly distinct - how do you feel about separating these tests into two workloads, something like `DensifyNumeric` and `DensifyDate`? This would make the workloads each more digestible, and also allow you to be more specific about document fields for each workload and the indexes you need to add for a covered $sort.

kaywux (changes requested):
I left a few larger comments, let me know if you have questions or would like to talk further about any of them!

src/workloads/execution/Densify.yml
davish: An hour will always be the same number of milliseconds (1000*60*60), but a month is a different number of milliseconds depending on the number of days in that month, which needs to be taken into account when determining whether a given value is on the step defined in the stage. There's different codepaths in the densify stage to account for this, and there's some potential for inefficiencies in the case for months, quarters and years, so I wanted to record metrics for both months and hours to find any large discrepancies.

This could certainly be clearer from the comment, and I'll add more details!

src/workloads/execution/Densify.yml
davish: To my knowledge $densify doesn't look for a gap of a certain size, but rather generates documents that are on-step relative to a lower bound that's either explicitly defined, or is the minimum value in the range/partition. So a collection with the values [2,5,7,9,13,14,15,19] and a densify operation with `{range: "full", step: 4}` would output [2,5,**6**,7,9,**12**,13,14,15,**16**,19] (inserted documents in bold) even though the gaps between documents are at their smallest there.

Step sizes will have an impact on the number of documents generated, though, which could impact runtime. It might be a good idea to split up the metrics into smaller and larger steps.

I haven't done noise/robustness testing though, and that's definitely something I missed! I'll be sure to queue up more builds to double check.

src/workloads/execution/Densify.yml
davish: Modifying the batch size makes sense to me -- will do!

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,421 @@
...
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - Repeat: 10
+    Database: *db
+    Operations:
+
+    # Metrics are grouped by the step unit they operate with (numiric, millisecond, hour, month).
```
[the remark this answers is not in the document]
davish: done.

src/workloads/execution/Densify.yml
davish: Increased batch size to 30k, since we're unlikely to generate more documents than that and since each document will only take up 5 bytes of space, this will be way less than 1mb.

src/workloads/execution/Densify.yml
kaywux: Ah got it, thanks for the explanation! An edited comment would be great.

src/workloads/execution/Densify.yml
davish: Would splitting up the workloads make it more difficult to compare the metrics between numeric and date workloads? While they're distinct in the code, I think it would be valuable to be able to compare the performance between both types, since I wouldn't expect them to be too different and a huge discrepancy would be something to look into.

src/workloads/execution/Densify.yml
kaywux: Thanks for clarifying that the step size is relative to the lower bound / min value specifically. Even with this, I agree that a widely-varying step size would have a not-insignificant impact on the latency results (the specifics of which I'm looking forward to seeing from your tests) so separate metrics would still be useful to have!

src/workloads/execution/Densify.yml
kaywux: That's a good point! My gut reaction is that apples-to-apples comparison across workloads would be fairly simple as long as we keep the relative document size and count the same. All in all it seems like a matter of preference, so as long as we don't find the length of this workload growing too unwieldy, I'm fine with either option.

src/workloads/execution/Densify.yml
davish: done.

src/workloads/execution/Densify.yml
davish: Just split up the tests into a small step (1-3) and a large step case (7-11). It blew up the size of the file, but I still think it's pretty readable. I could very well be misguided on that since I also put the file together, so let me know on that front

src/workloads/execution/Densify.yml
kaywux: Hmm after your last change this is growing pretty unwieldy; I think over 700 lines of a workload is a little too much for reading comprehension in my eyes. Could you split this into multiple workloads?

src/workloads/execution/Densify.yml
```diff
@@ -0,0 +1,749 @@
...
+  - *Nop
+  - Repeat: 10
+    Database: *db
+    Operations:
+
+    # Metrics are grouped by the step unit they operate with (numeric, millisecond, hour, month).
+    # Each group has a metric for each possible codepath (Partition/no partition, explicit/full/partitioned bounds).
+
+    # Numeric metrics. 
+    - OperationMetricsName: NumericFullDensifySmallStep
```
kaywux, line 79: Nitpicky comment (but perhaps an annoyingly manual one) - for ease of comprehension could you interleave the small/large step operations with the same densify type? Aka NumericFullDensifySmallStep, and then right after, NumericFullDensifyLargeStep. This way readers can easily see the difference between these two queries, instead of having to look through each $densify combination for a small step and then doing it again for a large step.

kaywux, line 79: Nitpicky comment (but perhaps an annoyingly manual one) - for ease of comprehension could you interleave the small/large step operations with the same densify type? Aka NumericFullDensifySmallStep, and then right after, NumericFullDensifyLargeStep. This way readers can easily see the difference between these two queries, instead of having to look through each $densify combination for a small step and then doing it again for a large step.

kaywux (commented):
A few more organizational requests! I'm going to hold off on approving this until we have some sense of the variance of the results, but this is looking pretty good to me :)

src/workloads/execution/Densify.yml
davish: After splitting up the workloads, one thing I'm worried about is for some of the common configuration getting out of sync. The metrics across different files won't be able to be usefully compared to each other unless the same documents are loaded in for both, and the queries are equivalent in both files. Some things like the step sizes for small steps and large steps and the batch size are factored out into YAML anchors, which could pretty easily get out of sync. Wondering if that should be a concern, or if the workloads will be changing so infrequently as to not really matter.

davish: Ended up splitting up into four workloads, one for each unit.

src/workloads/execution/DensifyHours.yml
```diff
@@ -0,0 +1,256 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload test tests the performance of the $densify stage. With step intervals of
+  different date units.
```
kaywux, line 5: Small suggestions to edit the wording of the description (and the ones following): "This workload tests the performance of the $densify stage with step intervals of different date units."

src/workloads/execution/DensifyHours.yml
```diff
@@ -0,0 +1,256 @@
...
+        createIndexes: Collection0
+        indexes: 
+        - key: {number: 1}
+          name: number
+        - key: {milliseconds: 1}
+          name: milliseconds
+        - key: {hours: 1}
+          name: hours
+        - key: {months: 1}
+          name: months
```
kaywux, line 50: We'd still want the generated documents to have the same fields across each workload to keep the doc size consistent, but as far as I can tell there won't be any benefit to having the non-hour indexes here (and the same for the other workloads), could you remove them?

kaywux (approved):
This LGTM mod a few small changes, thanks Davis! When you have a chance, could you link a summary of the results and the noise/robustness testing on this PR? Doesn't have to be anything fancy, here's a UnionWith one for reference: https://docs.google.com/spreadsheets/d/1Fq5U-wCjquGFTn3JDamr_8s4hahwgmggu7Wb-9nZ6Oo/edit?usp=sharing.

src/workloads/execution/DensifyHours.yml
davish: done

src/workloads/execution/DensifyHours.yml
davish: done

src/workloads/execution/DensifyHours.yml
```diff
@@ -0,0 +1,249 @@
...
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10000
+    BatchSize: &batchSize 30000 # Documents will be roughly 5 bytes in size, and we won't generate more than 30k documents, which is less than 1mb.
```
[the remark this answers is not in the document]
davish: Completely flubbed the math here -- thanks for the correction! I'll adjust the comment and also assumptions for the batchSize.

src/workloads/execution/DensifyHours.yml
```diff
@@ -0,0 +1,249 @@
...
+      milliseconds: {^RandomDate: {min: "2021-01-01T00:00:00.000Z", max: "2021-01-01T00:00:20.000Z"}} # Most basic date unit.
+      hours: {^RandomDate: {min: "2021-01-01", max: "2023-01-01"}} # Date unit with constant number of milliseconds.
+      months: {^RandomDate: {min: "2021-01-01", max: "4000-01-01"}} # Date unit with non-constant number of milliseconds.
+  - &Nop {Nop: true}
+  - *Nop
+  - *Nop
+
+# $densify stages always add a $sort on the field that is being densified, so this phase adds 
+# indexes on the date/number fields so that we aren't performing an in-memory sort.
+- Name: CreateIndex
```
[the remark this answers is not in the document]
davish: done.

davish:
Patch build: https://spruce.mongodb.com/version/61605ace57e85a5523d64a3d

PR #543, opened 2021-09-29
TIG-3324 Add documentation for query workloads

ksajme:


src/workloads/execution/PipelineUpdate.yml
```diff
@@ -1,7 +1,9 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
-Description: |
-  TODO: TIG-3324
+Description: >
+  This workload runs both classic updates and pipeline-based updates. It intends to record and
+  monitor the metrics of the two types of update operations on documents of different sizes and
```
mtandrei, line 5:
```suggestion
  monitors the metrics of the two update types on documents of different sizes and
```

mtandrei (changes requested):
Looks good! Just a couple of suggestions on wording and this should be good to go.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -1,7 +1,8 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
-Description: |
-  TODO: TIG-3324
+Description: >
+  This workload sequentially issues the same query in various combinations of projection and index
+  presence.
```
laplab, line 5:
```suggestion
  This workload measures the performance of queries with rich filters. First, we issue a bunch of plain queries with rich filter to get the performance baseline. After that, we issue the same query, but with projection. Finally, we create an index and issue the same query with projection, expecting it will use this index.
```

PR #547, opened 2021-10-05
TIG-3142: Improve Genny Loader index build efficiency

mridul-augustine:


src/cast_core/src/MonotonicLoader.cpp
```diff
@@ -109,27 +109,42 @@ void genny::actor::MonotonicLoader::run() {
...
-                        auto indexOpCtx = _indexBuild.start();
-                        collection.create_index(std::move(indexKey));
-                        indexOpCtx.success();
+                        std::string index_name = "";
+                        for (auto field : indexKey.view()) {
+                            index_name = index_name + field.key().to_string();
+                        }
+                        in_array = in_array << builder::stream::open_document
+                                            << "key" << indexKey.view()
+                                            << "name" << index_name
```
aggrand, line 133: The docs say the name field is optional and the server just does a similar concatenation to create it. I'd rather we cut this out and simplify this code.

src/cast_core/src/MonotonicLoader.cpp
```diff
@@ -109,27 +109,42 @@ void genny::actor::MonotonicLoader::run() {
...
                     }
                 }
+                auto after_array = in_array << builder::stream::close_array;
+                document::value doc = after_array << builder::stream::finalize;
+                if (_indexReq) {
+                    BOOST_LOG_TRIVIAL(debug)
+                            << "Building index" << to_json(doc.view());
+                    auto indexOpCtx = _indexBuild.start();
+                    config->database.run_command(doc.view());
+                    indexOpCtx.success();
```
aggrand, line 144: Also a note that this will change the nature of this recording, since this op is now tracking a single parallelized index creation rather than many separate ones. Might be worth mentioning in #performance-tooling-users so that people know / can object if needed.

src/cast_core/test/MonotonicLoader_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+            Threads: 1
+            DocumentCount: 10000
+            BatchSize: 3000
+            Document: {
+                field1: {^RandomInt: {min: 0, max: 100}},
+                field2: {^RandomInt: {min: 0, max: 100}},
+            }
+            Indexes:
+            - keys: {field1: 1}
+            - keys: {field2: 1}
```
aggrand, line 61: Can we add a test case for keys with options and for multikey indexes?

aggrand (changes requested):
Great job on your first genny PR! Thanks for digging into the details of the bsoncxx library and for making these changes!

Just had a few nitpicks but the gist is great!

src/cast_core/src/MonotonicLoader.cpp
mridul-augustine: Yes will do that once I merge the changes

aggrand (approved):
LGTM, thanks for this change!

mridul-augustine:
evergreen merge

PR #548, opened 2021-10-06
PERF-2547 Increase sleep time for balancer to finish balancing chunks in WouldChangeOwningShard workload

jz1242:
Increasing the timeout to avoid overlapping range deletion task.


jz1242:
evergreen merge

PR #549, opened 2021-10-08
Clarify install dir for genny

mvankeulen94:
I was installing genny on my workstation, and I got an error that the install dir didn't exist. /data/mci isn't a standard directory available on the workstation, so I got confused. I've added a note to our setup docs to clarify.


aggrand (approved):
Good catch!

PR #550, opened 2021-10-11
<NOMERGE> Perf 2078 state machine implementation

dalyd:
FSM Crud implementation. It's getting close. 
Notes: 

1. It mostly works
2. Setting a duration for the phase, the sleepBefores aren't ending. I may need to fiddle there, or just add a separate actor to hold the phase open and make the example non-blocking. 
3. I've played fast and loose with some parts, including 
     1. exposing intGenerator and doubleGenerator to everywhere
     2. Making most of documentGenerator namespace genny, because when I tried to do the minimal expansion it wouldn't compile. 
4. I completely don't grok when the use snake case, upper case, CamelCase, etc. Correct me and I'll do better.

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -32,6 +32,7 @@
 #include <gennylib/MongoException.hpp>
 #include <gennylib/context.hpp>
 #include <gennylib/conventions.hpp>
+#include <value_generators/DefaultRandom.hpp>
 
 using BsonView = bsoncxx::document::view;
 using CrudActor = genny::actor::CrudActor;
@@ -1132,6 +1133,17 @@ std::string getDbName(const PhaseContext& phaseContext) {
     return phaseDb ? *phaseDb : *actorDb;
 }
 
+// Represents one state in an FSM.
+struct State {
+    std::vector<std::unique_ptr<BaseOperation>> operations;
+    metrics::Operation metrics;
+    // Transitions are just next states with probabilities.
+    // There is an entry for every state, regardless of whether there is a transition.
+    // If we want to support multiple transitions to the same state (for different ops), then this
+    // will need to be re-written.
+    std::vector<double> transition_weights;
+};
+
 }  // namespace
 
 namespace genny::actor {
@@ -1164,6 +1176,8 @@ struct CrudActor::CollectionName {
 
 struct CrudActor::PhaseConfig {
     std::vector<std::unique_ptr<BaseOperation>> operations;
+    std::vector<std::unique_ptr<State>> states;
+    std::vector<double> initial_state_weights;
     metrics::Operation metrics;
     std::string dbName;
     CrudActor::CollectionName collectionName;
@@ -1206,23 +1220,55 @@ struct CrudActor::PhaseConfig {
                                    id);
         };
 
+        auto addState = [&](const Node& node) -> std::unique_ptr<State> {
+            BOOST_THROW_EXCEPTION(std::logic_error("Not implemented"));
+        };
         // Check if we have Operatopms or States. Through an error if we have both.
-        assert !(phaseContext["Operations"] and phaseContext["states"]) {}
-        operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
-            "Operation", "Operations", addOperation);
+        if (phaseContext["Operations"] and phaseContext["States"]) {
+            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                "CrudActor has Operations and States at the same time."));
+        }
+        if (phaseContext["Operations"]) {
+            operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
+                "Operation", "Operations", addOperation);
+        } else if (phaseContext["States"]) {  // Parse out the states}
+            states = phaseContext.getPlural<std::unique_ptr<State>>("State", "States", addState);
+        } else {  // Throw a useful error}
+            BOOST_THROW_EXCEPTION(
+                InvalidConfigurationException("CrudActor has neither Operations nor States "
+                                              "specified. Exactly one must be defined."));
+        }
     }
 };
 
 void CrudActor::run() {
+    int current_state = 0;
     for (auto&& config : _loop) {
         auto session = _client->start_session();
+        // TODO: If running without states, define a single state with the operations
+        if (!config->states.empty()) {
+            // pick the initial state for the phase
+            auto initial_distribution =
+                boost::random::discrete_distribution(config->initial_state_weights);
+            current_state = initial_distribution(_rng);
+        }
         for (const auto&& _ : config) {
             auto metricsContext = config->metrics.start();
 
-            for (auto&& op : config->operations) {
-                op->run(session);
+            if (!config->states.empty()) {
+                // pick next state
+                auto distribution = boost::random::discrete_distribution(
+                    config->states[current_state]->transition_weights);
+                current_state = distribution(_rng);
+                // run the operations for the next state.
+                for (auto&& op : config->states[current_state]->operations) {
+                    op->run(session);
+                }
+            } else {
+                for (auto&& op : config->operations) {
+                    op->run(session);
+                }
             }
-
             metricsContext.success();
         }
     }
@@ -1232,7 +1278,8 @@ CrudActor::CrudActor(genny::ActorContext& context)
     : Actor(context),
       _client{std::move(
           context.client(context.get("ClientName").maybe<std::string>().value_or("Default")))},
-      _loop{context, _client, CrudActor::id()} {}
+      _loop{context, _client, CrudActor::id()},
+      _rng{context.workload().getRNGForThread(CrudActor::id())} {}
 
 namespace {
 auto registerCrudActor = Cast::registerDefault<CrudActor>();
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1256,11 +1256,11 @@ struct CrudActor::PhaseConfig {
                 std::move(stateOperations), stateName, std::move(weights), std::move(next_states)});
         };
         // Check if we have Operations or States. Through an error if we have both.
-        if (phaseContext["Operations"] and phaseContext["States"]) {
+        if (phaseContext["Operations"] && phaseContext["States"]) {
             BOOST_THROW_EXCEPTION(InvalidConfigurationException(
                 "CrudActor has Operations and States at the same time."));
         }
-        if (phaseContext["Operations"]) {
+        if (phaseContext["Operations"] || phaseContext["Operation"]) {
             operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
                 "Operation", "Operations", addOperation);
         } else if (phaseContext["States"]) {  // Parse out the states}
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1135,13 +1135,19 @@ std::string getDbName(const PhaseContext& phaseContext) {
 
 // Represents one state in an FSM.
 struct State {
+
+public:
     std::vector<std::unique_ptr<BaseOperation>> operations;
-    metrics::Operation metrics;
+    //    metrics::Operation metrics;
     // Transitions are just next states with probabilities.
     // There is an entry for every state, regardless of whether there is a transition.
     // If we want to support multiple transitions to the same state (for different ops), then this
     // will need to be re-written.
+    std::string state_name;
     std::vector<double> transition_weights;
+    std::vector<int> transition_next_states;
+    // Combine the transition delay into a tuple with the next state in the vector above
+    // std::vector<> transition_delay; // this should be a time spec of some kind.
 };
 
 }  // namespace
@@ -1181,11 +1187,15 @@ struct CrudActor::PhaseConfig {
     metrics::Operation metrics;
     std::string dbName;
     CrudActor::CollectionName collectionName;
+    bool continue_current_state;
+
+    // min time of next operation
 
     PhaseConfig(PhaseContext& phaseContext, mongocxx::pool::entry& client, ActorId id)
         : dbName{getDbName(phaseContext)},
           metrics{phaseContext.actor().operation("Crud", id)},
-          collectionName{phaseContext} {
+          collectionName{phaseContext},
+          continue_current_state{false} {
         auto name = collectionName.generateName(id);
         auto addOperation = [&](const Node& node) -> std::unique_ptr<BaseOperation> {
             auto collection = (*client)[dbName][name];
@@ -1209,7 +1219,6 @@ struct CrudActor::PhaseConfig {
             // Node is convertible to bool but only explicitly so need to do the odd-looking
             // `? true : false` thing.
             const bool perPhaseMetrics = phaseContext["MetricsName"] ? true : false;
-
             auto createOperation = op->second;
             return createOperation(yamlCommand,
                                    onSession,
@@ -1220,8 +1229,24 @@ struct CrudActor::PhaseConfig {
                                    id);
         };
 
-        auto addState = [&](const Node& node) -> std::unique_ptr<State> {
-            BOOST_THROW_EXCEPTION(std::logic_error("Not implemented"));
+        auto addState =
+            [&](const Node& node,
+                const std::unordered_map<std::string, int> states) -> std::unique_ptr<State> {
+            //           State myState;
+            auto stateName = node["Name"].to<std::string>();
+            // Skipping repeat for now
+            // operations
+            auto stateOperations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
+                "Operation", "Operations", addOperation);
+            // Transitions
+            std::vector<double> weights;
+            std::vector<int> next_states;
+            for (auto [k, transitionYaml] : node["Transitions"]) {
+                weights.emplace_back(transitionYaml["Weight"].to<int>());
+                next_states.emplace_back(states.at(transitionYaml["To"].to<std::string>()));
+            }
+            return std::unique_ptr<State>(new State{
+                std::move(operations), stateName, std::move(weights), std::move(next_states)});
         };
         // Check if we have Operatopms or States. Through an error if we have both.
         if (phaseContext["Operations"] and phaseContext["States"]) {
@@ -1232,7 +1257,17 @@ struct CrudActor::PhaseConfig {
             operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
                 "Operation", "Operations", addOperation);
         } else if (phaseContext["States"]) {  // Parse out the states}
-            states = phaseContext.getPlural<std::unique_ptr<State>>("State", "States", addState);
+            // build the list of states, and then actuall process the states
+            int i = 0;
+            std::unordered_map<std::string, int> stateNames;
+            for (auto [k, state] : phaseContext["states"]) {
+                stateNames.at(state["name"].to<std::string>()) = i;
+                i++;
+            }
+            for (auto [k, state] : phaseContext["states"]) {
+                states.emplace_back(addState(state, stateNames));
+            }
+            // states = phaseContext.getPlural<std::unique_ptr<State>>("State", "States", addState);
         } else {  // Throw a useful error}
             BOOST_THROW_EXCEPTION(
                 InvalidConfigurationException("CrudActor has neither Operations nor States "
@@ -1243,27 +1278,37 @@ struct CrudActor::PhaseConfig {
 
 void CrudActor::run() {
     int current_state = 0;
+    int next_state = 0;
     for (auto&& config : _loop) {
         auto session = _client->start_session();
         // TODO: If running without states, define a single state with the operations
-        if (!config->states.empty()) {
+        if (!config->states.empty() and !config->continue_current_state) {
             // pick the initial state for the phase
             auto initial_distribution =
                 boost::random::discrete_distribution(config->initial_state_weights);
-            current_state = initial_distribution(_rng);
+            next_state = initial_distribution(_rng);
         }
         for (const auto&& _ : config) {
             auto metricsContext = config->metrics.start();
 
             if (!config->states.empty()) {
-                // pick next state
-                auto distribution = boost::random::discrete_distribution(
-                    config->states[current_state]->transition_weights);
-                current_state = distribution(_rng);
+                // Simplifying Assumption -- one shot operations on state enter
+                // We are here to fire those operations, and then pick the next state.
+                // next todo: add in the delay after picking the next state
+                // next next todo: Add support for repeated operations in the state
+
+                // transition to the next state
+                current_state = next_state;
                 // run the operations for the next state.
                 for (auto&& op : config->states[current_state]->operations) {
                     op->run(session);
                 }
+                // pick next state
+                auto distribution = boost::random::discrete_distribution(
+                    config->states[current_state]->transition_weights);
+                next_state =
+                    config->states[current_state]->transition_next_states[distribution(_rng)];
+                // pick the delay for the next state
             } else {
                 for (auto&& op : config->operations) {
                     op->run(session);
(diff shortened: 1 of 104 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,87 @@ std::string getDbName(const PhaseContext& phaseContext) {
     return phaseDb ? *phaseDb : *actorDb;
 }
 
+enum Units { NANOSECOND, MICROSECOND, MILLISECOND, SECOND, MINUTE, HOUR };
+// represent a random delay
+class Delay {
```
dalyd, line 1140: I nearly made this a ValueGenerator in DocumentGenerators, but I wasn't sure it made sense for generating bson, so I just put it here.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,87 @@ std::string getDbName(const PhaseContext& phaseContext) {
...
+public:
+    Delay(const Node& node, GeneratorArgs args)
+        : number_generator{doubleGenerator(node["^TimeSpec"]["value"], args)} {
+        auto unit_string = node["^TimeSpec"]["units"].maybe<std::string>().value_or("seconds");
+
+        // Use string::find here so plurals get parsed correctly.
+        if (unit_string.find("nanosecond") == 0) {
```
dalyd, line 1147: I suspect the real answer here is to make a class hierarchy, with a class per type. Or better yet a template, and add a factor function to produce them. I'm not sure I care enough to do that.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1295,129 @@ struct CrudActor::PhaseConfig {
...
+            next_state = initial_distribution(_rng);
+            BOOST_LOG_TRIVIAL(debug) << "Picking initial state " << next_state;
+        }
         for (const auto&& _ : config) {
             auto metricsContext = config->metrics.start();
-
-            for (auto&& op : config->operations) {
-                op->run(session);
+            BOOST_LOG_TRIVIAL(debug) << "In inner loop";
+            if (!config->states.empty()) {
```
dalyd, line 1392: It probably makes sense to create 1 state with a self loop when using operations instead of states. That would allow removing this if then else.

src/value_generators/include/value_generators/DocumentGenerator.hpp
```diff
@@ -57,6 +62,35 @@ struct GeneratorArgs {
...
+    }
+};
+
+template <class T>
+using UniqueGenerator = std::unique_ptr<Generator<T>>;
+const static boost::posix_time::ptime epoch{boost::gregorian::date(1970, 1, 1)};
+
+
+UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs);
+UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generatorArgs);
```
dalyd, line 93: I only needed to expose this line, but I think I need all of the stuff above it in order to expose. This is moved out from DocumentGenerator.cpp

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -34,34 +34,12 @@
...
 
-namespace {
-
-class Appendable {
-public:
-    virtual ~Appendable() = default;
-    virtual void append(const std::string& key, bsoncxx::builder::basic::document& builder) = 0;
-    virtual void append(bsoncxx::builder::basic::array& builder) = 0;
-};
+namespace genny {
```
dalyd, line 37: Namespaces are hard

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -652,18 +620,17 @@ class FormatStringGenerator : public Generator<std::string> {
         : _rng{generatorArgs.rng}, _format{node["format"].maybe<std::string>().value_or("")} {
         std::stringstream msg;
         if (!node["format"]) {
-            msg << "Malformed FormatString: format missing '" << _format << "'" << node
-                << "\n";
+            msg << "Malformed FormatString: format missing '" << _format << "'" << node << "\n";
```
dalyd, line 623: I didn't not intentionally edit these. I have clang-format in my emacs and it runs on save.

src/workloads/docs/CrudActorFSM.yml
```diff
@@ -153,5 +153,5 @@ Actors:
 
     # In this example we always start in the off State
     InitialStates:
-    - "Off":
-        Weight: 1
+    - State: "Off"
```
dalyd, line 156: Update to the syntax example. Encoding values in keys is hard, so I moved the state name into a proper field.

src/workloads/docs/CrudFSMTrivial.yml
```diff
@@ -0,0 +1,84 @@
+SchemaVersion: 2018-07-01
```
dalyd, line 1: This example runs!

src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,86 @@ std::string getDbName(const PhaseContext& phaseContext) {
     return phaseDb ? *phaseDb : *actorDb;
 }
 
+enum Units { NANOSECOND, MICROSECOND, MILLISECOND, SECOND, MINUTE, HOUR };
```
rtimmons, line 1138: 1. `enum class`
2. `kNanoseconds`

Could maybe also use `std::chrono` unit types, no?

src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,86 @@ std::string getDbName(const PhaseContext& phaseContext) {
...
+// represent a random delay
+class Delay {
+public:
+    Delay(const Node& node, GeneratorArgs args)
+        : number_generator{doubleGenerator(node["^TimeSpec"]["value"], args)} {
+        auto unit_string = node["^TimeSpec"]["units"].maybe<std::string>().value_or("seconds");
```
rtimmons, line 1144: Genny is effectively always `camelCase`. The `snake_case` comes in only from the stl (e.g. `value_or` is a function on `std::optional`).

src/cast_core/src/CrudActor.cpp
rtimmons: > Or better yet a template, and add a factor function to produce them. I'm not sure I care enough to do that.

Ya this is entirely in a `.cpp` so 🤷🏽, but this impl is straightforward enough and not really evaluated in a loop so clarity is probably better than raw speed and cleverness.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,86 @@ std::string getDbName(const PhaseContext& phaseContext) {
...
+};
+
+struct Transition {
+    int next_state;
+    Delay delay;
+};
+
+// Represents one state in an FSM.
+struct State {
+public:
```
rtimmons, line 1208: 💅🏽 in general prefer `class` to `struct` but that's no biggie. Either way, no need to put `public` in `struct`s 😸

src/cast_core/src/CrudActor.cpp
```diff
@@ -1132,6 +1135,86 @@ std::string getDbName(const PhaseContext& phaseContext) {
...
+    std::vector<std::unique_ptr<BaseOperation>> operations;
+    //    metrics::Operation metrics;
+    // Transitions are just next states with probabilities.
+    // There is an entry for every state, regardless of whether there is a transition.
+    std::string state_name;
+    std::vector<double> transition_weights;
+    std::vector<Transition> transitions;
```
rtimmons, line 1215: Does this want to be a `std::vector<std::pair<double, Transition>>` to allow the transitions and weights to be stored together? Avoid relying on same-indexing across containers.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1164,14 +1247,20 @@ struct CrudActor::CollectionName {
 
 struct CrudActor::PhaseConfig {
     std::vector<std::unique_ptr<BaseOperation>> operations;
+    std::vector<std::unique_ptr<State>> states;
+    std::vector<double> initial_state_weights;
     metrics::Operation metrics;
     std::string dbName;
     CrudActor::CollectionName collectionName;
+    bool continue_current_state;
```
rtimmons, line 1255: Consider a `StateConfig` "sub-struct" that holds all the states/weights info together distinct from the `PhaseConfig`.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1164,14 +1247,20 @@ struct CrudActor::CollectionName {
...
+
+    // min time of next operation
 
     PhaseConfig(PhaseContext& phaseContext, mongocxx::pool::entry& client, ActorId id)
         : dbName{getDbName(phaseContext)},
           metrics{phaseContext.actor().operation("Crud", id)},
-          collectionName{phaseContext} {
+          collectionName{phaseContext},
+          continue_current_state{phaseContext["Continue"].maybe<bool>().value_or(false)} {
```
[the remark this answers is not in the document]
rtimmons: The `PhaseContext` has access to the parent `ActorContext` which can then iterate over all the `Phase`s. That would be a good thing to add before this is used in production since new users are likely to mis-configure things the first time using it.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1294,129 @@ struct CrudActor::PhaseConfig {
...
+            std::vector<double> weights;
+            std::vector<Transition> transitions;
+            for (auto [k, transitionYaml] : node["Transitions"]) {
+                weights.emplace_back(transitionYaml["Weight"].to<double>());
+                BOOST_LOG_TRIVIAL(debug)
+                    << "Next state name is " << transitionYaml["To"].to<std::string>();
+                transitions.emplace_back(Transition{
+                    states.at(transitionYaml["To"].to<std::string>()),
+                    Delay(transitionYaml["SleepBefore"], GeneratorArgs{phaseContext.rng(id), id})});
+            }
```
rtimmons, line 1320: echoing a previous comment - tying the weights and the transitions to the same struct/pair would make this slightly easier since you could put that logic into a ctor:

```
for (const auto&& [k, xion] : node["Transitions"]) {
  transitions.emplace_back<WeightedTransition>(xion, states, phaseContext, id);
}
```

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1294,129 @@ struct CrudActor::PhaseConfig {
...
+        } else if (phaseContext["States"]) {  // Parse out the states}
+            if (!phaseContext["InitialStates"]) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                    "CrudActor has States, but has not specified InitialStates"));
+            }
+            BOOST_LOG_TRIVIAL(debug) << "Adding states";
+            // build the list of states, and then actuall process the states
+            int i = 0;
+            std::unordered_map<std::string, int> stateNames;
+            for (auto [k, state] : phaseContext["States"]) {
```
rtimmons, line 1341: consider a `States` class that encapsulates the name/id lookup and holds the `addState` results.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1294,129 @@ struct CrudActor::PhaseConfig {
...
+        // TODO: If running without states, define a single state with the operations
+        BOOST_LOG_TRIVIAL(debug) << "In CrudActor::run" << config->states.empty()
+                                 << config->continue_current_state;
+        if (!config->states.empty() and !config->continue_current_state) {
+            // pick the initial state for the phase
+            auto initial_distribution =
+                boost::random::discrete_distribution(config->initial_state_weights);
+            next_state = initial_distribution(_rng);
+            BOOST_LOG_TRIVIAL(debug) << "Picking initial state " << next_state;
+        }
```
rtimmons, line 1387: consider putting this logic into the `PhaseConfig` itself which will then potentially let you pull out the `StateManager` or whatever as a sub-member as mentioned in an earlier comment.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1294,129 @@ struct CrudActor::PhaseConfig {
...
         auto session = _client->start_session();
+        // TODO: If running without states, define a single state with the operations
+        BOOST_LOG_TRIVIAL(debug) << "In CrudActor::run" << config->states.empty()
+                                 << config->continue_current_state;
+        if (!config->states.empty() and !config->continue_current_state) {
+            // pick the initial state for the phase
+            auto initial_distribution =
+                boost::random::discrete_distribution(config->initial_state_weights);
+            next_state = initial_distribution(_rng);
+            BOOST_LOG_TRIVIAL(debug) << "Picking initial state " << next_state;
```
rtimmons, line 1386: Good to remove `debug` logging, but some level of logging around the initial state and transitions would be helpful. We log the transition between Phases iirc so following a similar pattern here would probably be helpful, but depending on how often transitions happen it could be very noisy. Your call here.

Either way, metrics on state transitions would be helpful.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -433,6 +433,14 @@ class ActorPhase final {
         return _currentPhase;
     }
 
+    /**
+     * Sleep for the duration, also waking up if the phase changes before the
```
rtimmons, line 437: Sleep for at most `duration` ?

src/value_generators/include/value_generators/DocumentGenerator.hpp
```diff
@@ -57,6 +62,35 @@ struct GeneratorArgs {
...
+public:
+    virtual ~Appendable() = default;
+    virtual void append(const std::string& key, bsoncxx::builder::basic::document& builder) = 0;
+    virtual void append(bsoncxx::builder::basic::array& builder) = 0;
+};
+
+using UniqueAppendable = std::unique_ptr<Appendable>;
+
+template <class T>
+class Generator : public Appendable {
```
rtimmons, line 75: do you actually need the def here or just to forward-declare it? I think you can get away with just the decl since you operate by ptr.

src/value_generators/src/DocumentGenerator.cpp
rtimmons: ya..if this has to be in `genny`, consider `genny::value_generators` to keep it a bit more tidy. I'm afraid of `genny::` becoming a `std::` behemoth.

src/value_generators/include/value_generators/DocumentGenerator.hpp
rtimmons: I'm afraid of having some decls here and others just in the cpp without any real rationale for what lives where beyond "shrug we needed this to compile."

I say this without a real suggestion for how to fix it / I would need to fight the compiler myself on this. I think leave this as a TODO or a comment here and in the cpp noting that there's not yet any strong intuition about what's forward-declared here versus compilation-unit-defined in the cpp.

src/value_generators/include/value_generators/DocumentGenerator.hpp
dalyd: Yeah, might be good to sit with you on this one. At the end of the day I just need to be able to forward declare
UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generatorArgs);
What is the minimum thing for that? 
UniqueGenerator is introduced by a using. 
Could I do

template <class t> class Generator;
using UniqueGenerator = std::unique_ptr<Generator<T>>;
UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs);
UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generatorArgs);

Or maybe 
class Appendable;
template <class t> class Generator : public Appendable;
using UniqueGenerator = std::unique_ptr<Generator<T>>;
UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs);
UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generatorArgs);

I supposed we could also pimpl it for exposing the intGenerator. That might be cleanest. What do you think?

src/value_generators/include/value_generators/DocumentGenerator.hpp
dalyd: Would a pimpl around these two things make all the other crap go away?

dalyd (commented):
Thanks for the feedback @rtimmons. I made some of the changes and have some follow-up questions for you. 

I have some tests for the FSM stuff in the Actor already. I'm not sure how to handle testing the delay stuff without adding a lot of code. I don't think other actors have anything like that. 

I'll add for the PhaseLoop changes, since I can see the proper ways to do those.

dalyd:
Also, @rtimmons big update to the sleeping functionality. I moved it into IterationChecker and I think it is fundamentally better/more correct now, and the phase ends after duration, as I expect it to. 

That addresses

> Setting a duration for the phase, the sleepBefores aren't ending. I may need to fiddle there, or just add a separate actor to hold the phase open and make the example non-blocking.

src/gennylib/include/gennylib/PhaseLoop.hpp
```diff
@@ -182,6 +183,36 @@ class IterationChecker final {
...
     }
 
+    // Set the minimum timepoint before continuning again
+    constexpr void setSleepUntil(SteadyClock::time_point SleepUntil) {
+        _sleepUntil = SleepUntil;
+    }
+
+    // Perform the sleep to get to specified timepoint, but ending phase if timepoint would be after
+    // the end of the phase.
+    void sleepForActor(Orchestrator& o,
```
dalyd, line 193: This grew some and it's worth having eyes on it to make sure it's correct.

src/gennylib/test/PhaseLoop_test.cpp
```diff
@@ -119,6 +119,62 @@ TEST_CASE("Correctness for N milliseconds") {
...
+            loop.sleepNonBlocking(chrono::milliseconds(1));
+        }
+        INFO("After loop");
+        INFO(i);
+        auto elapsed =
+            chrono::duration_cast<chrono::milliseconds>(chrono::system_clock::now() - start)
+                .count();
+
+        REQUIRE(i >= 9);
+        // REQUIRE(i <= 11);
```
dalyd, line 149: This require is failing and I'm completely flummoxed. the thing is running a few hundred iterations. I tried to add some fudge factors wondering if I had a race, but it didn't help. It does run for the right amount of time though.

src/cast_core/src/CrudActor.cpp
dalyd: It might make sense for this to be a TypeGenerator<Duration> and move to Generators. I don't think it makes sense to be Generator<>

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1296,125 @@ struct CrudActor::PhaseConfig {
...
+                    InvalidConfigurationException("Continue option not valid if not using States"));
+            }
+            operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
+                "Operation", "Operations", addOperation);
+        } else if (phaseContext["States"]) {  // Parse out the states}
+            // build the list of states, and then actuall process the states
+            int i = 0;
+            std::unordered_map<std::string, int> stateNames;
+            for (auto [k, state] : phaseContext["States"]) {
+                stateNames.emplace(state["Name"].to<std::string>(), i);
```
jimoleary, line 1337: Maybe this is where the state name validation needs to go. must be present, non-blank, unique etc ...

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1296,125 @@ struct CrudActor::PhaseConfig {
...
+                boost::random::discrete_distribution(config->stateConfig.initialStateWeights);
+            nextState = initial_distribution(_rng);
+            BOOST_LOG_TRIVIAL(debug) << "Actor " << id() << " picking initial state " << nextState;
+        }
         for (const auto&& _ : config) {
             auto metricsContext = config->metrics.start();
-
-            for (auto&& op : config->operations) {
-                op->run(session);
+            if (!config->stateConfig.states.empty()) {
```
jimoleary, line 1391: This seems like a big and possibly difficult to maintain change. 

At the very least *advanceFSM* would help hide it away (and maybe most of the FSM code ends up in stateConfig).

src/cast_core/src/CrudActor.cpp
```diff
@@ -1230,7 +1424,8 @@ CrudActor::CrudActor(genny::ActorContext& context)
     : Actor(context),
       _client{std::move(
           context.client(context.get("ClientName").maybe<std::string>().value_or("Default")))},
-      _loop{context, _client, CrudActor::id()} {}
+      _loop{context, _client, CrudActor::id()},
+      _rng{context.workload().getRNGForThread(CrudActor::id())} {}
```
jimoleary, line 1428: I'm kind of surprised that we need to have a _rng in crudactor now, what do the other implementations currently do (maybe the _rng should be in the stateConfig too).

src/lamplib/src/genny/tasks/dry_run.py
```diff
@@ -11,7 +11,7 @@
 def dry_run_workload(
     yaml_file_path: str, is_darwin: bool, genny_repo_root: str, workspace_root: str
 ):
-    if os.path.basename(yaml_file_path) in ["CrudActorFSM.yml", "CrudActorFSMAdvanced.yml"]:
+    if os.path.basename(yaml_file_path) in ["CrudActorFSMAdvanced.yml"]:
```
jimoleary, line 14: could we make this a constant with a semantic name
```python
WIP_YML_FILES = ["CrudActorFSMAdvanced.yml"]
""" A list of files to ignore. They are in WIP, not relevant etc...""" 
```

src/cast_core/src/CrudActor.cpp
dalyd: This could be viewed as a TypeGenerator<Duration>. where Duration is using Duration = typename TimeSpec::ValueT;  It would be a little bit of work, but I could probably make that a thing if desired. 

So, it is using TimeSpec.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1161,17 +1248,21 @@ struct CrudActor::CollectionName {
...
 };
-
 struct CrudActor::PhaseConfig {
     std::vector<std::unique_ptr<BaseOperation>> operations;
     metrics::Operation metrics;
+    StateConfig stateConfig;
     std::string dbName;
     CrudActor::CollectionName collectionName;
 
+    // min time of next operation
```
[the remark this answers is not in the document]
dalyd: Nope.  Deleted.

src/cast_core/src/CrudActor.cpp
```diff
@@ -1206,21 +1296,125 @@ struct CrudActor::PhaseConfig {
                                    id);
         };
 
-        operations = phaseContext.getPlural<std::unique_ptr<BaseOperation>>(
-            "Operation", "Operations", addOperation);
+        auto addState =
+            [&](const Node& node,
+                const std::unordered_map<std::string, int> states) -> std::unique_ptr<State> {
+            //           State myState;
+            auto stateName = node["Name"].to<std::string>();
```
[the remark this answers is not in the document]
dalyd: It  will  throw an error, but I don't know what kind I can add an assert that node["Name"] exists.

src/cast_core/src/CrudActor.cpp
dalyd: If I pull the addStates out into a proper function, I also have to pull the addOperations out (since addStates calls addOperations). I started to do that and realize I have another problem. addOperation is called in this line of code

```
            auto stateOperations = node.getPlural<std::unique_ptr<BaseOperation>>(
                "Operation", "Operations", addOperation);

```

I don't know how to pass the added arguments through that. The lambda function is allowing us to pass things through the environment. I can't do that with a proper function. If @rtimmons has an answer for that problem I'll put out addOperations and addStates into separate functions in the anonymous namespace above, replacing the two lambda functions.

rtimmons:
David and I paired on this and made some good 💅🏽 - made my own branch because 🤮
https://github.com/mongodb/genny/compare/master...rtimmons:pr550-rtimmons?expand=1
I'll spend another hour or so early next week to finish that and then lob it back over the fence for final 👀 and 👍🏽

PR #552, opened 2021-10-14
TIG-3351 Heap allocate threads

aggrand:


aggrand:
@vrachev in my patch the runtime seems within the historic range of variation of runtimes for this task.

PR #554, opened 2021-10-19
TIG-3408: Consolidate lookup genny workloads now that feature flag is on by default

HanaPearlman:
Evergreen: https://spruce.mongodb.com/version/616e213932f417267e2c69bf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

As discussed, this change will mean we lose perf history because the file names and collection sizes are changing. We decided this was acceptable, since we were able to use the previous history during the project, which has now been completed.

I won't post all of the results, since it's a ton and for the most part numbers have stayed identical. There were some changes when compared to previous patches due to collection sizes being increased (the size of documents and number of documents have changed for some queries). Beyond this, I attempted not to change any of the queries.

The key facts (which we may want to compile and share) are backed up by the numbers in the evergreen patch linked above.

For $lookup, going from unsharded local and unsharded foreign to exactly one of the collections sharded (ShardedToUnsharded or UnshardedToSharded) results in no execution strategy change. Going from exactly one collection sharded to both collections sharded (ShardedToSharded) allows us to parallelize the $lookup and gives better perf results!

Here's one example:
RunLookups.UntargetedLookupUnshardedToSharded 6.8 sec
RunLookups.UntargetedLookupShardedToUnsharded 7.9 sec
RunLookups.UntargetedLookupShardedToSharded 3.8 sec

For $graphLookup, going from unsharded local to sharded local, regardless of the foreign collection, allows us to parallelize the $lookup and gives better perf results!

Here's one example:
RunGraphLookups.TargetedGraphLookupUnshardedToShardedDepthUnlimited 2.7 sec
RunGraphLookups.TargetedGraphLookupShardedToShardedDepthUnlimited 1.6 sec


src/workloads/execution/GraphLookup.yml
```diff
@@ -1,18 +1,25 @@
...
+  GraphLookupInUnshardedEnvironment workload, so the query can be tested on replica sets as well. We
+  may want to compare results across the two files, so the loading stage (and some queries) should
+  be kept as similar as possible.
 
   The workload consists of the following phases:
     1. Creating empty sharded collections distributed across all shards in the cluster.
     2. Populating collections with data.
     3. Fsync.
-    4. Running $graphLookup into a sharded collection.
+    4. Running $graphLookups.
```
kaywux, line 14: Could you add some more detail to the 4th phase? Maybe state at a high level the cases we're testing (differing collections sharded, recursion depths, how many matches in the foreign collection for each one in the local coll, etc).

src/workloads/execution/Lookup.yml
```diff
@@ -1,13 +1,22 @@
...
+  LookupInUnshardedEnvironment workload, so the query can be tested on replica sets as well. We may
+  want to compare results across the two files, so the loading stage (and some queries) should be
+  kept as similar as possible.
 
   The workload consists of the following phases:
     1. Creating empty sharded collections distributed across all shards in the cluster.
     2. Populating collections with data.
     3. Fsync.
-    4. Running targeted and untargeted $lookups against a sharded foreign collection.
+    4. Running targeted and untargeted $lookups.
```
kaywux, line 14: Same as above, not only targeted / untargeted, but also prefix cacheing.

src/workloads/execution/GraphLookup.yml
```diff
@@ -1,18 +1,25 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  This test exercises the behavior of $graphLookup against a sharded foreign collection.
+  This test exercises the behavior of $graphLookup where either the local or foreign collection is
+  sharded. When both collections are unsharded, the pipeline should be moved to the
```
kaywux, line 5: Should we still have an UnshardedToUnsharded case in this workload? Same thing for the lookup sharded workload as well. It would be interesting to see the difference in this vs the unsharded environment, and useful to know for users who are using $lookup / $graphLookup with both collections unsharded but in a sharded cluster environment.

kaywux (commented):
Thanks for taking this on Hana, there are lots of moving parts in this patch! It looks great, just one request for additional tests.

src/workloads/execution/GraphLookup.yml
HanaPearlman: Seems like the `*UnshardedEnvironment` test files are confusingly named, sorry -- they do run in an unsharded environment, but they also run in a sharded environment too (see the AutoRun param at the bottom of the files). So, we would not be increasing coverage if we added the UnshardedToUnsharded case to the GraphLookup and Lookup files.  I can rename those files in the next patch.

Does that address your point?

kaywux: Ahh, gotcha - sorry I missed that! In that case, I'd agree that the file names are slightly confusing haha. Could you rename them to `LookupWithOnlyUnshardedCollections` (or something less wordy) and also add a line about where each workload is running (sharded environment, replica set, etc) to the description of each benchmark?

HanaPearlman:
I think I've responded to all comments! Let me know if I missed anything. 

I mentioned before that this change will mean we lose perf history because the file names and collection sizes are changing. We decided this was acceptable, since we were able to use the previous history during the project, which has now been completed.

Originally, I was worried about Perf BFs, due to workloads changing slightly while operation metric names stayed the same. I think we'll avoid that though, since now all of the files have been renamed!

src/workloads/execution/GraphLookup.yml
```diff
@@ -1,18 +1,29 @@
...
+  GraphLookupInUnshardedEnvironment workload, so the query can be tested on replica sets as well. We
```
kaywux, line 6: Renamed, now GraphLookupWithOnlyUnshardedColls!

src/workloads/execution/Lookup.yml
```diff
@@ -1,13 +1,26 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  This test exercises the behavior of $lookup against a sharded foreign collection.
+  This test exercises the behavior of $lookup where either the local or foreign collection is
+  sharded. When both collections are unsharded, the pipeline should be moved to the
+  LookupInUnshardedEnvironment workload, so the query can be tested on replica sets as well. We may
```
kaywux, line 6: Same as above, rename file reference.

kaywux (approved):
This LGTM mod two minor comment edits, thanks for all the changes Hana!

HanaPearlman:
evergreen merge

HanaPearlman:
evergreen merge

PR #557, opened 2021-10-21
TIG-3152 Fix quiesce support for sharded clusters

vrachev:


src/gennylib/src/quiesce.cpp
```diff
@@ -94,6 +94,65 @@ bool waitOplog(v1::Topology& topology) {
...
+bool CheckForDroppedCollections(v1::Topology& topology, std::string dbName, SleepContext sleepContext) {
+    class CheckForDroppedCollectionsVisitor : public v1::TopologyVisitor {
+    public:
+        CheckForDroppedCollectionsVisitor(std::string dbName, SleepContext sleepContext)
+            : _dbName{dbName},
+              _sleepContext{sleepContext} {}
+
+        void onBeforeReplSet(const v1::ReplSetDescription& desc) override {
+            mongocxx::client client(mongocxx::uri(desc.primaryUri));
+            _successAcc = checkCollectionsTestDB(client);
```
vrachev, line 106: We need to call this on all of the shards, not just the mongos

src/gennylib/src/quiesce.cpp
```diff
@@ -94,6 +94,65 @@ bool waitOplog(v1::Topology& topology) {
...
+            return _successAcc;
+        }
+
+    private:
+        bool checkCollections(mongocxx::database db) {
+            using bsoncxx::builder::basic::kvp;
+            using bsoncxx::builder::basic::make_document;
+
+            auto res = db.run_command(make_document(kvp("serverStatus", 1)));
+            if (!res.view()["storageEngine"] || !res.view()["storageEngine"]["dropPendingIdents"]) {
```
vrachev, line 119: We call this method on all of the mongo servers (mongos, mongod, configsvr), but only the mongod's themselves have collections that needed to be dropped

src/gennylib/src/quiesce.cpp
```diff
@@ -36,7 +36,7 @@
 
 namespace genny {
 
-const int DROPPED_COLLECTION_RETRIES = 1000;
+const int DROPPED_COLLECTION_RETRIES = 20;
```
vrachev, line 39: I encountered a very strange behavior. With the previous value of 1000 and a sleep time of 1s, the quiesce actor test hangs. This happens even though running the QuiesceActor workload has no issues (they both do the same thing). 

When you lower the retry to 20 it ceases to hang. I've stared at this for a long time and cannot reason what is happening. 

Note that this behavior occurs even without my changes.

src/gennylib/src/quiesce.cpp
```diff
@@ -94,6 +94,65 @@ bool waitOplog(v1::Topology& topology) {
...
+        }
+
+        bool checkCollectionsTestDB(mongocxx::client& client) {
+            auto db = client.database(_dbName);
+            int retries = 0;
+            while (checkCollections(db) && retries < DROPPED_COLLECTION_RETRIES) {
+                BOOST_LOG_TRIVIAL(debug)
+                    << "Sleeping 5 second while waiting for collection to finish dropping.";
+                retries++;
+                _sleepContext.sleep_for(std::chrono::seconds(5));
```
vrachev, line 134: Changed to sleep for 5 seconds instead of 1 as it still works (with the 20 retries) and gives a 100 second limit, instead of 20s.

vrachev (commented):
RFAL @aggrand 

The effect on Quiesce is largely the same as before, but I did some topology refactoring per Raiden's suggestion. 

I added an extra ConfigSvrDescription class to give us independent control over the {mongoses, shards, configsvr} in a sharded cluster. 

I also split out the onMongod visit method into these 3 - ```[onStandaloneMongod, onReplSetMongod, onConfigSvrMongod]```. This allowed me to only create a single visitor class, instead of needing one for standalone topology and another for replica-set + sharded cluster topologies. 

Patch here: https://spruce.mongodb.com/version/618c1fb9306615332c470f96/tasks?page=0&variant=%5Elinux-standalone%24

I also tested locally with single, replica-set and sharded cluster topologies. The QuiesceActor test still hangs so I commented it out.

src/gennylib/src/quiesce.cpp
```diff
@@ -94,51 +93,93 @@ bool waitOplog(v1::Topology& topology) {
...
 }
 
+bool CheckForDroppedCollections(v1::Topology& topology, std::string dbName, SleepContext sleepContext) {
+    class CheckForDroppedCollectionsVisitor : public v1::TopologyVisitor {
+    public:
+        CheckForDroppedCollectionsVisitor(std::string dbName, SleepContext sleepContext)
+            : _dbName{dbName},
+              _sleepContext{sleepContext} {}
+
+        void onBeforeReplSet(const v1::ReplSetDescription& desc) override {
```
vrachev, line 103: Handles replset + sharded cluster cases

src/gennylib/src/quiesce.cpp
```diff
@@ -94,51 +93,93 @@ bool waitOplog(v1::Topology& topology) {
...
         }
+
+        void onStandaloneMongod(const v1::MongodDescription& desc) override {
+            onMongod(desc);
+        };
+        void onReplSetMongod(const v1::MongodDescription& desc) override {
+            onMongod(desc);
+        };
+        void onConfigSvrMongod(const v1::MongodDescription& desc) override {
+            onMongod(desc);
```
aggrand, line 176: (nit) Instead of calling `onMongod` so many times, consider a visitor method `onMongod` in topology.hpp that is called unconditionally for a mongod regardless of type.

aggrand (approved):
LGTM, thanks for adding those topology types! They'll definitely be useful going forward. Just had a small suggestion but nothing to block on.

src/gennylib/src/quiesce.cpp
vrachev: Good idea, will do

vrachev:
evergreen merge

PR #562, opened 2021-10-26
TIG-3350 genny startup is slow when there are large number of threads

aggrand:
Cuts the startup time from 5 mins to 28 seconds for the workload in the ticket, for me.

This parallelizes the construction of actor contexts, which has the effect of parallelizing the construction of pools. I felt a bit wasteful constructing then tossing away these threads, so I went down a rabbit hole of using the same thread for context construction and actor-running, but it turned into a pretty messy refactor without much gain so I ended up keeping it like this.

Had to toss a few more locks in, but they're only in startup code that used to be serial.


brooookemiller:
do you know why the benchmark tests are failing on this, @aggrand?

aggrand:
@brooookemiller It's failing because of https://jira.mongodb.org/browse/TIG-3236

vrachev:
Hey @aggrand, I don't think I have the deep c++/genny expertise required to weigh in on the implications/trade-offs of this change. @rtimmons would you mind taking another look at this and signing off on it? 

In the mean-time I'm gonna spend some time sitting down with the original genny design and a c++ book. I'll hit Raiden or Ryan up at some point when I have questions.

src/gennylib/src/context.cpp
```diff
@@ -133,8 +141,10 @@ DefaultRandom& WorkloadContext::getRNGForThread(ActorId id) {
     if (this->isDone()) {
         BOOST_THROW_EXCEPTION(std::logic_error("Cannot create RNGs after setup"));
     }
+
+    std::lock_guard<std::mutex> lk(_rngLock);
     if (auto rng = _rngRegistry.find(id); rng == _rngRegistry.end()) {
-        auto [it, success] = _rngRegistry.try_emplace(id, _rng());
+        auto [it, success] = _rngRegistry.try_emplace(id, _rng + id);
```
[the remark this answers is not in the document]
aggrand: Ended up using a hash (not visible here since it's treating this as outdated). @dalyd any thoughts on whether that's random enough, or if there's a different approach we should use?

src/gennylib/src/context.cpp
```diff
@@ -67,21 +70,26 @@ WorkloadContext::WorkloadContext(const Node& node,
...
-            _actors.push_back(std::move(actor));
-        }
-    }
+    std::mutex actorsLock;
+    parallelRun(cbegin(_actorContexts),
+                   cend(_actorContexts),
+                   [&](const auto& actorContext) {
+                       return std::thread{[&]() {
+                           for (auto&& actor : _constructActors(cast, actorContext)) {
+                               std::lock_guard<std::mutex> lk(actorsLock);
```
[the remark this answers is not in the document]
aggrand: Turned out that one of the actors had a handy `Atomic` wrapper I was able to extract and generalize for this.

aggrand:
Another couple notes: I think we can be _relatively_ safe here as far as concurrency issues go. I think we want to make sure to lock-guard everything accessed through the workload context inside the context function, only allow context-owned resources to be accessed by context member functions so that we can enforce thread safety (don't return things), and enforce that no two locks are held at once (to prevent deadlocking). 

If we stick to the above I don't think actor-authors will really have to think about the fact that construction is done in parallel, except for `getActorSharedState`, which is already threadsafe and the existing actors seem to already enforce thread safety for their state types.

src/gennylib/src/context.cpp
```diff
@@ -133,8 +140,13 @@ DefaultRandom& WorkloadContext::getRNGForThread(ActorId id) {
     if (this->isDone()) {
         BOOST_THROW_EXCEPTION(std::logic_error("Cannot create RNGs after setup"));
     }
+
+    bool success = false;
+
+    std::lock_guard<std::mutex> lk(_rngLock);
     if (auto rng = _rngRegistry.find(id); rng == _rngRegistry.end()) {
-        auto [it, success] = _rngRegistry.try_emplace(id, _rng());
+        auto [it, returnedSuccess] = _rngRegistry.try_emplace(id, _rngSeed + std::hash<long>{}(id));
```
dalyd, line 148: Copying forward the discussion
This doesn't jump as far as the previous version for each RNG, but my understanding is that for a decent modern RNG this should be fine?

 
Member
@rtimmons rtimmons 7 days ago
This is clever. My instinct says to do a std::hash or something of the id rather than just straight addition, but I'd actually bring in David Daly here - he did some pretty extensive investigations into proper RNGs with his initial WLG proto-genny concept a few years ago.

Either way, this will cause workload behavior to change since Actors will get different seeds after this change. I'm not sure of the implications there but it may require a bit of optics-management when this goes out.

 
Member
Author
@aggrand aggrand 14 minutes ago
Ended up using a hash (not visible here since it's treating this as outdated). @dalyd any thoughts on whether that's random enough, or if there's a different approach we should use?

Let me see if I understand the issue. I'm guessing because we are creating these in parallel we can't call _rng() because it is singe threaded, so we are trying to find a way to get nicely distributed seeds for the random number generators. Is that correct? 

Assuming it is a few thoughts. 
1. Could we just generate the random seeds and put them in an array before parallelizing? That would allow (possibly) doing rng creation in parallel, but still using the existing seeding code, and would lead to the same per thread behavior before and after this change. Otherwise we do get a change in workload behavior, although hopefully an insignificant one. 
2. I'm not sure how sound my original code was, that seeded each thread with a random number from another rng. I'm going to dig and poke a bit.

dalyd: After further review I feel better about the original code, although I wish I had passed more random state into the constructor (longer seed sequence)

aggrand: Yeah pretty much, and if we make `_rng()` naively thread-safe it would be nondeterministic which thread gets which seed.

I didn't want to do option 1 because we don't necessarily want to assume how many actors we have to start with (would rather leave that up to the `ActorProducer`s).

One option would be, if actor if id `n` requests an rng, we produce all the rngs up to `n` using the existing code, then anyone else requesting an rng takes that. I guess everyone would be waiting if the last actor manages to request first, but they also don't have to construct anything themselves so maybe that balances out? (Not that initial rng generation was large in the flame graphs anyway.)

@dalyd if a workload behavior change would be bad I can take that approach instead! I'd assumed it would be okay if we change it one-off but if we want to avoid it at all I can take the above-mentioned approach

src/driver/src/DefaultDriver.cpp
```diff
@@ -163,35 +164,26 @@ DefaultDriver::OutcomeCode doRunLogic(const DefaultDriver::ProgramOptions& optio
...
-
-                               std::lock_guard<std::mutex> lk{reporting};
-                               ctx.success();
-                           }
-                       }};
-                   });
-
-    for (auto& thread : *threadsPtr)
-        thread.join();
+    parallelRun(workloadContext.actors(),
```
dalyd, line 167: @aggrand it looks like here should have the total number of threads, or at least could compute it (workloadContext.actors().size()) could we put a for loop (or something fancier) of the form

```
std::vector<int> RNGSeeds;
for (auto actor : workloadContext.actors())
    for (int i =; i < actor.size(): i++)
        {
          RNGSeeds.push_back(workloadContext._rng());
        }
```
Or directly construct the rng registry right here. 

You can't directly access the _rng today, but that part seems easy enough to expose.

dalyd: If this was a bottleneck in itself, you could try cloning the rng, and use discard to advance the rng as needed to match the serial execution.

src/gennylib/src/context.cpp
dalyd: Yes, could we give that a try? Let's see if it adds up or not. And note only the seed generation part needs to be serialized.

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -125,27 +149,28 @@ class AtomicContainer {
...
-        //std::lock_guard<std::mutex> lock(_mutex);
+        std::lock_guard<std::mutex> lock(_mutex);
         return _container.begin();
     }
+    typename T::iterator end() {
+        std::lock_guard<std::mutex> lock(_mutex);
+        return _container.end();
+    }
+
     typename T::const_iterator begin() const {
```
rtimmons, line 160: Not sure if you're ready for much scrutiny here yet, but as I was scrolling by I noticed the locks here. I _think_ the lock may need to be held for the entire iteration (i.e. until `it == end()`).

I honestly don't know for sure here, but intuitively we don't want to allow new items to be added in one thread while another is iterating.

aggrand: Agreed that holding the iterator for the entire iteration would be necessary for thread safety. I'm wary about baking that into an iterator wrapper definition because the semantics might be odd. While ranged-for loops and many STL algorithms might work with an iterator that holds the lock until it's compared to `end()`, that would open the door to deadlocks if someone has multiple iterators in one thread and might be otherwise unintuitive.

Right now we're only doing iteration in single-threaded contexts, so I'd rather leave this thread-unsafe until needed. It seems intuitive that multi-operation iteration wouldn't be atomic the same way single operations are, imho? I added a comment above saying so.

That said, please push back if you disagree!

rtimmons: The kicker is here:

> I'd rather leave this thread-unsafe until needed

I agree! But! I fear that if we implicitly don't solve the problem (or shrug and add a ⚠️ comment somewhere) that somebody in the future (maybe us) doesn't read or understand or _accidentally_ causes multi-threaded access during iteration. 

My suggestion:

Create an iterator wrapper not super different from what PhaseLoop already does. Have it set an "in use" flag in the collection upon `.begin()` and make the collection explicitly unset it:

```
for (const auto&& it : items) {
  do_something(it);
}
items.notify_done()
```

Then, in any member-functions that want the mutex, assert that flag isn't set and barf spectacularly if it is.

This "soft"-prevents other threads from modifying the state during iteration in a defensive-programming way, not in a compiler-ensured way.

I'll echo your sentiment to please push back. Perhaps this is being too conservative. But I personally am super conservative with doing anything "fancy" (or not strictly-necessary given the features) with concurrency in genny simply because any kind of data-race, deadlock, or generally hard-to-explain behavior will destroy a lot of trust in the utility Genny provides.

aggrand: That's a fair point. I was probably being too casual about pushing the responsibility of safety to users of the class when a defensive approach would be preferred.

I updated the class to meet the `BasicLockable` requirement and enforced that the lock is being held explicitly when iterating.

src/gennylib/test/context_test.cpp
```diff
@@ -136,16 +136,19 @@ SchemaVersion: 2018-07-01
         std::atomic_int calls = 0;
         auto foobar = genny::testing::toDocumentBson("foo: bar");
 
+        auto fromDocListAssert = false;
         auto fromDocList = std::make_shared<OpProducer>([&](ActorContext& a) {
             for (const auto&& [k, doc] : a["docs"]) {
-                auto docgen = doc.to<DocumentGenerator>(a, 0);
-                REQUIRE(docgen().view() == foobar.view());
+                auto docgen = doc.to<DocumentGenerator>(a, 1);
+                fromDocListAssert = docgen().view() == foobar.view();
```
aggrand, line 143: I had to make this and several similar changes because Catch2 isn't thread-safe. I'd considered having a flag to make execution serial for tests, but figured it'd be best to exercise the parallelism.

aggrand:
@rtimmons I think ready for more scrutiny when you get the chance!

rtimmons (commented):
I looked through some of the `AtomicContainer` stuff and it looks right on its face, but there's a lot going on here and this is relatively high-risk code. As it is, this change isn't _huge_ but it's surgical enough to want to scrutinize every line in the critical portions of the C++. I want to do these kinds of changes slowly and deliberately.

Can I be a jerk and ask that this PR be split into (at least) 2?

1. Python and doc changes; make tests multi-thread-capable; AtomicDeque; maybe add the parallelRun thing with tests but otherwise unused; maybe the changes from value-type to ptr-type...
2. Perhaps apart from (3) is maybe doing the ActorId and/or reserve things in `context` since that's not strictly related to multi-threading, it just makes certain parts of the multi-threading code easier. Perhaps include `AtomicContainer` and its wide gamut of testing in here as well (or yet another PR if you're so inclined). Not sure if this makes senseas being distinct from (3).
3. Changes to context/orchestrator that uses some of the functionality from 1,2.

Perhaps move the first parts out into a preliminary review and once merged merge master into this pr so the diff is small and focused on (3) and the existing comments are retained.

I also think it's worth gathering some data about how much of an improvement this is and some experimentation to ensure it doesn't cause any meaningful changes to the behavior of hot workloads (e.g. InsertRemove). Is it worth adding `genny::metrics` measurements around any of this so we have a canary to assert around this behavior?

I'm also still interested to see a flame-graph before/after so we can be sure this is tackling the root issue for large-thread case, but I realize that's easier said than done.

Either way still color me skeptical that adding this complexity is worth the benefit of slightly faster startup but only for pathological high-thread workloads (which may not be the best way to generate that kind of load, per upcoming project PM-2430).

Please push back wherever you disagree. It's not my goal to rain on a parade.

aggrand:
I split out to two other PRs. I broke it up a bit differently since the unittests for the parallelization primitives referred to each other, but lmk if the breakdown wasn't as helpful as would be preferred!

This PR will then have the context / context test / actor id changes that actually make the difference. This afternoon I'll see about posting before/after flame graphs. That said, the chronological time difference from straight-up executions is very noticeable.

Regarding the value in this change, @jimoleary might be able to comment a bit more? I feel like we've hit more scalability issues with highly-threaded workloads recently, suggesting this use is becoming more common.

jimoleary:
@aggrand We have been stress testing the server recently, so these issues have fallen out of these sorts of workloads.

CDRIVER-4002 has probably also been hampering this work but it looks like this is about to close soon. The combination of more genny scalabilty and the cdriver ticket will give us more options for stress testing.

aggrand:
@rtimmons ready for another look! Generalized the bucket container for our use.

I attached a .svg flame graph to the ticket showing the flame graph after the change. It looks pretty similar in overall shape to @jimoleary's graph from before the change, though the speedup is definitely there. I wonder if we're getting more CPU usage, so we still get similar stack trace samples / similar graph?

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -78,17 +116,16 @@ class ExceptionBucket {
         }
 
     private:
-        std::deque<std::exception_ptr> _caughtExceptions;
+        std::vector<std::exception_ptr> _caughtExceptions;
```
rtimmons, line 119: is `exception_ptr` really the right type to be using here? I think it's a global singleton pointing to what's returned by `current_exception` so the underlying objects may have gone away. Perhaps this wants to be a `std::vector<std::string> _whatMessages` or similar?

src/gennylib/src/context.cpp
```diff
@@ -133,15 +143,18 @@ DefaultRandom& WorkloadContext::getRNGForThread(ActorId id) {
...
-            // But invariants don't hurt we only call this during setup
-            throw std::logic_error("Already have DefaultRandom for Actor " + std::to_string(id));
-        }
+
+    if (id < 1) {
+        BOOST_THROW_EXCEPTION(std::logic_error("ActorId must be 1 or greater."));
     }
-    return _rngRegistry[id];
+
+    // Create RNGs for all ids up to our own.
```
rtimmons, line 151: does it make sense to eagerly do this when requesting the actorIds? Keeps from having to have another lock. Not sure of all pros/cons of this, but if it works out, I think make a DefaultRandom a `protected` member of `genny::Actor` similar to its `_id` field.

src/gennylib/src/context.cpp
```diff
@@ -116,6 +124,8 @@ GlobalRateLimiter* WorkloadContext::getRateLimiter(const std::string& name, cons
         BOOST_THROW_EXCEPTION(
             std::logic_error("Cannot create rate-limiters after setup. Name tried: " + name));
     }
+
+    std::lock_guard<std::mutex> lk(_limiterLock);
```
rtimmons, line 128: not that it matters a ton, but you can put these few items in a new scope so the `limiterLock` gets released sooner. I think it's only needed around the `if`:

```
{
  lock_guard lk{_limiter}
  if (_count(name) == 0) { _emplace.. }
}
```

rtimmons (approved):
This was pared down to a pretty safe and straightforward change.

A couple 💅🏽 ideas. Biggest gotcha I see is `exception_ptr` but it's possible that's just me needing to RTFM.

Good job and thanks for your patience in getting this through.

src/gennylib/include/gennylib/parallel.hpp
aggrand: The exception_ptr is a shared-ptr-like type that promises that the exception it manages exists as long as it does. That has its own problems, of course, but I think should be fine as long as this exception bucket and the exception thrown from it don't hang around forever?

src/gennylib/src/context.cpp
```diff
@@ -67,22 +73,23 @@ WorkloadContext::WorkloadContext(const Node& node,
...
-    // Default value selected from random.org, by selecting 2 random numbers
-    // between 1 and 10^9 and concatenating.
-    _rng.seed((*this)["RandomSeed"].maybe<long>().value_or(269849313357703264));
-
-    for (auto& actorContext : _actorContexts) {
-        for (auto&& actor : _constructActors(cast, actorContext)) {
-            _actors.push_back(std::move(actor));
-        }
-    }
+    ActorBucket bucket;
```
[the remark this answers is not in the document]
aggrand: I don't think so, unfortunately. ActorBucket is meant to only be used in a context where threadsafe insertions are required, and making it a vector wrapper in non-parallel contexts is unsafe.

src/gennylib/src/context.cpp
aggrand: The rate-limiters are name-specific not thread-specific, so I think we need to lock while interacting with both the container and the limiters themselves further on.

src/gennylib/src/context.cpp
aggrand: That's a good idea! In an earlier iteration of this I didn't want to, due to the producers having all the knowledge of how many actors there will be, but pre-claiming the IDs changes that.

aggrand:
evergreen merge

PR #566, opened 2021-11-01
Generator Updates

dalyd:
Adds ^Inc to the int generator list (it was missing)

Adds a int to string wrapper generator.

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -460,20 +460,6 @@ protected:
     std::vector<int64_t> _weights;
 };
 
-// This is a wrapper generator. It wraps an int generator to generate strings.
-class StringIntGenerator : public Generator<std::string> {
-public:
-    StringIntGenerator(const Node& node, GeneratorArgs generatorArgs)
-        : gen{std::move(intGenerator(node, generatorArgs))} {}
-
-    std::string evaluate() {
-        return (std::to_string(gen->evaluate()));
-    }
-
-protected:
-    UniqueGenerator<int64_t> gen;
-};
-
 // This is a a more specific version of ChooseGenerator that produces strings. It is only used
 // within the JoinGenerator.
 class ChooseStringGenerator : public Generator<std::string> {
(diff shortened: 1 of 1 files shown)


dalyd:
I wrote this code a couple weeks ago. Adding ^Inc generator to the intGenerator list we should just do. The rest seems nice, but isn't strictly needed. Let me know what you think.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -460,6 +460,19 @@ class ChooseGenerator : public Appendable {
     std::vector<int64_t> _weights;
 };
 
+// This is a wrapper generator. It wraps an int generator to generate strings.
+class StringIntGenerator : public Generator<std::string> {
```
rtimmons, line 464: I'm torn on this versus having expressions just use format-string generator which has a lot more power. I'll leave it to you and Jim, though, since it's subjective and may make certain kinds of generator expressions a lot more concise.

rtimmons (approved):
This seems fine.

I worry about needing combinatorially many generators to handle type conversions. For now we only seem to need str<->int so if we find we need...idk..doubles or specs or others...let's see about a more generalized approach (which will certainly be a thorny c++ type thing but that's the fun part).

src/value_generators/src/DocumentGenerator.cpp
dalyd: Yeah -- I go back and forth on this also. I used the format string generator for my actual use earlier, but it leads to stuff like this
`{^FormatString: {format: "%d", withArgs: [{^RandomInt:{min: 0, max: *NumItems}}]}}`
as opposed to 
`{^RandomInt:{min: 0, max: *NumItems}}}`
That example was in a ^Join node. There was also was a bit of surprise when I couldn't use an int as an argument to the ^Join generator. 

Anyways, up to you and @jimoleary whether to include this part or now.

dalyd:
I forgot that Jim is out this week. @rtimmons , what's your preference. I can strip out the StringIntGenerator part or leave it? Would like to merge something and move on either way.

rtimmons:
My preference would be to kill it until we have a few idiopathic uses of FormatGenerator that would be decidedly more clear with `StringIntGenerator`

dalyd:
Sounds good. I pulled them out. I'll let the checks run and then merge.

dalyd:
evergreen merge

PR #568, opened 2021-11-02
PERF-2687 Add benchmark to measure the performance of large document updates

laplab:


src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
...
+        Options:
+          WriteConcern:
+          Level: majority
+
+# TODO Configure which variants to run.
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - replica
```
laplab, line 83: I am not sure which other variants we should include. Basically, we need all variants which have replica set with at least 1 secondary. It could be beneficial to add some sharding variants as well, as long each shard has at least 1 secondary.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -982,6 +982,37 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+public:
+    ObjectGenerator(const Node& node,
+                   GeneratorArgs generatorArgs,
+                   std::map<std::string, Parser<UniqueAppendable>> parsers)
+        : _rng{generatorArgs.rng},
+          _node{node},
+          _generatorArgs{generatorArgs},
+          _keyGen{stringGenerator(node["key"], generatorArgs)},
+          _valueGen{valueGenerator<false, UniqueAppendable>(node["value"], generatorArgs, parsers)},
+          _nTimesGen{intGenerator(extract(node, "number", "^Object"), generatorArgs)} {}
```
rtimmons, line 996: This is a great generator. Syntax 💅🏽 - please consider the following syntax / key names:

```python
{^Object: {withNEntries: 1000, havingKeys: {^Foo}, andValues: {^Bar}}}
```

rtimmons (approved):
Nicely done! I'd still want to get the product-performance team's input on the actual workload, but from a "genny C++" perspective this seems 🌶️ 💯

src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This workload measures the performance of large document updates in a replica set. First, it
+  inserts documents with large embedded objects, containing 100k fields each. Then it runs a series
```
jimoleary, line 5: ```containing 100k fields each```

Aren't there 100 large documents, each document has 1K fields.

src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This workload measures the performance of large document updates in a replica set. First, it
```
jimoleary, line 4: Super pedantic but isn't ```the performance of updating large documents``` more accurate. The workload is updating a single field in large documents.

src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
...
+  ClientName: Default
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 1000
+    Threads: 50
+    Duration: 3 minutes
+    Collection: Collection0
+    Operations:
+    - OperationName: updateOne
```
jimoleary, line 70: Should we also be testing ```updateMany```?

src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
...
+      OperationCommand:
+        Filter: {_id: { ^RandomInt: { min: 0, max: *DocumentCount } }}
+        Update: {$set: {a: { ^RandomInt: { min: 0, max: 10000000 }}}}
```
jimoleary, line 73: Other questions:

- should there be a case where we update random strings or a random selection of different types (or at least non fixed size types like ints)?
- what about $unset ?
- what about different sized deltas (updating multiple fields)?

This ticket is specific to large documents,  do we also need a workload for different document sizes?

jimoleary (changes requested):
Can you add a test to src/value_generators/test/DocumentGeneratorTestCases.yml:
```
  - Name: ObjectGenerator
    GivenTemplate:
      a: {^Object: {key: {^RandomString: {length: 10}}, value: {^Inc: {}}, number: 10}}
    ThenReturns:
      - a:
          JGBEV0zQZI: { "$numberLong": "1" }
          eNbthBtmrZ: { "$numberLong": "2" }
          Odzj7TUCi+: { "$numberLong": "3" }
          SD6jA6Y+bp: { "$numberLong": "4" }
          kmFT2CuKVC: { "$numberLong": "5" }
          AFFXQ8ULbz: { "$numberLong": "6" }
          Naqsr/zBMQ: { "$numberLong": "7" }
          AzUNSxcpt0: { "$numberLong": "8" }
          xYYXCh/Ybl: { "$numberLong": "9" }
          EHcmXZX/d4: { "$numberLong": "10" }
```

src/workloads/execution/UpdateLargeDocuments.yml
laplab: Thanks! Where can I find setup description of each value? For instance, what does `shard-lite` mean (how many shards, how many replica-set members) and what is the difference from `shard`?

laplab: Found some details [here](https://github.com/10gen/performance-tooling-docs/blob/89f4a6e459ec3a3df6568af43f335081c886c8dd/build_variants.md), that was enough.

src/value_generators/src/DocumentGenerator.cpp
laplab: This syntax is not consistent with similar generator `^Array`, but I do not feel strongly about it. Done.

src/workloads/execution/UpdateLargeDocuments.yml
```diff
@@ -0,0 +1,83 @@
...
+  Type: Loader
+  # Loader doesn't support multiple threads for a single collection.
+  # TODO: TIG-2938 Use multiple threads to speedup insertion.
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &DB test
+    Threads: 1
+    CollectionCount: 1  # Collection name will be Collection0, this is not configurable.
+    DocumentCount: &DocumentCount 100
```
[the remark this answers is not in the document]
laplab: Yes, exactly.

src/workloads/execution/UpdateLargeDocuments.yml
laplab: > should there be a case where we update random strings or a random selection of different types (or at least non fixed size types like ints)?
> what about different sized deltas (updating multiple fields)?

In this workload, we would like to stress the extra overhead for the update operation. The update operation should be simple, so that we will be able to measure other parts of request execution (such as result document validation).

> what about $unset?

I am not sure we will get extra coverage from adding test with $unset operator. We are really interested in the non-update parts of the execution.

jimoleary (approved):
All looks good to me mod a question about the location of the new doc test. If the tests succeed in a patch than all is good. Otherwise the simplest fix is just to move the new test to the end of the file.

PR #571, opened 2021-11-08
TIG-3350 Python conveniences

aggrand:


src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -69,29 +71,44 @@ def _run_command_with_sentinel_report(
...
+def cmake_test(
+    genny_repo_root: str, workspace_root: str, regex: str = None, repeat_until_fail: int = 1
+):
     info = toolchain.toolchain_info(genny_repo_root=genny_repo_root, workspace_root=workspace_root)
     workdir = os.path.join(genny_repo_root, "build")
 
     xunit_dir = os.path.join(workspace_root, "build", "XUnitXML")
     os.makedirs(xunit_dir, exist_ok=True)
 
+    print("repeat until fail: ", str(repeat_until_fail))
```
rtimmons, line 83: 💅🏽 kill

src/lamplib/src/genny/cli.py
```diff
@@ -212,12 +212,25 @@ def clean(ctx: click.Context) -> None:
...
 @cli.command(name="cmake-test", help="Run genny's C++ unit tests.")
+@click.option(
+    "-g", "--regex", required=False, default=None, help=("Regex to match against tests."),
+)
+@click.option(
+    "-r",
+    "--repeat-until-fail",
+    required=False,
+    default=1,
+    help=("Repeat each test this many times until failing. Default is 1."),
```
rtimmons, line 223: This was hard for me to parse mentally. Really the flag is "repeat this many times unless it fails." Allows for tests to be flaky and dominated by passing or false-misses (didn't catch cuz only ran once and it only fails small-n% of the time). Not suggesting a change but I got confused by it, so perhaps an explanation of when/why you'd want to use this would be helpful.

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -69,29 +71,44 @@ def _run_command_with_sentinel_report(
...
+        "--parallel",
+        "4",
+        "--repeat-until-fail",
+        str(repeat_until_fail),
         "--label-exclude",
         "(standalone|sharded|single_node_replset|three_node_replset|benchmark)",
     ]
 
+    if regex is not None:
+        ctest_cmd += ["--tests-regex", regex]
```
rtimmons, line 97: do we need to `shlex.quote` or whatever? I forget how argv/quoting is handled downstream. Just double-checking.

rtimmons (approved):
Nice little change. An idea for added example text but nothing blocking.

src/lamplib/src/genny/tasks/run_tests.py
aggrand: Handled downstream, fortunately!

PR #572, opened 2021-11-08
TIG-3350 Parallel primitives

aggrand:
Split from here: https://github.com/mongodb/genny/pull/562

I went ahead and included both parallelrun and the atomic container in this one since the tests referred to each other, but let me know if splitting them apart would make the review a lot easier!

The remaining original PR would have the highest-level actual context and context test changes


src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+    AtomicContainer<T>& operator=(const AtomicContainer& rhs) {
+        std::lock_guard<lock_t> lock(rhs._mutex);
+        this->_container = rhs._container;
+        return *this;
+    }
+    AtomicContainer<T>& operator=(AtomicContainer&& rhs) {
+        std::lock_guard<lock_t> lock(rhs._mutex);
+        this->_container = std::move(rhs._container);
+    }
```
rtimmons, line 64: Do we gain anything by allowing copy/move construction or assignment? I think just inherit from boost::non_copyable and non_moveable or whatever they're called and we don't have to worry about the gotchas of these operations.

rtimmons (changes requested):
This seems about right.

This said, this is right at the edges of my ability in creating threadsafe library types, so we may want to consider recruiting an additional set of eyes from someone in #cxx-discuss or whatever that slack channel is.

Few ideas for improvement and one potential API gotcha but otherwise looks good to me.

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,245 @@
...
+     */
+    void lock() const {
+        // Paranoidly hold the lock here in case there are shenanigans
+        // between the flag flip and the unlock.
+        std::lock_guard<lock_t> lock(_mutex);
+        if (_forciblyHeld) {
+            BOOST_THROW_EXCEPTION(
+                std::logic_error("Cannot forcibly hold AtomicContainer lock multiple times."));
+        }
+        _mutex.lock();
```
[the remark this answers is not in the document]
aggrand: Since we're using `std::recursive_mutex`, my understanding is that at the end of this function, the `_mutex` has a lock count of two. When the guard is destructed the `unlock()` is called so when the function returns the lock count is one (representing the lock held by the caller).

We're effectively using the recursive mutex to allow multiple lockings to simulate a non-recursive lockable object with atomic operations.

src/gennylib/include/gennylib/parallel.hpp
aggrand: Good point. I'd originally made it copyable/moveable since the ActorVector that we plan to use this for needs them. That said, we could probably just pointer-wrap that so disallowing those would be a bit easier.

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,245 @@
...
+        return _container.cbegin();
+    }
+    typename T::const_iterator end() const {
+        std::lock_guard<lock_t> lock(_mutex);
+        _assertForciblyHeld();
+        return _container.end();
+    }
+    typename T::const_iterator cend() const {
+        std::lock_guard<lock_t> lock(_mutex);
+        _assertForciblyHeld();
```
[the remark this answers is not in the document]
aggrand: I think that's what the `AtomicContainer throws if iterating without holding lock` test asserts?

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,245 @@
...
+                }};
+            });
+
+    for (auto& thread : *threadsPtr) {
+        thread.join();
+    }
+
+    std::lock_guard<AtomicDeque<std::exception_ptr>> excLock(caughtExceptions);
+    for (auto&& exc : caughtExceptions) {
+        std::rethrow_exception(exc);
```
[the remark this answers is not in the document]
aggrand: I added a helper `ParallelException` type that manages this.

src/gennylib/include/gennylib/parallel.hpp
aggrand: Ended up having to undo the the removal of copy/move since the `ParallelException` needed it to be safely throwable, and I preferred to lean on this code rather than implement a thread-safe insertable but throwable type there. Defs push back if you disagree.

That said, I think most of the correctness of these operators will depend on the underlying type; we just need to make sure lock shenanigans aren't happening.

aggrand:
Good point about posting in `cxx-discuss`, will see if anyone is interested!

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,290 @@
...
+        this->_container = std::move(rhs._container);
+    }
+
+    // STL Methods
+    auto&& front() {
+        std::lock_guard<lock_t> lock(_mutex);
+        return _container.front();
+    }
+    template <typename... Args>
+    auto&& emplace_back(Args... args) {
```
chardan, line 77: Should this be "Args&&..."?

aggrand:
@rtimmons following the advice of people in the `#cxx-discuss` channel, I actually got rid of the `AtomicContainer`. As part of the other PR I'll add need-specific thread-safe containers for what we need (might split that change into a separate PRs, if needed). 

This PR is now focused on just extracting out the parallel execution, RFAL

src/gennylib/include/gennylib/parallel.hpp
aggrand: Good point! Ended up getting rid of this section anyway

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Class that represents a bundle of exceptions thrown together after
+ * ending parallel execution.
+ */
+class ParallelException : public std::exception {
+public:
+
+    ParallelException(std::deque<std::exception_ptr>&& exceptions) {
+        _caughtExceptions = std::move(exceptions);
```
rtimmons, line 41: init statement?

```c++
ParallelEx(deque&& exc)
: _caught{std::move(exc)} { .... }
```

(That `std::move` isn't strictly necessary of course but it's become an accepted piece of ceremony)

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,146 @@
...
+            std::logic_error("Tried to construct ParallelException, but no exceptions were given.");
+        }
+        std::stringstream ss;
+        ss << "Error in parallel execution. First exception's what(): ";
+
+        // We cannot just call what() through the exception pointer, unfortunately.
+        try {
+            std::rethrow_exception(_caughtExceptions[0]);
+        } catch (const std::exception& e) {
+            ss << e.what();
```
rtimmons, line 52: do you want to iterate over all `_caught`?

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <vector>
+#include <deque>
+
+namespace genny {
+
+/**
+ * Class that represents a bundle of exceptions thrown together after
+ * ending parallel execution.
+ */
+class ParallelException : public std::exception {
```
rtimmons, line 37: See if it makes sense to put this as an inner class of ExceptionBucket or similar.

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,146 @@
...
+    /*
+     * Add an exception to the bucket.
+     */
+    void addException(std::exception_ptr exc) {
+        std::lock_guard<std::mutex> lock(_mutex);
+        _caughtExceptions.push_back(exc);
+    }
+
+    /*
+     * If possible, exctract a ParallelException from the bucket.
```
rtimmons, line 92: I think mention here that this also "resets" the `ExceptionBucket`. I question the utility of being able to re-use this type--do we gain anything by `std::move`-ing out the exceptions versus just exposing the value as a `const &` ref?

src/gennylib/include/gennylib/parallel.hpp
```diff
@@ -0,0 +1,146 @@
...
+                        caughtExc.addException(std::current_exception());
+                    }
+                }};
+            });
+
+    for (auto& thread : *threadsPtr) {
+        thread.join();
+    }
+
+    auto parallelExc = caughtExc.extractExceptions();
```
rtimmons, line 138: consider moving this logic into the bucket.

```c++
for thread in thread: join
bucket.throwIfExceptions()
```

rtimmons (approved):
Couple ideas to consider but overall I'm much happier that the container's API is smaller.

src/gennylib/include/gennylib/parallel.hpp
aggrand: I'd figured not since in a many-threaded workload thread failure may be comorbid, and I figured we don't want hundreds or thousands of the same exception message printed from all the threads. Push back if you disagree though! Keeping the exceptions in the bundle exception and being able to retrieve them will let us do more with it in the future if necessary.

rtimmons (approved):

PR #578, opened 2021-11-18
PERF-2572 add warmup phase, lower initial/ secondary document count and name each index.

jimoleary:
There is a [patch](https://spruce.mongodb.com/version/61963c6a1e2d176241fc9742/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), but it is being blocked by a compile error.

### DIFF
--- a/src/workloads/transactions/LLTMixed.yml
+++ b/src/workloads/transactions/LLTMixed.yml
@@ -1,20 +1,20 @@
 SchemaVersion: 2018-07-01
+SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-perf"
 Description: |
   Workload to Benchmark the effect of LongLivedTransactions on a Mixed workload.
 
 GlobalDefaults:
   # Start: The following block gets updated for the other configurations.
-  # In-memory: Database size works out about 12GB.
-  InitialDocumentCount: &InitialNumDocs 10000000
-  SecondaryDocumentCount: &SecondaryNumDocs 10000000
+  InitialDocumentCount: &InitialNumDocs 6600000
+  SecondaryDocumentCount: &SecondaryNumDocs 6600000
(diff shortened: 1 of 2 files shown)


apodelko (dismissed):
LGTM, although I won't be surprised if it won't eliminate variability completely. But we probably need this step anyway even if we need to investigate it further.

apodelko:
Another option may be to run the patch multiple times and see how much the fix helps.

src/workloads/transactions/LLTMixed.yml
```diff
@@ -5,16 +5,15 @@ Description: |
 
 GlobalDefaults:
```
dalyd, line 6: By my reading this is: 
0. Queisce
1. Initial Load
2. Warmup
3. Second Load

Then repeating: quiesce, warmup, test. Is that correct? 
If it is I can't find the phase 5 warmup. 

Could we document this flow?

dalyd (changes requested):

jimoleary:
Updated the branch with PR comments. Started a new [patch](https://spruce.mongodb.com/version/61a7acd6850e610a956b7ffa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

*note:* I'm actually going to move the current LLTMixed.yml to LLTMixedWarmup.yml  and reinstate the old version before committing. This is for 2 reasons:
1. We want to keep the original workload, so keeping the task name keeps the history. This test will likely be noisy but it does provide a useful signal.
2. I wanted to be able to compare the new workload to the old one.

src/workloads/transactions/LLTMixed.yml
dalyd: Done below.

dalyd (approved):
Alright, my issues have been addressed. I'll leave it with the two of you to make sure it actually works and runs :)

PR #580, opened 2021-11-30
PERF-2613: Add perf test for locf 

madelinezec:


src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,257 @@
...
+    CollectionCount: 1
+    DocumentCount: 10000
+    BatchSize: &batchSize 100
+    Document:
+      part1: {^Choose: {from: [1, 2, 3, 4]}}
+      part2: {^Choose: {from: [1, 2, 3, 4]}}
+      numeric: {^Choose: {from: [
+        {^RandomInt: {min: -100, max: 100}}, 
+        {^RandomDouble: { min: 0.0, max: 500.0 }},
+        null], weights: [1, 1, 8]}}
```
TedTuckman, line 25: Any particular reason for these weights? Not saying it should be changed, just curious.

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,257 @@
...
+- Name: Quiesce
+  Type: QuiesceActor
+  Threads: 1
+  Database: *db
+  Phases:
+  - Nop: true
+  - Repeat: 1
+  - Nop: true
+
+- Name: locf
```
TedTuckman, line 50: I think actor names are supposed to be capitalized.

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,257 @@
...
+            partitionBy: {part1: "$part1", part2: "$part2"},
+            output: {
+                randomType: {
+                  $locf: "$randomType"}
+                  }
+          }
+        }]
+        cursor: {batchSize: *batchSize}
+        allowDiskUse: true
+
```
TedTuckman, line 243: Can you add a test or two with multiple outputs? Don't think we need all combinations, but one with all three types I think would be good to have.

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,257 @@
...
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq: 
+        - standalone
+        - replica
+        - replica-all-feature-flags
```
TedTuckman, line 250: Given that we're going to be removing the feature flag before this is committed, I don't think we need the all-feature-flag variant.

TedTuckman (changes requested):
Pretty happy with this. Minor changes requested

src/workloads/execution/locf.yml
madelinezec: done

src/workloads/execution/locf.yml
madelinezec: done

src/workloads/execution/locf.yml
madelinezec: done

src/workloads/execution/locf.yml
madelinezec: I think when I was discussing nulling documents with David Percy we mentioned making it 70% likely for a value to be null, so I was just trying to keep near that likelihood?

TedTuckman (approved):
LGTM. @dalyd Two questions for you:
1. Is there a formatter that should be run?
2. I don't think we need all the different tests for different types -- we don't do anything differently for them. Do you think there's any reason to include or remove them? I was thinking if they appear to be similar on the first run we could have each partition type match with one data type while keeping the multiple outputs in each situation.

dalyd:
> LGTM. @dalyd Two questions for you:
> 
> 1. Is there a formatter that should be run?
> 
Yes, please run YAPF. There should be configuration in the repo. This hasn't been turned on in CI, mostly because too many things would fail right now. https://github.com/mongodb/genny/blob/3cb776c6f5a21056c3c779ae8eafd0f47b8b5b94/docs/developing.md#L32
./run-genny lint-yaml

madelinezec:
> > LGTM. @dalyd Two questions for you:
> > 
> > 1. Is there a formatter that should be run?
> 
> Yes, please run YAPF. There should be configuration in the repo. This hasn't been turned on in CI, mostly because too many things would fail right now.
> 
> https://github.com/mongodb/genny/blob/3cb776c6f5a21056c3c779ae8eafd0f47b8b5b94/docs/developing.md#L32
> 
> 
> ./run-genny lint-yaml

I did run the linter and it passes - there's no other formatter I need to run though, correct? Thank you for all your help!

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
...
+  - Nop: true
+
+- Name: Locf
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Nop: true
+  - Nop: true
+  - Repeat: 10
+    Database: *db
```
dalyd, line 57: As written this will run each operation sequentially through, and then repeat that 10 times. Is that what you want? Or would you prefer to have the  operations in different phases with a quiesce between them? 

Also, is 10 repetitions sufficient? Would you prefer to run it for a period of time?

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
...
+  Type: QuiesceActor
+  Threads: 1
+  Database: *db
+  Phases:
+  - Nop: true
+  - Repeat: 1
+  - Nop: true
+
+- Name: Locf
+  Type: RunCommand
```
dalyd, line 51: Nothing for you to do, but I'm disappointed that the CrudActor doesn't support agg.

dalyd (changes requested):
First pass. I put most of my effort into the documentation and have some requests there. 

I looked less closely at the specifics of the actual operations as opposed to their structure. I'm assuming Ted will make sure the specifics make sense.

dalyd:
One more request: could you add a link to a sys-perf patch build with this workload? I'd like to see numbers.

dalyd:
> LGTM. @dalyd Two questions for you:

> 2. I don't think we need all the different tests for different types -- we don't do anything differently for them. Do you think there's any reason to include or remove them? I was thinking if they appear to be similar on the first run we could have each partition type match with one data type while keeping the multiple outputs in each situation.

No strong opinion from me on your second question here. If it looks like it provides value, let's include it. Otherwise let's run faster.

src/workloads/execution/locf.yml
madelinezec: We'd love to get your expert opinion here as we don't know what's best. We could definitely pivot to having each operation be in its own phase with a quiesce between them, if you think that's better

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
...
+        cursor: {batchSize: *batchSize}
+        allowDiskUse: true
+
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+        - replica
```
TedTuckman, line 288: Extra indent

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of $locf. The benchmark operations test numeric,
```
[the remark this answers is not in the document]
TedTuckman: Formatting of this may be off, but something like that?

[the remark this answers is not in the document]
madelinezec: done

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
...
+    Document:
+      part1: {^Choose: {from: [1, 2, 3, 4]}}
+      part2: {^Choose: {from: [1, 2, 3, 4]}}
+      numeric: {^Choose: {from: [
+        {^RandomInt: {min: -100, max: 100}},
+        {^RandomDouble: { min: 0.0, max: 500.0 }},
+        null], weights: [1, 1, 8]}}
+      string: {^Choose: {from: [
+        {^RandomString: {length: 15}},
+        null], weights: [2, 8]}}
```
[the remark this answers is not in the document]
madelinezec: done

src/workloads/execution/locf.yml
madelinezec: done

src/workloads/execution/locf.yml
dalyd: It depends what you are trying to learn about the system. I suspect you are more interested in the performance of the individual operations, as opposed to their interactions. If that is correct I would encourage you to put them into separate phases.

src/workloads/execution/locf.yml
```diff
@@ -0,0 +1,295 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of $locf. The benchmark operations test numeric,
+  string, date and randomized data types. There is a test without partitions, with a
+  single partition, and with multiple partitions for each data type.
```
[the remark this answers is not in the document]
TedTuckman:
```suggestion
  single partition, and with multiple partitions for each data type. Partitions require carry forward
  to be done individually on each partition, which requires additional tracking and may make queries
  slower.
```
Does this answer the question? Or is it still slightly too vague?

src/workloads/execution/locf.yml
TedTuckman: This is now done?

madelinezec:
evergreen build still running https://spruce.mongodb.com/version/61a8e6090305b97827bed971/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/execution/locf.yml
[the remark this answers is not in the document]
dalyd: That helps some, but I still don't think I know what a partition is here? 

The number of carry forwards is proportional to the number of partitions?

src/workloads/execution/locf.yml
dalyd: It looks like it. Nice. Is 10 reptitions giving you everything that you need here? Or would you prefer a duration? You can use both and it will wait for both to be satisfied. 

https://github.com/mongodb/genny/blob/3cb776c6f5a21056c3c779ae8eafd0f47b8b5b94/docs/HitchhikersGuide.md

> If both the Repeat and Duration keywords are specified, then it will run for the longer of the two (e.g. { Duration: 2 seconds, Repeat: 1 } the Duration will win unless the single iteration takes longer than 2 seconds).

src/workloads/execution/locf.yml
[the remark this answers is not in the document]
dalyd: Ah, could you just link to https://docs.mongodb.com/manual/reference/operator/aggregation/setWindowFields/ 
That explains what a partition is. There's a good chance someone from outside of query will need to look at a regression in these tests at some point, so that will help. 

And please include Ted's addition too.

dalyd (commented):
Starting to make sense to me now. Thanks for your patience with me. 

The patch build is purple. Could you reach out to #performance-tooling-users to try to figure out what's going on there? I wanted to see the results and durations.

src/workloads/execution/locf.yml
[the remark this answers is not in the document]
madelinezec: done

src/workloads/execution/locf.yml
madelinezec: We think 10 repetition is giving us sufficient info and we do not need a duration - thank you for providing us with this extra info, will be helpful in the future!

madelinezec:
new patch build: https://spruce.mongodb.com/version/61a916221e2d1765ffbebb05/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I ran previous patch locally and it failed so I was able to debug from there. Hoping the latest build doesn't have more issues!

src/workloads/execution/locf.yml
dalyd: The working patch build helps. Thanks. I can see that the operations are taking ~20-40ms on average, so running 10 repetitions is still under a second per operation, and the entire genny invocation is taking only 17 seconds. I think we should add a Duration for these. Maybe 30 seconds?  It just gives us a lot more data and chance to smooth out any weird behavior. 30 seconds is a fairly arbitrary number. For things that drive IO and checkpoints we would want each individual test to be some multiple of the WT checkpoint frequency (minute), but that's probably not necessary here.

dalyd (changes requested):
Cool -- we're in a good place here. To parrot my comment below, I think we should run these tests for longer and add a Duration. While the task took 30 minutes, the useful part was only 17 seconds, so we can safely increase each test time. 

Could you add a duration and submit another patch build? The patch build shouldn't block anything, but I am curious at the numbers at this point (particularly ratio between 50th and 95th Percentile on latency which would be an indicator of how much noise we see). 

Also, for patch builds for new genny workloads, if you pick the schedule_patch_auto_tasks task, it will run only the genny task you changed. No need to run anything else.

dalyd:
> Cool -- we're in a good place here. To parrot my comment below, I think we should run these tests for longer and add a Duration. While the task took 30 minutes, the useful part was only 17 seconds, so we can safely increase each test time.
> 
> Could you add a duration and submit another patch build? The patch build shouldn't block anything, but I am curious at the numbers at this point (particularly ratio between 50th and 95th Percentile on latency which would be an indicator of how much noise we see).
> 
> Also, for patch builds for new genny workloads, if you pick the schedule_patch_auto_tasks task, it will run only the genny task you changed. No need to run anything else.

I meant to say LGTM modulo longer run time.

madelinezec:
Duration based sys-perf patch passed! https://spruce.mongodb.com/version/61aa3a8b2a60ed5e66d0071a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

dalyd (approved):
Thanks for the chanegs and longer run. I can see them taking 30 seconds each, and the latencies seem really solid, with very little variation between average, median, 95th, 99th percentiles. 

The Max latency can come out up to 2x the min latency. I wonder if that is the first execution. We can pull that out, but I can't do it too easily. If it is the first execution, we should ask ourselves which we want to measure (first, or average). We should make sure that we aren't caching or memoizing the whole thing, but are really testing the desired feature. You two should know the feature well enough to hopefully answer that without any work.

PR #581, opened 2021-12-01
TIG-3599 Update genny to use new vcpkg/toolchain

richardsamuels:
(this is a revert of the revert)

sys-perf: https://evergreen.mongodb.com/version/61a806662a60ed340efabc6d


richardsamuels:
Broken test is TIG-3639

src/lamplib/src/genny/toolchain.py
```diff
@@ -199,6 +199,16 @@ def _can_ignore(self):
             self.ignore_toolchain_version or self._check_toolchain_githash()
         )
 
+    def _fetch_and_install_impl(self):
+        # Remove the local copy of the tarball if it exists. If we don't
+        # do this, toolchain upgrades can fail if a previous install was
+        # interrupted.
+        tarball = os.path.join(self._install_dir, self._name + ".tgz")
```
guoyr, line 206: [Optional] I'd personally move this to the parent class. Seems like a useful approach for any kinds of download. An aside: the curator downloader may be broken? We download .tar.gz [here](https://github.com/mongodb/genny/blob/c404181dd5f70d32307a898900a350650654e64c/src/lamplib/src/genny/curator.py#L217) but the tarball checks for .tgz [here](https://github.com/mongodb/genny/blob/db80302eb40d2b775c14af9ca21ad980864ee3e6/src/lamplib/src/genny/download.py#L97)

guoyr (approved):
LGTM! % ❓

src/lamplib/src/genny/toolchain.py
richardsamuels: When the tarball is downloaded locally, it names the local copy with .tgz, so the curator downloader should be fine. 

The reason I didn't implement this in the parent class is because of the logic that skips the download if it already exists. Now that I think of it, this bug could affect anything we download, so i'm going to move this logic to the parent class

richardsamuels:
Breakage in the test is caused [by this ticket](https://jira.mongodb.org/browse/SERVER-61791). Ignoring this and merging.

PR #582, opened 2021-12-02
PERF-2698 Block GraphLookup and Lookup workload from running before 5.1

alyacb:


HanaPearlman (approved):
Thanks Alya! I think I remember seeing a sys-perf project for v3.6. Do you know if that's still running and if so, do we need to exclude v3.6 here as well? Otherwise looks great.

alyacb:
Thank you for the quick review, Hana!

> I think I remember seeing a sys-perf project for v3.6. Do you know if that's still running and if so, do we need to exclude v3.6 here as well?

I believe 3.6 is EOL; the sys-perf 3.6 project doesn't seem to have run since October, and none of the other files in Genny reference 3.6 (v4.0 is the earliest they filter for), so I'm going to omit it.

The test failure on the patch is unrelated to these changes (has to do with https://jira.mongodb.org/browse/TIG-3638).

alyacb:
evergreen merge

PR #583, opened 2021-12-03
TIG-3638 Update pinned mongo version in Genny

aggrand:


vrachev (approved):
Thanks for the quick fix on this!

PR #584, opened 2021-12-06
PERF-2614: Add perf test for $linearFill

madelinezec:
sys-perf patch: https://spruce.mongodb.com/version/61ae6e2f9ccd4e07042ee473/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

There are two system failures that look to be due to AWS outage (checked with #performance-tooling-users). I have re-enqueued it today


src/workloads/execution/linearFill.yml
```diff
@@ -0,0 +1,308 @@
...
+      $eq:
+      - standalone
+      - replica
+    branch_name:
+      $neq:
+      - v4.0
+      - v4.2
+      - v4.4
+      - v5.0
+      - v5.1
```
TedTuckman, line 308: Also don't want to run on 5.2 right?

TedTuckman (approved):

src/workloads/execution/linearFill.yml
madelinezec: I guess we don't want to run on 5.2 because the purpose of the perf test is to not use a feature flag - right? Bc linearFill did get in yesterday before the release was cut

TedTuckman: Right, the feature flag would have to have been removed

PR #585, opened 2021-12-07
TIG-3657 Update Curator version in Genny

aggrand:


src/lamplib/src/genny/curator.py
```diff
@@ -178,9 +178,7 @@ class CuratorDownloader(Downloader):
     # These build IDs are from the Curator Evergreen task.
     # https://evergreen.mongodb.com/waterfall/curator
 
-    # Note that DSI also downloads Curator, the location is specified in defaults.yml.
```
aggrand, line 181: I removed this note about DSI, since the use of curator in DSI is deprecated with perf.send. I think later on in the Cedar migration project it'll behoove us to remove it from DSI entirely.

ancostas (changes requested):
Did this break the tests that are failing?

aggrand:
@ancostas that was an issue fixed on master, merged that into the branch.

But also there's TIG-3236 which will probably fail either way. Might be worth picking that up soon so we get better signal on the PR tests.

ancostas (approved):

aggrand:
evergreen merge

aggrand:
evergreen merge

PR #589, opened 2021-12-13
TIG-3236 t_benchmark_test is flaky

aggrand:
Templatizing phaseloop on a ClockSource would be a bit too invasive of a change to be worth it in the near-term, when this is red for basically every PR (affecting workload authors too). And even that wouldn't remove the thread-based nondeterminism. Arguably the thread jumble like this is expected and mocking it out would be a worse test of behavior anyway. (Plus we have separate unit tests for rate limiters anyway; the benchmark test I believe is happier with a more "real-world" situation.)

We could consider doing some stats about this, but the exact distribution would probably vary from machine to machine, and this is in line with how the other benchmark tests measure behavior.

Tested by removing the "benchmark" tag and running with `--repeat=500`, seemed good.


vrachev (approved):
LGTM, thanks for fixing this!

mridul-augustine (approved):

aggrand:
evergreen merge

PR #591, opened 2021-12-14
PERF-2728: Add perf test for $densify in timeseries collection

madelinezec:


apodelko:
Did you run a patch with this workload? If yes, please share a link. I believe that it is important to make sure that it works properly in sys-perf. Also it would be good to run it several times to make sure that we don't have high variability.

madelinezec:
first patch build https://spruce.mongodb.com/version/61b909b961837d1c6c017645/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC, currently running a second

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -0,0 +1,451 @@
...
+GlobalDefaults:
+  dbname: &db test
+  batchSize: &batchSize 30000
+  fieldName: &field "numeric"
+  unitName: &unit null
+  index: &index
+    keys: {numeric: 1}
+    options: {name: "numeric"}
+  smallStepSize: &smallStep {^RandomInt: {min: 3, max: 3}}
+  largeStepSize: &largeStep {^RandomInt: {min: 8, max: 8}}
```
TedTuckman, line 16: These numbers aren't ranges?

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -0,0 +1,451 @@
...
+  - Repeat: 1
+    Database: *db
+    Collection: *coll
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10000
+    BatchSize: *batchSize
+    Document:
+      partitionKey: {^RandomInt: {min: 1, max: 100}}
+      timestamp: {^RandomDate: {}}
```
TedTuckman, line 92: Should this have a range?

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -0,0 +1,451 @@
...
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: DensifyFullByPartitionSmallStep
```
TedTuckman, line 196: Full ignores partition. Don't need this and corresponding large tests

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -0,0 +1,451 @@
...
+  dbname: &db test
+  batchSize: &batchSize 30000
+  fieldName: &field "numeric"
+  unitName: &unit null
+  index: &index
+    keys: {numeric: 1}
+    options: {name: "numeric"}
+  smallStepSize: &smallStep {^RandomInt: {min: 3, max: 3}}
+  largeStepSize: &largeStep {^RandomInt: {min: 8, max: 8}}
+  explicitBounds: &bounds [{^RandomInt: {min: 5000, max: 15000}}, {^RandomInt: {min: 20000, max: 40000}}]
```
TedTuckman, line 17: The max value in the documents is 20k no? So we should lower the total bounds to cover just the range that the documents actually cover + a bit, but this seems excessive. Don't think we should make the max bounds always over the existing documents

TedTuckman (changes requested):
Looks good, minor comments

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: RandomDate doesn't require a range, and since we aren't worried about repeating values I wasn't worried about leaving it like this.  I think since we are densifying the meta field, this is fine?

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: Oh interesting, should I remove [the corresponding tests from DensifyNumeric](https://github.com/mongodb/genny/blob/master/src/workloads/execution/DensifyNumeric.yml#L145-L174) as well?

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: The syntax is strange but it help encode smallStep's type. For instance, if I change it to: `smallStepSize: &smallStep 3` it crashes the server (`{ "ok" : 0.0, "errmsg" : "Upper bound, lower bound, and step must all have the same type", "code" : 5876900, "codeName" : "Location5876900" }`) so its ugly but the wrapper is necessary so the value 3 isn't read as a string, double, etc. This kind of syntax also already exists in other workloads in the repo

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: Interesting, I got these bounds from DensifyNumeric. But I'll push changes like you described, bc you what you said does make more sense.

src/workloads/execution/DensifyTimeseriesCollection.yml
TedTuckman: Oh, so you wanted the step to always be 3 as opposed to 2-4? If so can you add a comment?

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -14,7 +14,7 @@ GlobalDefaults:
     options: {name: "numeric"}
   smallStepSize: &smallStep {^RandomInt: {min: 3, max: 3}}
   largeStepSize: &largeStep {^RandomInt: {min: 8, max: 8}}
-  explicitBounds: &bounds [{^RandomInt: {min: 5000, max: 15000}}, {^RandomInt: {min: 20000, max: 40000}}]
+  explicitBounds: &bounds [{^RandomInt: {min: 1000, max: 50000}}, {^RandomInt: {min: 10000, max: 20000}}]
```
TedTuckman, line 17: Not sure this will work either. The lower bound does have to be below the upper bound. I was getting at just having the upper bound start below the value in the documents.

madelinezec: Oh ha, that's a typo. Supposed to be 5k I'll change

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: done

madelinezec:
> first patch build https://spruce.mongodb.com/version/61b909b961837d1c6c017645/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC, currently running a second
second is almost finished https://spruce.mongodb.com/version/61ba05f3c9ec440a78f6342f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/execution/DensifyTimeseriesCollection.yml
TedTuckman: Still here?

src/workloads/execution/DensifyTimeseriesCollection.yml
TedTuckman: You raise an interesting point. Tbh I don't know much about the use cases of time series collections, but it feels like we should probably be operating on the timestamp field? Should we have at least one case where we densify on it?

TedTuckman (changes requested):

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: done

src/workloads/execution/DensifyTimeseriesCollection.yml
madelinezec: done

apodelko:
What is the reasoning behind running this workload on - replica and - replica-all-feature-flags?

madelinezec:
> What is the reasoning behind running this workload on - replica and - replica-all-feature-flags?

Thank you for catching that mistake. A lot of the fill was previously behind feature flag, so I accidentally had that in there. I will re-run the patch.

madelinezec:
another successful patch: https://spruce.mongodb.com/version/61bbae8da4cf47433e935bb6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

madelinezec:
(Ted has approved but he is flying and cant approve the PR on his phone)

apodelko (approved):

TedTuckman (approved):

PR #593, opened 2021-12-15
PERF-2729: Add perf test for densify/fill  combo

madelinezec:
patch nearing done: https://spruce.mongodb.com/version/61ba216e7742ae0dfb469b10/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/DensifyFillCombo.yml
+++ b/src/workloads/execution/DensifyFillCombo.yml
@@ -12,10 +12,15 @@ GlobalDefaults:
   batchSize: &batchSize 30000
   fieldName: &field "timestamp"
   index: &index
-    keys: {numeric: 1}
-    options: {name: "numeric"}
-  smallStepSize: &smallStep {^RandomInt: {min: 3, max: 3}}
+    keys: {timestamp: 1}
+    options: {name: "timestamp"}
+  stepSize: &step {^RandomInt: {min: 10, max: 10}}
   nop: &Nop {Nop: true}
+  explicitBounds: &bounds [
+    {^RandomDate: {min: "2021-01-01T00:00:02.000Z", max: "2021-01-01T00:00:10.000Z"}},
+    {^RandomDate: {min: "2021-01-01T00:00:10.000Z", max: "2021-01-01T00:00:20.000Z"}}
+  ]
+  numericBounds: &numericBounds [{^RandomInt: {min: 5000, max: 15000}}, {^RandomInt: {min: 20000, max: 30000}}]
 
 Actors:
 - Name: CreateTimeSeriesCollection
@@ -54,10 +59,10 @@ Actors:
     Database: *db
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 100
+    DocumentCount: 10000
     BatchSize: *batchSize
     Document:
-      partitionKey: {^RandomInt: {min: 1, max: 100}}
+      partitionKey: {^RandomInt: {min: 1, max: 10}}
       timestamp: {^RandomDate: {min: "2021-01-01T00:00:00.000Z", max: "2021-01-01T00:00:20.000Z"}}
       toFillRandomType: {^Choose: {from: [
         {^RandomInt: {min: -100, max: 100}},
@@ -69,7 +74,7 @@ Actors:
         {^RandomDouble: { min: 0.0, max: 2000.0 }}]}}
       numeric: {^Inc: {start: 0, multiplier: 1, step: 6}}
     # $densify stages always add a $sort on the field that is being densified, so this phase adds
-    # indexes on the date/number fields so that we aren't performing an in-memory sort.
+    # indexes on the timestamp field so that we aren't performing an in-memory sort.
     Indexes:
     - *index
   - *Nop
@@ -155,8 +160,8 @@ Actors:
           $densify: {
             field: *field,
             range: {
-              bounds: "full",
-              step: *smallStep,
+              bounds: *bounds,
+              step: *step,
               unit: "millisecond"
             }
           },
@@ -201,7 +206,7 @@ Actors:
             field: *field,
             range: {
               bounds: "partition",
-              step: *smallStep,
+              step: *step,
               unit: "millisecond"
             },
             partitionByFields: ["partitionKey"]
@@ -248,7 +253,7 @@ Actors:
             field: "numeric",
             range: {
               bounds: "partition",
-              step: *smallStep,
+              step: *step,
             },
             partitionByFields: ["partitionKey"]
           },
@@ -293,8 +298,8 @@ Actors:
           $densify: {
             field: "numeric",
             range: {
-              bounds: "full",
-              step: *smallStep
+              bounds: *numericBounds,
+              step: *step
             },
           },
           $fill: {
@@ -340,8 +345,8 @@ Actors:
           $densify: {
             field: "numeric",
             range: {
-              bounds: "full",
-              step: *smallStep
+              bounds: *numericBounds,
+              step: *step
             },
           },
           $fill: {
--- a/src/workloads/execution/DensifyFillCombo.yml
+++ b/src/workloads/execution/DensifyFillCombo.yml
@@ -2,11 +2,15 @@ SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
   This workload tests the performance of queries that combine $densify and $fill stages.
+  $densify creates new documents to eliminate the gaps in the "timestamp" or "numeric"
+  fields. The other fields in these generates documents however, will be missing.
+  $fill set values for the output field (in the context of this workload,
+  "toFillRandomType" or "toFillNumeric") when the value is null or missing.
 
 GlobalDefaults:
   dbname: &db test
   batchSize: &batchSize 30000
-  fieldName: &field "numeric"
+  fieldName: &field "timestamp"
   index: &index
     keys: {numeric: 1}
     options: {name: "numeric"}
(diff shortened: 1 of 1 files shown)


src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,181 @@
...
+            range: {
+              bounds: "full",
+              step: *smallStep
+            },
+            partitionByFields: ["partitionKey"]
+          },
+          $fill: {
+            output: {
+              toFillRandomType: {
+                method: "locf"
```
madelinezec, line 158: I wanted to mention that I couldn't find a way to reasonable combine densify and fill with linear method. We would want to not repeat values in the sort field, but generally the sort field is also where we $densify so we want to not repeat values and also for the values to have gaps (eg one doc contains numeric:5, the next numeric:10). I can see if this is possible in genny, but I thought it best to get this up now

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,181 @@
...
+            },
+            partitionByFields: ["partitionKey"]
+          }
+        ]
+        cursor: {batchSize: *batchSize}
+        allowDiskUse: true
+  - *Nop
+  - *Nop
+
+- Name: DensifyFillFullBoundsWithPartition
```
TedTuckman, line 129: Full/partition isn't necessary to include.

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,181 @@
...
+  batchSize: &batchSize 30000
+  fieldName: &field "numeric"
+  index: &index
+    keys: {numeric: 1}
+    options: {name: "numeric"}
+  smallStepSize: &smallStep {^RandomInt: {min: 3, max: 3}}
+  nop: &Nop {Nop: true}
+
+Actors:
+- Name: InsertData
```
TedTuckman, line 17: Should we do this with a timeseries collection?

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,181 @@
...
+  - *Nop
+  - Duration: 30 seconds
+    Database: *db
+    Operations:
+    - OperationMetricsName: TestDensifyFillWithoutPartition
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [
+          $densify: {
```
TedTuckman, line 73: We probably want these stages to interact in some way. Thinking about it now, its possible we should use the "arbitrary expression" fill method here (either as an additional test or instead of locf), as setting a default value makes sense to me as a use case for the combo. It also may make sense to take a look at the [PD](https://docs.google.com/document/d/1NFHnY2t8ThW2UVcIGuCHc5PJZpROZQ74lBiJFhqfSxI/edit) to see if Katya had cases either of us don't have off the top.

src/workloads/execution/DensifyFillCombo.yml
TedTuckman: Yeah I see the issue. Does the "increment" loader not make this doable?

TedTuckman (commented):
Some questions/comments

madelinezec:
successful patch: https://spruce.mongodb.com/version/61ba216e7742ae0dfb469b10/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/execution/DensifyFillCombo.yml
madelinezec: Would you like to? It wasn't my understanding that we needed to do this in a timeseries collection, but I can if you want?

src/workloads/execution/DensifyFillCombo.yml
madelinezec: Hmm I don't follow, what is "arbitrary expression" fill method?

src/workloads/execution/DensifyFillCombo.yml
madelinezec: done

src/workloads/execution/DensifyFillCombo.yml
madelinezec: done

src/workloads/execution/DensifyFillCombo.yml
madelinezec: done

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,372 @@
...
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 1
+    Database: *db
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 100
```
madelinezec, line 57: Because we are running hundreds of densify commands (we run each operation for 30 seconds), I put the initial document count very low and also increased the InternalQueryMaxAllowedDensifyDocs to 100k (default is 50k). I wonder if, for performance reasons, it would be better to change from duration based to repeat based @apodelko ?

apodelko: Could you please elaborate it a little further? I am not sure that I quite got what is going on here. If I understand correctly we have 8 different tests here and each has between 472 and 708 OperationsTotal (in the second patch) - so we have something like 5-6K OperationsTotal total.

madelinezec: Sorry, I misspoke. We are running thousands not hundreds of densify queries (each operation is one query that contains a densify and fill stage). Without the `internalQueryMaxAllowedDensifyDocs` manually increased to 100k, we run into an error. `PlanExecutor error during aggregation :: caused by :: Generated 500001 documents in $densify, which is over the limit of 500000. Increase the 'internalQueryMaxAllowedDensifyDocs' parameter to allow more generated documents: generic server error`

apodelko: What in this case is a operation? As I see in the patch results, we have something like 5-6K OperationsTotal total for 8x30 sec runs. How many documents do we generate per operation?

Generically speaking, I'd rather suggest a larger database (so the number of generated document is rather small comparing with the initial size - as database size usually impact results a lot and we may have a slow down during the test). As per duration vs. number of iteration, no clear preference - but it is definitely better to run it for some time. Any idea how long it takes to run 10 times in this case? Probably shouldn't be less than 30 sec.

madelinezec: I guess I was using test and operation synonymously since each [test](https://github.com/mongodb/genny/blob/PERF-2729/src/workloads/execution/DensifyFillCombo.yml#L225-L267) contains [one operation](https://github.com/mongodb/genny/blob/PERF-2729/src/workloads/execution/DensifyFillCombo.yml#L242-L263). But this was incorrect of me. 

I'm gonna get back to you on how many documents we generate per operation in these tests when I figure out how to measure that

madelinezec: @apodelko I am looking at the csv I exported from genny. My understanding is that counters.n is documents created -- however each operation for every test has 0 counters.n? I could create a jstest for a similar database and run it and see from there how many docs are outputted?

madelinezec: [DensifyTimestampFillWithoutPartition](https://docs.google.com/spreadsheets/d/1DrV8okba0Qa10M6JBQMMEr-I48ZzSkUUWbs8_6z8QVo/edit?usp=sharing) [DensifyTimestampFillValue](https://docs.google.com/spreadsheets/d/1mzW5in4yzo_Z-P9EYGlXr_T0fVztjrM14xzX7CPDZy4/edit#gid=917721369)  [DensifyNumericLinearFillNumeric](https://docs.google.com/spreadsheets/d/1AXyHpFV7TOu0M3IS4kkZ4DCtch1R3DcNPowrJtNpngI/edit?usp=sharing)  [DensifyNumericFillWithPartition](https://docs.google.com/spreadsheets/d/1bbbhTIWRoTjmmcPUKa94jucse_LMGNm_24oBpFBksrU/edit?usp=sharing)

TedTuckman: Is generating a large number of documents a problem? Would it make more sense to keep the densify limit at the default, and lower the ranges we're hitting? The limit was set as Bora felt that it was a good representation of the "average" users's largest operation, but it was very vague.

src/workloads/execution/DensifyFillCombo.yml
TedTuckman: Not done or not pushed?

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,372 @@
...
+            field: "numeric",
+            range: {
+              bounds: "full",
+              step: *smallStep
+            },
+          },
+          $fill: {
+            output: {
+              toFillRandomType: {
+                value: "0"
```
madelinezec, line 350: @TedTuckman arbitrary $fill method here

src/workloads/execution/DensifyFillCombo.yml
madelinezec: @TedTuckman would that mean making explicit bounds, rather than just doing "full"?

TedTuckman: Either that or having fields with a smaller range for full. I don't feel that I have a good grasp on this discussion, so I leave it up to you guys

src/workloads/execution/DensifyFillCombo.yml
TedTuckman: Sorry, I meant the part about the stages interacting. Did I miss that part in the review?

src/workloads/execution/DensifyFillCombo.yml
```diff
@@ -0,0 +1,372 @@
...
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - Duration: 30 seconds
+    Database: *db
+    Operations:
+    - OperationMetricsName: TestDensifyNumericLinearFillNumeric
```
TedTuckman, line 335: This isn't linearFill right? This is arbitrary fill?

TedTuckman (approved):

src/workloads/execution/DensifyFillCombo.yml
apodelko: @madelinezec I believe that we have each operation on a separate line in these csv files. So, for example, DensifyTimestampFillWithoutPartition has 62 lines - so we had 62 operations there. Apparently it is not a document per operations if we needed to increase  InternalQueryMaxAllowedDensifyDocs to 100K...

apodelko: Actually in your succesful patch it appears that we have 513 operations (lines) for DensifyTimestampFillWithoutPartition - which match the reported OperationThroughput metric of 17.12 ops/sec (and 30 sec duration).

madelinezec:
@apodelko I was able to increase the initial number of documents inserted into the collection to 10k by reducing the number of partitions and increasing the step size from 3 to 10. We still need to increase InternalQueryMaxAllowedDensifyDocs to 100k, but his seemed better than starting off with just 100 docs in the collection.  Do these seem like reasonable parameters for measuring performance?

apodelko:
Sounds good. Could we run the patch again with these new parameters?

madelinezec:
> Sounds good. Could we run the patch again with these new parameters?

patch test: https://spruce.mongodb.com/version/61c21d643627e025ab0aef96/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

apodelko (approved):
LGTM

PR #594, opened 2021-12-16
TIG-3319 Document storage workloads

luke-pearson:


src/workloads/docs/RollingCollections.yml
```diff
@@ -1,7 +1,12 @@
 SchemaVersion: 2018-07-01
 Owner: Storage Engines
 Description: |
-  TODO: TIG-3319
+  A workload to test/document the RollingCollections actor which has two modes of operation a
+  a "Setup" operation which creates CollectionWindowSize collections and populates them with
```
ddanderson, line 5: An extra "a" -- "two modes of operation a a Setup...".  Maybe a comment or period after "operation".

src/workloads/scale/LargeScaleLongLived.yml
```diff
@@ -1,7 +1,9 @@
 SchemaVersion: 2018-07-01
 Owner: Storage Engines
 Description: |
-  TODO: TIG-3319
+  This workload consists of two phases intended to test the basic long lived reader writer actors
+  created for the large scale workload automation project. It creates a databse with 10K collections
```
ddanderson, line 5: databse spelling.

src/workloads/scale/LargeScaleSerial.yml
```diff
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: Storage Engines
 Description: |
-  TODO: TIG-3319
+  An example large scale workload file created to prove that such a thing was possible.
```
ddanderson, line 4: I was going to ask for more info on this rather than a one-liner, but I see we didn't create this, but we are marked as the owner...?  It seems to be related to LargeScaleParallel.yml (which we are not marked as the owner).

ddanderson (approved):
Looks good - a couple minor nits and some puzzlement about whether we should be documenting LargeScaleSerial.yml ?

src/workloads/contrib/historystore/eMRCfBench.yml
```diff
@@ -1,18 +1,16 @@
...
-# Currently the workload is split across several yml files because we need to script
-# some things between different parts.  This is the grow phase.  It churns our dataset
-# with a lot of updates, inserts, and deletes.  It intent is to keep the logical size
-# of the data set the same (i.e., same number of documents with the same large/small mix)
-# but to cause the WT history store to grow when we run it on a degraded PSA replica set.
+  Currently the workload is split across several yml files because we need to script
+  some things between different parts.  This is the grow phase.  It churns our dataset
+  with a lot of updates, inserts, and deletes.  It intent is to keep the logical size
+  of the data set the same (i.e., same number of documents with the same large/small mix)
+  but to cause the WT history store to grow when we run it on a degraded PSA replica set.
```
keitharnoldsmith, line 13: Ooops.  I looks like I did a cut-and-paste of this comment from the "Grow" phase of the test.  Could you change this to say:

Currently the workload is split across several yml files because we need to script some things between the different parts.  This is the benchmark phase. We run it both before and after growing the WT history store so we can evaluate the effect of a large history store.  The benchmark has four phases. Each phase performs one million operations with different request mixes:
 1. 100% Inserts
 2. 50% Updates, 50% Reads
 3. 10% Updates, 90% Reads
 4. 100% deletes

keitharnoldsmith (approved):
I appears that one of my original workload descriptions was out of date. I've left a comment with the correct description.  

I'll LGTM since I don't think I need to review your cut-and-paste of my new text :-)

src/workloads/scale/LargeScaleSerial.yml
luke-pearson: I was wondering this too, I don't think we are the original author so instead of updating the description I'll change the ownership to whichever team we think owns it?

PR #595, opened 2021-12-17
TIG-2046 Random sampler actor refactor and associated workload tuning

luke-pearson:


luke-pearson:
@rtimmons I've been planning on making this change to the workload for a while, essentially instead of using $sample in the aggregation pipeline we chose a random integer and find the relevant document via that integer as it's ID. This should result in a much faster random document lookup. 

I've also tuned the collection count and the number of collections traversed per scan as per an email thread a while back.

luke-pearson:
Oh I also have a slight dependency on #594 as it updates some descriptions which would get changed by this PR so I'll wait for that to get in first.

src/cast_core/src/RandomSampler.cpp
```diff
@@ -73,12 +78,14 @@ void RandomSampler::run() {
...
                 ? config->readWithScanOperation.start()
                 : config->readOperation.start();
-            int index = config->collections.size() > 1 ? config->integerDistribution(_random) : 0;
-            auto cursor = config->collections[index].aggregate(config->pipeline,
-                                                               mongocxx::options::aggregate{});
-            for (auto doc : cursor) {
+            int collId = config->collections.size() > 1 ? config->collectionDistribution(_random) : 0;
+            int recordId = config->documentDistribution(_random);
+            auto maybe_result = config->collections[collId].find_one(bsoncxx::builder::stream::document{} << "_id" << recordId << bsoncxx::builder::stream::finalize);
+            if (maybe_result) {
```
rtimmons, line 84: 💅🏽 `mabyResult` (camelCase)

src/cast_core/src/RandomSampler.cpp
```diff
@@ -73,12 +78,14 @@ void RandomSampler::run() {
...
                 statTracker.addDocuments(1);
-                statTracker.addBytes(doc.length());
+                statTracker.addBytes(maybe_result->view().length());
+            } else {
+                throw;
```
rtimmons, line 88: can we throw a more explicit exception perhaps `std::domain_error` with a `what` indicating the id of the document that was expected to exist but didnt'?

src/cast_core/src/RandomSampler.cpp
```diff
@@ -73,12 +78,14 @@ void RandomSampler::run() {
             auto statTracker = _activeCollectionScannerInstances > 0
                 ? config->readWithScanOperation.start()
                 : config->readOperation.start();
-            int index = config->collections.size() > 1 ? config->integerDistribution(_random) : 0;
-            auto cursor = config->collections[index].aggregate(config->pipeline,
```
rtimmons, line 77: Can we get rid of `config->pipeline`?

rtimmons (approved):
Few ideas/questions but overall code looks reasonable. Approved assuming questions/ideas are either addressed or don't make sense and disagreed with as appopriate.

src/cast_core/src/RandomSampler.cpp
luke-pearson: Fixed in 33cd8e4

src/cast_core/src/RandomSampler.cpp
luke-pearson: Fixed in 33cd8e4

src/cast_core/src/RandomSampler.cpp
luke-pearson: Fixed in 33cd8e4

PR #596, opened 2021-12-17
TIG-3517 Add workload for normalized scale 1 TPC-H Benchmark

alyacb:
Link to 🌲 patch containing changes to Genny, DSI, and the evergreen task itself can be found [here](https://spruce.mongodb.com/task/sys_perf_linux_standalone_tpch_1_normalized_patch_fcd93d3668831df603399b2132260635cfe89ec7_61bc6079d1fe07301bb14132_21_12_17_10_05_35/logs?execution=0&sortBy=STATUS&sortDir=ASC).

Current structure:
- `src/phases/tpch/normalized/` contains the query definitions, each in its own file (this could easily be refactored to combine all queries into one file, but each query is quite complex, so I leave this question up to review). The queries must exist separately as phases so that scales 1 and 10 can reuse the same pipelines with different timeouts and possibly query parameters.
- `src/workloads/tpch/normalized/1` contains the workload definitions for scale 1. Because the corresponding test control in dsi needs to do some preparation before each individual query, each workload definition must live in its own file. Scale 10 will have its own directory (~1), and the denormalized schema will have a sibling directory to the normalized schema.

Note:
- This included 20/22 queries, which were validated against the scale 1 validation parameters.
- The queries use the validation parameters defined in the TPC-H benchmark for scale 1; however, this is configurable, as scale 10 may need to use different parameters.
- For now, the workloads are only configured to run on the standalone variant because sharding scenarios need to be handled in a different way in the DSI test control.
- The timeout for each query will be updated based on the timing results obtained from evergreen.

### DIFF
--- a/src/phases/tpch/normalized/Q1.yml
+++ b/src/phases/tpch/normalized/Q1.yml
@@ -4,7 +4,7 @@ Description: |
   Run TPC-H query 1 against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no
   documents remain, which ensures that the query executes in its entirety.
 
-batchSize: &batchSize 101 # The default batch size.
+batchSize: &batchSize 101  # The default batch size.
 query1Delta: &query1Delta -90
 
 TPCHNormalizedQuery1:
@@ -18,9 +18,7 @@ TPCHNormalizedQuery1:
         aggregate: lineitem
         pipeline:
           [
-            {$match: {$expr: {$lte:
-              ["$l_shipdate", {$dateAdd: {startDate:
-                {$dateFromString: {dateString: "1998-12-01"}}, unit: "day", amount: {^Parameter: {Name: "Query1Delta", Default: *query1Delta}}}}]}}},
+            {$match: {$expr: {$lte: ["$l_shipdate", {$dateAdd: {startDate: {$dateFromString: {dateString: "1998-12-01"}}, unit: "day", amount: {^Parameter: {Name: "Query1Delta", Default: *query1Delta}}}}]}}},
             {$group: {
               _id: {l_returnflag: "$l_returnflag", l_linestatus: "$l_linestatus"},
               sum_qty: {$sum: "$l_quantity"},
@@ -37,20 +35,20 @@ TPCHNormalizedQuery1:
               avg_price: {$avg: "$l_extendedprice"},
               avg_disc: {$avg: "$l_discount"},
               count_order: {$count: {}}}},
-          {$project: {
-            _id: 0,
-            l_returnflag: "$_id.l_returnflag",
-            l_linestatus: "$_id.l_linestatus",
-            sum_qty: 1,
-            sum_base_price: 1,
-            sum_disc_price: 1,
-            sum_charge: 1,
-            avg_qty: 1,
-            avg_price: 1,
-            avg_disc: 1,
-            count_order: 1}},
-          {$sort: {l_returnflag: 1, l_linestatus: 1}},
-        ]
+            {$project: {
+              _id: 0,
+              l_returnflag: "$_id.l_returnflag",
+              l_linestatus: "$_id.l_linestatus",
+              sum_qty: 1,
+              sum_base_price: 1,
+              sum_disc_price: 1,
+              sum_charge: 1,
+              avg_qty: 1,
+              avg_price: 1,
+              avg_disc: 1,
+              count_order: 1}},
+            {$sort: {l_returnflag: 1, l_linestatus: 1}},
+          ]
         cursor: {batchSize: *batchSize}
         allowDiskUse: true
       verbosity:
--- a/src/phases/tpch/normalized/Q1.yml
+++ b/src/phases/tpch/normalized/Q1.yml
@@ -1,11 +1,11 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/product-query"
 Description: |
-  Run TPC-H query 1 against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no
+  Run TPC-H query 1 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no
   documents remain, which ensures that the query executes in its entirety.
 
 batchSize: &batchSize 101  # The default batch size.
-query1Delta: &query1Delta -90
+query1Delta: &query1Delta {^Parameter: {Name: "Query1Delta", Default: -90}}
 
 TPCHNormalizedQuery1:
   Repeat: 1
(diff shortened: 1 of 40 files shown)


jimoleary (commented):
I have an initial question about using an explain with runCommand.

src/phases/tpch/normalized/Q1.yml
```diff
@@ -0,0 +1,55 @@
...
+query1Delta: &query1Delta -90
+
+TPCHNormalizedQuery1:
+  Repeat: 1
+  Database: tpch
+  Operations:
+  - OperationMetricsName: Query1
+    OperationName: RunCommand
+    OperationCommand:
+      explain:
```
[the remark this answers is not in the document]
alyacb: This is used to ensure that the query is executed completely, returning all documents at once (as opposed to specifying an aggregate at the top level with a batchSize of how many documents we are expecting).

There is another example of this usage for a genny workload [here](https://github.com/mongodb/genny/blob/a4a13df7d33a85b29160cab8c88f994fe7a50209/src/workloads/execution/ExternalSort.yml#L4-L7).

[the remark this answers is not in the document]
aggrand: Might be worth a comment _explaining_ that!

src/phases/tpch/normalized/Q1.yml
```diff
@@ -0,0 +1,55 @@
...
+        aggregate: lineitem
+        pipeline:
+          [
+            {$match: {$expr: {$lte: ["$l_shipdate", {$dateAdd: {startDate: {$dateFromString: {dateString: "1998-12-01"}}, unit: "day", amount: {^Parameter: {Name: "Query1Delta", Default: *query1Delta}}}}]}}},
```
aggrand, line 21: These parameters are good, but I think it'll be easy for the fact that the query is parameterized on these values to be lost in the length of the query. Really a flaw in our parameterization syntax.

Would it make sense to move them to the top and use anchors to place them in the query? Like [here](https://github.com/mongodb/genny/blob/69ee0b0391fb35b46b6fbb8d8fae1928157c7169/src/phases/scale/LargeScalePhases.yml#L42).

src/phases/tpch/normalized/Q10.yml
```diff
@@ -0,0 +1,39 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-query"
+Description: |
+  Run TPC-H query 10 against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no
```
aggrand, line 4: I don't know a ton about the system under test, but is the numbering system used here (e.g. "query 6") a property of the type of query being run? Or is it a numbering system used just to organize this workload?

If the latter, would it make sense to have more descriptive names of these queries that describe what they are and why they're tested?

(I might just be totally misunderstanding this query though!)

src/workloads/tpch/normalized/1/Q22.yml
```diff
@@ -0,0 +1,34 @@
...
+
+Actors:
+- Name: TPCHNormalizedQuery22
+  Type: RunCommand
+  Database: tpch
+  Phases:
+  - LoadConfig:
+      Path: ../../../../phases/tpch/normalized/Q22.yml
+      Key: TPCHNormalizedQuery22
+      Parameters:  # This is a hack, because yml parses these strings into ints before outputting the pipeline.
```
aggrand, line 19: Oof what behavior would be preferred here? The transformation might also be happening internal to genny too, we have our own set of types that are interpreted from the yaml.

aggrand (approved):
Looks good as far as using Genny, can't comment on the workload itself too much. Just had a few suggestions / questions but nothing I'd block on!

src/workloads/tpch/normalized/1/Q1.yml
```diff
@@ -0,0 +1,28 @@
...
+  Type: RunCommand
+  Database: tpch
+  Phases:
+  - LoadConfig:
+      Path: ../../../../phases/tpch/normalized/Q1.yml
+      Key: TPCHNormalizedQuery1
+      Parameters:
+        Query1Delta: -90
+
+AutoRun:
```
aggrand, line 22: Given that these workloads all run in one DSI test_control, you might want to remove the `AutoRun` section here. I'd expect otherwise the autorun scheduler would schedule all these as separate workload tasks.

aggrand:
Given that the parameterization of the DSI `test_control` file implicitly assumes the locations of these workloads, it might be worth putting a readme describing the layout of the workloads, what that means, and how DSI invokes it. Normally it's relatively safe to move Genny workloads around in the workloads dir since autorun picks up on them anyway, so we wouldn't want someone to assume that here.

Also, I don't like the fact that this needs doing. It might be worth filing a TIG ticket for us to improve on this. (In addition to the ticket for us to improve the "canned workload" functionality.)

src/phases/tpch/normalized/Q1.yml
```diff
@@ -0,0 +1,55 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-query"
+Description: |
+  Run TPC-H query 1 against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no
+  documents remain, which ensures that the query executes in its entirety.
```
timourk, line 5: It would be quite useful to add as a comment the equivalent SQL query.

src/phases/tpch/normalized/Q1.yml
RuoxinXu, line 21: Should the normalized query here be identical to the query in the jstest [here](https://github.com/10gen/mongo/pull/2061/files#diff-ce54cc54db72ff4e2b0cf7b85042ba4565ec425b64e514a2508f7dc5f0d1487d)? I see there is slight difference.

src/phases/tpch/normalized/Q1.yml
RuoxinXu: Or maybe attach the link to the TPC-H Specification as you did in the jstest?  (http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf)

RuoxinXu (changes requested):

src/phases/tpch/normalized/Q1.yml
alyacb: Is there a difference other than the parametrization?

src/phases/tpch/normalized/Q1.yml
alyacb: I added a link to the TPC-H spec to each query file.

alyacb (commented):
I responded to all comments and pushed some suggested changes.

src/phases/tpch/normalized/Q10.yml
aggrand: Gotcha, thanks for explaining!

src/workloads/tpch/normalized/1/Q22.yml
aggrand: Oof gotcha. Not sure whether that's a yaml thing or a genny thing, or both. I don't suppose some combination of internal single quotes or escapes helps? Feel free to file a TIG ticket if this is a pain! We're going to be thinking a lot about syntax stuff soon anyway.

jimoleary (commented):
Overall looks good to me.

I have a question about noise and actual averages.

jimoleary (approved):
Based on the feedback in the DSI review, the noise is <= 5% so this should be ok.

timourk (approved):

src/phases/tpch/normalized/Q1.yml
RuoxinXu: Oh it's just a minor one where you use `$dateAdd` here but `$dateSubtract` in the jstest IIRC

RuoxinXu (approved):

src/phases/tpch/normalized/Q1.yml
alyacb: I'll update the integration test to match this.

alyacb:
Thank you for the reviews! Sending this to the commit queue.

alyacb:
evergreen merge

alyacb:
evergreen merge

PR #597, opened 2021-12-21
TIG-3684 Genny Crud Actor support Per Operation metrics

dalyd:


dalyd:
This change works for me. If you guys like it, I'll add some tests for it. Pretty small and self contained.

dalyd:
Hm, I don't see the best way to add tests, as the CrudActor tests use a runner. I could copy, cut down, and adjust the RunCommandActor_test.cpp. Thoughts?

rtimmons (approved):
I'm surprised this ask hasn't come up sooner. Seems like a good idea.

**BUT**: This means there are now three ways to specify the metrics name:

1. In `Phase: {MetricsName:}`
2. In `OperationName`
3. In `OperationMetricsName`

The case of 2 and 3 seem easy to reason about, but it's not clear what the user would expect to happen if they specify both 1 and 3, for example. Really specifying more than one of these seems like a source of user-error and confusion.

So while I think this functionality is certainly useful, I think we may have reached a bit of a tipping-point in terms of the syntax/configuration here. There may even be a metrics/context member-funciton/primitive we're missing to make this more intuitive and consistent for Actor authors and users alike.

IMHO: As long as it's well-documented, opt-in, and backward-compatible this is fine for now (imho). But either way I think it's worth a comment here and/or in examples & documentation that we're not super happy with the number of slightly different ways to configure how to configure metrics names in the Crud Actor.

Others are fine to disagree and push back harder.

And unfortunately I don't have a good suggestion for how to test this. Something our testing infrastructure might want to account for better in the future. Metrics are the ultimate goal of Genny Actors but we don't have great facilities to assert the interactions between Actors and metrics...

dalyd:
FWIW, I was planning to use  `Phase: {MetricsName:} `and `OperationMetricsName` at the same time. I expect that they will stack, giving something like {MetricsName}.{OperationMetricsName}.{PhaseNumber}

I'll test it out and see what happens. 
Roger on adding more comments and documentation.

dalyd:
Hm, I tried using both and it doesn't work and that makes sense based on the code, although the results themselves don't make sense to me. 

With `OperationMetricsName: OtherInsertAndUpdate` I get
`CrudActor.OtherInsertAndUpdate.ftdc`
That makes sense to me. If I also set at the phase level `MetricsName: Phase3` I get
`CrudActor.Phase3.2.ftdc` but I have no idea how the `.2` is getting in there. 

It looks like the CrudActor is creating an operation in the phaseConfig. Could I use the actor operation constructor and just pass in the right stuff?

dalyd:
Alright, this last patch may be awful, but it now does what I expect with regards to combining. I now get 
```
-rw-r--r--  1 daviddaly  staff  342 Dec 21 12:02 CrudActor.Crud.ftdc
-rw-r--r--  1 daviddaly  staff  270 Dec 21 12:02 CrudActor.InsertAndUpdate.ftdc
-rw-r--r--  1 daviddaly  staff  276 Dec 21 12:02 CrudActor.Phase3.2.ftdc
-rw-r--r--  1 daviddaly  staff  264 Dec 21 12:02 CrudActor.Phase3.OtherInsertAndUpdate.ftdc
-rw-r--r--  1 daviddaly  staff  185 Dec 21 12:02 CrudActor.drop.ftdc
-rw-r--r--  1 daviddaly  staff  278 Dec 21 12:02 CrudActor.insertMany.ftdc
drwxr-xr-x  7 daviddaly  staff  224 Dec 21 12:02 internal
```

Before the last change I got 
```
-rw-r--r--  1 daviddaly  staff  342 Dec 21 11:07 CrudActor.Crud.ftdc
-rw-r--r--  1 daviddaly  staff  269 Dec 21 11:07 CrudActor.InsertAndUpdate.ftdc
-rw-r--r--  1 daviddaly  staff  345 Dec 21 11:07 CrudActor.Phase3.2.ftdc
-rw-r--r--  1 daviddaly  staff  185 Dec 21 11:07 CrudActor.drop.ftdc
-rw-r--r--  1 daviddaly  staff  279 Dec 21 11:07 CrudActor.insertMany.ftdc
drwxr-xr-x  7 daviddaly  staff  224 Dec 21 11:07 internal
```

src/cast_core/src/CrudActor.cpp
```diff
@@ -1455,6 +1455,20 @@ struct CrudActor::PhaseConfig {
         // Node is convertible to bool but only explicitly.
         const bool perPhaseMetrics = bool(phaseContext["MetricsName"]);
         auto opCreator = op->second;
+        // Special case if we have set MetricsName at the phase level, and OperationMetricsName on
+        // the operation.
+        if (auto metricsName = phaseContext["MetricsName"].maybe<std::string>())
```
rtimmons, line 1463: Please put below block in braces

rtimmons: This _may_ be clearer as:

```c++
if (auto metricsName = phaseContext["MetricsName"].maybe<std::string>();
     metricsName && metricsName != opName) {
    
}
```

rtimmons: Also..wait..we have `metricName` (singular) and `metricsName` (plural)? That's just waiting for a typo or mis-read.

rtimmons (approved):

src/cast_core/src/CrudActor.cpp
dalyd: I tried your syntax. I wasn't sure how to do it :) 
Also, good call on metricName/metricsName. Updated. My mind really wanted it to be singular, but it's plural in all existing places.

dalyd:
evergreen merge

PR #602, opened 2022-01-09
TIG-3702 ConfigSvrRetriever in Genny's Topology explorer never actually visits configsvers

aggrand:
Patch with affected workload: https://spruce.mongodb.com/version/61db24dd7742ae52361483a5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


aggrand:
Also a sys-perf patch on standalone: https://spruce.mongodb.com/version/61db279f2a60ed121efed998/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

vrachev (commented):
I'm having trouble grokking why the current code doesn't work / what the change is fixing. Mind explaining in a bit more detail?

aggrand:
The original fix was flawed, sorry about that!

The root issue is that the config sets sometimes have different names. Patch with affected workload: https://spruce.mongodb.com/version/61df1a10850e6151031f144e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/gennylib/src/topology.cpp
```diff
@@ -144,7 +144,7 @@ void Topology::computeDataMemberConnectionStrings(DBConnection& connection) {
 
     std::unique_ptr<ReplSetDescription> desc;
     auto setName = res.view()["setName"].get_utf8().value;
-    if (setName == "configSet") {
+    if (setName == "configSet" || setName == "configRepl") {
         desc = std::make_unique<ConfigSvrDescription>();
     } else {
         desc = std::make_unique<ReplSetDescription>();
```
vrachev, line 150: I can't comment on the exact line, but there's a few more places where we do the `setname` comparison that also need to be updated

vrachev (changes requested):
Thanks for the fix! I have another thought, but feel free to push back if it's not possible.

src/gennylib/src/topology.cpp
```diff
@@ -144,7 +144,7 @@ void Topology::computeDataMemberConnectionStrings(DBConnection& connection) {
 
     std::unique_ptr<ReplSetDescription> desc;
     auto setName = res.view()["setName"].get_utf8().value;
-    if (setName == "configSet") {
+    if (setName == "configSet" || setName == "configRepl") {
```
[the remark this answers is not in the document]
aggrand: Unfortunately, I think that's [derived](https://github.com/mongodb/mongo/blob/503acbbdd3e02fabb2fbc38c8e499df1d7559500/src/mongo/shell/servers.js#L948) from the shell intercepting the startup options for the server, which isn't something we have available here.

src/gennylib/src/topology.cpp
aggrand: Great catch! Updated.

src/gennylib/src/topology.cpp
```diff
@@ -159,7 +159,7 @@ void Topology::computeDataMemberConnectionStrings(DBConnection& connection) {
         for (auto member : hosts_view) {
             MongodDescription memberDesc;
             memberDesc.mongodUri = nameToUri(std::string(member.get_utf8().value));
-            if (setName == "configSet") {
+            if (setName == "configSet" || setName == "configRepl") {
```
vrachev, line 162: (nit) Maybe move these into a variable. Also fine if you keep as is.

vrachev (approved):
LGTM, but can you run another patch build with the latest change? Also had a nit, but not a blocker

aggrand:
Had some weird results in another patch run. Doing another round of testing.

aggrand:
evergreen merge

aggrand:
evergreen merge

aggrand:
evergreen merge

PR #604, opened 2022-01-12
TIG-3662 Create HTTP actor

richardsamuels:
I'm not sure if the appropriate way to document this is with the SampleHttpClient, as shown below, or just with text documentation. Please let me know your opinion on that.


src/cast_core/include/cast_core/actors/SampleHttpClient.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <gennylib/Actor.hpp>
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+
+#include <metrics/metrics.hpp>
+
+namespace genny::actor {
+
+/**
```
guoyr, line 30: All the boilerplate documentation can be removed

src/cast_core/include/cast_core/http.hpp
```diff
@@ -0,0 +1,85 @@
...
+    namespace ssl = net::ssl;       // from <boost/asio/ssl.hpp>
+    using tcp = net::ip::tcp;       // from <boost/asio/ip/tcp.hpp>
+
+    net::io_context ioc;
+
+    // The SSL context is required, and holds certificates
+    ssl::context ctx(ssl::context::tlsv11);
+
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32) && !defined(__CYGWIN__)
+    // TODO This does not work on Windows. If we ever care,
```
guoyr, line 46: I think you can remove the word "TODO", we don't perf on Windows

src/cast_core/src/SampleHttpClient.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+            // Gracefully close the stream
+            boost::beast::error_code ec;
+            stream.shutdown(ec);
+
+            try {
+
+                config->collection.insert_one(document.view());
+
+                inserts.addDocuments(1);
```
guoyr, line 249: instead of the `insert_one()` call, let's put the boost stuff here in the try block, then call `inserts.success()`. This will add a metric for each https call. The `addBytes()` and `addDocuments()` calls can be removed too since they're not applicable.

guoyr: Also you probably want to rename the `inserts` metrics context to something with http in the name.

src/cast_core/test/SampleHttpClient_test.cpp
```diff
@@ -0,0 +1,78 @@
...
+// an exercise in reading and testing your Actor. ⚠️
+//
+
+TEST_CASE_METHOD(MongoTestFixture, "SampleHttpClient successfully connects to a MongoDB instance.",
+          "[standalone][single_node_replset][three_node_replset][sharded][SampleHttpClient]") {
+
+    dropAllDatabases();
+    auto db = client.database("mydb");
+
+    NodeSource nodes = NodeSource(R"(
```
guoyr, line 44: Let's remove this file for now since the new actor doesn't have an associated workload. We can sync up with stakeholders first to see what kind of unit testing makes sense.

It may be tricky to write a test based on this template since this runs on Windows by default.

src/workloads/docs/SampleHttpClient.yml
```diff
@@ -0,0 +1,28 @@
+SchemaVersion: 2018-07-01
+Owner: TODO put your github team name here e.g. @10gen/dev-prod-tips
+Description: |
+  TODO describe your workload. For an example description, check out
+  src/workloads/selftests/GennyOverhead.yml.
+
+# TODO: delete this file or add a meaningful workload using or
```
guoyr, line 7: delete

src/cast_core/src/SampleHttpClient.cpp
```diff
@@ -0,0 +1,284 @@
...
+                //
+                BOOST_THROW_EXCEPTION(MongoException(e, document.view()));
+            }
+        }
+    }
+}
+
+SampleHttpClient::SampleHttpClient(genny::ActorContext& context)
+    : Actor{context},
+      _totalInserts{context.operation("Insert", SampleHttpClient::id())},
```
guoyr, line 266: Update name here too

guoyr (approved):
Lots of boilerplate code to delete. Otherwise LGTM! Happy to take another pass after the updates if you'd like.

src/cast_core/src/SampleHttpClient.cpp
```diff
@@ -0,0 +1,129 @@
...
+      // instance for each Phase.
+      //
+      _loop{context, (*_client)[context["Database"].to<std::string>()], SampleHttpClient::id()} {}
+
+namespace {
+//
+// This tells the "global" cast about our actor using the defaultName() method
+// in the header file.
+// Do not use simple-beast-client on macOS
+#ifndef __APPLE__
```
guoyr, line 125: 👍

guoyr (approved):
LGTM!

src/cast_core/include/cast_core/http.hpp
```diff
@@ -0,0 +1,85 @@
...
+    using tcp = net::ip::tcp;       // from <boost/asio/ip/tcp.hpp>
+
+    net::io_context ioc;
+
+    // The SSL context is required, and holds certificates
+    ssl::context ctx(ssl::context::tlsv11);
+
+#if defined(WIN32) || defined(_WIN32) || defined(__WIN32) && !defined(__CYGWIN__)
+    // This does not work on Windows. If we ever care,
+    // see https://stackoverflow.com/a/40710806
```
rtimmons, line 47: It seems like the SO question has the solution though. Did that not work?
It seems reasonable to leave this as-is since we don't run CI on Windows or readily have Windows dev boxes at our disposal. But if include additional context about why we didn't pursue this it will help future maintainers come to the same "yup not worth it" conclusion that we did here.

src/cast_core/src/SampleHttpClient.cpp
rtimmons, line 125: I need to push back here.

1. We've jumped through enormous hoops to make all of Genny cross-platform and easy to develop on the standard-issue macbooks provided by the company plus a number of other reasonable developer platforms like arch and Ubuntu although we don't run those in production.
5. Our ["build and install" page](https://github.com/mongodb/genny/blob/6a2a0648f36b02b4121d19789346a7a1ebba110a/docs/setup.md#build-and-install) says we support macOS. (It also says we support Windows which may in fact be true since we're pretty by-the-book with our code.)
2. The toolchain is built for macOS and other such platforms at considerable development and CI build cost.
3. The CMake is standard enough to be used by all manner of IDEs
4. We only use standards-compliant C++17 and don't use compiler-specific `#pragma`s for include guards.
6. We have explicit recommendations to avoid `#ifdef` in the [contributions guide](https://github.com/mongodb/genny/blob/master/CONTRIBUTING.md#dont-get-cute).
7. The reason why this code doesn't work on macOS isn't well-understood according to my reading of this situation. Does this extend to the other platforms that we support as well? This means there could be a latent bug in the code (or the libraries we're using) that could surface at some unknown time on some arbitrary platform in the future.  I.e., are we relying on undefined or accidental behavior of linux? Or is it a "leak" in the toolchain that causes us to link to a system library that behaves differently on macOSand/or other systems?

This ["breaks the first window"](https://www.psychologytoday.com/us/basics/broken-windows-theory) of the consistent cross-platform nature of the product.

Our "definition of done" for Genny work includes ensuring that all features work across all the platforms we support. If this weren't the case we'd have abandoned macOS, Arch, RHEL, and ubuntu long ago since our actual perf tests only need to run on Amazon-Linux... (None of this is explicit anywhere so this mis-reading of the situation is totally understandable. I'll cut a ticket to ensure our docs require x-platform).

Of course this must be balanced with the need to wrap up this work and move on to other things.

May I suggest:

1. Time-boxing a debugging session with a macOS debugger to chase down what the underlying issue actually is. I had to do this with the Catch2 integration and would be happy to provide a hand or a shoulder to cry on. If we've given sufficient effort into debugging we can provide additional context and justification in the code so that someone who has more time/energy in the future can pick up the work if/when it's deemed necessary.
2. Explicitly not supporting DNS and requiring URLs to be IP based. This may not work for the endpoints in question, but it does explicitly exclude known edge-cases.
3. Reworking the way we invoke DNS to require users to provide the DNS mappings in some manner. (I have no idea how feasible this is, but ideally the internals would handle DNS via vtables which we could hook into.)

Other options exist too of course. Not trying to be unreasonable here. I'm happy to discuss further on slack/zoom.

src/third_party/cppcodec/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/catch"]
+	path = test/catch
+	url = https://github.com/catchorg/Catch2.git
```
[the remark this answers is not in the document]
richardsamuels: No, the test suite for cppcodec requires catch, and the vendored it as a submodule. We don't run their test suite. (I copied the cppcodec repo w/o change into this directory)

src/cast_core/src/SampleHttpClient.cpp
guoyr: Hey Ryan, I'm going to continue the conversation around macOS support as part of https://jira.mongodb.org/browse/TIG-3843 and resolve this thread. The code currently builds on macOS and anyone not developing this actor would not be affected. Richard mentioned the IFDEF has been removed as well.

rtimmons (approved):
LGTM. Please add a TODO to the http client somewhere with a link to the bug ticket to fix dns resolving on macOS. Good work getting this in - fighting CMake and the toolchain are known sore spots and we didn't expect to hit dns issues!

PR #607, opened 2022-01-19
TIG-3391 Upgrade C++ Driver for Genny

aggrand:
I think we can expect RHEL 7 build to fail due to to the errors on the toolchain, but that's unrelated to our recent work and shouldn't affect production anyway because of [TIG-2860](https://jira.mongodb.org/browse/TIG-2860). I think we should go ahead with the upgrade because [TIG-3753](https://jira.mongodb.org/browse/TIG-3753) is affecting production anyway and that fix would be lumped into this upgrade. 

If this is agreed to merge, I can cut a followup ticket to either deal with the RHEL7 issue or just stop building the toolchain on it. Based on a conversation with @richardsamuels  it seems to be a rhel licensing issue which we'll either have to file a dependent build ticket for, or else find a hacky workaround.

### DIFF
--- a/src/lamplib/src/genny/toolchain.py
+++ b/src/lamplib/src/genny/toolchain.py
@@ -162,7 +162,7 @@ class ToolchainDownloader(Downloader):
     # If we were 💅 we could do the string logic here in python, but we're not that fancy.
     #
 
-    TOOLCHAIN_BUILD_ID = "82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21"
+    TOOLCHAIN_BUILD_ID = "c3ecb94b7ec0c080dd71253722b75fe1b94f86f1_22_01_19_17_37_44"
     TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split("_")[0]
     TOOLCHAIN_ROOT = "/data/mci"  # TODO BUILD-7624 change this to /opt.
 
--- a/src/lamplib/src/genny/toolchain.py
+++ b/src/lamplib/src/genny/toolchain.py
@@ -28,14 +28,14 @@ def _create_compile_environment(
 
     # For cmake and ctest
     cmake_bin_relative_dir = {
-        "linux": "downloads/tools/cmake-3.13.3-linux/cmake-3.13.3-Linux-x86_64/bin",
-        "osx": "downloads/tools/cmake-3.13.3-osx/cmake-3.13.3-Darwin-x86_64/CMake.app/Contents/bin",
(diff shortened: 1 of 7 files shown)


aggrand:
Patch sanity-checking an affected task: https://spruce.mongodb.com/version/61e864122fbabe701b13a434/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

aggrand:
Also a patch running microbenchmarks to make sure we're fine without rhel7 (the historic use of it, nowadays seems to only be rhel6)

vrachev (approved):
rhel7 not working should break m60 I think ([based on this](https://github.com/10gen/dsi/blob/41f94803c9287ec124002375cf29c56c45565ff7/configurations/workload_setup/workload_setup.common.yml#L44-L45)). I'm ok with fixing that in a later ticket since we need to unblock genny users now.

richardsamuels (approved):

aggrand:
Another patch after some updates: https://spruce.mongodb.com/version/61faf40757e85a0bcfa14097/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/lamplib/src/genny/toolchain.py
```diff
@@ -157,12 +157,12 @@ class ToolchainDownloader(Downloader):
     # These build IDs are from the genny-toolchain Evergreen task.
     # https://evergreen.mongodb.com/waterfall/genny-toolchain
     # Find a compile task (for any build variant) and modify the URL:
-    # genny_toolchain_amazon2_t_compile_cf1c50fab5291cd583a5036dc9ade265bca165a3_21_10_20_20_05_26
-    # =>                                cf1c50fab5291cd583a5036dc9ade265bca165a3_21_10_20_20_05_26
+    # genny_toolchain_archlinux_t_compile_82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
+    # =>                                  82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
```
richardsamuels, line 161: Random question, but why did this change to something else? Are we keeping this up to date with `TOOLCHAIN_BUILD_ID`?

richardsamuels (approved):
LGTM aside from my comment

src/lamplib/src/genny/toolchain.py
aggrand: It changed during the unrevert. I didn't think we were keeping it up-to-date, but I can do that.

aggrand:
evergreen merge

PR #613, opened 2022-02-04
PERF-2789 Add $lookup pushdown workloads

yun-soo:
Document: https://docs.google.com/document/d/1KijhnCvbfhfWtZJXUSrreNK-vcke7xofdkuUrVrnjDo/edit#

### DIFF
--- a/src/phases/execution/LookupCommands.yml
+++ b/src/phases/execution/LookupCommands.yml
@@ -1,0 +1,140 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  Defines common configurations for 'LookupSBEPushdown' and 'LookupSBEPushdownMisc' workloads.
+
+# Defines value generators.
+
+# Generates a random integer value from 1 to 'card' parameter of which the default is 9500.
+RandIntUpToTemplate: &RandIntUpTo {^RandomInt: {min: 1, max: {^Parameter: {Name: "card", Default: 9500}}}}
+# Generates a monotonically increasing integer value from 1 to 'card' parameter of which the default
+# is 9500.
+IncIntUpToTemplate: &IncIntUpTo {^Inc: {start: 1, end: {^Parameter: {Name: "card", Default: 9500}}}}
+# Generates a random 32bit integer value.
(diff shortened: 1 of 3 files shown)


src/phases/execution/LookupCommands.yml
```diff
@@ -0,0 +1,140 @@
...
+Description: |
+  Defines common configurations for 'LookupSBEPushdown' and 'LookupSBEPushdownMisc' workloads.
+
+# Defines value generators.
+
+# Generates a random integer value from 1 to 'card' parameter of which the default is 9500.
+RandIntUpToTemplate: &RandIntUpTo {^RandomInt: {min: 1, max: {^Parameter: {Name: "card", Default: 9500}}}}
+# Generates a monotonically increasing integer value from 1 to 'card' parameter of which the default
+# is 9500.
+IncIntUpToTemplate: &IncIntUpTo {^Inc: {start: 1, end: {^Parameter: {Name: "card", Default: 9500}}}}
```
dalyd, line 12: I don't think the ^Inc generator has an end field. We could add one. What should it do when it gets there? Should it wrap around back to start? https://github.com/mongodb/genny/blob/2ed9a81c20ef0f4a76420f67f94017a9b3dd8365/src/value_generators/src/DocumentGenerator.cpp#L1017

src/workloads/execution/LookupSBEPushdownMisc.yml
```diff
@@ -0,0 +1,358 @@
...
+  This workload conducts testing for the following test matrix:
+  - LocalCollectionCardinality: 100k / 300k / 500k / 700k
+  - DocumentSize: Small document size vs Large document size
+  - JoinFieldType: string type / sub-field type / array type
+  - NumberOfLookups: 1 / 2 / 3
+
+  But this workload does not try to measure performances for full combination of the above test
+  dimensions to avoid too lengthy test execution.
+
+  While measuring performances, this workload collects numbers for both the SBE HJ and the classic
```
dalyd, line 13: I'd encourage simplifying the workload to just do one at a time, and we can cover the classic versus SBE with different variants or patch tests as needed.

PR #616, opened 2022-02-08
Fix BF-23993

madelinezec:
A stark improvement was observed after removing some tasks from DensifyNumeric. After reviewing densify behavior, we realized we need to add back those tasks for fill with full bounds. 

This is the BF: https://jira.mongodb.org/browse/BF-23993

This is this PR's patch:  https://spruce.mongodb.com/version/620297e99ccd4e4171723f35/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/DensifyNumeric.yml
+++ b/src/workloads/execution/DensifyNumeric.yml
@@ -127,6 +127,36 @@ Actors:
           }
         }]
         cursor: {batchSize: *batchSize}
+    - OperationMetricsName: FullByPartitionSmallStep
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [{
+          $densify: {
+            field: *field,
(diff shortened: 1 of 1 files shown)


src/workloads/execution/DensifyNumeric.yml
```diff
@@ -127,6 +127,36 @@ Actors:
           }
         }]
         cursor: {batchSize: *batchSize}
+    - OperationMetricsName: FullByPartitionSmallStep
```
TedTuckman, line 130: These are copied from what was removed right?

TedTuckman (approved):
Adding these back looks good, but a couple meta points.

1. I believe this has to be merged under a PERF ticket. I think you can just file one, assign it to yourself, then change the PR title.
2. I'm still a little suspicious about this whole thing. Removing individual tests shouldn't cause other tests to speed up. Did you get any more clarity on this?

PR #617, opened 2022-02-09
PERF-2796 New workload to measure effect of constant folding optimizations

davish:


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -982,17 +982,59 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+    ConcatGenerator(const Node& node,
+                    GeneratorArgs generatorArgs,
+                    std::map<std::string, Parser<UniqueAppendable>> parsers)
+        : _rng{generatorArgs.rng}, _id{generatorArgs.actorId} {
+        if (!node["arrays"].isSequence()) {
+            std::stringstream msg;
+            msg << "Malformed node for join array. Not a sequence " << node;
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg.str()));
+        }
+        for (const auto&& [k, v] : node["arrays"]) {
```
dalyd, line 997: I think you should create a function UniqueGenerator<bsoncxx::array::value> BsonArrayGenerator similar to IntGenerator https://github.com/mongodb/genny/blob/89ce68b88ad27c35f80f98cf11078564e20c74c1/src/value_generators/src/DocumentGenerator.cpp#L1373, doubleGenerator, etc. 

Then you can pass each element in node["arrays"] to that function. That will enable nested concat generators, and also support the two generators here. And if anyone adds a new array generator you will automatically be able to use it within a concat generator.

dalyd (commented):
Comments just on the new generator.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -551,6 +551,40 @@ Tests:
     - a: "Hello random person: Huey"
     - a: "Hello random person: IonKmglImH"
 
+  - Name: Concat two singletons
+    GivenTemplate:
+      a: {^Concat: {arrays: [["It"], ["worked"]]}}
```
aggrand, line 556: Nice! Thanks for adding so many tests to it! Can you also add an example and basic docs to the `Generators.yml` to show this off?

aggrand (approved):
+1 to @dalyd 's point but looking good! Not much comment on the workload itself; seems fine w.r.t. genny usage!

src/value_generators/test/DocumentGeneratorTestCases.yml
davish: Done!

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1507,6 +1503,39 @@ UniqueGenerator<std::string> stringGenerator(const Node& node, GeneratorArgs gen
...
+/**
+ * @param node
+ *   Any node type that evaluates to an array (array literal, literal array of generators, ^Concat,
+ * ^Array)
+ * @return
+ *   The proper generator based on the parsed input.
+ */
+UniqueGenerator<bsoncxx::array::value> bsonArrayGenerator(const Node& node,
+                                                          GeneratorArgs generatorArgs) {
+    // Set of parsers to look when we request an double parser
```
dalyd, line 1515: double parser?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1507,6 +1503,39 @@ UniqueGenerator<std::string> stringGenerator(const Node& node, GeneratorArgs gen
...
+    const static std::map<std::string, Parser<UniqueGenerator<bsoncxx::array::value>>> arrayParsers{
+        {"^Concat",
+         [](const Node& node, GeneratorArgs generatorArgs) {
+             return std::make_unique<ConcatGenerator>(node, generatorArgs);
+         }},
+        {"^Array",
```
dalyd, line 1521: I think you also want LiteralArrayGenerator for ^Verbatim keyword.

src/value_generators/src/DocumentGenerator.cpp
davish: Deleted

src/value_generators/src/DocumentGenerator.cpp
davish: Added some tests to make sure this works properly.

jimoleary (approved):

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -551,6 +551,56 @@ Tests:
...
+    GivenTemplate:
+      a: {^Concat: {arrays: [{^Verbatim: [{^RandomInt: "Hello"}]}, ["World"]]}}
+    ThenReturns:
+    - a: [{^RandomInt: "Hello"}, "World"]
+
+  - Name: Concat with ^Verbatim object is invalid.
+    GivenTemplate:
+      a: {^Concat: {arrays: [{^Verbatim: {^RandomInt: "Hello"}}, ["World"]]}}
+    ThenThrows: InvalidValueGeneratorSyntax
+
```
nzolnierzmdb, line 603: My admittedly brief reading of the code is that the one case you're missing is nested concats? Is that allowed?

src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -0,0 +1,113 @@
...
+  This workload tests the impact of the constant folding optimization in the aggregation pipeline to better
+  understand the impact of SERVER-63099 and hopefully mitigate the performance regression with other optimizations.
+
+  The first pipeline is the control and should be unaffected by changes to the associativity of multiplication in the optimizer
+  because it only contains constants.
+
+  The three pipelines in this workload that contain a fieldpath should have different performance characteristics depending
+  on the level of associativity in the optimizer.
+
+  - Full associativity (current situation on master):
```
nzolnierzmdb, line 14: nit: I'd refrain from saying "current situation", since its bound to get stale (for instance even after SERVER-63099)

src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -0,0 +1,113 @@
...
+
+  - No associativity (simple correctness fix):
+    All three pipelines should have similar performance that is significantly worse than full associativity, because
+    each document in the pipeline will require 1000 multiplication operations because no constant folding will happen
+    with $multiply marked as non-associative.
+
+  - Left associativity (implementation of a correct optimization):
+    StartingFieldpath should see no regression from the "no associativity" case.
+    MiddleFieldpath should see a speedup that puts its performance in between "no associativity" and "full associativity".
+    TailingFieldpath should speed up back to its performance in the "full associativity" case.
```
nzolnierzmdb, line 26: Hmm I'm confused on this one, why isn't it more comparable to middle field path?

src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -0,0 +1,113 @@
...
+
+
+GlobalDefaults:
+  dbname: &db test
+  batchSize: &batchSize 3000
```
nzolnierzmdb, line 31: Arbitrary value? Or do you want to test getMore as well?

src/phases/execution/RunLargeArithmeticOp.yml
```diff
@@ -0,0 +1,25 @@
...
+  Operations:
+  - OperationMetricsName: {^Parameter: {Name: "Name", Default: "RunMultiply"}}
+    OperationName: RunCommand
+    OperationCommand:
+      aggregate: {^Parameter: {Name: "Collection", Default: "Collection0"}}
+      pipeline: [
+        {
+          $group: {
+            _id: {
+              $multiply: {^Parameter: {Name: "Expression", Default: []}}
```
nzolnierzmdb, line 20: Curious why the $group? Would it be sufficient to project a new field with the computation?

src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -0,0 +1,113 @@
...
+  - Repeat: 1
+  - *Nop
+  - Repeat: 1
+  - *Nop
+  - Repeat: 1
+  - *Nop
+  - Repeat: 1
+  - *Nop
+
+- Name: Multiply
```
nzolnierzmdb, line 75: nit: I think this may be a better name for the shared actor, since its the one who's actually running the $mult

nzolnierzmdb (approved):
LGTM! My comments are all either nits or optional but let me know if you want another look

src/workloads/execution/ConstantFoldArithmetic.yml
davish: The TailingFieldpath case looks like `[c1, c2, c3, c4, "$v"]` and optimizes to `[c', "$v"]`, which means that only one multiplication is performed per document. In the full associative and commutative case, that same expression would be optimized to `["$v", c']` (commutative constants are arbitrarily put at the end right now) which is also one multiplication per document

src/workloads/execution/ConstantFoldArithmetic.yml
davish: It's an arbitrary value, I wasn't sure if there was any issues with extremely large batch sizes, but I'll poke around the other workloads and see.

src/phases/execution/RunLargeArithmeticOp.yml
davish: This matches the BFs that were reported by the fuzzer -- right now $project is pushed down fully to SBE and that might be a reason there are no fuzzer BFs with $project. One thing I'm thinking about is if it makes sense to run this workload on a variant that forces the classic engine. One of my todos still is reaching out to QE to see how SBE handles this and if it makes sense to be testing SBE too.

src/value_generators/test/DocumentGeneratorTestCases.yml
davish: done

src/workloads/execution/ConstantFoldArithmetic.yml
davish: done

src/workloads/execution/ConstantFoldArithmetic.yml
davish: Done

PR #618, opened 2022-02-18
PERF-2754 Add perf tests for lastpoint optimization use case 1

kaywux:
From local testing, going from `featureFlagLastPointQuery=false` to `true` results in a 622x speedup. Will update with preliminary Evergreen results when they are available.

### DIFF
--- a/src/workloads/execution/TimeSeriesLastpoint.yml
+++ b/src/workloads/execution/TimeSeriesLastpoint.yml
@@ -121,6 +121,7 @@ AutoRun:
     mongodb_setup:
       $eq:
       - standalone
+      - standalone-all-feature-flags
       - replica
       - replica-all-feature-flags
       - shard-lite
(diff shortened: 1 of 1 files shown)


starzia (approved):
Nice work, Katherine!  It's helpful to see how that ad hoc JS perf test gets translated to Genny.

rusabd1 (approved):

kaywux:
Results from the Evergreen patch build:
- On standalone and replica set environments, this optimization has around a 500x speedup
- On the sharded cluster environment, there is around a 400x speedup (still great but perhaps a decrease from above due to added overhead with sharding). Note that this is with an unsharded collection, we should add additional tests for sharded collections as the perf results are expected to be significantly different.

Patch build: https://spruce.mongodb.com/version/62166298d6d80a05849d8be0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #620, opened 2022-02-23
TIG-3651 Respect OnSession for operations inside CrudActor's WithTransaction op

richardsamuels:
Testing remains a sore point for this.
- There are no unit tests for actors at this point, so I'd be starting from scratch there. This would be incredibly painful
- The changes I've made to the integration test are at the very least an assertion that OnSession won't crash.
- There does not appear to be a reasonable way to assert

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1008,7 +1008,7 @@ private:
         }
         auto createOp = opConstructor->second;
         auto& yamlCommand = txnOp["OperationCommand"];
-        // operations can override withTransaction level
+        // operations can override withTransaction's OnSession value
         bool onSession = yamlCommand["OnSession"].maybe<bool>().value_or(_onSession);
         _txnOps.push_back(createOp(yamlCommand, onSession, _collection, _operation, context, id));
     }
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -868,6 +868,8 @@ private:
 };
 
 /**
+ * Warning: Commands that should be apart of the transaction must declare
+ * the following value under OperationCommand: `OnSession: true`
  * Example usage:
  *    Operations:
  *    - OperationName: startTransaction
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -1009,6 +1009,10 @@ private:
         auto createOp = opConstructor->second;
         auto& yamlCommand = txnOp["OperationCommand"];
         // operations can override withTransaction's OnSession value
+        // This behavior is hard to test.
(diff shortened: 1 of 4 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -1008,7 +1008,9 @@ struct WithTransactionOperation : public BaseOperation {
         }
         auto createOp = opConstructor->second;
         auto& yamlCommand = txnOp["OperationCommand"];
-        _txnOps.push_back(createOp(yamlCommand, _onSession, _collection, _operation, context, id));
+        // operations can override withTransaction level
+        bool onSession = yamlCommand["OnSession"].maybe<bool>().value_or(_onSession);
+        _txnOps.push_back(createOp(yamlCommand, onSession, _collection, _operation, context, id));
```
rtimmons, line 1013: Since this isn't tested (admittedly hard to do) let's add a comment here for future maintainers. Something like

```
// This behavior is hard to test.
// The CrudActorYamlTests suite exercises both branches of this boolean but doesn't assert the behavior.
// Be careful when making changes around this code.
```

Also I wonder if it would be possible to test by hand. Write a dummy workload that does a whole bunch of inserts on a session and compare the behavior with a similar workload without the session? It's almost certainly not going to be worth the effort for this one bug, but that may be an interesting avenue to explore in the future.

rtimmons (approved):

src/cast_core/src/CrudActor.cpp
rtimmons: And yes, we intentionally did not invest in lots of machinery to unit-test Actors. Feedback or ideas welcome, but it quickly went into template/macro meta-programming hell to allow for monkey-patching at test time. It was beyond my expertise so just decided to have an integration testing setup that was as robust and easy-to-use as possible. Transactions and failure cases are admittedly a blind-spot in this approach.

aggrand:
Drive by comment, feel free to ignore:

Would it be possible to use `serverStatus` or similar to assert that a transaction occurred? Might be hard to plug into the CrudActor tests, so up to you whether it's worth it.

richardsamuels:
@aggrand Are you thinking of `transactions.totalCommitted` from serverStatus? As written, that could assert that a transaction occurred, but it can't help us assert that one of the commands ran on the transaction session, while another did not. 

Or is there another field in serverStatus that you're thinking of?

aggrand:
@richardsamuels yeah that was the field I was thinking of, maybe it would be possible to check the pre/post `transactions.totalCommitted` of an actor run that has onSession set, and then another assert of pre/post for a run that doesn't have it set.

I don't know if the crudactor yaml tests have that flexibility, but a dedicated written actor test probably (hopefully) can with two separate ActorHelper runs. Up to you if it's worth trying!

richardsamuels:
evergreen merge

PR #622, opened 2022-03-01
PERF-2814 Add remaining cases for lastpoint type queries

alyacb:
Evergreen patch: https://spruce.mongodb.com/version/621cf72b3627e06fd97624fa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/phases/execution/TimeSeriesLastpoint.yml
+++ b/src/phases/execution/TimeSeriesLastpoint.yml
@@ -1,0 +1,64 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  These are the phases used to measure performance of the last-point-in-time query optimization on timeseries collections.
+
+CreateIndex:
+  Repeat: 1
+  Database: &db test
+  Operations:
+  - OperationName: RunCommand
+    OperationMetricsName: CreateIndex
+    OperationCommand:
+      createIndexes: {^Parameter: {Name: "Collection", Default: "Collection0"}}
+      indexes:
+      - {key: {^Parameter: {Name: "IndexPattern", Default: {}}}, name: {^Parameter: {Name: "IndexName", Default: ""}}}
+
+Quiesce:
+  Repeat: 1
+  Database: admin
+  Operations:
+  - OperationName: RunCommand
+    OperationMetricsName: Quiesce
+    OperationCommand:
+      fsync: 1
+
+RunLastPointQuery:
+  Repeat: 10
+  Database: *db
+  Operations:
+  - OperationMetricsName: LastPointQuery
+    OperationName: RunCommand
+    OperationCommand:
+      aggregate: &coll {^Parameter: {Name: "IndexPattern", Default: "Collection0"}}
+      pipeline:
+        [
+          &sortPattern {$sort: {^Parameter: {Name: "SortPattern", Default: {}}}},
+          &groupPattern {$group: {^Parameter: {Name: "GroupPattern", Default: {}}}}
+        ]
+      cursor: &batchSize {batchSize: {^Parameter: {Name: "BatchSize", Default: 100}}}
+      # In cases when we cannot optimize for a covered index sort, allow for external sorting.
+      allowDiskUse: true
+  - OperationMetricsName: LastPointQueryWithPrecedingPredicate
+    OperationName: RunCommand
+    OperationCommand:
+      aggregate: *coll
+      pipeline:
+        [
+          {$match: {"metadata.sensorId": {$gt: 10}}},
+          *sortPattern,
+          *groupPattern
+        ]
+      cursor: *batchSize
+      # In cases when we cannot optimize for a covered index sort, allow for external sorting.
+      allowDiskUse: true
+
+DropIndex:
+  Repeat: 1
+  Database: *db
+  Operations:
+  - OperationMetricsName: DropIndex
+    OperationName: RunCommand
+    OperationCommand:
+      dropIndexes: {^Parameter: {Name: "Collection", Default: "Collection0"}}
+      index: {^Parameter: {Name: "IndexName", Default: ""}}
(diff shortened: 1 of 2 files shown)


src/workloads/execution/TimeSeriesLastpoint.yml
```diff
@@ -3,9 +3,11 @@ Owner: "@mongodb/query"
 Description: |
   This test exercises the behavior of lastpoint-type queries on time-series collections. The
   currently supported lastpoint aggregate pipelines that are tested here include:
-    1. a $sort on a meta field (ascending) and time (descending) and $group with _id on the meta
+    1. a $sort on a meta field (ascending, descending) and time (descending) and $group with _id on the meta
```
kaywux, line 6: nit: The meta phrasing is a little confusing, maybe something like `meta field (both ascending and descending)`?

src/workloads/execution/TimeSeriesLastpoint.yml
```diff
@@ -46,75 +61,191 @@ Actors:
...
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
 
-- Name: CreateIndex
+# Lastpoint query with ascending index on meta subfield and descending index on time.
```
kaywux, line 78: Another wording nit: this comment sounds like we have two separate indexes, one on meta and one on time. Maybe something like "compound index with meta subfield ascending and time descending"?

src/phases/execution/TimeSeriesLastpoint.yml
```diff
@@ -0,0 +1,64 @@
...
+
+DropIndex:
+  Repeat: 1
+  Database: *db
+  Operations:
+  - OperationMetricsName: DropIndex
+    OperationName: RunCommand
+    OperationCommand:
+      dropIndexes: {^Parameter: {Name: "Collection", Default: "Collection0"}}
+      index: {^Parameter: {Name: "IndexName", Default: ""}}
```
kaywux, line 64: I've never created a genny phase before - using this to parameterize the workload is super clever, nice job!

src/workloads/execution/TimeSeriesLastpoint.yml
```diff
@@ -46,75 +61,191 @@ Actors:
...
   - *Nop
-  - Repeat: 1
-    Database: *db
-    Operations:
-    - OperationName: RunCommand
-      OperationCommand:
-        createIndexes: Collection0
-        indexes:
-        - {key: {"metadata.sensorId": 1, timestamp: -1}, name: "MetaSubfieldAscendingTimeDescending"}
+  - LoadConfig:
```
kaywux, line 85: Could you add a phase here (and for the later queries as well) to test for when this query doesn't have an accompanying backing index? Note that after this the comments should be changed to reflect this.

For some context, according to Michael G the use cases we should care about for this project are (in order of importance):
1. when the collection has {meta: 1, time: 1} index
2. when the collection doesn't have any indices at all
3. all other index formulations - he doesn't think these will be very common, since MongoDB officially recommends config 1

kaywux (changes requested):
A few requested changes but the overall structure looks great to me, thanks Alya!

alyacb (commented):
Thank you for taking a look, Katherine! I made the changes you requested.

src/workloads/execution/TimeSeriesLastpoint.yml
kaywux: Good question - I think what you've done here is sufficient for this ticket scope but might be good to talk to Mike and see if there are any other indexes we should test against for the project.

src/workloads/execution/TimeSeriesLastpoint.yml
```diff
@@ -46,75 +111,309 @@ Actors:
...
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
 
-- Name: CreateIndex
+# Lastpoint query with a sort and group with meta subfield descending and time ascending, but no index.
+- Name: RunLastpointQueryWithMetaSubfieldAscendingTimeDescendingNoIndex
```
kaywux, line 135: Question: was there a reason for having the index/no index as separate actors? Would it be possible to instead do something like below instead? If so it would simplify the workload and help with readability.

```
- Name: RunLastpointQueryWithMetaSubfieldAscendingTimeDescending
    # No index operations (Quiesce, LastpointQueryMetaAscTimeDesc)
    # With index (CreateIndex, Quiesce, LastpointQueryMetaAscTimeDesc, DropIndex)
```

kaywux (commented):
Great job on this Alya! I have one remaining comment but overall looks good to me! I'll leave the formal approval to whoever picks up the review after me :)

src/workloads/execution/TimeSeriesLastpoint.yml
alyacb: I don't think there is any clean way to do this because of phases, but I'll ask in perf-tooling-users to see if anyone knows of a better way to do this.

There are ActorTemplates which allow us to parametrize entire actors; however, because each actor has an array of phases such that the actual workload begins at a different index, there is no way to parametrize on the actor level without having all the workloads occur simultaneously (hence the giant array of Nops).

Similarly, there are phases; however, afaik, we can't parametrize arrays of phases, only single phases. This makes the repetition necessary (and unfortunate). Writing a dedicated actor to handle this seemed like overkill for this ticket.

kaywux: Ah gotcha, thanks for expanding on that - in that case the workload structure make sense to me.

starzia (approved):
LGTM!

alyacb:
Thank you for the reviews! Off to the commit queue.

PR #624, opened 2022-03-04
PERF-2783 Validate base performance of Multi:true doc by doc deletions

haleyConnelly:


haleyConnelly:
[Evergreen](https://spruce.mongodb.com/version/6221f3c757e85a1b06a7688d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

This is just for base analysis. We can create another ticket to expand the parameters we want

src/workloads/execution/MultiDeletesDocByDoc.yml
```diff
@@ -0,0 +1,27 @@
...
+Description: >
+  Runs a workload that inserts and then deletes a large range of documents on one collection while
+  performing writes on another collection in the same database. Serves as a base line for comparing
+  legacy document by document deletes versus efficiently batched multi deletes.
+
+# Increase the maxPoolSize to enable faster initial data population
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 300
```
haleyConnelly, line 12: This was done to speed up the inserts significantly. When I added 2 separate MonotonicSingleLoader phases, I had to raise it from 200 - 300

src/workloads/execution/MultiDeletesDocByDoc.yml
```diff
@@ -0,0 +1,27 @@
...
+      # Allow for longer duration since the loading of the data may take a while.
+      socketTimeoutMS: 3_600_000  # = 1 hour
```
haleyConnelly, line 14: The deletes right now take about 7 minutes. This is to prevent timeouts with minimal information in case something gets really broken and a delete takes forever

src/workloads/execution/MultiDeletesDocByDoc.yml
```diff
@@ -0,0 +1,27 @@
...
+
+LoadConfig:
+  Path: "../../phases/execution/MultiDeletes.yml"
+  Parameters:
+    DocumentSizePaddingBytes: 1024 #1KB
+    DocumentCount: 5_000_000
```
haleyConnelly, line 20: 7 minutes for doc by doc deletions with concurrent writes in initial patch testing

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,163 @@
...
+    - OperationName: bulkWrite
+      OperationCommand:
+        WriteOperations:
+        - WriteCommand: deleteMany
+          Filter: {_id: {$gte: 0}}
+          OperationOptions:
+            WriteConcern: { Level: "majority" }
+
+# Phase 5: Perform some write operations - see how write latency is impacted by the large delete.
+- Name: Insert
```
haleyConnelly, line 142: We can create a ticket to make this more robust --- with other CRUD operations, just trying to get vanilla implementation out right now

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,163 @@
...
+Owner: "@mongodb/server-execution"
+Description: >
+  Runs inserts on one collection while a mass deletion takes place on another collection in the same database.
+
+GlobalDefaults:
+  # Configurable parameters
+  DocumentSizePaddingBytes: &DocumentSizePaddingBytes {^Parameter: {Name: "DocumentSizePaddingBytes", Default: 0}}
+  DocumentCount: &DocumentCount {^Parameter: {Name: "DocumentCount", Default: 100_000_000}}
+
+# Static variables.
```
haleyConnelly, line 11: s/Static/Constant

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,163 @@
...
+          OperationOptions:
+            WriteConcern: { Level: "majority" }
+  - *Nop
+  - *Nop
+
+# Phase 4: Populate the collection for mass deletion again. This time concurrent writes will be
+# issued to another collection in the same database.
+# NOTE: The MonotonicSingleLoader counter will start where it left off in the first instance of this
+# Actor in a previous phase.
+- Name: PopulateInitialData1
```
josefahmad, line 90: I'm seeing a problem while running it locally. Phase 1 through phase 3 completed. However phase 4 doesn't seem to run, and phase 5 seems to have run for a tiny fraction of time, since I found just 1 document (and only 7 documents in another attempt) inserted in `test.Collection1`. Do you see the same on your end?

apodelko:
Looking at the evergreen results, it appears that nothing was inserted during phase 4 (DocumentsTotal=0).

If I read correctly, we delete all documents. May delete all documents be a special case - vs. remove random 90% of documents? Just curious.

As per phase 5 - we inserting documents in parallel with removing all document. How does it work? Is it continue removing documents as we are inserting them? If not - what mechanism we use here to not remove the new documents?

I don't see anything related to sharding. What is the point of using a shard config without actual sharding data?

src/phases/execution/MultiDeletes.yml
haleyConnelly: Phase 5 runs as long as it takes to delete all the documents. The inserts into Collection1 only run for a fraction of time since the delete didn't have much work to do. It appears the issue is with phase 4

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,161 @@
...
+          Filter: {_id: {$gte: 0}}
+          OperationOptions:
+            WriteConcern: { Level: "majority" }
+  - *Nop
+  - *Nop
+
+# Phase 4: Populate the collection for mass deletion again. This time concurrent writes will be
+# issued to another collection in the same database. TODO Make multithreaded once TIG-2938 is resolved.
+- Name: PopulateInitialData1
+  Type: Loader
```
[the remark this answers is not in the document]
haleyConnelly: Update: we believe this is because normal index scans have type bounds. Which means GTE _id: 0 won't incorporate id's with type OID (see [key_string.cpp](https://github.com/10gen/mongo/blob/c176492ec7c149a081585ad52416cdc630edab81/src/mongo/db/storage/key_string.cpp#L75) for type encoding)

haleyConnelly:
@apodelko we've been running shardlite for when we do extend performance to sharding. If you'd like, we can add a TODO and only run on linux replica sets for now.

Additionally, we test concurrent writes on Collection1, while there is a multi delete on Collection0 (both on the same database)

"If I read correctly, we delete all documents. May delete all documents be a special case - vs. remove random 90% of documents? Just curious." --- The idea is to delete large amounts of documents, a number of documents we can control. It would be possible to delete a subsection of them, but perhaps once issues with the MonotonicSingleLoader used twice in one test are resolved.

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,162 @@
...
+  - Repeat: 1
+    MetricsName: DuringMassDeletion0
+    Collection: *MassDeletionsColl
+    Operations:
+    - OperationName: bulkWrite
+      OperationCommand:
+        WriteOperations:
+        - WriteCommand: deleteMany
+          # Filter by the smallest OID so all documents are included.
+          Filter: {_id: {$gte: {^ObjectId: "000000000000000000000000"}}}
```
haleyConnelly, line 81: We could filter by a new field "intToFilterBy" that is not_id. It could be randomly generated between 0 and DocumentCount, but there would be no uniqueness guarantee. It seems like setting the ObjectId to all 0's is sufficient.

haleyConnelly (commented):
https://spruce.mongodb.com/version/62264b0f9ccd4e76a6555c7d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/phases/execution/MultiDeletes.yml
josefahmad: Looks resolved in the current revision. Good job!

josefahmad (commented):
Great work! I've added some suggestions to make it easier to interpret the results.

haleyConnelly (commented):
[evergreen with profiling](https://spruce.mongodb.com/version/6227747861837d2db73b8f77/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

apodelko:
Looking into it... Wonder why do we report both DocumentsTotal and DocumentThroughput as 0 in Trend Charts for both PopulateInitialData0 and PopulateInitialData1 - while it appears that it did insert 10M documents. [in the first patch we saw these metrics for PopulateInitialData0]

@josefahmad I guess profiling may be an option for debugging only - I'd say that we need to see it in metrics. We should see it, I guess, in the duration of mass deletion. Maybe we indeed need to increase the real insert rate for the final stage to see the effect - maybe make it multi-threaded?

By the way, are we interested on how deletes impact inserts? If we do - maybe we need to have first a baseline insert phase without deletion.

As per sharding - if we plan to add it in the future, maybe we should add the sharding config then too? Two considerations for that 1) first running additional configuration is an expense if it doesn't provide anything useful. 2) When you actually shard it, metrics will probably change significantly - triggering false data change points/BFs.

apodelko:
Never mind about DocumentsTotal and DocumentThroughput. My bad. It looks like this actor reports OperationsTotal and OperationsThroughput - which it does.

src/phases/execution/MultiDeletes.yml
```diff
@@ -0,0 +1,162 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: >
+  Runs inserts on one collection while a mass deletion takes place on another collection in the same database.
+
+GlobalDefaults:
+  # Configurable parameters
+  DocumentSizePaddingBytes: &DocumentSizePaddingBytes {^Parameter: {Name: "DocumentSizePaddingBytes", Default: 0}}
+  DocumentCount: &DocumentCount {^Parameter: {Name: "DocumentCount", Default: 100_000_000}}
```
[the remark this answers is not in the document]
josefahmad: Thanks for testing with the verbose setProfilingLevel. Unfortunately that has the drawback of making the log file unwieldily large. And perhaps more relevantly, I figured out how we can get highly accurate per-operation [latency metrics in genny](https://github.com/10gen/performance-tooling-docs/blob/main/getting_started/intrarun_data_generation.md), so there's no need for increased logging server side. @haleyConnelly let's revert the setProfilingLevel change, and I'm happy approving this peer review!

josefahmad:
@apodelko agree, we're removing the custom setProfilingLevel and we're going to analyse the client-side latencies recorded by genny. 

I'm okay with the current insert rate during mass deletion. We're interested in the latency distributions of inserts that happen concurrently with mass deletions. Our goal is to evaluate the latency distribution of these inserts with the _classic_ delete stage (this PR) vs a new, more efficient delete stage (upcoming PERF-2824). The rate of inserts is sufficient to fulfil this goal.

josefahmad (approved):
LGTM % let's revert the custom profiling level. Great job!

apodelko (approved):
LGTM if we getting what we want here.

PR #626, opened 2022-03-08
TIG-3937 bulkWrite should not ignore OnSession

richardsamuels:
In a similar vein to [this PR](https://github.com/mongodb/genny/pull/620), we need to respect OnSession when it's on bulkWrite. [Example affected workload](https://github.com/mongodb/genny/blob/master/src/workloads/scale/UpdateMillionDocsInTxn.yml#L78).

### DIFF
--- a/src/cast_core/test/CrudActorYamlTests.yml
+++ b/src/cast_core/test/CrudActorYamlTests.yml
@@ -597,6 +597,25 @@ Tests:
         Count: 1
     AssertNumTransactionsCommitted: 0
 
+  - Description: simple findOneAndReplace in a withTransaction without OnSession
+    Operations:
+      - OperationName: withTransaction
+        OperationCommand:
+          OperationsInTransaction:
+            - OperationName: insertOne
+              OperationCommand:
+                Document: {a: 100}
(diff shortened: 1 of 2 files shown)


src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -615,3 +615,22 @@ Tests:
...
+              WriteOperations:
+                - WriteCommand: insertOne
+                  Document: { a: 1 }
+                - WriteCommand: deleteOne
+                  Filter: { a: 1 }
+                - WriteCommand: insertOne
+                  Document: { a: 2 }
+    OutcomeData:
+      - {a: 2}
+    AssertNumTransactionsCommitted: 1
```
rtimmons, line 636: worth a test where onSession=false?

rtimmons (approved):
Thanks for crossing the Ts here.

src/cast_core/test/CrudActorYamlTests.yml
richardsamuels: Added

richardsamuels:
evergreen merge

PR #633, opened 2022-03-15
TIG-3957 add expansion support to LoadConfig Path

jimoleary:
This allows the horror that is [Large.yml](https://github.com/10gen/PrivateWorkloads/blob/production/src/workloads/customer_representative/coinbase/balance/Large.yml) be parameterized more easily like [LargeWfD.yml](https://github.com/10gen/PrivateWorkloads/blob/production/src/workloads/customer_representative/coinbase/balance/LargeWfD.yml)


aggrand (approved):
LGTM regarding the exact change, but agreed regarding a test case! Might also be worth running a couple of the more preprocessor-intensive workloads (e.g. MixedWorkloadsGenny) through this to make sure the results are the same.

jimoleary:
:+1: on testing and I will need to run a full set of patches too.

PR #635, opened 2022-03-17
TIG-3421 print out actor info in phase loop

vrachev:
Took me a bit to get my head around genny internals, but I feel like I have a much better understanding of how the phaseloop/orchestrator/context stuff work (though some of the c++ stuff still confuses me). 

This PR does much of what my other one does, but prints stuff out in the phaseloop instead of in each individual class. I'm running a patch to see if performance is affected. 

One thing that I couldn't achieve was to print out the actorId, but I'm pretty sure it's possible though. I think it's a matter of changing the [PhaseLoop constructor](https://github.com/mongodb/genny/blob/25a5f5737671be446e240a71cf6cd28f3829cedd/src/gennylib/include/gennylib/PhaseLoop.hpp#L673-L675) and [constructPhaseMap()](https://github.com/mongodb/genny/blob/25a5f5737671be446e240a71cf6cd28f3829cedd/src/gennylib/include/gennylib/PhaseLoop.hpp#L700) to expect an explicit `ActorId id` parameter, then moving the [id](https://github.com/mongodb/genny/blob/25a5f5737671be446e240a71cf6cd28f3829cedd/src/cast_core/src/Insert.cpp#L64) parameter here (as an example actor) to be the second parameter. I sorta tried that but ran into compiler errors. I didn't wanna go too far into that rabbit hole in case I was wrong about that approach since there's a bunch of invocations to change. Any thoughts on whether this would work? I didn't dive into how templating works so part of my issue is that I don't understand the arg forwarding stuff super well.


vrachev:
Hey @aggrand, I made a new PR with the PhaseLoop approach. Take a look when you can! I ran some perf patches and (using the perf-analyzer tool) it doesn't look like this change affects things negatively.

aggrand (changes requested):
Re: the adding actorid, I do think adding it as the first arg to phaseloop in addition to the context would work. (Or, following the suggestion below, doing an ActorInfo struct that's passed in would work). That would be in addition to the id you see from InsertRemove, since that id is being passed on to the phasemaps.

I'd be happy to pair-debug on it! I think getting the actor ids in would be really valuable, since right now people would just see a mass of identical lines whenever a many-threaded actor starts/stops.

Alternatively, if we could have the start/stop message be printed "globally" instead of one per thread, that might even be better. I feel like wanting per-thread data on that is rare.

Totally up to you either route! Happy to sit down and pair-debug (bang heads against the compiler) if you're free sometime!

vrachev:
Made some more changes. Output looks like this now

```
[curator] 2022/03/25 10:07:50 [p=info]: starting poplar gRPC service at 'localhost:2288'
[2022-03-25 10:07:52.000777] [0x000070000eea1000] [info]    Constructing pool with MongoURI 'mongodb://localhost:27017/?appName=Genny&maxPoolSize=100'
[2022-03-25 10:07:52.023741] [0x000070000f6d1000] [debug]   Beginning phase 0
[2022-03-25 10:07:52.023778] [0x000070000f6d1000] [debug]   Starting Insert::InsertWithNopTest::3::isNop
[2022-03-25 10:07:52.023822] [0x000070000f64e000] [debug]   Starting Insert::InsertWithNopTest::2::isNop
[2022-03-25 10:07:52.023849] [0x000070000f5cb000] [debug]   Starting Insert::InsertWithNopTest::1::isNop
[2022-03-25 10:07:52.023872] [0x000070000f5cb000] [debug]   Ended phase 0
[2022-03-25 10:07:52.023894] [0x000070000f5cb000] [debug]   Ended Insert::InsertWithNopTest::1::isNop
[2022-03-25 10:07:52.023903] [0x000070000f64e000] [debug]   Ended Insert::InsertWithNopTest::2::isNop
[2022-03-25 10:07:52.023917] [0x000070000f6d1000] [debug]   Ended Insert::InsertWithNopTest::3::isNop
[2022-03-25 10:07:52.023934] [0x000070000f6d1000] [debug]   Beginning phase 1
[2022-03-25 10:07:52.023950] [0x000070000f6d1000] [debug]   Starting Insert::InsertWithNopTest::3
[2022-03-25 10:07:52.023957] [0x000070000f5cb000] [debug]   Starting Insert::InsertWithNopTest::1
[2022-03-25 10:07:52.023974] [0x000070000f64e000] [debug]   Starting Insert::InsertWithNopTest::2
[2022-03-25 10:07:52.024001] [0x000070000f6d1000] [info]     Inserting { "a" : 55 }
[2022-03-25 10:07:52.024009] [0x000070000f64e000] [info]     Inserting { "a" : 59 }
[2022-03-25 10:07:52.024021] [0x000070000f5cb000] [info]     Inserting { "a" : 55 }
[2022-03-25 10:07:52.040463] [0x000070000f64e000] [debug]   Ended phase 1
[2022-03-25 10:07:52.040499] [0x000070000f64e000] [debug]   Ended Insert::InsertWithNopTest::2
[2022-03-25 10:07:52.040523] [0x000070000f5cb000] [debug]   Ended Insert::InsertWithNopTest::1
[2022-03-25 10:07:52.040543] [0x000070000f6d1000] [debug]   Ended Insert::InsertWithNopTest::3
[2022-03-25 10:07:52.040590] [0x000070000f6d1000] [debug]   Beginning phase 2
[2022-03-25 10:07:52.040635] [0x000070000f6d1000] [debug]   Starting Insert::InsertWithNopTest::3::isNop
[2022-03-25 10:07:52.040657] [0x000070000f64e000] [debug]   Starting Insert::InsertWithNopTest::2::isNop
[2022-03-25 10:07:52.040675] [0x000070000f5cb000] [debug]   Starting Insert::InsertWithNopTest::1::isNop
[2022-03-25 10:07:52.040731] [0x000070000f5cb000] [debug]   Ended phase 2
[2022-03-25 10:07:52.040780] [0x000070000f5cb000] [debug]   Ended Insert::InsertWithNopTest::1::isNop
[2022-03-25 10:07:52.040790] [0x000070000f64e000] [debug]   Ended Insert::InsertWithNopTest::2::isNop
[2022-03-25 10:07:52.040802] [0x000070000f6d1000] [debug]   Ended Insert::InsertWithNopTest::3::isNop
````


It can get spammy with a lot of threads but I think it would be a lot of work to make it print once per global actor, and having per thread information could be useful! 

I've made some unit tests flakey, I'll investigate those when I'm back on Monday. Also need to run another perf patch. But the functionality itself is RFAL @aggrand

aggrand (changes requested):
Looking great! Thanks for threading this id info through! Main requested change is the flaky test

PR #637, opened 2022-03-18
PERF-2841 Create baseline performance test for timeseries sort in preparation for BUS

mattBoros:
Posting this just to get some feedback on manually written pipelines/ideas for new tests/any comments. We don't want to commit manual rewrites so I've marked this as WIP.

It can be hard to tell what's going on in this test since there's so many phases, so I added a summary at the top of TimeSeriesSort.yml

[results](https://docs.google.com/spreadsheets/d/1FmiFekimHteQgIylTjo5oXg3z1EYYuZ9mChwE70xEyU/edit#gid=2115490241)


src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,512 @@
...
+                     bucketMaxSpanSeconds: 3600,
+                     assumeNoMixedSchemaData: true
+                   }
+                   },
+                   {$_internalBoundedSort: {
+                     sortKey: {t: 1},
+                     bound: 3600,
+                   }
+                   }]
+        cursor: {batchSize: *batchSize}
```
dpercy, line 293: I think this will only return the first batch of results.  I assume RunCommand doesn't have any special knowledge about cursors or getMore, so it would send this command as-is and wait for the response.

If that's right, we can force the whole query to run by adding some extra stage at the end.  Maybe {$skip: 1e10} or {$_internalInhibitOptimization: {}}, {$skip: 1e10}.  There must be examples in other workloads.

src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,512 @@
...
+        pipeline: [{$sort: {t: 1}}]
+        cursor: {batchSize: *batchSize}
+  - Repeat: *numQueryRuns
+    Database: *db
+    Operations:
+    - OperationMetricsName: SortQueryBoundedSorter
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: system.buckets.Collection0
+        pipeline: [{$sort: {"control.min.t": 1}},
```
dpercy, line 279: Currently this $sort will do a blocking sort, because the query planner doesn't realize that _id contains control.min.time.  So a more realistic test would be to remove the $sort stage and add `hint: {$natural: 1}` to the command.  Then it will scan the collection in {_id: 1} order, unpack, bounded-sort.

src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,512 @@
...
+  - Repeat: &numQueryRuns 50
+    Database: *db
+    Operations:
+    - OperationMetricsName: SortQueryBlockingSort
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *coll
+        pipeline: [{$sort: {t: 1}}]
+        cursor: {batchSize: *batchSize}
+  - Repeat: *numQueryRuns
```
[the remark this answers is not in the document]
mattBoros: Oh yea that was my original intent. I can change that

src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,512 @@
...
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
+  - Repeat: &numQueryRuns 50
```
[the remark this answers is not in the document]
mattBoros: In one of the [comments](https://github.com/mongodb/genny/pull/580#discussion_r761205779) in the PR for locf.yml, it was recommended to put operations in different phases and quiesce in between the phases if we're looking to measure performance of each operation (as opposed to interactions between operations). Also some operations need to be done in between queries, such as adding more data to force a disk spill, or creating/dropping an index.

src/workloads/execution/TimeSeriesSort.yml
mattBoros: Offline we decided on $skip for this one

src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,594 @@
...
+  - *Nop
+  - Repeat: *numQueryRuns
+    Database: *db
+    Operations:
+    - OperationMetricsName: SelectiveMetaPointPredicateTimeSortBlockingSort
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *coll
+        pipeline: [
+          {$match: {m: 1002}},  # this will match ~10 docs
```
dpercy, line 472: This will match exactly 10 docs, right?  Because the only insert phase where m is 1002 generates 10 docs.

mattBoros: Oh yea I can change that

src/workloads/execution/TimeSeriesSort.yml
```diff
@@ -0,0 +1,395 @@
...
+      - replica
+      - replica-all-feature-flags
+    branch_name:
+      $neq:
+      - v4.0
+      - v4.2
+      - v4.4
+      - v5.0
+      - v5.1
+      - v5.2
```
dpercy, line 395: Maybe also exclude 5.3 for now?

dpercy (approved):

mattBoros:
evergreen merge

PR #639, opened 2022-03-22
CLOUDP-116026: Create a new Genny workload to trigger serverless scaling events

betsybutton:
JIRA ticket: [CLOUDP-116026](https://jira.mongodb.org/browse/CLOUDP-116026)

Yesterday during manual testing, this workload successfully triggered serverless scaling events! I'll add a successful patch here shortly, but this code should be fine to merge.

### DIFF
--- a/src/workloads/serverless/IndexStress.yml
+++ b/src/workloads/serverless/IndexStress.yml
@@ -1,0 +1,271 @@
+SchemaVersion: 2018-07-01
+Owner: Atlas Serverless
+Description: |
+  This workload is an extension of scale/LargeIndexedIns.yml to run the workload
+  on multiple serverless tenants. On each cluster, it sends bursts of finds that
+  use large $ins while running a low rate update workload.  The workload causes high CPU load as the server
+  becomes bottlenecked on tcmalloc spinlocks during the find operations.
+  Improvements or regressions in this aspect of the allocator should
+  be measurable by the average CPU usage during this test as well as
+  the latency of the find operations.
+
+  In this workload, large arrays of random strings are generated to use
+  for the $in queries.  To avoid a CPU bottleneck on the workload client,
+  it uses a ^Once generator to generate the arrays once during initialization.
+
+  IMPORTANT NOTE: You must match the number of clients with the value for the
+  clustersPerConfig key specified in the atlas-serverless-local.yml file in DSI. This will ensure
+  that x serverless instances are created to run workloads on x serverless tenants, using x clients.
+
+Clients:
+  Client1:
+    QueryOptions:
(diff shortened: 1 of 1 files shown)


JamesHeppenstall (approved):
LGTM but wait for @aggrand before merging 😄

src/workloads/serverless/IndexStress.yml
```diff
@@ -0,0 +1,269 @@
...
+      OperationCommand:
+        Filter: {id: {^RandomInt: {min: 0, max: 1000}}}
+        Update:
+          $inc: {a: 1}
+
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq: replica
```
aggrand, line 269: This will run for normal replica sets in sys-perf as well. Is that okay, considering the URIs wouldn't be overridden for them?

src/workloads/serverless/IndexStress.yml
```diff
@@ -0,0 +1,269 @@
+SchemaVersion: 2018-07-01
+Owner: Atlas Serverless
```
aggrand, line 2: Is there an @-mentionable github username that could be put here for owner?

aggrand (changes requested):
Looks great, just had a couple questions!

src/workloads/serverless/IndexStress.yml
betsybutton: Ah, thank you for pointing this out! Siyuan and I chatted about whether or not we wanted to use the AutoRun feature for our tests, and we're not ready to automatically schedule them to run yet. I'm just going to remove this section.

src/workloads/serverless/IndexStress.yml
betsybutton: As far as I know, Atlas teams don't have team github usernames like the server teams. I saw a few examples of workloads where the team name was just written out (like [this one](https://github.com/mongodb/genny/blob/0c154851e154663b5cdbb804f001019f7901bd7a/src/workloads/scale/ScanWithLongLived.yml#L2)), so I thought that would be appropriate here. @JamesHeppenstall , can you confirm that we don't have a team username like this?

JamesHeppenstall: Yeah afaik the teams in the `10gen` org are being used to determine permissions and not used for anything else. Also, not sure if we'd even be able to reference a team from another GitHub org here.

aggrand: Gotcha fair, just wanted to check!

aggrand (approved):

betsybutton:
evergreen merge

PR #640, opened 2022-03-23
PERF-2883 Temporarily disable the SBE lookup testing

yun-soo:
Given that the SBE lookup algorithms have not been yet fully implemented and we're enabling the feature on all-feature-flags build variant for correctness testing coverage, I'm temporarily disabling the SBE lookup perf testing.

As soon as we have enough confidence in feature's correctness, will re-enable the perf testing. Until then, we can run manually the SBE lookup perf testing by re-enabling the following part.

```
# Reenable this setup if you want to enable the SBE $lookup perf testing on all-feature-flags
# build variant on Evergreen patch build
      # - standalone-all-feature-flags
```

### DIFF
--- a/src/workloads/execution/LookupSBEPushdown.yml
+++ b/src/workloads/execution/LookupSBEPushdown.yml
@@ -437,4 +437,4 @@ AutoRun:
       - standalone
 # Reenable this setup if you want to enable the SBE $lookup perf testing on all-feature-flags
 # build variant on Evergreen patch build
-      # - standalone-all-feature-flags
+# - standalone-all-feature-flags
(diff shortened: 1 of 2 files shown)


dalyd (approved):
Remind me what the test was doing before the functionality was enabled? Was it taking another pass? So potentially going from correct to incorrect behavior for now?

yun-soo:
> Remind me what the test was doing before the functionality was enabled? Was it taking another pass? So potentially going from correct to incorrect behavior for now?

It was run the classic $lookup on both standalone and standalone-all-feature-flags build variants because the SBE $lookup (`featureFlagSBELookupPushdown`) was in `fully_disabled_feature_flags.yml`.

The intention was to compare numbers from `standalone` to ones from `standalone-all-feature-flags`.

Does this answer your question?

dalyd:
Yeah, that makes sense. Thanks.

PR #642, opened 2022-03-25
PERF-2890 remove tpch q15 drop metrics

jimoleary:
[patch](https://spruce.mongodb.com/version/623def6a850e6138026725f3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) test

### DIFF
--- a/src/phases/tpch/denormalized/Q15.yml
+++ b/src/phases/tpch/denormalized/Q15.yml
@@ -54,7 +54,6 @@ TPCHDenormalizedQuery15:
         allowDiskUse: true
       verbosity:
         executionStats
-  - OperationMetricsName: Query15DropView
-    OperationName: RunCommand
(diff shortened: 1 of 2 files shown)


alyacb (approved):
The change looks LGTM, assuming the patch succeeds! Just a note- the TPCH tasks aren't triggered by the genny auto-generate tasks because they rely on a separate DSI test control; I believe you will have to select them directly from evergreen tasks, e.g. on Linux Standalone:
Screenshot 2022-03-25 at 16 41 10

jimoleary:
>  The change looks LGTM, assuming the patch succeeds! Just a note- the TPCH tasks aren't triggered by the genny auto-generate tasks because they rely on a separate DSI test control; I believe you will have to select them directly from evergreen tasks, e.g. on Linux Standalone:

Fixed, thanks!

PR #643, opened 2022-03-29
PERF-2828 Document RandomSeed

jimoleary:


rtimmons (approved):
Thanks for adding this.

dalyd (approved):
Nice use of LoadConfig also.

PR #644, opened 2022-03-29
CLOUDP-116968: Increase client count and ignore find operation failures in serverless index stress workload

JamesHeppenstall:


src/workloads/serverless/IndexStress.yml
```diff
@@ -28,6 +28,24 @@ Clients:
   Client4:
     QueryOptions:
       maxPoolSize: 250
+  Client5:
+    QueryOptions:
+      maxPoolSize: 250
```
betsybutton, line 33: Was this the maxPoolSize when we ran it successfully yesterday? I remember decreasing it to 50 at some point, but I don't remember at what point..

betsybutton (commented):
Thanks for getting this in! Just had one question :)

src/workloads/serverless/IndexStress.yml
JamesHeppenstall: I think we can safely keep the maxPoolSize at 250. As long as it's lower than 500 (which is the max connections per serverless instance), we won't hit any limits.

betsybutton: Sounds good!

betsybutton (approved):
LGTM!!

JamesHeppenstall:
evergreen merge

PR #646, opened 2022-04-05
PERF-2892 Extend initial performance tests to include overlapping bucket scenarios

mattBoros:
[patch](https://spruce.mongodb.com/version/624b5f2156234338da7ae0a9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

[results](https://docs.google.com/spreadsheets/d/1UDHMTJCmtR_6kzVOab0kUWZ65g6_JOzpEiXH1XLV9vQ/edit#gid=1759546507)


src/phases/execution/TimeSeriesSortOverlappingBucketsCommands.yml
```diff
@@ -0,0 +1,31 @@
...
+      metaField: "m",
+      bucketMaxSpanSeconds: 3600,
+      assumeNoMixedSchemaData: true
+    }},
+    {$_internalBoundedSort: {
+      sortKey: {t: 1},
+      bound: {base: "min"},
+    }},
+    {$_internalInhibitOptimization: {}},
+    {$skip: 1e10}
```
samontea, line 27: Possibly dumb comment but you may want to do this as a limit query to simulate time to first result in addition. While I am happy we do see a marginal speed up in the non-degenerate case. It would be interesting to seem time to first result specifically because that is the case that we're optimizing for.

samontea (commented):
I think this all looks good. you may want to add my feedback case and then just send the doc about performance in the channel and then close this pr. i don't think we want to commit any of this code as this case isn't one we want to measure perf for (moving forward).

src/phases/execution/TimeSeriesSortOverlappingBucketsCommands.yml
mattBoros: Yea that's a good idea. I don't think that will be a difficult change either.

mattBoros:
[results](https://docs.google.com/spreadsheets/d/1UDHMTJCmtR_6kzVOab0kUWZ65g6_JOzpEiXH1XLV9vQ/edit#gid=1176086019)  for time to first result (100x - 140x throughput)

samontea:
LGTM mode the potentially misleading language that we discussed in standup...

samontea:
If you could just reduce to a single data shape and to a single command (the nonbounded sort aggregate command) this looks good.

samontea:
If you could just reduce to a single data shape and to a single command (the nonbounded sort aggregate command) this looks good.

mattBoros:
evergreen merge

mattBoros:
evergreen merge

mattBoros:
evergreen merge

PR #651, opened 2022-04-07
PERF-2839 Conditionally shard ClusteredCollection*.yml workload

ykangpark:
https://spruce.mongodb.com/version/6250331bc9ec442a24f74b5f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


josefahmad (approved):
I've reviewed the yaml workload and it looks good to me, I've just added some comments. Let's get someone from TIG or PERF to review the work on the actor. I've asked for a volunteer in #performance-tooling-users.

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,110 @@ class DatabaseOperation {
...
+                    break;
+                }
+            }
+        }
+        return _onlyRunInInstancesContext.size() == 0 || _typeFoundInConfig;
+    }
+
+    // The output of the "hello" command is parsed to check if we are running in a Mongos, a replica
+    // set, or a standalone instance.
+    static std::string_view getInstanceType(mongocxx::database &db) {
```
aggrand, line 286: Would it be possible to make use of the toplogy visitor: https://github.com/mongodb/genny/blob/5a525da35c8153ae59bc20af190b0dd454ac1975/src/gennylib/include/gennylib/v1/Topology.hpp#L61

If we used the above, we could change `OnlyRunInInstances` to instead be a `RunThroughTopologyVisitor` or something, constructing a topology visitor using a mapping between visitor names and visitor functions. That would be a lot more general, and re-use existing code.

To see the topology types in use, see here: https://github.com/mongodb/genny/blob/master/src/gennylib/src/quiesce.cpp

Definitely push back if you don't think that'd be worth it!

src/phases/execution/ClusteredCollection.yml
```diff
@@ -7,120 +7,153 @@ Description: >
...
+# Phase 1: Shard the collection
+- Name: ShardCollection
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        OnlyRunInInstance: sharded
```
aggrand, line 46: Would it make sense to configure this on a per-operation basis, rather than on the phase-level? I could see someone wanting to always run one operation, and only sometimes run another.

(If we went with the topology visitor method above, we could construct one visitor with multiple ops on it.)

aggrand (changes requested):
I have a suggestion for a different approach here that may be more flexible, but please push back if it wouldn't be worth!

src/phases/execution/ClusteredCollection.yml
```diff
@@ -7,120 +7,153 @@ Description: >
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 3 minutes
+        Collection: *Collection
+        Operations:
+        - OperationName: deleteOne
+          OperationCommand:
+            Filter:
+              # 1000000 docs / 16 threads = 62500
+              {_id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: 1, multiplier: 62500}}]}}, *PaddingGenerator]}}}
```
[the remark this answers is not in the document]
josefahmad: Oh I see, I'm not aware myself. @aggrand might know?

src/cast_core/src/RunCommand.cpp
ykangpark: @aggrand I have been trying to implement this using the Topology and TopologyVisitor, but I ran into issues with it. I filed  [TIG-4060](https://jira.mongodb.org/browse/TIG-4060). Given the amount of time we already spent with this ticket, we would like to leave it as it is for now. We could open a follow up ticket to change this to use the Topology, and also allow configuration on a per operation basis.

ykangpark:
Latest patch with PointDeleter fix:
https://spruce.mongodb.com/version/6255a596d6d80a1fd4bb2cba/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/phases/execution/ClusteredCollection.yml
```diff
@@ -7,120 +7,153 @@ Description: >
...
+          OperationCommand:
+            Filter: {a: {^FastRandomString: {length: 6, alphabet: "0123456789"}}}
 
-# Phase 6: Point deletion. TODO convert to range deletions once TIG-3707 is implemented.
+
+# Phase 6: Point deletion.
 - Name: PointDeleter
-  Type: Deleter
-  Database: test
+  Type: CrudActor
```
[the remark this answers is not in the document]
ykangpark: With the latest commit, on the replica set variant, we see a document throughput of ~3400 vs Deleter's ~800. In total during the 3 minutes it runs, ~600k deleted documents vs ~140k.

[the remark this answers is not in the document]
josefahmad: To clarify: do you mean that the new `CrudActor/deleteOne` is deleting 3400 docs/s, vs the older `Deleter/findOneAndDelete` deletes 800 docs/s?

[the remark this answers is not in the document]
ykangpark: Yes, the new one is 3400 docs/s vs the old 800 docs/s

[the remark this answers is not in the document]
ykangpark: [Patch for old](https://spruce.mongodb.com/task/sys_perf_linux_3_node_replSet_disabled_feature_flags_clustered_collection_large_record_ids_patch_7d0634a03461f22ef0c0461a011bde9425f00a3b_6257e8d4d6d80a2c5c7e6a8d_22_04_14_09_27_29/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) findOneAndDelete, [Patch for new ](https://spruce.mongodb.com/task/sys_perf_linux_3_node_replSet_disabled_feature_flags_clustered_collection_large_record_ids_patch_7d0634a03461f22ef0c0461a011bde9425f00a3b_6255a596d6d80a1fd4bb2cba_22_04_12_16_16_02/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) deleteOne

src/phases/execution/ClusteredCollection.yml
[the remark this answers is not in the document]
ykangpark: With the latest change we use a parametrised thread count as the step.

src/phases/execution/ClusteredCollection.yml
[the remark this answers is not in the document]
josefahmad: Cool, do you know the reason for this improvement?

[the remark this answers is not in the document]
ykangpark: By looking at FTDC data, turns out that this is due to a huge amount of write conflicts in `findOneAndDelete`, while with the `deleteOne` with _id specified, there are 0 write conflicts! `findOneAndDelete` tries to delete the first entry by _id, so if there are 16 threads doing the same, they will be in constant conflict. 

See write conflicts for `findOneAndDelete`
Screenshot 2022-04-14 at 18 14 17

And `deleteOne`
Screenshot 2022-04-14 at 18 15 20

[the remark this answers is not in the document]
ykangpark: What I am not sure about is why the performance degrades with time.

src/cast_core/src/RunCommand.cpp
aggrand: Yeah a followup ticket sounds great!

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,110 @@ class DatabaseOperation {
...
+                }
+            }
+            return true;
+        }
+    };
+
+    InstanceTypeFilter(PhaseContext &context, mongocxx::pool::entry& client) {
+        try {
+            _onlyRunInInstancesContext = context.getPlural<std::string>(
+                "OnlyRunInInstance", "OnlyRunInInstances", [&](const Node& node) {
```
aggrand, line 244: So there's a singular and plural version? I'd say we just keep the plural version; people can make it a one-item list with `key: [value]` and it both makes this implementation easier and makes the overall syntax simpler.

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,110 @@ class DatabaseOperation {
...
+            _adminDB = (*client)["admin"];
+        }
+    }
+
+    // Indicates if operations should be run. If there is no OnlyRunInInstance configuration or one
+    // of the specified options matches the client instance type, returns true. Otherwise returns
+    // false.
+    bool shouldRun() {
+        if(_type.length() == 0 && _onlyRunInInstancesContext.size()) {
+            _type = getInstanceType(_adminDB);
```
aggrand, line 272: If shouldRun gets evaluated during the run(), then this `getInstanceType` gets evaluated during the phase rather than during the context setup? We should ideally only run this during context setup, the instance shouldn't change after that point so that's safe.

For some workloads, running this during the phase may be prohibitively expensive.

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,110 @@ class DatabaseOperation {
...
+            _typeFoundInConfig = false;
+            for(const auto &configInstanceType : _onlyRunInInstancesContext) {
```
aggrand, line 274: Should we go ahead and quit out if there is no configured instance type, before we make the hello query to determine it (possibly pointlessly?)

aggrand (changes requested):
Looks great overall, just had a few details I'd recommend. Definitely push back where you disagree!

I also recommend adding an actor integration test to sanity-check this: https://github.com/mongodb/genny/blob/master/src/cast_core/src/RunCommand.cpp

src/cast_core/src/RunCommand.cpp
ykangpark: That's what I tried doing in the first place. But it was failing the PR's compilation task while doing a dry-run. I tried running dry locally, and that worked fine. But I wasn't able to find out why it failed in evergreen. See [patch](https://spruce.mongodb.com/version/624f04e53066151c1c109b0d/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and [error](https://evergreen.mongodb.com/task_log_raw/genny_amazon2_t_compile_patch_25a5f5737671be446e240a71cf6cd28f3829cedd_624f04e53066151c1c109b0d_22_04_07_15_36_06/0?type=T#L899)

The conclusion I reached was that dry mode was supposed to be able to evaluate the configuration, regardless of connection to the database. And that we should not query the database in context setup. So I ended up moving the query to the server out of context setup.

Any idea of why it fails?

src/cast_core/src/RunCommand.cpp
ykangpark: This is already what is done. See line [271](https://github.com/mongodb/genny/pull/651/files#diff-05612e944a021e56717199e577fcac940d40eb7b838cf41da8361ed32fd9ed35R271). getInstanceType is only called when _onlyRunInInstancesContext.size() is not 0.

src/cast_core/src/RunCommand.cpp
ykangpark: I have seen this singular/plural scheme used in the existing actors, which is why I decided to follow this pattern. Is this no longer the recommended way? I just want to be consistent with the rest of actors.

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,110 @@ class DatabaseOperation {
...
+    };
+
+    InstanceTypeFilter(PhaseContext &context, mongocxx::pool::entry& client) {
+        try {
+            _onlyRunInInstancesContext = context.getPlural<std::string>(
+                "OnlyRunInInstance", "OnlyRunInInstances", [&](const Node& node) {
+                    return node.to<std::string>();
+            });
+        } catch(...) {
+            // Exception might be due to keys not found or other errors. If its due to keys not
```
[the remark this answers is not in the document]
ykangpark: Makes sense to only catch InvalidKeyException.

The getPlural function always throws if the specified keys are not found. We wanted this to be optional, but also leverage existing code. A way to do that is to use the existing getPlural and filter out the error if it is due to the keys not being found.

src/cast_core/src/RunCommand.cpp
ykangpark: I changed it any ways to do an early return, so the code is easier to understand.

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,135 @@ class DatabaseOperation {
...
+// Class to parse and hold the OnlyRunInInstance configuration. This option may be used to only
+// run under certain configurations Available configurations are specified in Options.
+class InstanceTypeFilter {
+public:
+    struct Options {
+        enum Code {
+            kStandalone,
+            kSharded,
+            kReplicaSet,
+            _size,          // Must before kNotInitialized
```
aggrand, line 220: This is clever

src/cast_core/src/RunCommand.cpp
aggrand: Ah, fair in that case! I don't like it personally but I can deal (and won't block on it!)

src/cast_core/src/RunCommand.cpp
```diff
@@ -208,13 +208,135 @@ class DatabaseOperation {
...
+        std::vector<std::string> stringOptions;
+        try {
+            stringOptions = context.getPlural<std::string>(
+                "OnlyRunInInstance", "OnlyRunInInstances", [&](const Node& node) {
+                    return node.to<std::string>();
+            });
+        } catch(const InvalidKeyException&) {
+            // Exception might be due to keys not found or other errors. If its due to keys not
+            // found ignore the expception. Otherwise rethrow.
+            if(context["OnlyRunInInstance"] || context["OnlyRunInInstance"]) {
```
aggrand, line 275: Should there be two different things on either side of this OR? Also, is this rethrow needed now that we're only catching the invalid key?

src/cast_core/src/RunCommand.cpp
aggrand: Thank you!

src/cast_core/src/RunCommand.cpp
aggrand: yeah dry run doesn't run against a server, so makes sense that it'll fail here.

I recommend moving it to the context and putting your workload here to be excluded: https://github.com/mongodb/genny/blob/c38a522e44a12fec43558527f110998eaa817465/src/lamplib/src/genny/tasks/dry_run.py#L28

We have [TIG-3290](https://jira.mongodb.org/browse/TIG-3290) in the future to improve this!

aggrand (changes requested):
Thanks for bearing with me! Main thing I'm blocking on is figuring out the instance type in the context setup. I'm definitely concerned that not doing so may impact other workloads.

Definitely happy to pair on it sometime if needed too!

src/cast_core/src/RunCommand.cpp
ykangpark: You are right, one of those should be the plural, thanks! 

The `getPlural` uses the same InvalidKeyExpcetion for various cases. One of them is none of the keys (singular or plural) being specified, in which case we don't want to throw and proceed with default behaviour. The same exception is used to signal that both keys exist, which is incorrect and should throw. That is why the rethrow is needed.

src/cast_core/src/RunCommand.cpp
ykangpark: In that case every workload using this feature would have to be added to the exclusion list in dry_run.py, wouldn't that be quite confusing and cumbersome? 

The result of hello is cached after the first time. Do you still think that this is prohibitively expensive?

ykangpark: I think existing workloads using this actor should see no effect because the configuration for this option will be empty and `shouldRun` will return without doing the query.

But given the concerns for now I will do as suggested, move the query to context and excluding the workload from dry run.

Thanks for all the feedback!

ykangpark:
Latest patch https://spruce.mongodb.com/version/6266c6b91e2d17672db1deba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/cast_core/src/RunCommand.cpp
```diff
@@ -232,11 +354,17 @@ struct actor::RunCommand::PhaseConfig {
...
     bool throwOnFailure;
     std::vector<std::unique_ptr<DatabaseOperation>> operations;
+    InstanceTypeFilter onlyRunInInstancesFilter;
 };
 
 void actor::RunCommand::run() {
     for (auto&& config : _loop) {
         for (auto&& _ : config) {
+            // Run ops when there is no OnlyRunInInstances option or when one of the options matches
+            if(!config->onlyRunInInstancesFilter.shouldRun()) {
```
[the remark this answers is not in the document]
aggrand: Also, thanks for adding such great test coverage!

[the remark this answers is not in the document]
ykangpark: @aggrand Hey, I also thought about that and tried it. Was getting failures in NOP phases, because the config cannot be dereferenced in nop phases. So I decided to move it back in. I just discovered that there's a `isNop` check, so I moved it out again! 

Quick test: https://spruce.mongodb.com/version/62696fed32f41723a8313d38/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/cast_core/test/RunCommandActor_test.cpp
```diff
@@ -803,5 +803,291 @@ TEST_CASE_METHOD(MongoTestFixture,
         ah.runDefaultAndVerify(verifyFn);
     }
 }
+
+TEST_CASE_METHOD(MongoTestFixture,
```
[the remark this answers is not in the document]
ykangpark: Makes sense! I added an explicit test case for that.

PR #652, opened 2022-04-07
TIG-3421 Add actor debug information to actors

vrachev:
Ok here we go again.. 

This is pretty similar to the first PR. Originally added the actor information to the phaseContext and had actors add it to their phaseConfig structs so that we could print out whether we nop or not, but the phaseConfig [does not allow accessing members](https://github.com/mongodb/genny/blob/8746510d1f97a7e9990710a4285d1a59434028c9/src/gennylib/include/gennylib/PhaseLoop.hpp#L431) when it's a nop, so I reverted to putting this info in the actor class and not printing out nop info. 

Here's a [patch](https://spruce.mongodb.com/version/624f01cc2fbabe3975799a3a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). I've added these log lines to the create-new-actor script as well. 

Example log:

```
[2022-04-07 10:45:25.666051] [0x0000700000d13000] [info]    Beginning phase 2
[2022-04-07 10:45:25.666076] [0x0000700000d13000] [debug]   Starting HelloWorld::HelloWorld::1 execution
[2022-04-07 10:45:25.666095] [0x0000700000d96000] [debug]   Starting HelloWorld::HelloWorld::2 execution
[2022-04-07 10:45:25.666105] [0x0000700000d13000] [info]    Hello Phase 2
[2022-04-07 10:45:25.666122] [0x0000700000d96000] [info]    Hello Phase 2
[2022-04-07 10:45:25.666159] [0x0000700000d13000] [info]    Counter: 1017
[2022-04-07 10:45:25.666180] [0x0000700000d96000] [info]    Counter: 1018
[2022-04-07 10:45:25.666218] [0x0000700000d13000] [info]    Hello Phase 2
[2022-04-07 10:45:25.666236] [0x0000700000d96000] [info]    Hello Phase 2
[2022-04-07 10:45:25.666248] [0x0000700000d13000] [info]    Counter: 1019
[2022-04-07 10:45:25.666263] [0x0000700000d96000] [info]    Counter: 1020
[2022-04-07 10:45:25.666278] [0x0000700000d13000] [debug]   Ended HelloWorld::HelloWorld::1 execution
[2022-04-07 10:45:25.666317] [0x0000700000d96000] [debug]   Ended HelloWorld::HelloWorld::2 execution
[2022-04-07 10:45:25.666349] [0x0000700000d96000] [info]    Ended phase 2
```

### DIFF
--- a/src/cast_core/src/HelloWorld.cpp
+++ b/src/cast_core/src/HelloWorld.cpp
@@ -25,6 +25,7 @@ namespace genny::actor {
 /** @private */
 struct HelloWorld::PhaseConfig {
     std::string message;
+    std::string actorInfo;
 
     /** record data about each iteration */
     metrics::Operation operation;
@@ -34,11 +35,13 @@ struct HelloWorld::PhaseConfig {
     explicit PhaseConfig(PhaseContext& context, ActorId actorId)
         : message{context["Message"].maybe<std::string>().value_or("Hello, World!")},
           operation{context.operation("DefaultMetricsName", actorId)},
-          syntheticOperation{context.operation("SyntheticOperation", actorId)} {}
+          syntheticOperation{context.operation("SyntheticOperation", actorId)},
+          actorInfo{context.actorInfo(actorId)} {}
 };
 
 void HelloWorld::run() {
     for (auto&& config : _loop) {
+        BOOST_LOG_TRIVIAL(debug) << "Starting " << config->actorInfo << " execution";
         for (auto _ : config) {
             auto ctx = config->operation.start();
             BOOST_LOG_TRIVIAL(info) << config->message;
@@ -51,6 +54,7 @@ void HelloWorld::run() {
             config->syntheticOperation.report(metrics::clock::now(),
                                               std::chrono::milliseconds{_helloCounter});
         }
+        BOOST_LOG_TRIVIAL(debug) << "Ended " << config->actorInfo << " execution";
     }
 }
 
--- a/src/cast_core/src/HelloWorld.cpp
+++ b/src/cast_core/src/HelloWorld.cpp
@@ -25,7 +25,6 @@ namespace genny::actor {
 /** @private */
 struct HelloWorld::PhaseConfig {
     std::string message;
-    std::string actorInfo;
 
     /** record data about each iteration */
     metrics::Operation operation;
@@ -35,13 +34,12 @@ struct HelloWorld::PhaseConfig {
     explicit PhaseConfig(PhaseContext& context, ActorId actorId)
         : message{context["Message"].maybe<std::string>().value_or("Hello, World!")},
           operation{context.operation("DefaultMetricsName", actorId)},
-          syntheticOperation{context.operation("SyntheticOperation", actorId)},
-          actorInfo{context.actorInfo(actorId)} {}
+          syntheticOperation{context.operation("SyntheticOperation", actorId)} {}
 };
 
 void HelloWorld::run() {
     for (auto&& config : _loop) {
-        BOOST_LOG_TRIVIAL(debug) << "Starting " << config->actorInfo << " execution";
+        BOOST_LOG_TRIVIAL(debug) << "Starting " << this->actorInfo() << " execution";
         for (auto _ : config) {
             auto ctx = config->operation.start();
             BOOST_LOG_TRIVIAL(info) << config->message;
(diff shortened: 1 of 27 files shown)


src/lamplib/src/genny/tasks/create-new-actor.sh
```diff
@@ -338,6 +338,11 @@ void ${actor_name}::run() {
     // Actors so that all Actors start and end at the right time.
     //
     for (auto&& config : _loop) {
+        //
+        // This will print out some information about the actor beginning execution.
+        // This will occur for every actor thread, and will print out the actor type,
+        // name and id. Debug mode is used to not spam logs for high thread actors.
+        BOOST_LOG_TRIVIAL(debug) << "Starting " << this->actorInfo() << " execution";
```
aggrand, line 345: Worth noting that I think the rate limiter is not accounted for here. I.e. an actor may print "Starting ... execution", then be rate limited and not act for a very long time, since rate limiting is part of the inner actor iteration.

Not that I think there's anything that can be done about that at this level, just probably worth noting here.

aggrand (approved):
Thanks for doing this! I'm so sorry we had to go through so many iterations here, but hopefully this end result is useful!

One question, I thought we weren't actually gonna modify any actors, just document where to insert these log lines? This is great too just concerned about:

1. Since putting the lines inside phase loop may have had a slight perf impact, is there an impact here? Should we run some patches and check?
2. People use the `-v` option to see phase transitions, and flooding the logs with many print statements for highly-threaded actors might make that tricky. This one won't hurt prod data at least, and we could always revert if people don't like it.

That said, I'm happy merging this assuming point 1 above is good. We can iterate further on this logging experience in the future. LGTM

src/gennylib/src/Orchestrator.cpp
```diff
@@ -86,7 +86,7 @@ PhaseNumber Orchestrator::awaitPhaseStart(bool block, int addTokens) {
         for (auto&& cb : _prePhaseHooks) {
             cb(this);
         }
-        BOOST_LOG_TRIVIAL(debug) << "Beginning phase " << currentPhase;
+        BOOST_LOG_TRIVIAL(info) << "Beginning phase " << currentPhase;
```
vrachev, line 89: This was specifically requested in the ticket

vrachev (commented):
Removed log stmt from all except HelloWorld. I'll put up another PR with a link to the docs. Want to use a hard link to a github commit.

src/gennylib/src/Orchestrator.cpp
aggrand: I think this is the only thing I'm slightly skeptical about. Right now phase transitions are pretty speedy, and I think @dalyd had mentioned before that there are workloads that run one phase, then transition to another, and try to measure the "hangover" from one phase to the next?

Not that printing a log line is the most expensive operation in the world or anything, but considering that it seemed to slow down unit tests enough to be noticeable in the last PR, there's likely some degree of effect here.

Main ask here is to run a faster workload like mixedworkloads and see if there's a visible difference. Sorry for the paranoia, I know it's annoying! Hopefully this sort of check will be easier as we improve patch-analysis workflows.

src/cast_core/src/HelloWorld.cpp
```diff
@@ -39,6 +39,7 @@ struct HelloWorld::PhaseConfig {
 
 void HelloWorld::run() {
     for (auto&& config : _loop) {
+        BOOST_LOG_TRIVIAL(debug) << "Starting " << this->actorInfo() << " execution";
```
aggrand, line 42: bump on the previous comment about mentioning that this doesn't account for rate limiting

aggrand (approved):
LGTM, thanks for bearing with the several rounds of review! Just had one paranoia comment, but happy to merge if you feel confident on that perf point

src/cast_core/src/HelloWorld.cpp
vrachev: Done

src/gennylib/src/Orchestrator.cpp
vrachev: Ran some patches with mixedworkloads and I think this should be fine (used the perf-viz app to do some comparisons)

vrachev:
evergreen merge

vrachev:
evergreen merge

vrachev:
Tests are passing but macos compile is broken. Manually merging

PR #653, opened 2022-04-07
PERF-2534 and PERF-2756 Time-series metric indexes (ascending/descending, 2dsphere)

dpercy:
I started a patch build here: https://spruce.mongodb.com/version/624f530661837d6b5e6c9067/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/TimeSeries2dsphere.yml
+++ b/src/workloads/execution/TimeSeries2dsphere.yml
@@ -4,10 +4,12 @@ Description: |
   This test exercises the behavior of querying for data in a timeseries collection using a 2dsphere index.
 
   The phases are:
-  1. Create collection
-  2. Insert events + create index
-  3. Quiesce
-  4. Query
+  0. Create collection
+  1. Insert events + create index
+  2. Quiesce
+  3. $geoWithin
+  4. Quiesce
+  5. $geoNear
 
 GlobalDefaults:
   # Schema constants.
@@ -72,6 +74,8 @@ Actors:
   - *Nop
   - *Nop
   - *Nop
+  - *Nop
+  - *Nop
 
 - Name: InsertData
   Type: Loader
@@ -98,6 +102,8 @@ Actors:
     - *index
   - *Nop
   - *Nop
+  - *Nop
+  - *Nop
 
 - Name: Quiesce
   Type: QuiesceActor
@@ -105,6 +111,8 @@ Actors:
   Database: *db
   Phases:
   - *Nop
+  - *Nop
+  - Repeat: 1
   - *Nop
   - Repeat: 1
   - *Nop
@@ -116,7 +124,7 @@ Actors:
   - *Nop
   - *Nop
   - *Nop
-  - Duration: 10 seconds  # TODO more
+  - Duration: 30 seconds
     Database: *db
     Operations:
     - OperationMetricsName: GeoWithinQuery
@@ -128,6 +136,28 @@ Actors:
             {^RandomDouble: { min: *minX, max: *maxX }},
             {^RandomDouble: { min: *minY, max: *maxY }},
           ], 0.001]}}}},
+          {$skip: 1e5},
+        ]
+        cursor: {}
+  - *Nop
+  - Duration: 30 seconds
+    Database: *db
+    Operations:
+    - OperationMetricsName: GeoNearQuery
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [
+          {$geoNear: {
+            key: 'location',
+            near: [
+              {^RandomDouble: { min: *minX, max: *maxX }},
+              {^RandomDouble: { min: *minY, max: *maxY }},
+            ],
+            spherical: true,
+            distanceField: 'd',
+            maxDistance: 0.001,
+          }},
           {$skip: 1e5},
         ]
         cursor: {}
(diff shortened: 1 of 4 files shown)


src/workloads/execution/TimeSeries2dsphere.yml
```diff
@@ -0,0 +1,145 @@
...
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: *db
+    Operation:
+      OperationMetricsName: CreateTimeSeriesCollection
+      OperationName: RunCommand
+      OperationCommand:
+        {create: &coll Collection1, timeseries: {timeField: "timestamp", metaField: "meta"}}
```
mattBoros, line 61: I think it might be more convenient to name the collection "Collection0" because of how the Loader works (more details about this below)

src/workloads/execution/TimeSeries2dsphere.yml
```diff
@@ -0,0 +1,145 @@
...
+  - *Nop
+  - *Nop
+  - *Nop
+  - Duration: 10 seconds  # TODO more
+    Database: *db
+    Operations:
+    - OperationMetricsName: GeoWithinQuery
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
```
mattBoros, line 121: Just a note, this is querying from Collection0 and the createTimeSeriesCollection phase created Collection1, so these queries might not be running against a time series collection.

src/workloads/execution/TimeSeries2dsphere.yml
```diff
@@ -0,0 +1,145 @@
...
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: *db
+    # Collection: *coll
+    AreYOuEvenListening: false
```
mattBoros, line 74:
```suggestion
```

src/workloads/execution/TimeSeries2dsphere.yml
```diff
@@ -0,0 +1,145 @@
...
+  - *Nop
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: *db
+    # Collection: *coll
```
mattBoros, line 73: In the [Loader cpp file](https://github.com/mongodb/genny/blob/master/src/cast_core/src/Loader.cpp#L127) the Loader assumes you want collection names "Collection0" through "Collection{CollectionCount}" and doesn't allow custom names. I looked around and a few other tests specify the collection name in the loader, but still work anyway because the Loader starts with Collection0 and the queries run against Collection0. For example [TimeSeriesSort](https://github.com/mongodb/genny/blob/master/src/workloads/execution/TimeSeriesSort.yml#L188) (one that I wrote) and [TimeSeriesLastPoint](https://github.com/mongodb/genny/blob/master/src/workloads/execution/TimeSeriesLastpoint.yml). I'll make sure to file a ticket for this, because if these tests change it could lead to us thinking we query against a collection with data when really it's an empty collection.

mattBoros (changes requested):
I requested a few changes for TimeSeries2dsphere. The generator and the secondary index test look good!

src/workloads/execution/TimeSeries2dsphere.yml
dpercy: Oops, thanks. 😀  I think I was confused about the behavior of Loader--it must be ignoring that `Collection:` argument, because it ignores other unexpected arguments.

src/workloads/execution/TimeSeries2dsphere.yml
```diff
@@ -0,0 +1,145 @@
...
+  - *Nop
+  - *Nop
+  - Duration: 10 seconds  # TODO more
+    Database: *db
+    Operations:
+    - OperationMetricsName: GeoWithinQuery
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: Collection0
+        pipeline: [
```
mattBoros, line 122: From what I can tell, this pipeline doesn't match any results right now because of the 2dsphere index bug. We could wait for the bug fix to merge, or just merge this now and check that its matching results once the bug fix merges. I think worst case we'll get a BF when the bug fix merges but not sure

mattBoros (commented):
LGTM, just one concern related to the 2dsphere bug.

mattBoros:
Oh sorry one more comment. It might be worthwhile to add a 2dWalk generator example to [Generators.yml](https://github.com/mongodb/genny/blob/master/src/workloads/docs/Generators.yml) and maybe write a test case for it in [DocumentGeneratorTestCases.yml](https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocumentGeneratorTestCases.yml). I've looked at the data added to the collection in the TimeSeries2dSphere test and it looks correct but it doesn't hurt to write a test for it just in case.

dpercy:
I added a test to [DocumentGeneratorTestCases.yml](https://github.com/mongodb/genny/blob/master/src/value_generators/test/DocumentGeneratorTestCases.yml) with some explanation of the expected behavior.  I'm not sure that adding an example to [Generators.yml](https://github.com/mongodb/genny/blob/master/src/workloads/docs/Generators.yml) would add much, because this generator is more special-purpose than the others, and there's already the geo workload and the test to demonstrate its behavior.

I also added a `$geoNear` query with the same parameters as `$geoWithin`.

mattBoros (approved):
Nice! LGTM

PR #654, opened 2022-04-08
PERF-2936 Add smaller, more realistic queries to the constant folding workload

davish:


src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -56,6 +57,14 @@ Actors:
   - *Nop
   - *Nop
   - *Nop
+  - *Nop
```
mattBoros, line 60: One way to avoid a list of no-ops in the InsertData actor is to restructure it like this:

```
- Name: InsertData
  Type: Loader
  Threads: 1
  Phases:
    OnlyActiveInPhases:
      Active: [0]
      NopInPhasesUpTo: 16
      PhaseConfig:
        Repeat: 1
        Database: *db
        Threads: 1
        CollectionCount: 1
        DocumentCount: 10000
        BatchSize: *batchSize
        Document: {
          _id: {^Inc: {}},
          x: {^RandomInt: {min: -1000, max: 1000}},
        }
```
(The phases are 0-indexed so even though there are 17 phases we write `NopInPhasesUpTo: 16` which can be confusing)

This restructuring can also be done with the QuiesceActor, but the active phases argument ends up being something like `Active: [1, 3, 5, 7, ..., 15]` which might be harder to read than the original list of phases.

mattBoros (approved):
LGTM! I just have one suggestion below

PR #662, opened 2022-04-19
PERF-2948 compound sort workload

dpercy:


src/workloads/execution/TimeSeriesSortCompound.yml
```diff
@@ -0,0 +1,117 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of the time series bounded sorter with a compound sort.
```
mattBoros, line 4: It might be helpful to add a comment about how the data looks, or mention that its the same as the description in TimeSeriesSortOverlappingBuckets.yml

src/workloads/execution/TimeSeriesSortCompound.yml
```diff
@@ -0,0 +1,117 @@
...
+          }},
+          {$_internalBoundedSort: {
+            sortKey: {m: 1, t: 1},
+            bound: {base: "min"},
+          }},
+          {$_internalInhibitOptimization: {}},
+          {$skip: 1e10},
+        ]
+        cursor: {}
+        # allowDiskUse: true # TODO enable if committing without $_internal stages?
```
mattBoros, line 99: Yea this seems right. Maybe enable it when we commit, and then when we're able to rewrite the blocking sort version of this into a bounded sort we can come back and disable it.

mattBoros (approved):
LGTM

PR #663, opened 2022-04-22
Add ^NumExpr preprocessing directive to perform arithmetic operations on YAML configuration files

ykangpark:
Many times it would be nice to have arithmetic with parameters in YAML files.

This is a proposal to add this functionality. Not sure if the syntax is the most appropriate, since it might be confusing with doc generators. Given that ^Parameter also follows this ^xxxx style, I used the same.

### DIFF
--- a/src/lamplib/src/genny/tasks/preprocess.py
+++ b/src/lamplib/src/genny/tasks/preprocess.py
@@ -291,20 +291,20 @@ class _WorkloadParser(object):
             msg = "Invalid value for 'Expr', which must be a string," f" in following node: {input}"
             raise ParseException(msg)
 
-        parsedDict = {}  # Pass empty dict to avoid yaml to access context of this function
+        parsed_values = {}  # Pass empty dict to avoid yaml to access context of this function
         if "Dict" in input:
-            inputDict = input["Dict"]
-            parsedDict = self._recursive_parse(inputDict)
-            if not all(type(value) == int or type(value) == float for value in parsedDict.values()):
+            input_values = input["Dict"]
+            parsed_values = self._recursive_parse(input_values)
+            if not all(type(value) == int or type(value) == float for value in parsed_values.values()):
                 msg = (
                     "Invalid values for 'Dict' in '^NumExpr', only numerical values are allowed.\n"
-                    f"Node source: {inputDict}\n"
-                    f"Node eval: {parsedDict}\n"
+                    f"Node source: {input_values}\n"
+                    f"Node eval: {parsed_values}\n"
                 )
                 raise ParseException(msg)
 
         try:
-            return numexpr.evaluate(input["Expr"], parsedDict).tolist()
+            return numexpr.evaluate(input["Expr"], parsed_values).tolist()
         except KeyError as e:
             msg = (
                 "Key used in 'Expr' not found in 'Dict'\n"
--- a/src/lamplib/src/genny/tasks/preprocess.py
+++ b/src/lamplib/src/genny/tasks/preprocess.py
@@ -312,6 +312,13 @@ class _WorkloadParser(object):
                 f"Faulting key: {e}\n"
             )
             raise ParseException(msg)
+        except Exception as e:
+            msg = (
+                "Failure trying to parse ^NumExpr node.\n"
+                f"Node source: {input}\n"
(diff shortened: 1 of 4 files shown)


aggrand (commented):
I don't have too strong of an opinion to this, if people would find it useful! That said, I feel like the desire for this points to a deeper need for more flexible workload configuration.

also cc @rtimmons  who might have opinions on new workload-level syntax!

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -279,6 +282,44 @@ def _replace_param(self, input):
...
+        if "Expr" not in input:
+            msg = "Invalid keys for '^NumExpr', please set 'Expr'" f" in following node: {input}"
+            raise ParseException(msg)
+
+        if type(input["Expr"]) != str:
+            msg = "Invalid value for 'Expr', which must be a string," f" in following node: {input}"
+            raise ParseException(msg)
+
+        parsedDict = {}  # Pass empty dict to avoid yaml to access context of this function
+        if "Dict" in input:
```
aggrand, line 295: Could we call this "args" or "values"? I'm not sure it's obvious how a dict would necessarily be used.

rtimmons (changes requested):
I'm not quite sure why this is necessary and fear as Raiden suggested that this might be papering over other syntax shortcomings. This said, I think it makes sense to see where this goes based on actual usage and to consider further syntax changes with real-world usage in mind.

A couple things here:

1. Please use `snake_case`
2. The `Dict` name is ambiguous. Consider `{^Expr: {withExpression:"a - b", andValues:{a:1,b:2}}`. We haven't been super rigorous with our syntax, so it will be mostly subjective. My point is to make the name at least somewhat meaningful.
3. A test-case that shows this is evaluated recursively and/or potentially within the context of a value-generator would be useful for regressions e.g. `{Document: {^FastRandomString:{length:{^Expr:...}}`.
4. We have a special `^Verbatim` syntax that allows for values to be passed through without any `^`-handling. I don't know if the existing Python is aware of this, but if so please ensure an expression like `{^Verbatim: {^Expr:...}}` gets evaluated to literal `{^Expr:...}` in the preprocessor output.

Otherwise this code looks straightforward and well-tested so not a lot of change to get this over the finish line. Please push back or ping back with any questions or disagreements.

aggrand (changes requested):
Mostly had a nitpick, and also a request that an example of this be added to a HelloWorld.yml workload in `src/workloads/docs`.

Also a point of discussion, this is getting pretty close to just evaluating arbitrary python. Is there an argument for going all-in on that to pre-empt the next time something like this is needed?

src/lamplib/src/genny/tasks/preprocess.py
ykangpark: Thanks for the feedback. Please see latest commits where this has been changed. Also added an example to HelloWorld.yml

ykangpark:
> I'm not quite sure why this is necessary and fear as Raiden suggested that this might be papering over other syntax shortcomings. This said, I think it makes sense to see where this goes based on actual usage and to consider further syntax changes with real-world usage in mind.
> 
> A couple things here:
> 
> 1. Please use `snake_case`
> 2. The `Dict` name is ambiguous. Consider `{^Expr: {withExpression:"a - b", andValues:{a:1,b:2}}`. We haven't been super rigorous with our syntax, so it will be mostly subjective. My point is to make the name at least somewhat meaningful.
> 3. A test-case that shows this is evaluated recursively and/or potentially within the context of a value-generator would be useful for regressions e.g. `{Document: {^FastRandomString:{length:{^Expr:...}}`.
> 4. We have a special `^Verbatim` syntax that allows for values to be passed through without any `^`-handling. I don't know if the existing Python is aware of this, but if so please ensure an expression like `{^Verbatim: {^Expr:...}}` gets evaluated to literal `{^Expr:...}` in the preprocessor output.
> 
> Otherwise this code looks straightforward and well-tested so not a lot of change to get this over the finish line. Please push back or ping back with any questions or disagreements.

Thanks for the feedback. 
I addressed all points except point 4, given that ^Verbatim is processed at DocumentGenerator.cpp and the Python is not aware.

rtimmons (approved):

PR #665, opened 2022-04-28
TIG-3722 Document WiredTiger genny workloads

korteland:
Just some docs, hopefully nothing too controversial.

### DIFF
--- a/src/workloads/scale/LargeScaleSerial.yml
+++ b/src/workloads/scale/LargeScaleSerial.yml
@@ -1,7 +1,21 @@
 SchemaVersion: 2018-07-01
 Owner: STM
 Description: |
-  TODO: TIG-3722
+  This config simulates a "typical" large-scale OLTP workload running in parallel with a
+  slower analytical operation. This is a use-case intended to improve with the addition of
+  durable history.
+
+  The first phase loads data while performing updates, lasts for one hour, and expects an
+  overall update throughput of ten updates per millisecond.
+
+  The second phase is a no-op.
+
+  The third phase lasts for another hour, and runs three parallel query operations: a
+  warmup that queries 100 collections, and two larger ones that query all 10K
+  collections. These aren't performance-sensitive, they just serve to keep various "older"
+  records around (hence the rate-limiting). The queries run in parallel with a full
+  collection scan, which represents a conventional performance-sensitive OLTP workload
+  that shouldn't be too affected by the long-running queries.
 
 Clients:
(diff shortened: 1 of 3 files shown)


korteland:
I've looked at the CI failures, I don't believe they're related to this change.

src/workloads/scale/LargeScaleSerial.yml
```diff
@@ -1,7 +1,21 @@
...
+  overall update throughput of ten updates per millisecond.
+
+  The second phase is a no-op.
+
+  The third phase lasts for another hour, and runs three parallel query operations: a
+  warmup that queries 100 collections, and two larger ones that query all 10K
+  collections. These aren't performance-sensitive, they just serve to keep various "older"
+  records around (hence the rate-limiting). The queries run in parallel with a full
+  collection scan, which represents a conventional performance-sensitive OLTP workload
+  that shouldn't be too affected by the long-running queries.
```
jimoleary, line 18: It looks like you have covered the what and the why. 

Is is possible to describe what metrics are important for these workloads?

Can we also get some 'Keywords:' in here and maybe duplicated into [LargeScaleParallel.yml](https://github.com/mongodb/genny/pull/665/files#diff-b5b53a513b927e3d5fa30c0ff5f7d95b17e0b74880c037c10516094d63d48705)?

E.g. See [BigUpdate.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/BigUpdate.yml#L9-L15) for some inspiration but whatever you think would help others discover these workloads.

jimoleary (commented):
Is STM the correct owner for these workloads?

Some general comments / suggestions.

src/workloads/scale/LargeScaleSerial.yml
korteland: > Is STM the correct owner for these workloads?

I think so, assuming it's short for "storage (engines) team".

> Is is possible to describe what metrics are important for these workloads?

Sure, I'll try to chase this up. These were added before my time, so it might be a question for Don.

> Can we also get some 'Keywords:' in here and maybe duplicated into LargeScaleParallel.yml?

Yep, I've pushed that. Thanks

jimoleary (approved):
LGTM as is, even better if we can get some idea of the target metrics.

korteland:
Thanks. Merged as-is, it's easy enough to update if @ddanderson gets back to me with some metrics (or any other comments).

PR #666, opened 2022-04-28
PERF-2696: Adjust actor and phase threads to match in mixed_workloads_stress

dalyd:
Patch build: https://spruce.mongodb.com/version/626adf410ae6067e047fdcb2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

It looks like this workload is still inserting documents, but I don't know how with that setting of threads. This should make it sane.

### DIFF
--- a/src/workloads/scale/MixedWorkloadsGennyStress.yml
+++ b/src/workloads/scale/MixedWorkloadsGennyStress.yml
@@ -148,7 +148,7 @@ Actors:
   Phases:
   - Repeat: 1
     BatchSize: 100
-    Threads: 8
+    Threads: 1
     DocumentCount: *NumDocs
     Database: *dbname
     CollectionCount: *NumColls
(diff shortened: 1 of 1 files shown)


src/workloads/scale/MixedWorkloadsGennyStress.yml
```diff
@@ -148,7 +148,7 @@ Actors:
   Phases:
   - Repeat: 1
     BatchSize: 100
-    Threads: 8
+    Threads: 1
```
brad-devlugt, line 151: I'm a bit confused why this is necessary if the workload is correctly loading documents.  Should we add some documentation with this change about difference between the Threads setting on phases and Actors? Will this reduce the speed of the load phase?

dalyd: I misread my own code again. I'm going to push an update, so they are both 8. This workload has 512 collections, so 8 threads can insert into 64 collections a piece. 

What i think is happening today is 1 thread is pushing documents into 64 collections, rather than 8 threads pushing into 512 collections total.

dalyd:
Updated to use the 8 loader threads. There's a perf impact and it will create a change point. I still think it's the right thing to do.

brad-devlugt (approved):

dalyd:
evergreen merge

PR #667, opened 2022-04-28
[NOMERGE] Fix loader

dalyd:
I think this does what is needed. For discussion. It probably needs a little more love and care, and we should clear out the phase threads from all the other workloads. 

Patch build: https://spruce.mongodb.com/version/626adf5c61837d2d352c1153/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -53,7 +53,7 @@ struct Loader::PhaseConfig {
           numCollections{multipleThreadsPerCollection
                              ? 1
                              : context["CollectionCount"].to<IntegerSpec>() /
-                                 context["Threads"].to<IntegerSpec>()},
+                                 totalThreads},
           numDocuments{context["DocumentCount"].to<IntegerSpec>()},
           batchSize{context["BatchSize"].to<IntegerSpec>()},
           documentExpr{context["Document"].to<DocumentGenerator>(context, id)},
@@ -77,11 +77,6 @@ struct Loader::PhaseConfig {
         };
 
         if (multipleThreadsPerCollection) {
-            if (context["Threads"]) {
-                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
-                    "Phase Config 'Threads' parameter is not supported if "
-                    "'MultipleThreadsPerCollection' is true."));
-            }
             uint collectionCount = context["CollectionCount"].to<IntegerSpec>();
             if (totalThreads % collectionCount != 0) {
                 std::ostringstream ss;
@@ -100,10 +95,10 @@ struct Loader::PhaseConfig {
             }
         } else {
             createIndexes();
-            if (thread == context["Threads"].to<int>() - 1) {
+            if (thread == totalThreads - 1) {
                 // Pick up any extra collections left over by the division
                 numCollections +=
-                    context["CollectionCount"].to<uint>() % context["Threads"].to<uint>();
+                    context["CollectionCount"].to<uint>() % totalThreads();
             }
         }
     }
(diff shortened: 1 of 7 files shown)


src/cast_core/src/Loader.cpp
```diff
@@ -77,11 +77,6 @@ struct Loader::PhaseConfig {
         };
 
         if (multipleThreadsPerCollection) {
-            if (context["Threads"]) {
-                BOOST_THROW_EXCEPTION(InvalidConfigurationException(
-                    "Phase Config 'Threads' parameter is not supported if "
```
aggrand, line 82: What was the original rationale for this? Does it shed any light on whether there's a reason for both phase and actor level config?

src/workloads/docs/RandomSampler.yml
```diff
@@ -13,12 +13,11 @@ Clients:
 Actors:
 - Name: Loader
   Type: Loader
-  Threads: 50
+  Threads: 1
```
jimoleary, line 16: This seems totally different to me?

src/workloads/scale/MixedWorkloadsGennyStress.yml
```diff
@@ -144,7 +144,7 @@ ActorTemplates:
 Actors:
 - Name: Setup
   Type: Loader
-  Threads: 1
+  Threads: 8
   Phases:
   - Repeat: 1
     BatchSize: 100
```
jimoleary, line 150: I'm pretty sure the next line does nothing now (and is never even accessed).

src/workloads/scale/OutOfCacheScanner.yml
```diff
@@ -16,20 +16,18 @@ Clients:
 Actors:
 - Name: Loader
   Type: Loader
-  Threads: 60
+  Threads: 1
```
jimoleary, line 19: I don't see how this is the same as before.

jimoleary (commented):
Doesn't this change completely remove the 'Threads' setting from the phase?

It's certainly not being used anywhere in the code AFAICS.

Plus I am confused about the changes from 50 to 1 and 60 to 1.

Have we tried a patch?

src/cast_core/src/Loader.cpp
dalyd: @jimoleary this was your code I think

src/workloads/docs/RandomSampler.yml
dalyd: So, worth discussion. I set this to 1 because previously the Phase.Threads was 1, so I think only 1 thread was doing something. I could be wrong on that. And maybe 50 was closer to the real intent.

src/workloads/scale/MixedWorkloadsGennyStress.yml
dalyd: Yes -- I meant to delete it. Will push that on up. I did not go and remove all the vestigal thread settings though.

dalyd:
> Doesn't this change completely remove the 'Threads' setting from the phase?
> 
Yes, that was the intent. Is there reason to keep Threads in the phase? 

> It's certainly not being used anywhere in the code AFAICS.
> 
> Plus I am confused about the changes from 50 to 1 and 60 to 1.

Those two cases had mismatch between actor threads and phase threads. I picked the lower. 

> 
> Have we tried a patch?
Yep, it's linked in the description, although I haven't inspected it yet.

src/cast_core/src/Loader.cpp
jimoleary: This code ensures that the multiple threads code doesn't use the phase threads configuration as it doesn't really need it.

I think that if we were going forward with this version of the code we should throw an error when there is phase thread configured. Although I am not sure if we have covered all the cases especially w.r.t some of the workloads

dalyd:
@aggrand after talking to Jim, I think this is going to be more work than I can commit right now to get this over the line. I'm going to add a comment to TIG-3016 and link to this PR. Hopefully someone can pick this up and get it over the line.

PR #669, opened 2022-04-29
PERF-2969: Add assert if Loader phase threads > Loader Actor threads

dalyd:
Patch build: https://evergreen.mongodb.com/patch/626c3d8a1e2d1750d3d66217?redirect_spruce_users=true

### DIFF
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -79,7 +79,7 @@ struct Loader::PhaseConfig {
         if (context["Threads"] && context["Threads"].to<IntegerSpec>() > totalThreads) {
             BOOST_THROW_EXCEPTION(
                 InvalidConfigurationException("Phase Config 'Threads' parameter must be less than "
-                                              "or equal to Actor Config 'Threads' in Loader actor"))
+                                              "or equal to Actor Config 'Threads' in Loader actor"));
         }
 
         if (multipleThreadsPerCollection) {
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -76,6 +76,12 @@ struct Loader::PhaseConfig {
             }
         };
 
+        if (context["Threads"] && context["Threads"] > totalThreads) {
+            BOOST_THROW_EXCEPTION(
+                InvalidConfigurationException("Phase Config 'Threads' parameter must be less than "
+                                              "or equal to Actor Config 'Threads' in Loader actor"))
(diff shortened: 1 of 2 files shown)


dalyd:
This is meant to be the simplest thing which would have caught the bad behavior I just fixed.

src/cast_core/src/Loader.cpp
```diff
@@ -76,6 +76,12 @@ struct Loader::PhaseConfig {
             }
         };
 
+        if (context["Threads"] && context["Threads"].to<IntegerSpec>() > totalThreads) {
```
jimoleary, line 79: This should be on line 108. 

(Otherwise you will get a series of InvalidConfigurationException("Phase Config 'Threads' parameter must be less than  or equal to Actor Config 'Threads' in Loader actor")); followed by InvalidConfigurationException("Phase Config 'Threads' parameter is not supported if  MultipleThreadsPerCollection' is true.")); when multipleThreadsPerCollection is set)

jimoleary (changes requested):

src/cast_core/src/Loader.cpp
dalyd: Ah, good catch. Thanks. Moving

src/cast_core/src/Loader.cpp
```diff
@@ -99,6 +100,11 @@ struct Loader::PhaseConfig {
                 createIndexes();
             }
         } else {
+            if (context["Threads"] && context["Threads"].to<IntegerSpec>() > totalThreads) {
```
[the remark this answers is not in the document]
dalyd: Oh gosh, that's a different error though that will already fail early. If we want that I think it should be a separate check with a different BOOST_THROW_EXCEPTION. I can do it, but I'm not sure that's the right thing to do. I was aiming at the smallest thing in which we could get silent errors. This additional change would make a non-silent error clearer.

jimoleary (approved):
ok, let's just make some incremental improvements.

Can we raise a ticket to consider how we handle the 2 different threads definitions (or link to an existing one)?

TBH, the fact that it is defined twice and handled in different / surprising ways is the main issue with this actor.

I'm lgtm with a linked ticket.

jimoleary:
> I've used the existing https://jira.mongodb.org/browse/TIG-3016 for the follow-up and linked my nomerge PR against it. That should completely get rid of the 2 thread definitions. Does that sound good?

Yeah, that works for me. LGTM!

PR #672, opened 2022-05-03
PERF-2984  remove create view metrics

jimoleary:
[patch](https://spruce.mongodb.com/version/627151291e2d170bc6a17622/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/phases/tpch/denormalized/Q15.yml
+++ b/src/phases/tpch/denormalized/Q15.yml
@@ -11,8 +11,7 @@ TPCHDenormalizedQuery15:
   Repeat: 1
   Database: tpch
   Operations:
-  - OperationMetricsName: Query15CreateView
-    OperationName: RunCommand
+  - OperationName: RunCommand
     OperationCommand:
       create: revenue
       viewOn: customer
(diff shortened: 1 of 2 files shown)


jimoleary:
@alyacb  this is similar to the drop view issue.

src/phases/tpch/denormalized/Q15.yml
```diff
@@ -11,8 +11,7 @@ TPCHDenormalizedQuery15:
   Repeat: 1
   Database: tpch
   Operations:
-  - OperationMetricsName: Query15CreateView
-    OperationName: RunCommand
+  - OperationName: RunCommand
```
alyacb, line 14: Just one question (non-blocking)- as I understand it this still gets tracked as DatabaseOperation.4.4 or similar in OperationMetrics even once the name is removed. Is that intended?

alyacb (approved):
LGTM! One small question for my own understanding.

src/phases/tpch/denormalized/Q15.yml
jimoleary: > Just one question (non-blocking)- as I understand it this still gets tracked as DatabaseOperation.4.4 or similar in OperationMetrics even once the name is removed. Is that intended?

Oh you are correct, so that's not good! I have just moved the problem to a very obscurely named metric. 

I'll try and clarify the situation with genny. I'll open a feature request if it is not possible to exclude a phase metric. Although I'll probably merge / close this as well.

Thanks!

PR #677, opened 2022-05-12
PERF-2985 Add query workloads for columnar indexes

yun-soo:
This PR is to add a query workload for columnar indexes and a companion PR with other PRs which have been submitted separately.
- system_perf.yml changes: https://github.com/10gen/mongo/pull/5229
- DSI test_control: https://github.com/10gen/dsi/pull/1053

@jimoleary,
Decided to not measure either cold cache or warm cache scenarios but just measure hot cache scenarios after discussion with the team.

As of now, all queries in the workload can leverage columnar indexes except `TopErrorsQuery` which uses a regular scan even though there is a columnar index.

Evergreen patch build:
https://spruce.mongodb.com/version/627d69980ae6067da38f8de6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/ColumnstoreIndexes.yml
+++ b/src/workloads/execution/ColumnstoreIndexes.yml
@@ -36,7 +36,6 @@ TotalTenantCreatedQuery: &TotalTenantCreatedQuery
       {"$project": {"value": "$__alias_0", "_id": 0}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 ActiveCloudUsersLastMonthQuery: &ActiveCloudUsersLastMonthQuery
@@ -56,7 +55,6 @@ ActiveCloudUsersLastMonthQuery: &ActiveCloudUsersLastMonthQuery
       {"$project": {"value": "$__alias_0", "target": "$__alias_1", "_id": 0}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 VersionsHistogramQuery: &VersionsHistogramQuery
@@ -81,7 +79,6 @@ VersionsHistogramQuery: &VersionsHistogramQuery
       {"$project": {"x": "$__alias_0", "y": "$__alias_2", "color": "$__alias_1", "_id": 0}},
       {"$limit": 50000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 50000}
 
 UniqueUsersPerMonthQuery: &UniqueUsersPerMonthQuery
@@ -134,7 +131,6 @@ UniqueUsersPerMonthQuery: &UniqueUsersPerMonthQuery
       {"$sort": {"x.year": 1, "x.month": 1}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 TopErrorsQuery: &TopErrorsQuery
@@ -182,7 +178,6 @@ TopErrorsQuery: &TopErrorsQuery
       },
       {"$limit": 50000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 50000}
 
 NumberOfChartsPerTypeQuery: &NumberOfChartsPerTypeQuery
@@ -202,7 +197,6 @@ NumberOfChartsPerTypeQuery: &NumberOfChartsPerTypeQuery
       {"$project": {"x": "$__alias_1", "y": "$__alias_0", "_id": 0}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 EventsByVersionQuery: &EventsByVersionQuery
@@ -246,7 +240,6 @@ EventsByVersionQuery: &EventsByVersionQuery
       {"$sort": {"x.year": 1, "x.month": 1, "x.date": 1, "x.hours": 1}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 BrowserUsageByDistinctUserQuery: &BrowserUsageByDistinctUserQuery
@@ -281,7 +274,6 @@ BrowserUsageByDistinctUserQuery: &BrowserUsageByDistinctUserQuery
       {"$project": {"label": "$__alias_0", "value": "$__alias_1", "_id": 0}},
       {"$limit": 5000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 5000}
 
 TopTenantsByNoOfEventsQuery: &TopTenantsByNoOfEventsQuery
@@ -310,7 +302,6 @@ TopTenantsByNoOfEventsQuery: &TopTenantsByNoOfEventsQuery
       {"$project": {"color": "$__alias_0", "x": "$__alias_2", "y": "$__alias_1", "_id": 0}},
       {"$limit": 50000}
     ]
-  allowDiskUse: true
   cursor: {batchSize: 50000}
 
 # Defines test operations.
--- a/src/workloads/execution/ColumnstoreIndexes.yml
+++ b/src/workloads/execution/ColumnstoreIndexes.yml
@@ -20,33 +20,12 @@ Database: &db charts-metrics
 # Defines the collection name.
 Collection: &Collection events
 
-DateFieldProjectionQuery: &DateFieldProjectionQuery
-  aggregate: *Collection
-  pipeline:
-    [{"$project": {"_id": 0, "created_at": 1}}]
-  allowDiskUse: true
-  cursor: {batchSize: 2000}
-
-DistinctStringFieldProjectionQuery: &DistinctStringFieldProjectionQuery
-  aggregate: *Collection
-  pipeline:
-    [{"$project": {"_id": 0, "metadata.tenant_id": 1}}]
-  allowDiskUse: true
-  cursor: {batchSize: 2000}
-
-DuplicatedStringFieldProjectionQuery: &DuplicatedStringFieldProjectionQuery
-  aggregate: *Collection
-  pipeline:
-    [{"$project": {"_id": 0, "action": 1}}]
-  allowDiskUse: true
-  cursor: {batchSize: 2000}
-
 CountPerActionQuery: &CountPerActionQuery
   aggregate: *Collection
   pipeline:
-    [{"$group": {"_id": "$action", o: {$sum: 1}}}]
+          [{"$group": {"_id": "$action", o: {$sum: 1}}}, {"$limit": 5000}]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 TotalTenantCreatedQuery: &TotalTenantCreatedQuery
   aggregate: *Collection
@@ -59,7 +38,7 @@ TotalTenantCreatedQuery: &TotalTenantCreatedQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 ActiveCloudUsersLastMonthQuery: &ActiveCloudUsersLastMonthQuery
   aggregate: *Collection
@@ -79,7 +58,7 @@ ActiveCloudUsersLastMonthQuery: &ActiveCloudUsersLastMonthQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 VersionsHistogramQuery: &VersionsHistogramQuery
   aggregate: *Collection
@@ -104,7 +83,7 @@ VersionsHistogramQuery: &VersionsHistogramQuery
       {"$limit": 50000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 50000}
 
 UniqueUsersPerMonthQuery: &UniqueUsersPerMonthQuery
   aggregate: *Collection
@@ -157,7 +136,7 @@ UniqueUsersPerMonthQuery: &UniqueUsersPerMonthQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 TopErrorsQuery: &TopErrorsQuery
   aggregate: *Collection
@@ -205,7 +184,7 @@ TopErrorsQuery: &TopErrorsQuery
       {"$limit": 50000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 50000}
 
 NumberOfChartsPerTypeQuery: &NumberOfChartsPerTypeQuery
   aggregate: *Collection
@@ -225,7 +204,7 @@ NumberOfChartsPerTypeQuery: &NumberOfChartsPerTypeQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 EventsByVersionQuery: &EventsByVersionQuery
   aggregate: *Collection
@@ -269,7 +248,7 @@ EventsByVersionQuery: &EventsByVersionQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 BrowserUsageByDistinctUserQuery: &BrowserUsageByDistinctUserQuery
   aggregate: *Collection
@@ -304,7 +283,7 @@ BrowserUsageByDistinctUserQuery: &BrowserUsageByDistinctUserQuery
       {"$limit": 5000}
     ]
   allowDiskUse: true
-  cursor: {batchSize: 2000}
+  cursor: {batchSize: 5000}
 
 TopTenantsByNoOfEventsQuery: &TopTenantsByNoOfEventsQuery
   aggregate: *Collection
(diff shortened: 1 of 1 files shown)


src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,467 @@
...
+  - Repeat: *TestRepeatCount
+    Database: *db
+    Operations: *TestOperations
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-all-feature-flags
+    branch_name:
```
jimoleary, line 460: This is to run on master only? Would $eq be simpler (and work)

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,467 @@
...
+TestRepeatCount: &TestRepeatCount 10
+
+Actors:
+# The 'WarmupCache1' actor warms up the buffer cache.
+- Name: WarmupCache1
+  Type: RunCommand
+  Database: *db
+  Thread: 1
+  Phases:
+  - Repeat: 2
```
jimoleary, line 391: Suggestion: [OnlyActiveInPhases](https://github.com/mongodb/genny/blob/master/src/workloads/scale/LoadTest.yml#L46-L58) might cut down on the verbosity a bit (and insulate you from any later changes).

jimoleary (commented):
generally lgtm with some suggestions.

How noisy are these tests likely to be?

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,467 @@
...
+
+ActiveCloudUsersLastMonthQuery: &ActiveCloudUsersLastMonthQuery
+  aggregate: *Collection
+  pipeline:
+    [
+      {"$addFields": {"target": 3000}},
+      {
+        "$match": {
+          "metadata.target": {"$in": ["cloud"]},
+          "created_at": {"$gte": {"$date": "2020-03-09T01:10:04Z"}},
```
borosaurus, line 72: $date isn't actually an expression, so I think this is doing comparisons with the object literal `{$date: "blah blah"}` which prevents it from returning any results. Is there a way to check that the server returns X (or even greater than 0) results from the query?

borosaurus (changes requested):
Thanks for putting this all together! One point about some of the queries.

One other thought: I'm not totally sure about running the query over and over now that I think about it, since this feature is more targeting one-off queries that people are running manually. If we only ran each query once, would that give us enough signal?

Also curious what @RedBeard0531 since he's already done some work on perf testing this.

src/workloads/execution/ColumnstoreIndexes.yml
yzhang1991: I didn't follow this, what is the problem with comparing with object literal `{"$date": "2020-03-09T01:10:04Z"}`?

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,467 @@
...
+
+# Defines the collection name.
+Collection: &Collection events
+
+DateFieldProjectionQuery: &DateFieldProjectionQuery
+  aggregate: *Collection
+  pipeline:
+    [{"$project": {"_id": 0, "created_at": 1}}]
+  allowDiskUse: true
+  cursor: {batchSize: 2000}
```
yzhang1991, line 28: Can you help me better understand why we are setting the batch size here?

src/workloads/execution/ColumnstoreIndexes.yml
yzhang1991: Good to know such thing exists!

src/workloads/execution/ColumnstoreIndexes.yml
yzhang1991: I think Yoonsoo meant master and release version > 6.0 here.
True that we want this on master now, but if we do $eq master here then we need to come back and update this after we have 6.1 and later branched.

yzhang1991 (dismissed):

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: Exactly Ethan!

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: AFAIK, the default batch size is 201. Here, I'm trying to increase the batch size so that returning result is more efficient.

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: @borosaurus, you're right. It should be `{"$gte": ISODate("2020-03-09T01:10:04Z")}`. I don't think that there's $date operator. Instead, we have $dateFromString. Fixed.

borosaurus: Yeah, I wish the behavior there would be different too. What's even worse is that the shell will sometimes print values using the `$date` syntax.

Anyway, one last question: is there any way to do this check for number of results within genny, rather than doing it manually?

borosaurus:
Hmm, something odd is going on.

I'm looking at the patch build latency results [here](https://spruce.mongodb.com/task/sys_perf_linux_standalone_all_feature_flags_queries_on_columnstore_indexes_patch_203430ac9559f82ca01e3cbb3b0e09149fec0835_627d69980ae6067da38f8de6_22_05_12_20_11_36/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) and am seeing that the column index is sometimes 10x slower than the coll scan. For example with the UniqueUsersPerMonthQuery, I see this for Latency50.

Do we know what exactly "latency50" is measuring? Also, it is possible there's a bug in the column index itself, but a difference this big is suspicious.

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: I don't know. It looks like that Alya is adding the assertive actor as part of TPC-H benchmarks:

https://github.com/mongodb/genny/pull/674

BTW, all queries are copied and pasted from https://docs.google.com/document/d/12Q_-w1PYDJD8nkRIXKSkGGSjJVsSn6JWrC3_wXl51vc/edit#

Better to update those queries too. I have neither comment access nor edit access to the doc.

Why are you asking? Is it a suggestion or a pure question out of curiosity?

yun-soo:
> Hmm, something odd is going on.
> 
> I'm looking at the patch build latency results [here](https://spruce.mongodb.com/task/sys_perf_linux_standalone_all_feature_flags_queries_on_columnstore_indexes_patch_203430ac9559f82ca01e3cbb3b0e09149fec0835_627d69980ae6067da38f8de6_22_05_12_20_11_36/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) and am seeing that the column index is sometimes 10x slower than the coll scan. For example with the UniqueUsersPerMonthQuery, I see this for Latency50.
> 
> Do we know what exactly "latency50" is measuring? Also, it is possible there's a bug in the column index itself, but a difference this big is suspicious.

AFAIK, Latency99, Latency95, LatencyXX are 99 percentile, 95 percentile and etc. So, 50 should mean median.

yun-soo (commented):
Addressed PR comments. Please, review!

yun-soo:
Evergreen patch build:
https://spruce.mongodb.com/version/628beee43066155cf6933093/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/execution/ColumnstoreIndexes.yml
borosaurus: I'm just asking because if we change this benchmark in the future I don't want it to break without our knowing. But if there's no way to do basic correctness tests here, there's not much to be done :/

yun-soo: I see. It sounds like a suggestion then. That's a valid concern and it's a current limitation of Genny. As I replied in the previous comment, Alya is adding an assertive actor and I think the actor can be used for result verification. Will wait for Alya's PR to be merged and add result verification if this PR is not urgent. LMK if that's desirable.

borosaurus: Up to you. That PR is 19 days old so it may take awhile. We can always merge this and add the assertion later.

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,495 @@
...
+        "$project": {
+          "_id": 0,
+          "__alias_0": "$_id.__alias_0",
+          "__alias_1": "$_id.__alias_1",
+          "__alias_2": {
+            "$size": "$__alias_2"
+          }
+        }
+      },
+      {"$project": {"x": "$__alias_0", "y": "$__alias_2", "color": "$__alias_1", "_id": 0}},
```
borosaurus, line 155: [nit] any reason for the odd formatting here?

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,495 @@
...
+    [
+      {
+        "$match": {
+          "resource": {"$in": ["DashboardItem"]},
+          "action": {"$in": ["added"]},
+          "metadata.version": {"$nin": [null, "0.0.0"]},
+          "created_at": {"$gte": ISODate("2020-01-09T01:13:37Z")}
+        }
+      },
+      {"$group": {"_id": {"__alias_0": "$metadata.chart_type"}, "__alias_1": {"$sum": 1}}},
```
borosaurus, line 222: [nit] same question about formatting

borosaurus (approved):
LGTM, many thanks for putting this all together!

I looked at the numbers in the last patch and it appears to be roughly even between collscan and column scan, which relieves my concern from the last patch build where column scan was much slower.

The one last thing I would confirm is that we're actually using the column index when we expect to. With the recent changes in flags etc this has become very fragile.

Also, can you file follow on tickets for the larger dataset and using the assertion actor?

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: I'm not sure what you mean by "odd formatting". If you meant why a separate line for $project, then my answer would be "to reduce number of lines" if a stage can be fit into a line.

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: Same answer as before

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: ok, then will merge this and add verification later.

yun-soo:
> LGTM, many thanks for putting this all together!
> 
> I looked at the numbers in the last patch and it appears to be roughly even between collscan and column scan, which relieves my concern from the last patch build where column scan was much slower.
> 
Could you look at numbers at https://spruce.mongodb.com/version/628c47700305b97f309e834c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC?

The patch build that you took a look at had a bug that we didn't turn off internalQueryForceClassicEngine.

> The one last thing I would confirm is that we're actually using the column index when we expect to. With the recent changes in flags etc this has become very fragile.
> 
> Also, can you file follow on tickets for the larger dataset and using the assertion actor?

Filed two PERF tickets:
- https://jira.mongodb.org/browse/PERF-3059
- https://jira.mongodb.org/browse/PERF-3060

yun-soo (commented):
Replying to Ian's comments

borosaurus:
> Could you look at numbers at https://spruce.mongodb.com/version/628c47700305b97f309e834c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC? Are these numbers explainable?
> 
> The patch build that you took a look at had a bug that we didn't turn off internalQueryForceClassicEngine.

It looks like the column index is much much slower than the collscan. ~5x, maybe more. If this is the reality today, then I'm still good with merging these changes. Do you have the data/testing locally set up? (Would it be easy to verify that this isn't a bug in the genny test, and that the column index is actually slower?)

> Filed two PERF tickets:
Thanks!

yun-soo:
> > Could you look at numbers at https://spruce.mongodb.com/version/628c47700305b97f309e834c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC? Are these numbers explainable?
> > The patch build that you took a look at had a bug that we didn't turn off internalQueryForceClassicEngine.
> 
> It looks like the column index is much much slower than the collscan. ~5x, maybe more. If this is the reality today, then I'm still good with merging these changes. Do you have the data/testing locally set up? (Would it be easy to verify that this isn't a bug in the genny test, and that the column index is actually slower?)

Yeah~ I'm trying to get a sense of whether it's a bug in Genny workload or a real columnstore index perf. I don't have the definite answer yet. I have a lot of things on my plate right now and it's hard to find quality focus time on this.

Will try to confirm either direction soon. Maybe the dataset is just too small to show benefits of columnstore indexes.

yun-soo:
> > > Could you look at numbers at https://spruce.mongodb.com/version/628c47700305b97f309e834c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC? Are these numbers explainable?
> > > The patch build that you took a look at had a bug that we didn't turn off internalQueryForceClassicEngine.
> > 
> > 
> > It looks like the column index is much much slower than the collscan. ~5x, maybe more. If this is the reality today, then I'm still good with merging these changes. Do you have the data/testing locally set up? (Would it be easy to verify that this isn't a bug in the genny test, and that the column index is actually slower?)
> 
> Yeah~ I'm trying to get a sense of whether it's a bug in Genny workload or a real columnstore index perf. I don't have the definite answer yet. I have a lot of things on my plate right now and it's hard to find quality focus time on this.
> 
> Will try to confirm either direction soon. Maybe the dataset is just too small to show benefits of columnstore indexes.

I'm getting consistent results on both Evergreen and local environment. I think that the dataset is just too small. Merging this PR.

yun-soo:
evergreen merge

yun-soo:
@jimoleary, is there any chance for you to be able to take a second look?

yun-soo:
@RedBeard0531, @jimoleary, @yzhang1991, @cswanson310,
Can anyone of you approve this PR? Interestingly, merging is blocked though @borosaurus approved this PR.

jimoleary (dismissed):
LGTM

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,463 @@
...
+  aggregate: *Collection
+  pipeline:
+    [
+      {"$match": {"metadata.target": "cloud"}},
+      {"$group": {"_id": {}, "__alias_0": {"$addToSet": "$metadata.tenant_id"}}},
+      {"$project": {"_id": 0, "__alias_0": {"$size": "$__alias_0"}}},
+      {"$project": {"value": "$__alias_0", "_id": 0}},
+      {"$limit": 5000}
+    ]
+  allowDiskUse: true
```
cswanson310, line 39: This is the default now, right? Consider just removing this throughout to reduce the noise for the reader.

src/workloads/execution/ColumnstoreIndexes.yml
cswanson310: Actually, are you sure that you are exhausting the cursor? It looks like possibly you are just running the 'aggregate' command which generates a cursor and the first batch, but I don't see any logic that would make sure that that cursor is iterated to completion. Could you check on this or let me know if I'm missing something?

cswanson310 (changes requested):

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: sure, done. Some people wanted to see it explicitly in other PRs. Everyone has different preference. So...

yun-soo (commented):
Thanks for review, @cswanson310! Addressed your comments. And in the latest commit, I matched the batchsize to the limit size. LMK if this still may cause issues.

src/workloads/execution/ColumnstoreIndexes.yml
cswanson310: Ah! Sorry to flip flop on it! Yes I had the same thought about multi version and that column store wouldn't make sense in that scenario. But I am happy with either - consistency is also valuable.

src/workloads/execution/ColumnstoreIndexes.yml
```diff
@@ -0,0 +1,463 @@
...
+  cursor: {batchSize: 5000}
```
cswanson310, line 39: Regarding the batch size, I do think this is not a good pattern to use for future testing or for larger data sets. As we discussed on slack:
> What I am worried about is that we are not exhausting the cursor. It is related to batch size but I don't think can be fixed by only specifying or not specifying that parameter. 
We are only generating aggregate commands in this workload and never getMore commands, so there is no guarantee that we see all results even if we match the batch size to the limit, because
> To clarify why it can't be fixed with batch size: the batch size is an upper bound on each batch. Fundamentally the wire protocol uses BSON and BSON is capped at 16MB. If you request 50,000 things but we can only fit 10,000 then we will send 10,000 and a cursorId

Could you please file a follow-up ticket to either introduce some actors that can exhaust cursors or do what we do in some other workloads like [add a trailing $skip](https://github.com/mongodb/mongo-perf/blob/905cd5185fc520d6952a292b04ecb3bddee2c2bb/testcases/pipelines.js#L94-L128) or [use explain to execute the query](https://github.com/mongodb/genny/commit/6ccaf2baa6b030d731bec31a36310d1964fd64c9) - looks like that was not favored and phased out actually. 

This does seem to be a more pervasive problem, as I notice a bunch of tests using the batchSize in this way.

cswanson310 (approved):

src/workloads/execution/ColumnstoreIndexes.yml
yun-soo: Thanks for suggestion.
Filed https://jira.mongodb.org/browse/PERF-3066

yun-soo:
Thanks all for thorough reviews! Merging this PR.

PR #678, opened 2022-05-17
PERF-2998: Add ^ChooseFromDataset Generator to Genny

miguelangelnieto:
- Added a new generator named ChooseFromDataset.
- Added tests for the new generator.
- Modified the Generators.yml to show an example.
- Added RandomDataFromDataset.yml to docs/.
- Added four datasets: 

1. names.txt with 2000 names (from [SecLists](https://github.com/danielmiessler/SecLists)).
2. familynames.txt with 1000 family names (from [SecLists](https://github.com/danielmiessler/SecLists)).
3. airport_codes.txt with all airport codes.
4. empty_test.txt with no data for unit testing.


miguelangelnieto:
Looks like there is an issue I don't know how to solve it right now. We need to specify the relative path to the datasets, but this will work or not depending on your current working directory. That it could be `/home/ubuntu/dev/genny` on local dev environment, `/data/mci/26970c17bacfeeab3478d615953f022e` on evergreen or any other possibility.

So, tests will pass or fail depending on the cwd. For example, now all pass in local environment, but they fail in evergreen.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +739,67 @@ class StringGenerator : public Generator<std::string> {
...
+                }
+        }
+
+    std::string evaluate() override {   
+        auto distribution = boost::random::uniform_int_distribution<size_t>{0, dataSetContent.size() - 1};
+        return dataSetContent[distribution(_rng)];
+    }
+private:
+    std::ifstream ifs;
+    std::vector<std::string> dataSetContent;
```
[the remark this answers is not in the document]
dalyd: +1 on static map somehow. There are a couple of levels of this. 
At its simplest form, we should make sure that each instantiation within an actor shares state. That likely can be done with a static member (maybe?)
Then there is possible sharing between different actors or different parts of the same actor (e.g, multiple operations). I'm not sure the best way to do that.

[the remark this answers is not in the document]
miguelangelnieto: Right now, every time the generator is called and the object is created the file is read into memory. Once the generator has finished, the object is destroyed and memory released. This means that if a particular genny workload has two or more ^ChooseFromDataset, data in memory will be loaded once per generator. I don't know if this is going to be in parallel, since I don't really know if the document generation is multithreaded.

I will prototype something quick when I have time to use a static map, to avoid re-reading the file into memory multiple times.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +739,67 @@ class StringGenerator : public Generator<std::string> {
...
+          _id{generatorArgs.actorId},
+          _path{node["path"].maybe<std::string>().value()},
+          _pathLength{_path.size()} {
+        if (_pathLength <= 0) {
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+                "Data Set Generator requieres non-empty path"));
+        }
+    }
+
+protected:
```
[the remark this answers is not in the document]
miguelangelnieto: To be honest, I don't think so. I just tried to follow the convention used by all the other generators. I can get rid of it.

src/value_generators/src/DocumentGenerator.cpp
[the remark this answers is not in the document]
brad-devlugt: Yes the evaluate method could be called from multiple threads here. For example, if we have 32 threads set on the loader actor, there will be 32 copies of the actor working in parallel to insert documents into a collection, and each will be generating documents. This means that if we have static state to save memory, we'll have to handle thread-safety. One option is a static unordered_map and a static mutex in the generator class.

src/value_generators/src/DocumentGenerator.cpp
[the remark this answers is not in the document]
miguelangelnieto: This is done, parent class removed.

src/value_generators/src/DocumentGenerator.cpp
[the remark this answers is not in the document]
miguelangelnieto: Thank you for your comments, this is done in the last commit.

- Using a static unordered_map to store the file content, where the key is the file path.
- If the key exists, we don't open the file again.
- If the key doesn't exist, we enter a section protected by a mutex that opens the file and stores the data as a vector of strings for that particular key (path).

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -745,6 +747,73 @@ class StringGenerator : public Generator<std::string> {
     const size_t _alphabetLength;
 };
 
+/** `{^ChooseFromDataset:{...}` */
+class RandomDataFromDataset : public Generator<std::string> {
```
aggrand, line 751: Should this have `string` somewhere in its name to clarify that it works for string datasets, not numeric ones?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -745,6 +747,73 @@ class StringGenerator : public Generator<std::string> {
...
+            }
+
+            /* To avoid reading and storing the datasets multiple times
+               we check if the path already exists in the map (as key). If
+               it doesn't exist, we open the file and store its contents. 
+               Section locked with a mutex */
+            
+            _dataset_mutex.lock();
+            if (_dataset.count(_path) == 0) {
+                std::cout << "I enter to open " << _path << std::endl;
```
aggrand, line 774: Is this a debug statement or one to leave in? If it's one to leave in, please use `BOOST_LOG_TRIVIAL`, example [here](https://github.com/mongodb/genny/blob/fb48f0ea6056d920ab6b3a93c5470a977c03c288/src/value_generators/src/DocumentGenerator.cpp#L991).

aggrand: (nit) also consider a non-first-person statement like `Loading dataset at <PATH>` or so. Also consider making it a debug-level statement.

aggrand (changes requested):
Had a few requested changes, but this is a great start! Definitely push back where you disagree!

src/workloads/docs/Generators.yml
```diff
@@ -125,6 +126,9 @@ Actors:
             # Generate the actor ID as a string
             actorString: {^ActorIdString: {}}
 
+            # Generate a random airport code from a dataset
+            airportDataSet: {^ChooseFromDataset: {"path": "src/genny/src/workloads/datasets/airports_codes.txt"}}
```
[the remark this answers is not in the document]
dalyd: cc: @rtimmons who I think had opinions on paths and things. It's an important thing to get right.

src/workloads/datasets/names.txt
```diff
@@ -0,0 +1,2000 @@
+JAMES
```
dalyd, line 1: This and the previous are really cool. Three questions: 
1. Can we document these datasets somewhere? 
2. Should they be under workloads or in some reference directory? 
3. Why are all the names all caps?

src/workloads/docs/RandomDataFromDataset.yml
```diff
@@ -0,0 +1,21 @@
...
+  If running in evergreen, the relative path needs to be: src/genny/src/workloads/datasets/.
+
+Actors:
+- Name: Insert
+  Type: Insert
+  Threads: 1
+  Database: "test2"
+  Phases:
+  - Collection: "inserts"
+    Document:
```
dalyd, line 17: Could you increase the threads and add a use of the form 
`{^Join: {array: [{^ChooseFromDataset: {"path": "src/genny/src/workloads/datasets/names.txt"}}, " ", {^ChooseFromDataset: {"path": "src/genny/src/workloads/datasets/familynames.txt"}}]}}`
In addition to these examples, to show off the idea that these datasets are loaded only once into memory?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -745,6 +747,73 @@ class StringGenerator : public Generator<std::string> {
...
+    
+private:
+    DefaultRandom& _rng;
+    ActorId _id;
+    std::string _path;
+    const size_t _pathLength;
+    std::ifstream _ifs;
+    std::vector<std::string> _dataSetContent;
+    std::string _current_line;
+    static inline std::unordered_map<std::string, std::vector<std::string>> _dataset;
```
dalyd, line 813: Are you using the full path as the key here? That probably makes sense. My first reaction though was to put in an explicit key. That said, I can't come up with a valid reason why there would ever be a non 1-1 relationship from keys to paths.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -745,6 +747,73 @@ class StringGenerator : public Generator<std::string> {
...
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+            "RandomDataFromDataset requieres non-empty path"));
+            }
+
+            /* To avoid reading and storing the datasets multiple times
+               we check if the path already exists in the map (as key). If
+               it doesn't exist, we open the file and store its contents. 
+               Section locked with a mutex */
+            
+            _dataset_mutex.lock();
```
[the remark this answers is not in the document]
dalyd: +1 on std::lock_guard. 

Do we need to lock the the mutex before checking the _dataset.count? It seems like we could do an early exit if (_dataset.count(_path) > 0). Obviously if false, we need to recheck after getting the lock. This is likely a bad idea unless we end up stalling a lot of threads here.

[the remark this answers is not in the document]
dalyd: FWIW, unordered_map::count is a const function and therefore threadsafe for the container. So is the boolean function contains, which might be a better fit for the condition below. 

We should be able to safely insert into two different entries in the unordered map at the same time from different threads, so the locking could be done by path name. 
Again, I'm likely over thinking this.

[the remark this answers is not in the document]
aggrand: I think checking the count makes sense; it'd be best to not queue everyone up waiting on the lock to check if the data is present if it actually is present. IIRC from the slow startup times ticket, evaluating generators all this was a nontrivial part of initialization and serializing it too much would make it far worse.

Though I imagine if we have 100 threads in an actor configured similarly, they'll all end up queuing up here at about the same time?

src/workloads/docs/Generators.yml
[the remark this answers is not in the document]
rtimmons: Ack. "Best" solution (imho without poking too deeply here yet) is to use cwd-relative paths, and remember that we check out all repos in `./src`. So `./src/genny/src/workloads/....`.

I'd like to take a look once the above comments are addressed. Added myself to the reviewers list. Thanks for the ping.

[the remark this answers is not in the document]
miguelangelnieto: Changed Generators.yml to use the cwd-relative paths, that is ./src/genny/src/workloads/....

src/value_generators/src/DocumentGenerator.cpp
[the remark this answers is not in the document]
miguelangelnieto: Yep, all datasets. Renamed.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Renamed to RandomStringFromDataset.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: I forgot to remove that line, thank you :)

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Yes, the key is the path, since there won't be two different files with the same path. I think this method is simple enough and it works for what we need, since we already have the path and it is unique.

src/workloads/datasets/names.txt
miguelangelnieto: 1. I have added the documentation to `using.md`.
2. I think that since this is part of a generator, workloads/datasets makes more sense. But let me know if you have any other idea.
3. Because this is how they are distributed. The datasets for names and familiynames are from https://github.com/danielmiessler/SecLists. In any case, I modified both datasets so instead of JAMES it will be James.

src/workloads/docs/RandomDataFromDataset.yml
miguelangelnieto: I have added the join with 8 threads.

src/value_generators/src/DocumentGenerator.cpp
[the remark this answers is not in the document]
miguelangelnieto: - Modified the code so it uses std::lock_guard instead of locking/unlocking.
- Moved the load to its own method.
- Now, during evaluate() call there is a check to see if the path is present in memory. If so we just continue, if not, we call loadDataset method protected by a mutex.

It is true that when we start the load the 100 threads will queue about the same time while one is loading the data, but with the current code it will only happen once per dataset loaded. Then, once it is in memory, the mutex and the loadDataset() is not called anymore. I think this is the best approach here.

docs/using.md
```diff
@@ -1,39 +1,40 @@
...
-    5.  [Connecting to the Server](#orgd6b0450)
-        1.  [Connection Strings and Pools](#orgd2659db)
-        2.  [Multiple Connection Strings](#orga591018)
-        3.  [Default](#org65830c2)
-    6.  [Creating an Actor](#org7e6c6bd)
-5.  [Pitfalls](#org3aaae9e)
-    1.  [pipe creation failed (24): Too many open files](#orga7ab911)
-    2.  [Actor integration tests fail locally](#orgb084b49)
-    3.  [The Loader agent requires thread count set on both Actor and phase level](#org97681a9)
+- [Table of Contents](#table-of-contents)
```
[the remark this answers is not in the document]
dalyd: Are there any changes here beyond adding the introduction?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
 
+/** `{^ChooseFromDataset:{...}` */
+class RandomStringFromDataset : public Generator<std::string> {
+
+public:
+    RandomStringFromDataset(const Node& node, GeneratorArgs generatorArgs)
+        : _rng{generatorArgs.rng},
+          _id{generatorArgs.actorId},
+          _path{node["path"].maybe<std::string>().value()},
+          _pathLength{_path.size()}{}
```
dalyd, line 752: I don't think you need this member. Why not just check _path.size() <= 0 on line 756 and remove the member?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+
+        {
+            if (_all_datasets.count(_path) == 0){
+                _ifs.open(_path, std::ifstream::in);
+                if (_ifs.is_open()){
+                    while(std::getline(_ifs, _current_line))
+                    {
+                        // Ignore emtpy lines
+                        if (_current_line.size()!=0){
+                            _all_datasets[_path].push_back(_current_line);
```
dalyd, line 769: You could read this in and assemble everything locally, and then put it into the dataset. That feels cleaner to me, but maybe it doesn't make a difference.

docs/using.md
```diff
@@ -569,6 +570,13 @@ It is often necessary to use Genny to operate with large amounts of data which w
 
 Value generators are not a builtin feature of Genny, but must be integrated by each Actor for the configuration values that accept them. For examples of using value generators, see [./src/workloads/docs/Generators.yml](../src/workloads/docs/Generators.yml). To integrate generators into an Actor, use the [DocumentGenerator](../src/value_generators/include/value_generators/DocumentGenerator.hpp) with the yaml node you intend to generate documents from. (And see [here](#org7e6c6bd) for more details on creating an Actor in the first place.)
 
+Some Value Generators, like `ChooseFromDataset`, can read a dataset from the disk that store the values to choose from. Some example datasets are stored in [./src/workloads/datasets](../src/workloads/datasets). Four are included in the repo:
+
+- **airport_codes.txt:** includes all airport codes
```
dalyd, line 575: Could you add a README.md file to the datasets directory. Maybe move this content there, and replace with a link.

src/workloads/docs/Generators.yml
```diff
@@ -125,6 +126,9 @@ Actors:
             # Generate the actor ID as a string
             actorString: {^ActorIdString: {}}
 
+            # Generate a random airport code from a dataset
+            airportDataSet: {^ChooseFromDataset: {"path": "./src/genny/src/workloads/datasets/airports_codes.txt"}}
```
dalyd, line 130: Could you include a comment explaining how to properly construct the path?

dalyd: Maybe something short here and reference to the description in RandomStringFromDataset.yml, or just comment "See RandomStringFromDataset.yml for full documentation on this generator".

dalyd (changes requested):

docs/using.md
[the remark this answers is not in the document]
miguelangelnieto: I didn't change all of that. It has to be vscode. I will try to revert that.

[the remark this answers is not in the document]
miguelangelnieto: Done, reverted.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+            "ChooseFromDataset requieres non-empty path"));
+            }
+
+        {
+            if (_all_datasets.count(_path) == 0){
+                _ifs.open(_path, std::ifstream::in);
+                if (_ifs.is_open()){
+                    while(std::getline(_ifs, _current_line))
+                    {
```
rtimmons, line 766: Please ensure you run your changed files through `clang-format`. Annoyingly this isn't part of the toolchain and we have no CI checking for it.

```sh
brew install clang-format
clang-format -i ./src/**/DocumentGenerator.*
```

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+                        // Ignore emtpy lines
+                        if (_current_line.size()!=0){
```
rtimmons, line 768: `if (!currentLine.empty())`

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+                            _all_datasets[_path].push_back(_current_line);
+                        }
+                    }
+                }
+            
+                else {
+                    BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+                        "The specified file for ChooseFromDataset cannot be opened or it does not exist")); }
+
+                if (_all_datasets[_path].size() == 0) {
```
rtimmons, line 778: 1. Another place to use `.empty()`?
2. In both of these error messages, please include the indicated file/path and cwd

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+                    BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+                        "The specified file for ChooseFromDataset is empty")); }
+            }
+        }
+      }
+    }
+
+    std::string evaluate() {
```
rtimmons, line 786: This means the first invocation of the generator will (presumably) be orders of magnitude slower than subsequent invocations. Can you prime this at ctor-time?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+        /* If path the doesn't exist in memory we load it.
+           To avoid reading and storing the datasets multiple times
+           we check if the path already exists in the map (as key). If
+           it doesn't exist, we open the file and store its contents. 
+        */
+
+        if (_all_datasets.count(_path) == 0){
```
rtimmons, line 793: If two threads fire at the same time they'll both see count=0. Then one will lock, load the dataset, and unlock. Then the other thread will lock, load the dataset (for a second time) and then unlock. I think you want the lock around the size-check:

```c++
{
    std::lock_guard lk{mutex};
    if(sets.count(key) <= 0) {
      load()
    }
}
```

src/value_generators/src/DocumentGenerator.cpp
rtimmons: Ya that's what I would do as well, but there's definitely a matter of personal preference there.

(If it speaks to you, you might also consider a `DataSetHolder` helper class that encapsulates the logic of "load dataset unless loaded and do so thread-safe-ly" and then have a `static DataSetHolder _holder` in the value-generator. This is bike-shedding, however, so please don't feel compelled to do so if you don't want to. Biggest thing imho is removing as much member-variable state as possible and ensuring the locks/mutexes have correct bookkeeping.)

docs/using.md
miguelangelnieto: Added the README.md and a link to in the usage.md

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+          _id{generatorArgs.actorId},
+          _path{node["path"].maybe<std::string>().value()},
+          _pathLength{_path.size()}{}
+
+    void loadDataset() {
+      {
+        if (_pathLength <= 0) {
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+            "ChooseFromDataset requieres non-empty path"));
+            }
```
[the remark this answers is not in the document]
miguelangelnieto: Thank you, good advice. Done, now the error if the file does not exist looks like this:

The specified file for ChooseFromDataset cannot be opened or it does not exist. Specified path: ./src/workloads/datasets/NAMES.txt. Current Working Directory: /home/ubuntu/dev/genny

src/workloads/docs/Generators.yml
miguelangelnieto: Done, I added the comment to people check the already documented yml.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: I get not output, so I guess that means it is ok?

```
dev/genny [PERF-2998●] » clang-format -i ./src/**/DocumentGenerator.* 
dev/genny [PERF-2998●] » echo $? 
0
```

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+    std::string evaluate() {
+        /* If path the doesn't exist in memory we load it.
+           To avoid reading and storing the datasets multiple times
+           we check if the path already exists in the map (as key). If
+           it doesn't exist, we open the file and store its contents. 
+        */
+
+        if (_all_datasets.count(_path) == 0){
+            std::lock_guard<std::mutex> lk(_dataset_mutex);
+            loadDataset();
```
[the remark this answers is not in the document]
miguelangelnieto: I have moved the loadDataset call to the constructor and placed the lock inside. Thank you.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Thank you. I changed the error messages to include paths and used .empty().

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: I moved the lock just before the check of count, as you explained. Makes more sense, thank you.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +741,76 @@ class StringGenerator : public Generator<std::string> {
...
+        return _all_datasets[_path][distribution(_rng)];
+    }
+    
+private:
+    DefaultRandom& _rng;
+    ActorId _id;
+    std::string _path;
+    const size_t _pathLength;
+    std::ifstream _ifs;
+    std::vector<std::string> _dataSetContent;
```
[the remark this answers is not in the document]
miguelangelnieto: it is not needed, it was part of a previous implementation and I forgot to remove it. thank you!

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: changed, thank you!

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Based on a comment from David I moved the loadDataset to the constructor instead of being part validate().

dalyd (dismissed):
LGTM. Thank you.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -748,67 +747,73 @@ class RandomStringFromDataset : public Generator<std::string> {
...
+               To avoid reading and storing the datasets multiple times
+               we check if the path already exists in the map (as key). If
+               it doesn't exist, we open the file and store its contents.
+            */
+
+            {
+                std::mutex _dataset_mutex;
+                std::lock_guard<std::mutex> lk(_dataset_mutex);
+
+                if (_all_datasets.count(_path) == 0) {
```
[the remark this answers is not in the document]
miguelangelnieto: 1. I will add it. I had it before, but lost if after moving the code around.
2. I don't think this will work. If I add an empty vector, some of the threads will continue to evaluate() as if the data was already there and the data could still be not loaded, or only partially, by the first thread. The results could be difficult to predict. And we won't have repeatable outcomes on different runs.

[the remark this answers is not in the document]
dalyd: There won't be any calls to evaluate until after all the actors and phases and everything are constructed. There's a hard barrier in there. So, no risk of a call to evaluate with an empty vector.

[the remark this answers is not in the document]
miguelangelnieto: Oh, didn't know that. Then, I will implement that too. Thanks.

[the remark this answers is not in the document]
dalyd: My understanding (could be wrong) ...

The first time you call `_all_datasets[_path].push_back(currentLine);`  for a given `_path` a new `std::vector<std::string>` is implicitly created an emplaced in the vector `_all_datasets` at key `_path`. I'm not suggesting calling `_all_datasets[_path].push_back(std::vector<std::string>());` but rather doing something like `_all_datasets[_path] = std::vector<std::string>()`. The end result should be _all_datasets[_path] temporarily having an empty vector, and no difference once you start emplacing things. Does that make sense? Hopefully it's even correct. 

Honestly, the implicit creation surprised me a bit, as I probably would have made a local `auto local_vector = std::vector<std::string>`, loaded in all the values, and then done `_all_datasets[_path] = std::move(local_vector)` or some such.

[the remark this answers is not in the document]
miguelangelnieto: Mmm well, I think that I can just remove the first element of the vector after the load has finished, and then there won't be problems.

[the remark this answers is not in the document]
miguelangelnieto: Ok, so, what I am going to do is add an empty vector. Then in the last line of the method, once all the data is loaded, I remove the empty vector that is at the first position.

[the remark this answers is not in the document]
dalyd: Looks beautiful. Thanks.

dalyd (approved):

miguelangelnieto:
@aggrand @rtimmons if you have some time, please review the latest changes. I will also probably need someone to merge, since the compilation on macos fails and I guess it won't allow me to merge it. thank you!

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +740,93 @@ class StringGenerator : public Generator<std::string> {
...
+               If the number of items is more than 0, it means the data is
+               loaded, so we can skip everything, including the mutex.
+            */
+
+            if (_all_datasets.count(_path) > 0) {
+                return;
+            }
+
+            {
+                std::mutex _dataset_mutex;
```
aggrand, line 775: I might be missing something, but is this lock being declared locally on the stack? So every thread that construct a `RandomStringFromDataset` will be creating a separate mutex, locking its own lock, and getting no cross-thread coordination? I think the mutex should be a shared static var

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +740,93 @@ class StringGenerator : public Generator<std::string> {
...
+                std::lock_guard<std::mutex> lk(_dataset_mutex);
+
+                if (_all_datasets.count(_path) == 0) {
+                    std::string currentLine;
+                    std::ifstream ifs;
+
+                    /* Add a first empty vector as soon as possible
```
aggrand, line 782: This only works if we have multiple locks, right? There's the `_all_datasets` lock that should be locked when adding a new dataset, and a dataset-specific lock that should be used when populating a specific dataset. If we hold the `_all_datasets` lock while populating a specific dataset, then all threads are blocked on every dataset load, even if they don't care about that dataset.

It seems like we should have the `_all_datasets` lock that is used when adding a new dataset vector, and a dataset-specific lock used to populate that dataset. I imagine we'd take the `_all_datasets` lock, add the vector and dataset-specific lock as a pair, release the lock, take the dataset-specific lock, and populate the vector, then release it.

Also, we might want to consider a reader-writer lock system like the orchestrator uses [here](https://github.com/mongodb/genny/blob/24e68966b9846d0ba0615c8bc4503008e04b58c1/src/gennylib/src/Orchestrator.cpp#L37).

aggrand (changes requested):
A few more safety things, but locking / thread-coordinating can get messily complicated so please lmk if I missed something!

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Oh, yes, that is a mistake when I was restructuring and moving variables. Will revert it to shared static.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Hello @aggrand. That makes sense, but how can I create locks at runtime? I don't know in advance how many and which datasets are going to be loaded, so I cannot define in the code the lock names. Maybe I am misunderstanding something?

brad-devlugt: I think we should only ever have one thread populating the vector. My thought in pseudo-code is we should have something like:
```
results = []

lock(_map)
if(path in _map)
  // someone else is already loading, nothing to do
  return
else
 // this thread is going to load the dataset
 _map.add(path, results)
unlock(_map)

for(item in file) {
  results.add(item)
}
```

If another two threads are loading the same dataset, the first one to enter here would hit the map.add path, and the second would immediately bail out. If two threads are loading different datasets, both get a unique results list that they are populating after unlocking the map. Let me know if I'm missing something here; I think if this is the case we only need a single mutex.

aggrand: Correct me if I'm wrong, but with the above pseudocode, a thread might take the lock, populate the map with the empty vector, drop the lock, then start populating the vector. Another thread might come along, take the lock, see that the map contains the vector, then assume it's populated with no way of knowing that the population is in-progress, then try to use the vector while the population is incomplete.

aggrand: Hence the two levels of locking, for the multi-dataset map and for the dataset.

dalyd: @aggrand I don't think we ever have reading and writing at the same time of the actual datasets. We should write them during initialization, and read them during workload execution. 

@brad-devlugt I think your code is technically correct, but a bit scary :) If someone bungled the early return such that it just exited that scope rather than the function, we would have multiple threads installing the dataset at the same time. I think it makes sense to have a lock guard around updating the items in the dataset. Does that make sense? I might go a step further and say the dataset vector should be built locally, and then moved into the _all_datasets maps. That way, if we have a lock on it, but we still somehow try to add it twice (bug), we end up with the right value.

aggrand: i.e. we have two resources: a global registry of datasets and the contents of a specific dataset. We can ensure safety at the cost of efficiency by guarding them both with the same lock, but if we want both resources to be interact-able independently it seems like we need a lock for each.

dalyd: @aggrand yes.

brad-devlugt: Yeah @aggrand that's a concern if the evaluate methods can be called in other threads at the same time as the constructor. Does genny construct the generators lazily or up front? Is the constructing stage multi or single threaded?

miguelangelnieto: Then, can it be something like this? This way the threads will be waiting on LockOne (checking is the path's vector is there) for a short time, just until the first thread adds the empty vector. This way, we won't have more than one thread loading the data, and the time waiting on the lockone will be short.

```
   {      
   GetLockOne()
   if (_all_datasets.count(_path) > 0) {
                return;
   }

            {
                GetLockTwo()
                if (_all_datasets.count(_path) == 0) {
                    std::string currentLine;
                    std::ifstream ifs;

                    /* Add a first empty vector as soon as possible
                       so new threads that are being created do not
                       have to wait to acquire the mutex */

                    _all_datasets[_path] = std::vector<std::string>();
		    UnlockLockOne()
...
```

aggrand: @brad-devlugt I believe the generators are constructed upfront because they are part of the actor classes, e.g. [here](https://github.com/mongodb/genny/blob/24e68966b9846d0ba0615c8bc4503008e04b58c1/src/cast_core/src/Loader.cpp#L59), and as of [this](https://github.com/mongodb/genny/pull/562
) commit constructing actors is indeed [parallelized](https://github.com/mongodb/genny/blob/24e68966b9846d0ba0615c8bc4503008e04b58c1/src/gennylib/src/context.cpp#L83).

@miguelangelnieto I think your code there has the lock two attained inside the scope of lock one, based on the indent / still open paranetheses. I think we should really treat adding to the registry and populating the vector as basically two independent operations, which are sequential

brad-devlugt: thanks - so it sounds like, if we construct them up-front, we only need to lock the map. This is since no thread would try to use the vector in evaluate() until all the generators are constructed. And if we are constructing actors in parallel, that has the nice benefit of letting us load different datasets in parallel.

aggrand: We construct them up front, but as I see it if lock only the map during the whole vector population and don't get that parallelism benefit. And if we want to prevent multiple writers to the same vector we need to lock it, right?

Or wait actually we could set a flag when we are the thread to create the vector to also say we are the thread to populate it, to associate the two events. So any thread that doesn't create the mapping just returns?

miguelangelnieto: Thank you. So, based on your conversation with Brad, and based in his pseudocode, I understand that this is what we want. Lock the map, and then only the first thread will populate the vector. Let me know if my understanding is not correct.

```
std::vector<std::string> data;

{
    std::lock_guard<std::mutex> lk(_dataset_mutex);    
    if (_all_datasets.count(_path) > 0) {
        return;
    }

    else {
    _all_datasets[_path] = data;
    }
}

std::string currentLine;
std::ifstream ifs;

ifs.open(_path, std::ifstream::in);
if (ifs.is_open()) {
    while (std::getline(ifs, currentLine)) {
        if (!currentLine.empty()) {
            _all_datasets[_path].push_back(currentLine);
        }
    }
}
```

aggrand: yup I think the above sounds good! Maybe a comment on the return to make it clearer that the return MUST happen

dalyd: That quote hit a formatting problem. The exception is std::vector\<bool\>. I'll see if I can fix it.

brad-devlugt: Yeah I agree, the elements can be modified concurrently, this is why we can push to the vector via the pointer. I think the lookup via `operator []` isn't thread-safe for associative containers, however. Also from https://en.cppreference.com/w/cpp/container#Thread_safety:
>  and, except in associative containers, operator[], behave as const for the purposes of thread safety

I think with unordered_map, for example, an insert could trigger a re-hashing and then one thread could be in the middle of that when the read comes in.

dalyd: Ah, okay. And better safe than sorry here. Thanks Brad.

brad-devlugt: No problem! I had to debug a really tricky problem once before that ended up being related to the thread-safety of `operator []`.


@miguelangelnieto I don't think it is worth 1-2 weeks of time, so my vote is to either time-box it at an hour or move forward with your proposed fix and keep the pattern in mind as an option for next time. The time-savings with the pattern comes from catching errors at compile time vs runtime/code review, and in time savings when working in the area of code next time around.

miguelangelnieto:
@brad-devlugt @rtimmons and @aggrand if you have some time, check the changes. It should address all the issues reported.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -739,6 +740,108 @@ class StringGenerator : public Generator<std::string> {
...
 };
 
+
+class DataSetCache {
+public:
+    DataSetCache() {}
+
+    /* The dataset is stored in a map, using the path of the file as the key and a vector of strings
+    with file's content as the value. In this method the empty vector is created and return its
+    pointer */
```
brad-devlugt, line 750: For the docs, one suggestion is a note on this class that says the getDataSetForPath method doesn't need the lock because loading and evaluating are separate stages in genny.

brad-devlugt (approved):
LGTM, just one small thought about documentation in the new DataSetCache class

aggrand (commented):
+1 lgtm as well. Thanks for hammering out the threading issues!

brad-devlugt (approved):

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: This is done.

dalyd (approved):
Just hitting the button to get out of the way. Nice work @miguelangelnieto @aggrand since you said LGTM on the ticket could you hit approve also. 

@rtimmons are you actively in the loop at this point checking things, or should we let Miguel go ahead and merge this one?

aggrand (approved):
Oof sorry hit the wrong review button!

PR #679, opened 2022-05-18
PERF-3040 Create a genny workload for measuring replication overhead

mtrussotto:
This is a workload that just does a simple multi-update on a large number of documents.  This ends up measuring the impact of secondary replication on the primary.

@dalyd 

@mridul-augustine

### DIFF
--- a/src/workloads/scale/ManyUpdate.yml
+++ b/src/workloads/scale/ManyUpdate.yml
@@ -1,0 +1,71 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-replication"
+Description: |
+  This workload loads a large number of small documents into a single collection, and then does a
+  small multi-update which touches all of them.  It is intended to measure the impact of the load of
+  replication of many oplog entries on the primary.  Number of documents should be significantly
+  greater than the maximum replication batch size of 5K (50K is a good minimum).
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 1000
+
+Actors:
+- Name: Setup
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &DB test
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand: {dropDatabase: 1}
+  - &Nop {Nop: true}
+  - *Nop
+
+- Name: Loader
+  Type: Loader
+  Threads: 10
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: *DB
+    CollectionCount: 1
+    Threads: 10
+    DocumentCount: 5e4
+    BatchSize: 1000
+    Document:  # Documents are approximately 100B in size
+      i: 0
+      string0: {^FastRandomString: {length: 100}}
+  - *Nop
+
+- Name: DocumentUpdater
+  Type: CrudActor
+  Threads: 1
+  Database: *DB
+  Phases:
+  - *Nop
+  - *Nop
+  - Duration: 1 minutes
+    MetricsName: Update
+    Collection: Collection0
+    Operations:
+    - OperationName: updateMany
+      OperationCommand:
+        Filter: {}
+        Update: {$inc: {i: 1}}
+        Options:
+          WriteConcern:
+            Level: 1
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - atlas
+      - replica
+      - replica-noflowcontrol
+      - replica-1dayhistory-15gbwtcache
+      - replica-all-feature-flags
+      - single-replica
--- a/src/workloads/scale/ManyUpdate.yml
+++ b/src/workloads/scale/ManyUpdate.yml
@@ -15,6 +15,8 @@ Keywords:
 - CrudActor
 - updateMany
 - update
+- replication
+- oplogSourceOverhead
 
 Clients:
   Default:
@@ -42,7 +44,6 @@ Actors:
   - Repeat: 1
     Database: *DB
     CollectionCount: 1
-    Threads: 10
     MultipleThreadsPerCollection: true
     DocumentCount: 5e4
     BatchSize: 1000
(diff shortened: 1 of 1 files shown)


src/workloads/scale/ManyUpdate.yml
```diff
@@ -0,0 +1,72 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-replication"
+Description: |
+  This workload loads a large number of small documents into a single collection, and then does a
+  small multi-update which touches all of them.  It is intended to measure the impact of the load of
+  replication of many oplog entries on the primary.  Number of documents should be significantly
+  greater than the maximum replication batch size of 5K (50K is a good minimum).
```
dalyd, line 7: Is there a chance flow control will kick in on this workload due to the w:1 below? Then we are measuring flow control behavior rather than replication.

src/workloads/scale/ManyUpdate.yml
```diff
@@ -0,0 +1,72 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-replication"
+Description: |
```
dalyd, line 3: Could you add some "Keywords" also? https://github.com/mongodb/genny/blob/2ae299767d0c870a57d1e80301616ed0309e7cac/docs/using.md#L173

We don't enforce anything on it yet, but we hope to use the keywords for workload discovery. A search of the code base will show other keywords people have used. Try to not add duplicate variation of keywords, but if you do we will fix them when we start linting the keywords.

src/workloads/scale/ManyUpdate.yml
```diff
@@ -0,0 +1,72 @@
...
+  - *Nop
+
+- Name: Loader
+  Type: Loader
+  Threads: 10
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: *DB
+    CollectionCount: 1
```
dalyd, line 34: With CollectionCount of 1, only one thread will do anything. It's best to either drop the Threads to 1 (in both places) or set `multipleThreadsPerCollection` to true.

mridul-augustine:
I see the multipleThreadsPerCollection issue mentioned by @dalyd to be the only concern. @aggrand has more experience in genny. Adding him as a reviewer.

src/workloads/scale/ManyUpdate.yml
mtrussotto: Added some keywords

src/workloads/scale/ManyUpdate.yml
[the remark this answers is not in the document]
mtrussotto: Done

src/workloads/scale/ManyUpdate.yml
mtrussotto: Done

mtrussotto:
New patch build:
https://evergreen.mongodb.com/version/628665331e2d173b302c1e71

Two older ones:

https://evergreen.mongodb.com/version/6283e7bc3627e075d1bd65d2

https://evergreen.mongodb.com/version/6282b8312fbabe7554f130be

src/workloads/scale/ManyUpdate.yml
dalyd: Thanks! Could you add one or two more for the idea that this is meant to test oplog processing overhead? I'm bad with names, but could imagine oplogOverhead primaryOverhead or something along those lines. I'm sure we do not have any such keywords now, so that's all up to you.

src/workloads/scale/ManyUpdate.yml
dalyd: Sounds great. thanks!

src/workloads/scale/ManyUpdate.yml
dalyd: I think you may need to remove the phase level threads statement with MultipleThreadsPerCollection: true.

src/workloads/scale/ManyUpdate.yml
```diff
@@ -0,0 +1,83 @@
...
+  - *Nop
+
+- Name: DocumentUpdater
+  Type: CrudActor
+  Threads: 1
+  Database: *DB
+  Phases:
+  - *Nop
+  - *Nop
+  - Duration: 1 minutes
```
dalyd, line 61: It looks like the most recent patch build is still running, but an older one shows you got 22 operations completed, which is barely enough to get a 95th percentile. Do you want to run this longer?

dalyd (approved):
This looks good. There are a few small things, but they are small and you can deal with them without further review from me. Thanks Matthew. And if I haven't given you a sticker before (or need another one), you deserve a performance leafie sticker.

aggrand (approved):
Can't comment much on the workload contents/value but LGTM w.r.t. genny usage!

src/workloads/scale/ManyUpdate.yml
```diff
@@ -0,0 +1,83 @@
...
+
+- Name: Loader
+  Type: Loader
+  Threads: 10
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: *DB
+    CollectionCount: 1
+    Threads: 10
```
aggrand, line 45: Ah I think this phase-level spec is giving you issues here

aggrand: (Oh saw that was mentioned above)

mridul-augustine (approved):

src/workloads/scale/ManyUpdate.yml
mtrussotto: Ah, good point, I had it short mostly to reduce the development cycle.  5 minutes would give me a 99th percentile (barely) and a decent 95th percentile, do you think that makes sense?

dalyd: That sounds reasonable to me. There's a fair amount of overhead in setting up the system, so a 5 minute test isn't bad. 10 would likely be fine also.

mtrussotto:
evergreen merge

mtrussotto:
evergreen merge

mtrussotto:
evergreen merge

mtrussotto:
evergreen merge

PR #681, opened 2022-05-20
TIG-4166: Update code owners for workloads.

ancostas:


.github/CODEOWNERS
```diff
@@ -1,10 +1,13 @@
 # https://help.github.com/articles/about-code-owners/
 
 # global match unless overridden below
-* @10gen/dev-prod-tips
+* @mongodb/dev-prod-tips
+
+# everyone in kernel contributors can work with workloads, unless overridden below
+src/workloads/ @mongodb/kernel
```
aggrand, line 7: We should do something similar for `src/cast_core` to avoid blocking other peoples' actors.

aggrand (changes requested):

aggrand (approved):
LGTM with the cast-core changes!

PR #687, opened 2022-05-26
PERF-3062: Change test data in DensifyTimeSeriesCollection to have more than one document per bucket

mattBoros:
This change will create a pretty diverse set of buckets. We'll have buckets that close, buckets with a small amount of documents due to the previous bucket containing the meta/time range being full. Then we'll have buckets that have close to 1000 documents and are still open.

[patch](https://spruce.mongodb.com/version/62964f852a60ed2b207ddc7a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/execution/DensifyTimeseriesCollection.yml
+++ b/src/workloads/execution/DensifyTimeseriesCollection.yml
@@ -83,7 +83,7 @@ Actors:
     Collection: *coll
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 10000
+    DocumentCount: 100000
     BatchSize: *batchSize
     Document:
       partitionKey: {^RandomInt: {min: 1, max: 100}}
--- a/src/workloads/execution/DensifyTimeseriesCollection.yml
+++ b/src/workloads/execution/DensifyTimeseriesCollection.yml
@@ -87,7 +87,7 @@ Actors:
     BatchSize: *batchSize
     Document:
       partitionKey: {^RandomInt: {min: 1, max: 100}}
-      timestamp: {^RandomDate: {min: "2021-01-01T00:00:00.000", max: "2021-01-01T00:01:00.000"}}
+      timestamp: {^IncDate: {start: "2021-01-01T00:00:00.000", step: 400}}
       # Test a subset of units that have different characteristics.
       numeric: {^RandomInt: {min: 0, max: 100}}  # Separate codepath for numeric DensifyValues.
     # $densify stages always add a $sort on the field that is being densified, so this phase adds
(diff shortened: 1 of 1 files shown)


src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -83,13 +83,13 @@ Actors:
     Collection: *coll
     Threads: 1
     CollectionCount: 1
-    DocumentCount: 10000
+    DocumentCount: 100000
     BatchSize: *batchSize
     Document:
       partitionKey: {^RandomInt: {min: 1, max: 100}}
-      timestamp: {^RandomDate: {min: "2021-01-01", max: "2023-01-01"}}
+      timestamp: {^RandomDate: {min: "2021-01-01T00:00:00.000", max: "2021-01-01T00:01:00.000"}}
```
samontea, line 90: You may want to run a perf test just to double check this recovers the lost performance & produces reasonable bucketing.  Going backwards in time relative to the first observed time still causes a bucket closure.

mattBoros: Sounds good I think I'll change it to ^IncDate

mattBoros: Also checked the bucketing for a few different runs, it seems reasonable and definitely much better than before

src/workloads/execution/DensifyTimeseriesCollection.yml
```diff
@@ -87,7 +87,7 @@ Actors:
     BatchSize: *batchSize
     Document:
       partitionKey: {^RandomInt: {min: 1, max: 100}}
-      timestamp: {^RandomDate: {min: "2021-01-01T00:00:00.000", max: "2021-01-01T00:01:00.000"}}
+      timestamp: {^IncDate: {start: "2021-01-01T00:00:00.000", step: 400}}
       # Test a subset of units that have different characteristics.
       numeric: {^RandomInt: {min: 0, max: 100}}  # Separate codepath for numeric DensifyValues.
```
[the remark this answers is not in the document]
mattBoros: I checked the spread inside a workload (intra-workload spread i guess), and also the variance between workloads (inter-workload spread? idk). I ran four workloads.

The spread inside of a workload i found the standard deviation / mean of the workload ranged from 1.3% to 2.8%. The spread between workloads for stdev / mean was 1.5% which seems pretty good.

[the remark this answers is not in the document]
samontea: cool sounds good to me.

samontea (approved):
LGTM

PR #689, opened 2022-05-30
Writing-11168 Add workload for $dateTrunc expression in $group id calculation for time-series collection

lriuui0x0:


src/workloads/execution/TimeSeriesGroupDateTrunc.yml
```diff
@@ -0,0 +1,113 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  Timeseries performance test
```
jimoleary, line 4: If you are going to merge, can we get a more complete description, including what the workload is for, what it does, what the phases are doing.

Also some keyword search terms  under 'Keywords' attribute.

src/workloads/execution/TimeSeriesGroupDateTrunc.yml
```diff
@@ -0,0 +1,113 @@
...
+
+Actors:
+- Name: Setup
+  Type: RunCommand
+  Thread: 1
+  Phases:
+  - Repeat: 1
+    Database: *db
+    Operations:
+      - OperationName: RunCommand
```
jimoleary, line 28: Can you add a comment that the drop command is only here as a convenience to support running the test locally.  A sys-perf run would have an empty database.

jimoleary (changes requested):
some minor documentation issues

jimoleary (dismissed):
LGTM.

I created https://jira.mongodb.org/browse/TIG-4182 for the long branch name config, but I expect it will be a while before it is implemented,

jimoleary (commented):
actually there are some linter errors.

lriuui0x0:
evergreen merge

PR #673, opened 2022-05-03
Compiler 💅

rtimmons:
Gets rid of some compiler and clang-tidy errors.

PR #690, opened 2022-06-02
Use binary numpy on darwin

rtimmons:
Successor to #673 but much smaller (for now).
Fixes compile on macOS Big Sur Intel. Untested on Monterey or M1 or anything else as of yet.


vrachev (commented):
Looks like it's still failing

aggrand:
Heads up, to compile successfully on M1 I had to bump the python version to 3.9, and bump the numpy version to 1.22.1. That build works without the `--only-binary=numpy` flag, I don't suppose leaving that off and updating those works here too? Fine if not just checking

aggrand:
@vrachev my earlier comment can be ignored for the moment. I'd commented that w.r.t. the numpy issues specifically. The mongodbtoolchain isn't supported on macos arm, so I think dedicating more time to fixing genny build on it is in order.

rtimmons:
evergreen merge

rtimmons:
evergreen merge

PR #693, opened 2022-06-10
Revert "Writing-11168 Add workload for $dateTrunc expression in $grou…

lriuui0x0:
…p id calculation for time-series collection (#689)"

This reverts commit 9bea242e1c4a13b62a519546295420f5336899f5.


jimoleary (commented):
Would it make more sense to commit a PR with the file being deleted? A revert makes it look like this causes an issue.

jimoleary (approved):
LGTM: Discussed this directly, revert is there to make this an explicit undo.

PR #695, opened 2022-06-13
SERVER-66900 Add a classic-engine variant to sys-perf

leozzx:


dalyd (dismissed):
LGTM -- discussion on the other review of whether we should use single-replica or not. Let's figure that out and update here if appropriate.

src/workloads/execution/LookupSBEPushdown.yml
```diff
@@ -436,6 +436,9 @@ AutoRun:
       $eq:
       - standalone
       - standalone-all-feature-flags
+      - standalone-classic-query-engine
```
dstorch, line 439: These aren't the only `AutoRun` configurations that would have to change, are they? I'm curious as to how these two files were chosen, since grepping for `AutoRun` in the genny repository I see 142 hits. I can't tell at a glance which other genny workloads might be relevant to the tasks we intend to run on the classic engine build variant...

leozzx: These two files contains "SBE" in file names, so I guess these are the only queries that could be run on SBE engine. 
There are only four tasks run on "standalone-all-feature-flags", the other two are:

- ColumnStoreIndex.yml
- TimeSeriesLastpoint.yml

dstorch: > These two files contains "SBE" in file names, so I guess these are the only queries that could be run on SBE engine.

I'm not sure that is true. When `featureFlagSbeFull` is on, as it is in the "all feature flags" build variants, then most queries should be using SBE. I think we want to make sure that pretty much all sys-perf benchmarks that are relevant to query processing run in the new classic engine build variant. Therefore, I suspect that there are additional `AutoRun` configurations in the genny repository that need to change.

@dalyd do you have any tips for how to track down the right set of files? Should we just independently triage all Genny workloads and try to figure out which ones should be running on the classic engine build variant?

Another question: I think some Genny workloads run in sys-perf and some in performance, is that right? Assuming that is correct, can this patch preemptively modify all of the `AutoRun` sections for benchmarks related to query processing, even for those workloads that run in the performance project rather than sys-perf?

dalyd: @allenwux has a nice if experimental tool to view documentation for all the Genny workloads. That might be helpful here. 

Yes, some genny tests run on performance. It makes sense to me to update those also. I'm not entirely sure how to find them. At least some of them are explicitly defined in perf.yml. https://github.com/mongodb/mongo/blob/fecef7a1f75e196a24715fabb0721124e71e170b/etc/perf.yml#L475 
I do see a task generation code in perf.yml but I don't see a matching task, so maybe it's not used. That might be worth checking in #performance-tooling-users.

dstorch (changes requested):
Hmm, @leozzx what was your methodology for finding the set of files modified by this patch? I still can't tell if it's the right subset. I'm wondering if you, me, and David Daly should sit down together to determine the right subset of files to modify. @dalyd do you think that woul be a productive exercise?

My understanding is that we should try to include any query-oriented workload which runs queries that could conceivably execute in SBE. Some of the window function or densification ones, for example, are not super relevant because they never execute in SBE.

dalyd:
That would be fine by me

leozzx:
Discussed with @dstorch, following workloads are included

- ConstantFoldArithmetic.yml
- ExpressiveQueries.yml
- ExternalSort.yml
- GraphLookupWithOnlyUnshardedColls.yml
- LookupSBEPushdown.yml
- LookupSBEPushdownINLJMisc.yml
- LookupWithOnlyUnshardedColls.yml
- UnionWith.yml
- scale/LargeIndexedIns.yml
- docs/RunCommand-Simple.yml
- [scale/MixedWorkloadsGenny.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGenny.yml) (Need @dalyd to confirm)

A few workloads we're not sure, @dalyd could you take a look?

- [scale/LargeScaleLongLived.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/LargeScaleLongLived.yml) 
- [scale/MajorityReads10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MajorityReads10KThreads.yml)
- [scale/MajorityWrites10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MajorityWrites10KThreads.yml) 
- [scale/Mixed10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/Mixed10KThreads.yml) 
- [scale/MixedWorkloadsGennyRateLimited.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGennyRateLimited.yml) 
- [scale/MixedWorkloadsGennyStress.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGennyStress.yml)

dstorch (dismissed):
LGTM with a comment! Please wait for David Daly to sign off on this as well.

dalyd:
> Discussed with @dstorch, following workloads are included
> 
> * ConstantFoldArithmetic.yml
> * ExpressiveQueries.yml
> * ExternalSort.yml
> * GraphLookupWithOnlyUnshardedColls.yml
> * LookupSBEPushdown.yml
> * LookupSBEPushdownINLJMisc.yml
> * LookupWithOnlyUnshardedColls.yml
> * UnionWith.yml
> * scale/LargeIndexedIns.yml
> * docs/RunCommand-Simple.yml
> * [scale/MixedWorkloadsGenny.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGenny.yml) (Need @dalyd to confirm)
> 
> A few workloads we're not sure, @dalyd could you take a look?
> 
> * [scale/LargeScaleLongLived.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/LargeScaleLongLived.yml)
> * [scale/MajorityReads10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MajorityReads10KThreads.yml)
> * [scale/MajorityWrites10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MajorityWrites10KThreads.yml)
> * [scale/Mixed10KThreads.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/Mixed10KThreads.yml)
> * [scale/MixedWorkloadsGennyRateLimited.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGennyRateLimited.yml)
> * [scale/MixedWorkloadsGennyStress.yml](https://github.com/mongodb/genny/blob/master/src/workloads/scale/MixedWorkloadsGennyStress.yml)

MixedWorkloadsGenny does inserts, updates, findOne, and remove in parallel. The updates do an $inc and a $set. All the filters use a random int id value. It's been an interesting test for us. 

After that, MixedWorkloadsGennyStress is just  a scaled up version of MixedWorkloadGenny that hurts the system. I don't think we need to include it. MixedWorkloadsGennyRateLimited.yml is then a rate limited version of it the original. It's subsaturation so the different component latencies can move independently. Because the first two saturate the system, if one component gets faster, other components usually get slower and vice versa. 

I'd skip LargeScaleLongLived. It's focused at the storage engine. 
The other three all appear to be stress tests for large number of connections. That doesn't seem quite relevant to me. Please chime up if you disagree. 

I think you have a nice collection of tests right now.

dalyd (dismissed):

leozzx:
evergreen merge

dstorch (approved):

PR #696, opened 2022-06-14
PERF-3039: Add long running operation workloads from stress testing

brad-devlugt:
This commit adds two new workloads from sharded stress testing

### DIFF
--- a/src/workloads/sharding/DistinctCommands.yml
+++ b/src/workloads/sharding/DistinctCommands.yml
@@ -1,0 +1,64 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  This workload tests distinct commands with large strings to repro SERVER-43096
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 20000
+
+Keywords:
+  - Distinct
+  - Large Strings
+
+Actors:
+- Name: Loader
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: test
+    CollectionCount: 1
+    Threads: 1
+    DocumentCount: 10000
+    BatchSize: 50000
+    Document:
+      shardKey: &RandomString {^FastRandomString: {length: 22}}
+  - {Nop: true}
+  - {Nop: true}
+
+- Name: Distinct
+  Type: RunCommand
+  Threads: 32
+  Phases:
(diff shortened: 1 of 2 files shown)


src/workloads/sharding/DistinctCommands.yml
```diff
@@ -0,0 +1,64 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  This workload tests distinct commands with large strings to repro SERVER-43096
```
jimoleary, line 4: Are there important metrics to look at?

src/workloads/sharding/MongosMerging.yml
```diff
@@ -0,0 +1,151 @@
...
+- Mongos
+- Sample
+- Unwind
+- Sort
+
+ActorTemplates:
+- TemplateName: Aggregation
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "Aggregation"}}
+    Type: RunCommand
```
jimoleary, line 31: Do you care about exhausting the cursor here? If so, then the collection scanner actor will work.

src/workloads/sharding/MongosMerging.yml
```diff
@@ -0,0 +1,151 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  This workload runs different types of aggregations where the query results
```
jimoleary, line 4: What metrics should we be looking at?

jimoleary (commented):
Question about metrics to look at and cursor exhaustion.

src/workloads/sharding/MongosMerging.yml
brad-devlugt: Updated the docs to clarify that we're looking for a failure here

src/workloads/sharding/MongosMerging.yml
brad-devlugt: I'd prefer not to change the workload since we know it repros SERVER-29446 in the current state. I think this could be worth investigating after SERVER-29446 is fixed.

src/workloads/sharding/DistinctCommands.yml
brad-devlugt: Good thought, updated

brad-devlugt:
evergreen merge

PR #697, opened 2022-06-14
TIG-3320 Add documentation for workloads owned by sharding team

sphanse99:


src/workloads/docs/MonotonicSingleLoader.yml
```diff
@@ -1,7 +1,8 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/sharding"
 Description: |
-  TODO: TIG-3320
+  Loads a large set of documents in test.mycoll (Collection0 by default if mycoll is not specified)
```
jz1242, line 4: I'd probably keep it a bit more high level in case the behavior changes. Something like: "Loads a large set of documents with random value assigned to `field`" is probably good enough. The nitty gritty details can be left for the reader in the comments.

src/workloads/execution/CreateIndexSharded.yml
```diff
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/sharding"
 Description: |
-  TODO: TIG-3320
+  Shards collection test.Collection0 and creates a series of indexes on the collection.
```
jz1242, line 4: ditto

jz1242 (changes requested):
looks good just a few comments, also I think the linter was not happy with the formatting in the yaml.

sphanse99:
The Mojave variant is broken right now, see https://mongodb.slack.com/archives/C01VD0LQZED/p1655477546346909/.

jz1242 (changes requested):
Looks good final comment

jz1242 (approved):
LGTM

src/workloads/docs/MonotonicSingleLoader.yml
```diff
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/sharding"
 Description: |
-  TODO: TIG-3320
+  Loads a large set of documents with a random value assinged to `field`.
```
brad-devlugt, line 5: There's a keywords field we could also add to these workloads that we are starting to parse in our workload documentation. See for example: https://github.com/mongodb/genny/blob/master/src/workloads/scale/BigUpdate.yml#L9

brad-devlugt (commented):
Thanks for adding the docs! I just had one suggestion around adding keywords lists for each of the workloads.

src/workloads/docs/MonotonicSingleLoader.yml
sphanse99: Sure! Thanks for the example. I've included some keywords, let me know if there are any others I should add.

brad-devlugt (approved):

jz1242 (approved):
LGTM

sphanse99:
evergreen merge

PR #698, opened 2022-06-16
Document task re-generation

aggrand:


docs/using.md
```diff
@@ -489,7 +489,8 @@ AutoRun searches in `./src/*/src/workloads` and assumes that all repos, includin
...
 
-Both of the above tasks will have a dependency on `schedule_global_auto_tasks`, which invokes `./run-genny auto-tasks` to create all possible tasks, viewable in the `TaskJson` directory of that task's DSI artifacts. The variant-specific task generator task will then schedule the appropriate task, based on the workload configurations described below.
+Both of the above tasks will have a dependency on `schedule_global_auto_tasks`, which invokes `./run-genny auto-tasks` to create all possible tasks, viewable in the `TaskJson` directory of that task's DSI artifacts. The variant-specific task generator task will then schedule the appropriate task, based on the workload configurations described below. Note that task execution only occurs once per patch, even if a task is re-executed. If testing changes to task generation configuration, creating a new patch will be necessary.
```
vrachev, line 492: I think it would be helpful to log the re-execution pitfall [here](https://github.com/mongodb/genny/blob/9febdaf2aae009cdb32775bd4a4a7c443e16ccdd/src/lamplib/src/genny/tasks/auto_tasks.py#L459).

aggrand:
(I think it requested the perf team because I temporarily pushed a commit that had a workload yaml change as I tested patch-tasks, pls ignore.)

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -461,6 +463,11 @@ def write(self, tasks: List[GeneratedTask], write: bool = True) -> Configuration
                     f"{'Succeeded' if success else 'Failed'} to write to {output_file} from cwd={os.getcwd()}."
                     f"{raised if raised else ''}"
                 )
+                if self.op.execution != 0:
+                    SLOG.warning(
```
vrachev, line 467: Thoughts on making this an explicit error?

vrachev (changes requested):
Awesome change, this will be very helpful!

Left a pair of comments, and also need to fix a test https://evergreen.mongodb.com/test_log/genny_amazon2_t_cmake_test_patch_9febdaf2aae009cdb32775bd4a4a7c443e16ccdd_62b36c35850e61240425cd44_22_06_22_19_23_34/0?test_name=550f791c8ca7c8fe4aa799dfb73ba97b&group_id=#L1

Also would be great to get a sys-perf patch where you try to re-execute task generation. Thanks!

src/lamplib/src/genny/tasks/auto_tasks.py
aggrand: I'd argue not an explicit error so that if someone restarts a whole patch, they don't get confusing errors here. I think that's the evergreen rationale for not errorring on it too.

vrachev: That's a great point!

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -133,6 +134,7 @@ def create(
         mode = OpName.ALL_TASKS
         variant = None
 
+        execution = reader.load(workspace_root, "expansions.yml")["build_variant"]
```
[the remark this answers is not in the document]
vrachev: Haha it happens :)

vrachev (approved):
LGTM! Great QOL change

PR #702, opened 2022-06-21
Remove spurious autorun

dalyd:
Genny ignores autorun in the phases directory. This definition has one, and it's surprising and confusing. This change removes it. 

The workload itself still runs. The workload is defined in the workloads directory and uses LoadConfig to load this whole file.


vrachev (approved):

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #704, opened 2022-06-27
Comment out macos temporarily

aggrand:
Until we pull in the ticket, we can comment this out. The test is just noise at this point and makes people merge past the commit queue.


anjani-bhat (approved):

aggrand:
evergreen merge

PR #706, opened 2022-07-01
consolidate the keywords format

allenwux:
consolidate the keywords format to use camel case

### DIFF
--- a/src/workloads/scale/InsertRemove.yml
+++ b/src/workloads/scale/InsertRemove.yml
@@ -8,7 +8,7 @@ Description: |
 
 Keywords:
 - docs
-- InsertRemove
+- insertRemove
 - insert
 - delete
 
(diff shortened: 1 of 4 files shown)


src/workloads/scale/InsertRemove.yml
```diff
@@ -8,7 +8,7 @@ Description: |
 
 Keywords:
 - docs
-- InsertRemove
```
dalyd, line 11: This one uses the InsertRemove actor, which is instantiated with "InsertRemove" in the yaml. I think all the actors are named CamelCase as opposed to lowerCamelCase. I don't have a strong opinion, but that's why this one was different.

allenwux: If this keyword indicates the actor name, can we change it to "actorInsertRemove"? Just trying to consolidate the naming convention since all other keywords are using lowerCamelCase.

dalyd: Yeah, that sounds great!

dalyd (approved):

PR #707, opened 2022-07-03
PERF-3059 Remove cache warmup phases

yun-soo:
Companion PRs:
- https://github.com/10gen/dsi/pull/1063
- https://github.com/10gen/mongo/pull/6151

### DIFF
--- a/src/workloads/execution/ColumnStoreIndex.yml
+++ b/src/workloads/execution/ColumnStoreIndex.yml
@@ -445,12 +445,10 @@ TestRepeatCount: &TestRepeatCount 1
 
 SetupPhase: &SetupPhase 0
 BuildColumnStoreIndexPhase: &BuildColumnStoreIndexPhase 1
-WarmupCache1Phase: &WarmupCache1Phase 2
-ColumnstoreScanPhase: &ColumnstoreScanPhase 3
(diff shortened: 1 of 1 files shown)


jimoleary (changes requested):
It looks like you still need to update the workload description to change the phases / actors.

Do these changes have an effect on performance, i.e. do we need to let the barons know that this will cause a BF?

yun-soo:
> It looks like you still need to update the workload description to change the phases / actors.

Thanks for review, Jim. Updated the description.
> 
> Do these changes have an effect on performance, i.e. do we need to let the barons know that this will cause a BF?

I think it would effect on perf to a certain degree but I asked the perf-build-baroning channel to ignore the perf suite until I'll say otherwise. Since then, I've never asked to monitor the perf suite. Is there any systematic way to mark a test suite as "should be ignored" state for perf build baroning other than asking the channel to do so? As of now, the feature is still under active development.

jimoleary:
> ... but I asked the perf-build-baroning channel to ignore the perf suite until I'll say otherwise. 

Of course. At the moment, there is no better way to do this.

jimoleary (approved):

yun-soo:
Abandoning this PR after talking with @YuhongZhang98.

PR #709, opened 2022-07-07
PERF-3132 simple ^Date generator

jimoleary:
Leverage the existing code for the simplest possible implementation.

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -938,6 +938,8 @@ const static auto formats = {
                 new boost::local_time::local_time_input_facet("%Y-%m-%dT%H:%M:%s%ZP")),
     std::locale(std::locale::classic(),
                 new boost::local_time::local_time_input_facet("%Y-%m-%d %H:%M:%s%ZP")),
+    std::locale(std::locale::classic(),
+                new boost::local_time::local_time_input_facet("%Y-%m-%dT%H:%M:%S%F%ZP *")),
 };
 
 class DateToIntGenerator : public Generator<int64_t> {
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -929,7 +929,7 @@ public:
     }
 };
 
-// The 2 formats also cover "%Y-%m-%d". Timezones require local_time_input_facet AND local_date_time
+// The following formats also cover "%Y-%m-%d". Timezones require local_time_input_facet AND local_date_time
 // see https://www.boost.org/doc/libs/1_75_0/doc/html/date_time/date_time_io.html.
 // We strive to use smart pointers where possible. In this case this is not possible
 // but not a huge deal as these objects are statically allocated.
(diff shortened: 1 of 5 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -938,6 +938,8 @@ const static auto formats = {
                 new boost::local_time::local_time_input_facet("%Y-%m-%dT%H:%M:%s%ZP")),
     std::locale(std::locale::classic(),
                 new boost::local_time::local_time_input_facet("%Y-%m-%d %H:%M:%s%ZP")),
+    std::locale(std::locale::classic(),
+                new boost::local_time::local_time_input_facet("%Y-%m-%dT%H:%M:%S%F%ZP *")),
```
dalyd, line 942: Update comment above please. Is says "2 formats".

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1889,6 +1910,10 @@ UniqueGenerator<int64_t> dateTimeGenerator(const Node& node, GeneratorArgs gener
          [](const Node& node, GeneratorArgs generatorArgs) {
              return std::make_unique<RandomDateGenerator>(node, generatorArgs);
          }},
+        {"^Date",
```
dalyd, line 1913: While here, shoudl  IncDateGenerator be in this list also?

dalyd (dismissed):
That is fairly elegant. It did take me a while to work through why it works.

src/value_generators/src/DocumentGenerator.cpp
jimoleary: good spot!

src/value_generators/src/DocumentGenerator.cpp
jimoleary: it's already in the array on line 1509.

We should probably be better in the way we group the items in the list but I am going to leave this for now.

dalyd: That is different. That list is for all generators and does not assume a type. This is for the things that can return a date type. I'm trying to think of their uses -- it might not make sense to include it here, but I can't come up with a reason right now.

jimoleary: :+1: this doesn't really make sense for {^Date} but it should have been in here (and there are maybe other cases that do make sense). I've added it and created a new test to exercise it.

jimoleary:
@rtimmons I think I need another LGTM for this. It's already been reviewed (and actually most of the changes are either test or unrelated).

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -704,6 +704,61 @@ Tests:
...
+      date: {^Date: "1990-01-01T00:00:00.858+06:00"}
+    ThenReturns:
+      - { "date" : { "$date" : { "$numberLong" : "631130400858" } } }
+
+  # This doesn't make a lot of sense, but it should be possible.
+  - Name: Date with IncDate
+    GivenTemplate:
+      date: {^Date: {^IncDate: {start: "1990-01-01T00:00:00", step: 400}}}
+    ThenReturns:
+      - { "date" : { "$date" : { "$numberLong" : "631152000000" } } }
```
rtimmons, line 760: I think a couple more `ThenReturn` items would illustrate how `^IncDate` works (statefully).

PR #713, opened 2022-07-15
TIG-2766 Genny warn on unused yaml

aggrand:
Almost verbatim the main parts from @rtimmons 's PR (#424), with a more recent commit / some line noise removed. Spent a bit of time trying to get this to work with the STL containers but got a bit tangled up in templates, so agree with pushing this out for now and filing a separate bug ticket to catch it later, and maybe a separate ticket to make it barf explicitly.

Fwiw, I think the stl containers don't see a ton of use extracted from YAML like that in actor code.

Right now it's always on by default in dry runs and workload runs, with warning verbosity if there's anything caught by it. Hopefully people won't find that too annoying.

Patch: https://spruce.mongodb.com/version/62d1afc5c9ec4462ba43fdba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Updated patch: https://spruce.mongodb.com/version/62e93199c9ec445a8443bdb5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Updated updated patch: https://spruce.mongodb.com/version/62eed22bc9ec443061c7d57e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -90,6 +90,10 @@ void reportMetrics(genny::metrics::Registry& metrics,
 
 /**
  * Filter to remove any nodes inside dedicated "ignore" values.
+ *
+ * Basically re-implementing the following python logic:
+ *  ignored = set("foo","bar")
+ *  return [u for u in unused if not any(ignore in u for ignore in ignored)]
  */
 UnusedNodes removeIgnored(const UnusedNodes& unused, const std::vector<std::string>& ignored) {
     UnusedNodes out{};
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -131,6 +131,9 @@ void reportUnused(const NodeSource& nodeSource, const bool dryrun) {
         message << "All YAML structures appear to have been used when " << action
                 << " this workload." << std::endl;
     }
+    message << "The following nodes were ignored in this analysis:" << std::endl;
+    message << "\n\t" << boost::algorithm::join(ignored, "\n\t") << std::endl;
+
     message << "Incorrect results are possible. "
             << "Please file a TIG ticket on the TIPS backlog, or otherwise let us know in the "
             << "#performance-tooling-users slack channel if this looks wrong." << std::endl;
--- a/src/driver/src/DefaultDriver.cpp
+++ b/src/driver/src/DefaultDriver.cpp
@@ -17,6 +17,7 @@
 #include <thread>
 #include <vector>
 
+#include <boost/algorithm/string/join.hpp>
 #include <boost/exception/diagnostic_information.hpp>
 #include <boost/exception/exception.hpp>
 #include <boost/filesystem.hpp>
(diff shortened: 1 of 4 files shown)


src/gennylib/include/gennylib/Node.hpp
```diff
@@ -21,6 +21,7 @@
 #include <utility>
 #include <variant>
 #include <vector>
+#include <typeinfo>
```
rtimmons, line 24: Is this needed?

src/driver/src/DefaultDriver.cpp
```diff
@@ -87,6 +88,63 @@ void reportMetrics(genny::metrics::Registry& metrics,
     actorSetup.report(std::move(finishTime), std::move(duration), std::move(outcome));
 }
 
+/**
+ * Filter to remove any nodes inside dedicated "ignore" values.
+ */
+UnusedNodes removeIgnored(const UnusedNodes& unused, const std::vector<std::string>& ignored) {
```
rtimmons, line 94: Even though I wrote this I still can't easily visually+mentally parse it. I shudder to think what this would have looked like in C++11 though 😂

I think my original impl had the python equivalent explaining the logic. Subjective so fine to ignore (and I'm just lurking on this PR anyway) but consider explaining what this is doing just a bit.

src/gennylib/include/gennylib/Node.hpp
aggrand: Ah remnant from my hack to detect STLs at runtime and to special logic lol, good catch

src/driver/src/DefaultDriver.cpp
```diff
@@ -87,6 +88,67 @@ void reportMetrics(genny::metrics::Registry& metrics,
...
+        message << "\n\t" << boost::algorithm::join(unused, "\n\t") << std::endl;
+        
+    } else {
+        message << "All YAML structures appear to have been used when " << action
+                << " this workload." << std::endl;
+    }
+    message << "The following nodes were ignored in this analysis:" << std::endl;
+    message << "\n\t" << boost::algorithm::join(ignored, "\n\t") << std::endl;
+
+    message << "Incorrect results are possible. "
```
mridul-augustine, line 141: Shouldn't this be displayed only if unused is not empty?

src/driver/src/DefaultDriver.cpp
```diff
@@ -87,6 +88,67 @@ void reportMetrics(genny::metrics::Registry& metrics,
...
+    return out;
+}
+
+/**
+ * Log a message showing the unused yaml structures.
+ */
+void reportUnused(const NodeSource& nodeSource, const bool dryrun) {
+    auto raw = nodeSource.unused();
+
+    std::vector<std::string> ignored{".yml/Description", ".yml/Owner", ".yml/AutoRun", ".yml/Keywords"};
```
mridul-augustine, line 116: Based on patch run add - /Clients/Default/QueryOptions/maxPoolSize to ignore list

mridul-augustine (changes requested):
Couple of minor comments

src/driver/src/DefaultDriver.cpp
aggrand: The incorrect results are possible part? I think that could be the case either way? e.g. maybe unused is empty and we were wrong about that?

mridul-augustine (approved):

aggrand:
evergreen merge

PR #714, opened 2022-07-20
[nomerge] Update to latest vcpkg versions

rtimmons:
Will this just work without also updating the toolchain at the same time?


ancostas:
@rtimmons converting to draft until you're ready to merge.

PR #715, opened 2022-07-25
TIG-3687: run-genny resmoke-test command requires local modifications…

mridul-augustine:
… to run correctly on Evergreen virtual workstations

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -55,10 +55,10 @@ buildvariants:
   - amazon2-build
   expansions:
     distro: amazon2
-    # latest successful build from master as of 2020-10-01 on "Amazon Linux 2" compile task "Binaries" file
+    # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
-      https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/amazon/cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3/binaries/mongo-mongodb_mongo_master_amazon_cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3_21_12_03_09_47_48.tgz
+      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_amazon2_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz
   tasks:
   - name: tg_compile_and_test_with_server
 
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -260,7 +260,7 @@ functions:
     params:
       directory: src/genny
       revisions:
-        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
+        mongo: e61bf27c2f6a83fed36e5a13c008a32d563babe2
 
   ##
   # Compile the project in src/build.
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -123,7 +123,7 @@ buildvariants:
 tasks:
 - name: t_compile
   commands:
-  # - func: f_fetch_source
+  - func: f_fetch_source
   - func: f_compile
   - func: f_dry_run_workloads
 
@@ -259,8 +259,8 @@ functions:
   - command: git.get_project
     params:
       directory: src/genny
-  #      revisions:
-  #        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
+      revisions:
+        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
 
   ##
   # Compile the project in src/build.
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -123,7 +123,7 @@ buildvariants:
 tasks:
 - name: t_compile
   commands:
-  - func: f_fetch_source
+  # - func: f_fetch_source
   - func: f_compile
   - func: f_dry_run_workloads
 
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -58,7 +58,7 @@ buildvariants:
     # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
-      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_amazon2_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz
+      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz
   tasks:
   - name: tg_compile_and_test_with_server
 
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -55,8 +55,6 @@ buildvariants:
   - amazon2-build
   expansions:
     distro: amazon2
-    # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
-    # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
       https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz
   tasks:
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -62,15 +62,6 @@ buildvariants:
   tasks:
   - name: tg_compile_and_test_with_server
 
-- name: archlinux
-  display_name: Arch Linux
-  modules: [mongo]
-  expansions:
-    distro: archlinux
-  run_on:
-  - archlinux-large
-  tasks:
-  - name: tg_compile
 
 - name: rhel70
   display_name: RHEL 7
@@ -89,16 +80,6 @@ buildvariants:
     distro: rhel8
   run_on:
   - rhel80-test
-  tasks:
-  - name: tg_compile
-
-- name: ubuntu1604
-  display_name: Ubuntu 16.04
-  modules: [mongo]
-  expansions:
-    distro: ubuntu1604
-  run_on:
-  - ubuntu1604-large
   tasks:
   - name: tg_compile
 
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -283,7 +283,7 @@ functions:
     params:
       directory: src/genny
       revisions:
-        mongo: cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3
+        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
 
   ##
   # Compile the project in src/build.
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -59,7 +59,6 @@ buildvariants:
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
       https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_amazon2_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz
-      
   tasks:
   - name: tg_compile_and_test_with_server
 
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -55,8 +55,6 @@ buildvariants:
   - amazon2-build
   expansions:
     distro: amazon2
-    mongodb_archive_url:
-      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz
   tasks:
   - name: tg_compile_and_test_with_server
 
@@ -261,8 +259,8 @@ functions:
   - command: git.get_project
     params:
       directory: src/genny
-      revisions:
-        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
+  #      revisions:
+  #        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
 
   ##
   # Compile the project in src/build.
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -59,6 +59,7 @@ buildvariants:
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
       https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_amazon2_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz
+      
   tasks:
   - name: tg_compile_and_test_with_server
 
--- a/src/lamplib/src/genny/cli.py
+++ b/src/lamplib/src/genny/cli.py
@@ -46,16 +46,7 @@ def cli(ctx: click.Context, verbose: bool) -> None:
     "--linux-distro",
     required=False,
     default="not-linux",
-    type=click.Choice(
-        [
-            "ubuntu1804",
-            "ubuntu2004",
-            "rhel8",
-            "rhel70",
-            "amazon2",
-            "not-linux",
-        ]
-    ),
+    type=click.Choice(["ubuntu1804", "ubuntu2004", "rhel8", "rhel70", "amazon2", "not-linux",]),
     help=(
         "Specify the linux distro you're on; if your system isn't available,"
         " please contact us at #workload-generation. The not-linux value is useful on macOS."
@@ -454,7 +445,7 @@ def resmoke_test(
     create_new_actor_test_suite: bool,
     mongo_dir: Optional[str],
     mongodb_archive_url: Optional[str],
-    mongodb_commit: Optional[str]		
+    mongodb_commit: Optional[str],
 ):
     from genny.tasks import run_tests
 
@@ -466,7 +457,7 @@ def resmoke_test(
         mongo_dir=mongo_dir,
         env=os.environ.copy(),
         mongodb_archive_url=mongodb_archive_url,
-        mongodb_commit=mongodb_commit
+        mongodb_commit=mongodb_commit,
     )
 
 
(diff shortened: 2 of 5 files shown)


src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -163,10 +163,14 @@ def out(cmd_output: str) -> bool:
 # See the logic in _setup_resmoke.
 # These are the "Binaries" evergreen artifact URLs for mongodb-mongo compile tasks.
 # The binaries must be compatible with the version of the mongo repo checked out in use for resmoke,
-# which is the sha "cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3" mentioned below.
+# which is the sha "6e2dcc5a39eb2de9d2e8209271115c078ae16470" mentioned below.
 _canned_artifacts = {
-    "osx": "https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/macos/cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3/binaries/mongo-mongodb_mongo_master_macos_cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3_21_12_03_09_47_48.tgz",
-    "amazon2": "https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/amazon/cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3/binaries/mongo-mongodb_mongo_master_amazon_cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3_21_12_03_09_47_48.tgz",
+    "osx": "https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/macos/6e2dcc5a39eb2de9d2e8209271115c078ae16470/binaries/mongo-mongodb_mongo_master_macos_6e2dcc5a39eb2de9d2e8209271115c078ae16470_21_12_03_09_47_48.tgz",
```
aggrand, line 168: I see that osx is in the mciuploads bucket, is that a straight-up evg build artifact? If so, should we put it in dsi-donot-remove as well?

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -163,10 +163,14 @@ def out(cmd_output: str) -> bool:
 # See the logic in _setup_resmoke.
 # These are the "Binaries" evergreen artifact URLs for mongodb-mongo compile tasks.
 # The binaries must be compatible with the version of the mongo repo checked out in use for resmoke,
-# which is the sha "cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3" mentioned below.
+# which is the sha "6e2dcc5a39eb2de9d2e8209271115c078ae16470" mentioned below.
```
aggrand, line 166: Should we use a release version? That would minimize the chances of random server bugs being in it, since the release process does some good testing.

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -163,10 +163,14 @@ def out(cmd_output: str) -> bool:
...
+    "osx": "https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/macos/6e2dcc5a39eb2de9d2e8209271115c078ae16470/binaries/mongo-mongodb_mongo_master_macos_6e2dcc5a39eb2de9d2e8209271115c078ae16470_21_12_03_09_47_48.tgz",
+    "amazon2": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_amazon2_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz",
+    "ubuntu1804": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_enterprise_ubuntu1804_64_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz",
+    "ubuntu2004": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_ubuntu2004_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz",
+    "rhel70": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_rhel70_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz",
+    "rhel8": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongo-mongodb_mongo_v6.0_enterprise_rhel_80_64_bit_6e2dcc5a39eb2de9d2e8209271115c078ae16470_22_06_29_21_31_44.tgz",
```
aggrand, line 173: It's further down and unchanged, but I think the automatic search for a mongo repo and binary is happening still? I think we want to remove that logic and make sure we only pick up the mongo repo and existing binaries if users explicitly specify them with a flag.

src/lamplib/src/genny/tasks/run_tests.py
aggrand: Also, can we document either in a comment here or in one of the .md files how to update these canned artifacts if needed?

aggrand (changes requested):
Looks like a good start! Just a few notes!

src/lamplib/src/genny/tasks/run_tests.py
mridul-augustine: Looking into https://github.com/mongodb/mongo/commits/r6.0.0 "e61bf27" is the release version with no actual code changes or evergreen runs. "6e2dcc5" (which is currently being used) is the next commit. Is there another release version?

src/lamplib/src/genny/tasks/run_tests.py
mridul-augustine: Yeah missed this part of the ticket.

src/lamplib/src/genny/tasks/run_tests.py
mridul-augustine: Added official 6.0 build to canned artifacts

evergreen.yml
```diff
@@ -55,22 +55,13 @@ buildvariants:
...
-    # latest successful build from master as of 2020-10-01 on "Amazon Linux 2" compile task "Binaries" file
+    # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
-      https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/amazon/cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3/binaries/mongo-mongodb_mongo_master_amazon_cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3_21_12_03_09_47_48.tgz
+      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz
   tasks:
   - name: tg_compile_and_test_with_server
 
-- name: archlinux
```
aggrand, line 65: Bump about asking about these variants in #performance-tooling-users! I think @mvankeulen94 was also looking into discussions about genny platform support, and might want to have a say? I might be misremembering though.

evergreen.yml
```diff
@@ -283,7 +264,7 @@ functions:
     params:
       directory: src/genny
       revisions:
-        mongo: cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3
+        mongo: 6e2dcc5a39eb2de9d2e8209271115c078ae16470
```
aggrand, line 267: Same here about specifying in evergreen, ideally the genny python can handle all of this now?

src/lamplib/src/genny/cli.py
```diff
@@ -430,6 +418,16 @@ def canaries(ctx: click.Context, canary_args: List[str]):
...
     ),
 )
+@click.option(
+    "--mongodb-commit",
+    type=str,
+    required=False,
+    default=None,
+    help=(
+        "The commit corresponding to the mongodb binary passed with -mongodb-archive-url option."
+        "Not needed if you are not passing --mongodb-archive-url."
```
aggrand, line 428: I'm confused, these lines say it has to correspond with the commit of the binary used with that option, but it shouldn't be used with that option? Is that contradictory?

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -201,80 +208,43 @@ def _setup_resmoke(
...
-            cmd=["git", "rev-parse", "HEAD"], check=False, cwd=mongo_repo_path, capture=False,
-        )
-
-    # Look for mongod in
-    # build/opt/mongo/db/mongod
-    # build/install/bin/mongod
-    # bin/
-    opt = os.path.join(mongo_repo_path, "build", "opt", "mongo", "db", "mongod")
-    install = os.path.join(mongo_repo_path, "build", "install", "bin", "mongod")
+    cmd_runner.run_command(
```
aggrand, line 211: Is this unconditionally checking out the mongo repo? I think if someone passing `--mongo-repo-path` that means they have an existing mongo repo they want to use, possibly on a specific commit, and we shouldn't do a checkout?

We could probably eliminate the mongo commit option and just always use the pinned commit or whatever the user has checked out. I'm wary about running git inside someone's repo and maybe messing things up

aggrand (changes requested):
Looking great so far! Just have a few more questions / nitpicks!

evergreen.yml
```diff
@@ -55,22 +55,13 @@ buildvariants:
   - amazon2-build
   expansions:
     distro: amazon2
-    # latest successful build from master as of 2020-10-01 on "Amazon Linux 2" compile task "Binaries" file
+    # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
     # commit should match that in get-mongo-source.sh
     mongodb_archive_url:
-      https://mciuploads.s3.amazonaws.com/mongodb-mongo-master/amazon/cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3/binaries/mongo-mongodb_mongo_master_amazon_cf8ddbdf99ae7fb2d96ca9f350224ffa193e40b3_21_12_03_09_47_48.tgz
+      https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz
```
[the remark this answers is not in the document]
mridul-augustine: I'm not sure what's the reasoning for this being different. But I feel it's a good idea for one of the variants to use mongodb_archive_url to ensure this option is not broken.

evergreen.yml
mridul-augustine: archlinux is not under [supported platforms](https://www.mongodb.com/docs/manual/installation/#supported-platforms) for 6.0.

src/lamplib/src/genny/cli.py
mridul-augustine: It says: - '--mongodb-commit' option is not needed if a user is not passing '--mongodb-archive-url'. 
It doesn't say it shouldn't be used with that option?

evergreen.yml
```diff
@@ -55,22 +55,13 @@ buildvariants:
   - amazon2-build
   expansions:
     distro: amazon2
-    # latest successful build from master as of 2020-10-01 on "Amazon Linux 2" compile task "Binaries" file
+    # latest successful build from master as of 2022-06-29 on "Amazon Linux 2" compile task "Binaries" file
     # commit should match that in get-mongo-source.sh
```
mridul-augustine, line 59: get-mongo-source.sh is not used anymore

src/lamplib/src/genny/cli.py
mridul-augustine: removing this line to avoid confusion.

src/lamplib/src/genny/cli.py
```diff
@@ -430,6 +418,15 @@ def canaries(ctx: click.Context, canary_args: List[str]):
...
 )
+@click.option(
+    "--mongodb-commit",
+    type=str,
+    required=False,
+    default=None,
+    help=(
+        "The commit corresponding to the mongodb binary passed with -mongodb-archive-url option."
+    ),
+)
```
aggrand, line 429: I'm not 100% convinced that the `--mongodb-commit` flag is necessary. I think someone who needs it could just clone the repo, check out whatever commit they want, and specify that dir. 

But won't block the PR on that, up to you either way!

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -10,6 +10,20 @@
...
+# which is the sha "6e2dcc5a39eb2de9d2e8209271115c078ae16470" mentioned below.
+CANNED_ARTIFACTS = {
+    "osx": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-macos-x86_64-6.0.0.tgz",
+    "amazon2": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-amazon2-6.0.0.tgz",
+    "ubuntu1804": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-ubuntu1804-6.0.0.tgz",
+    "ubuntu2004": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-ubuntu2004-6.0.0.tgz",
+    "rhel70": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-rhel70-6.0.0.tgz",
+    "rhel8": "https://dsi-donot-remove.s3.us-west-2.amazonaws.com/compile_artifacts/mongodb-linux-x86_64-rhel80-6.0.0.tgz",
+}
+MONGO_COMMIT = "6e2dcc5a39eb2de9d2e8209271115c078ae16470"
```
aggrand, line 25: Is it possible to derive the commit from the binary, using `./mongod --version` or something? Super cool if not but that'd be one less config option to track

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -203,78 +210,41 @@ def _setup_resmoke(
...
+        capture=False,
+        check=True,
+    )
+    cmd_runner.run_command(
+        cmd=["tar", "--strip-components=1", "-zxf", "mongodb.tgz"],
+        cwd=mongo_repo_path,
+        capture=False,
+        check=True,
+    )
+    mongod = from_tarball
```
aggrand, line 247: Is this downloading and extracting unconditionally? I think we want to check if it's already there, and reuse it if so (making sure it's the right version, so that python pinned version updates will overwrite the one there).

aggrand (changes requested):
Only blocking thing is the conditional download thing. Sorry I didn't see that on an earlier pass.

We use the downloader.py class here in the genny repo to handle that for the toolchain and such, maybe that'd be useful here?

src/lamplib/src/genny/cli.py
mridul-augustine: I agree. Removing the option

evergreen.yml
mridul-augustine: This is required and is not part of resmoke. This is required for the run-genny used in evergreen.yml.

src/lamplib/src/genny/tasks/run_tests.py
mridul-augustine: How do I know what is the right version without extracting the artificat?

aggrand: I think we would check if there is already a downloaded and extracted artifact in the build dir. If so, we can check the version on it and use it. If there is not a binary already in the build dir, we download and extract. (Not sure if the intermediate tarball would be kept, probably not needed).

So ideally the first time I run genny it'll download and extract, then the second time it'll use the binary already there.

aggrand (changes requested):
Adding a request-changes review based on the comment added. Feel free to slack me and we can chat about it too!

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -201,82 +227,63 @@ def _setup_resmoke(
...
-    else:
-        mongod = None
-
-    if mongod is not None and mongodb_archive_url is not None:
-        SLOG.info(
-            "Found existing mongod so will not download artifacts.",
-            existing_mongod=mongod,
-            wont_download_artifacts_from=mongodb_archive_url,
-        )
+    download_required = True
```
aggrand, line 234: Beautiful!

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -201,82 +227,63 @@ def _setup_resmoke(
...
+    if os.path.exists(from_tarball):
+        mongodb_commit = _get_mongo_commit(from_tarball, workspace_root)
+        if mongodb_commit == expected_commit:
+            SLOG.info("Mongo binary exist and is the correct version")
+            download_required = False
+        else:
+            SLOG.info(
+                "Mongo binary exist, but is not the correct version. Mongo will be dowloaded from the canned artifact."
```
aggrand, line 242: Maybe log the versions for debugging purposes?

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -201,82 +227,63 @@ def _setup_resmoke(
...
+            cwd=mongo_repo_path,
+            capture=False,
+            check=True,
+        )
+        cmd_runner.run_command(
+            cmd=["tar", "--strip-components=1", "-zxf", "mongodb.tgz"],
+            cwd=mongo_repo_path,
+            capture=False,
+            check=True,
+        )
```
aggrand, line 274: Do these both work if there's already something there? Just wanted to check!

src/lamplib/src/genny/tasks/run_tests.py
```diff
@@ -201,82 +227,63 @@ def _setup_resmoke(
...
-            cwd=mongo_repo_path,
-            check=True,
-            capture=False,
-        )
-    else:
-        SLOG.info("Using existing mongo repo checkout", mongo_repo_path=mongo_repo_path)
-        cmd_runner.run_command(
-            cmd=["git", "rev-parse", "HEAD"], check=False, cwd=mongo_repo_path, capture=False,
-        )
+        checkout_required = True
```
aggrand, line 230: Why do we checkout if using the existing repo? Especially since a comment further down says "only required if the local repo didn't exist and had to be cloned"

In particular, this could be the user-specified repo, right? In that case I think we agreed we wouldn't touch it?

I do think we wanna check if this is the repo at the default location or the user-specified one, and only checkout with the former (so that genny updates to the checked-out repo are applied)

(Sorry for being so overfocused on that! I made a mistake in this file that deleted someone's local mongo repo and want to avoid a similar issue.)

aggrand (changes requested):
Only blocker was the checkout thing

aggrand (approved):

PR #719, opened 2022-08-04
PERF-3181: Incorrect comment in DocumentGenerator.cpp

miguelangelnieto:
Updated the randomDouble generator comments to reflect that we are returning double, not int

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -1707,7 +1707,7 @@ UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorA
 
 /**
  * @param node
- *   a top-level document value i.e. either a scalar or a `^RandomInt` value
+ *   a top-level document value i.e. either a scalar or a `^RandomDouble` value
  * @return
  *   either a `^RandomDouble` generator (etc--see `doubleParsers`)
  *   or a constant generator if given a constant/scalar.
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -1586,8 +1586,8 @@ UniqueGenerator<bsoncxx::array::value> literalArrayGenerator(const Node& node,
 
 /**
  * @param node
- *   the *value* from a `^RandomInt` node.
- *   E.g. if higher-up has `{^RandomInt:{v}}`, this will have `node={v}`
+ *   the *value* from a `^RandomDouble` node.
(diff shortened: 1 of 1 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1711,7 +1709,7 @@ UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorA
  * @param node
  *   a top-level document value i.e. either a scalar or a `^RandomInt` value
```
dalyd, line 1710: Should this ^RandomInt also be a ^RandomDouble?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1711,7 +1709,7 @@ UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorA
...
  * @return
- *   either a `^RantomInt` generator (etc--see `intParsers`)
+ *   either a `^RandomDouble` generator (etc--see `intParsers`)
```
dalyd, line 1712: Shouldn't the parenthetical comment be updated also. hopefully see `doubleParsers`

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: As far as I have seen from the workloads where randomDouble is used, the parameter we pass are integers, so that's why I didn't change that.

src/value_generators/src/DocumentGenerator.cpp
miguelangelnieto: Oops, forgot that one.

rtimmons (approved):

dalyd (approved):
I think that one other line should be changed too. I'm sure the origin of this was my copy paste on code, and forgetting to update as appropriate.

miguelangelnieto:
@dalyd I get it now, thank you! I have changed that. Unfortunately, that means you need to approve it again.

src/value_generators/src/DocumentGenerator.cpp
dalyd: This one still needs to be updated, doesn't it?

miguelangelnieto: That's fully updated now.

dalyd (approved):

miguelangelnieto:
@rtimmons sorry but I need your approval again.

rtimmons (approved):

PR #720, opened 2022-08-04
PERF-3196: Large index insert test does not perform any update operation after the index is created.

miguelangelnieto:
Updated the workload to remove the update operations and unused fields from the index.


dalyd (approved):

miguelangelnieto:
evergreen merge

PR #724, opened 2022-08-11
Update python deps; don't rely on /data for run-genny

rtimmons:
WIP. We had an assumption that the toolchain was not relocatable, but I think that is wrong. That means we no longer have to fear latest macOS which doesn't let us write to `/data`.

Also

1. updated to later versions of a few python deps that don't install on M1.
2. don't build cppcodec's tests which don't compile on M1 (maybe Monterey?) due to (vendored) catch2 things

### DIFF
--- a/src/lamplib/requirements.txt
+++ b/src/lamplib/requirements.txt
@@ -1,4 +1,4 @@
-wheel==0.36.2
+wheel==0.37.1
 pymongo>=3.9
 click==7.1.2
 click-option-group==0.5.2
@@ -15,7 +15,7 @@ pytest==6.0.2
 
 # Force a binary distribution of numpy on macOS:
 # https://github.com/numpy/numpy/issues/15947
-numpy==1.20.1 --only-binary=numpy; platform_system == 'darwin'
-numpy==1.20.1;                     platform_system != 'darwin'
+numpy==1.23.1 --only-binary=numpy; platform_system == 'darwin'
+numpy==1.23.1;                     platform_system != 'darwin'
 
-numexpr==2.8.1
+numexpr==2.8.3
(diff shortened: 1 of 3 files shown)


src/lamplib/requirements.txt
```diff
@@ -15,7 +15,7 @@ pytest==6.0.2
 
 # Force a binary distribution of numpy on macOS:
 # https://github.com/numpy/numpy/issues/15947
-numpy==1.20.1 --only-binary=numpy; platform_system == 'darwin'
-numpy==1.20.1;                     platform_system != 'darwin'
+numpy==1.23.1 --only-binary=numpy; platform_system == 'darwin'
+numpy==1.23.1;                     platform_system != 'darwin'
 
-numexpr==2.8.1
+numexpr==2.8.3
```
ancostas, line 21: Can you cut a tech-debt tix to make this into a pyproject.toml?

src/lamplib/src/genny/toolchain.py
```diff
@@ -164,7 +164,7 @@ class ToolchainDownloader(Downloader):
 
     TOOLCHAIN_BUILD_ID = "fc5ec55493f12c0791739e66bd9ffc6db78468e1_22_01_31_16_45_23"
     TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split("_")[0]
-    TOOLCHAIN_ROOT = "/data/mci"  # TODO BUILD-7624 change this to /opt.
+    # TOOLCHAIN_ROOT = "/data/mci"  # TODO BUILD-7624 change this to /opt.
```
ancostas, line 167: Just delete? Or why do we still need this?

src/third_party/cppcodec/CMakeLists.txt
```diff
@@ -66,9 +66,9 @@ set_target_properties(cppcodec PROPERTIES LINKER_LANGUAGE CXX)
 add_subdirectory(tool)
 add_subdirectory(example)
 
-if (BUILD_TESTING)
-    add_subdirectory(test)
-endif()
+# if (BUILD_TESTING)
+#     add_subdirectory(test)
+# endif()
```
ancostas, line 71: Same

PR #726, opened 2022-08-18
PERF-3186 Extend Genny to add FLE support

erwee:


src/gennylib/src/PoolManager.cpp
```diff
@@ -38,9 +39,20 @@ auto createPool(const std::string& name,
...
+    poolFactory.setEncryptionContext(std::move(encryption));
     return poolFactory.makePool();
 }
 
+auto setupEncryption(const std::string& name, const Node& context, bool dryRun) {
+    auto mongoUri = context["Clients"][name]["URI"].to<std::string>();
+    auto encryption = std::make_shared<EncryptionContext>(
+        context["Clients"][name]["EncryptionOptions"], std::move(mongoUri));
+    if (!dryRun) {
+        encryption->setupKeyVault();
```
jimoleary, line 51: if `setupKeyVault` took a `mongoUri` as a parameter then the EncryptionContext wouldn't need to store the value and then maybe the new Pools struct could be rolled back?

src/gennylib/src/PoolManager.cpp
```diff
@@ -62,9 +74,13 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
...
 
-    auto& pool = pools[instance];
+    if (!pools.encryption) {
+        pools.encryption = setupEncryption(name, context, _dryRun);
+    }
+
+    auto& pool = pools.instances[instance];
     if (pool == nullptr) {
-        pool = createPool(name, this->_apmCallback, context);
+        pool = createPool(name, this->_apmCallback, pools.encryption, context);
```
jimoleary, line 83: the context is being passed to createPool, would it not make more sense to let it create and configure encryption?

src/gennylib/src/PoolManager.cpp
```diff
@@ -21,6 +21,7 @@ namespace {
 
 auto createPool(const std::string& name,
                 PoolManager::OnCommandStartCallback& apmCallback,
+                std::shared_ptr<EncryptionContext> encryption,
```
jimoleary, line 24: It feels a bit odd that the other configuration is read from the context here, but the encryption is passed in?

I think that reading the EncryptionContext here would significantly simplify the code and lower the scale of the changes.

jimoleary (changes requested):
This change is quite large so I haven't fully scanned it all yet .... but 


I feel that if `PoolManager::createPool ` / https://github.com/mongodb/genny/pull/726/files#diff-b934d03673bad68251834d6b9bfff6722ca0ded5e9960eb4cd6a94166e684ae0R24:~:text=std%3A%3Astring%26%20name%2C-,auto%20createPool(const%20std%3A%3Astring%26%20name%2C,std%3A%3Ashared_ptr%3CEncryptionContext%3E%20encryption%2C,-Member  was responsible for reading the EncryptionContext from the Node context, then the scale of the changes would be smaller / simpler.

src/gennylib/src/PoolManager.cpp
erwee: The reason that `setupEncryption()` is being done here, instead of within `createPool()` is that the initialization that's being done in `setupEncryption()` (e.g. creating the key vault & data keys) must only happen once, when the very first "instance" of this Pool is created.

When subsequent instances of this pool name are initialized, they would need to reuse the existing EncryptionContext to set up their auto-encryption options.  These subsequent instances cannot just re-create the EncryptionContext from the YAML node, because if they did, they will lose the data key IDs that were created during `setupKeyVault()`.

src/gennylib/include/gennylib/v1/PoolFactory.hpp
```diff
@@ -66,6 +67,8 @@ class PoolFactory {
      */
     void setOption(OptionType type, const std::string& option, std::string value);
 
+    void setEncryptionContext(std::shared_ptr<EncryptionContext> encryption);
```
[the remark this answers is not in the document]
erwee: I decided to use shared pointer here because it's PoolManager that actually needs to maintain the EncryptionContext for all instances of this client pool name, and so that makes PoolManager the real "owner".

src/gennylib/src/PoolFactory.cpp
```diff
@@ -192,28 +194,59 @@ std::string PoolFactory::makeUri() const {
 }
 
 mongocxx::options::pool PoolFactory::makeOptions() const {
-    mongocxx::options::tls tlsOptions;
+    mongocxx::options::client clientOptions;
```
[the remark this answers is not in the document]
erwee: I'll revert the `std::move` so it remains true to the original. 

As for setting the `apm_opts`, the original code was creating a copy of the entire `options::client` object because it cannot just simply add the new options to the existing `options::client` within `options::pool` as it is `const`ed & read only. Since I've unified all the options setting in this function, I can just set the options directly in the `clientOptions`, before constructing the `options::pool`.

src/gennylib/src/PoolFactory.cpp
```diff
@@ -222,30 +255,19 @@ std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
...
-    return std::make_unique<mongocxx::pool>(uri, clientOpts);
+    return std::make_unique<mongocxx::pool>(uri, poolOptions);
 }
 
 void PoolFactory::setOption(OptionType type, const std::string& option, std::string value) {
     _config->set(type, option, value);
 }
 
+void PoolFactory::setEncryptionContext(std::shared_ptr<EncryptionContext> encryption) {
+    _config->encryptionCtxt = std::move(encryption);
```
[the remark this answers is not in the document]
erwee: I wouldn't consider this an issue, because `std::move`ing a shared pointer just means the atomic updates to the reference counter are avoided.

src/gennylib/src/PoolFactory.cpp
```diff
@@ -192,28 +194,59 @@ std::string PoolFactory::makeUri() const {
...
 
-    // Just doing CAFile and PEMKeyFile for now, it's reasonably trivial to add other options
-    // Note that this is entering as a BSON string view, so you cannot delete the config object
-    auto caFile = *_config->get(OptionType::kAccessOption, "CAFile");
-    if (!caFile.empty()) {
-        BOOST_LOG_TRIVIAL(debug) << "Using CA file '" << caFile << "' for TLS";
-        tlsOptions = tlsOptions.ca_file(caFile.data());
+    if (useEncryption) {
+        const auto& encryption = *_config->encryptionCtxt;
+        mongocxx::options::auto_encryption autoEncryptionOptions{};
```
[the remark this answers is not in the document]
erwee: done

src/gennylib/src/PoolFactory.cpp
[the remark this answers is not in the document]
erwee: removed it anyways avoid confusion

src/gennylib/src/encryption.cpp
```diff
@@ -0,0 +1,446 @@
...
+ * Splits a dot-separated path string into its constituent parts &
+ * returns a list.
+ *
+ * @param path the dotted path string
+ * @return a non-empty ordered list of parsed path components
+ * @throws InvalidConfigurationException
+ *   if path is empty, or has empty path segments
+ */
+std::vector<std::string> splitDottedPath(const std::string& path) {
+    std::vector<std::string> parts;
```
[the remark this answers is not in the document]
erwee: done

src/gennylib/src/encryption.cpp
```diff
@@ -0,0 +1,527 @@
...
+    _type = yaml["type"].to<std::string>();
+
+    // verify path is valid
+    splitDottedPath(_path);
+
+    if (yaml["keyId"]) {
+        auto keyId = yaml["keyId"].to<std::string>();
+        try {
+            auto jsonDoc = R"({"v": {"$uuid": ")" + keyId + R"("}})";
+            auto uuidDoc = bsoncxx::from_json(jsonDoc);
```
jimoleary, line 292: would something like boost::format make this a bit more readable:

```
auto uuidDoc = bsoncxx::from_json(boost::format(R"({"v": {"$uuid": "%1%" }})" % keyId));
```

src/gennylib/src/encryption.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+using bsoncxx::builder::basic::kvp;
+using bsoncxx::builder::basic::make_document;
+using bsoncxx::builder::basic::sub_array;
+using bsoncxx::builder::basic::sub_document;
+using bsoncxx::types::bson_value::view_or_value;
+
+EncryptedField::EncryptedField(const Node& yaml) {
+    _path = yaml.key();
+    _type = yaml["type"].to<std::string>();
```
jimoleary, line 283: what happens if type is missing?

src/gennylib/test/PoolFactory_test.cpp
```diff
@@ -239,6 +239,55 @@ TEST_CASE("PoolFactory behavior") {
         REQUIRE(pool);
     }
 
+    SECTION("Make a pool with client-side encryption enabled") {
```
jimoleary, line 242: It looks like there are tests for valid combinations of FLE fields in encryption_test.cpp.


Do we need a test with a field with an empty keyId? It looks like this is a valid permutation.

src/gennylib/test/encryption_test.cpp
```diff
@@ -0,0 +1,422 @@
...
+    EncryptionContext encryption(ns.root(), kSourceUri);
+    auto doc = encryption.generateExtraOptionsDoc();
+    REQUIRE(doc.view()["mongocryptdBypassSpawn"]);
+    REQUIRE_NOTHROW(doc.view()["mongocryptdBypassSpawn"].get_bool());
+    REQUIRE(doc.view()["mongocryptdBypassSpawn"].get_bool());
+    REQUIRE(doc.view()["cryptSharedLibRequired"]);
+    REQUIRE_NOTHROW(doc.view()["cryptSharedLibRequired"].get_bool());
+    REQUIRE(doc.view()["cryptSharedLibRequired"].get_bool() == false);
+}
+TEST_CASE("EncryptionContext outputs correct schema map document") {
```
jimoleary, line 319: Is there need for a unit test without a keyId?

jimoleary (commented):
Looks mostly ok (sorry for the delay).

I have some questions about keyId tests and / or examples/

src/gennylib/src/PoolManager.cpp
aggrand: So you're saying encryption settings are global? I think we could still create this inside createPool. We have a few instances in Genny of creating static vars that we can expect to be reused across different calls, e.g. [here](https://github.com/mongodb/genny/blob/d5e219bbbf7d182e8872abe7df40d082e7cb0ca8/src/metrics/include/metrics/v2/event.hpp#L117).

I'm imagining something like the following inside `createPool`:

    static auto encryption = setupEncryption(name, context, _dryRun);

I think it would be worth doing that to keep this construction internal to the pool construction, and avoid some of the pointer ownership issues / var piping mentioned further up.

src/gennylib/src/encryption.cpp
erwee: it will throw an exception, with an error saying `"Invalid key 'type': Tried to access node that doesn't exist. On node with path ..."`

src/gennylib/src/PoolManager.cpp
aggrand: Also, if it's only the results of `setupKeyVault()` specifically that need to be shared, maybe that could be constructed statically and everything else still treated as pool-specific?

src/gennylib/src/encryption.cpp
```diff
@@ -0,0 +1,527 @@
+// Copyright 2022-present MongoDB Inc.
```
aggrand, line 1: Before digging into this, would it make sense to split this into two PRs, one for this functionality and one that integrates it with the rest of Genny?

src/gennylib/src/PoolManager.cpp
```diff
@@ -62,9 +77,15 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
 
     Pools& pools = lap.second;
 
-    auto& pool = pools[instance];
+    if (pools.instances.empty() && !pools.encryption) {
+        // set up encryption context when the first instance of this
+        // pool is about to be created.
+        pools.encryption = setupEncryption(name, context, _dryRun);
+    }
```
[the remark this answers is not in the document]
aggrand: Also not sure how this would handle different pools with different encryption setups.

src/workloads/docs/CrudActorEncrypted.yml
```diff
@@ -0,0 +1,166 @@
...
+    algorithm - the encryption algorithm: either "random" or "deterministic"
+                "deterministic" means that the ciphertext is always the same for the same
+                plaintext value, which means that this field can be used on queries.
+
+  Under QueryableEncryptedFields, each object mapped to a path must have the following fields:
+    type      - the BSON type of values for this field
+    queries   - an array of objects that specify "queryType" and other parameters for QE.
+
+  An Actor that wishes to perform encrypted operations on an encrypted collection must
+  be using the Client pool under which the schema for that encrypted collection is defined.
```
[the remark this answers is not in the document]
erwee: The way this is currently structured does not allow for 2 pools to share the same schema, and if two pools on the same URI happen to have the same copy of an encrypted collection's schema, then they would each generate their own data keys, which is not good.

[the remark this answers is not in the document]
erwee: I think to accomplish sharing between two pools, we'd need to have a global `EncryptionManager` that will handle all the global initialization (key vault & data key creation). Afterwards, pools can reference any schema owned by the EncryptionManager in order to use encrypted operations.

src/gennylib/src/encryption.cpp
aggrand: (Also saw there's now comments on this file, so feel free to ignore this point!)

src/gennylib/src/PoolManager.cpp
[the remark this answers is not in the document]
erwee: No, this setup will happen on all Pools with encryption enabled. The pools.instances map is empty if the very first instance pool with the given `name` is about to be created.  Since `Default` and `EncryptedPool` are different pool names, there will always be a setupEncryption call when a pool named `EncryptedPool` is first initialized.

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -77,14 +102,20 @@ class PoolManager {
     /** callback passed into ctor */
     OnCommandStartCallback _apmCallback;
 
-    using Pools = std::unordered_map<size_t, std::unique_ptr<mongocxx::pool>>;
+    struct Pools {
+        std::shared_ptr<EncryptionContext> encryption;
+        std::unordered_map<size_t, std::unique_ptr<mongocxx::pool>> instances;
+    };
```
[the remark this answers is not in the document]
erwee: The way I have it structured at the moment requires an associated EncryptionContext per Pool "name". There is no single global `EncryptionContext`, so a static variable will not work.

src/gennylib/src/PoolFactory.cpp
```diff
@@ -192,28 +193,47 @@ std::string PoolFactory::makeUri() const {
...
 
-    // Just doing CAFile and PEMKeyFile for now, it's reasonably trivial to add other options
-    // Note that this is entering as a BSON string view, so you cannot delete the config object
-    auto caFile = *_config->get(OptionType::kAccessOption, "CAFile");
-    if (!caFile.empty()) {
-        BOOST_LOG_TRIVIAL(debug) << "Using CA file '" << caFile << "' for TLS";
-        tlsOptions = tlsOptions.ca_file(caFile.data());
+    if (_config->encryptionCtxt) {
+        BOOST_LOG_TRIVIAL(debug) << "Adding encryption options to pool...";
+        clientOptions.auto_encryption_opts(_config->encryptionCtxt->getAutoEncryptionOptions());
```
aggrand, line 227: (nit) If this is the only use of this object here, would it make sense to pass in the encryption options "raw" to this class? Optional up to you.

src/workloads/docs/CrudActorEncrypted.yml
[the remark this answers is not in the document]
aggrand: Gotcha. Yeah whether that's worth it or not I couldn't say, but maybe someone on product perf would have an idea of whether that'd constrain the tests they run by too much?

If that doesn't get implemented, I'd recommend pointing out explicitly in the docs that it's not supported.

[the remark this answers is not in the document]
erwee: I implemented EncryptionManager anyways since we'll probably going run into this exact situation you described when we're writing our QE workloads. Plus, this meant that I no longer need to touch the Pools maps in PoolManager.

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -21,12 +21,34 @@
...
+    void setupKeyVault();
+
+    bsoncxx::document::value generateKMSProvidersDoc() const;
+    bsoncxx::document::value generateSchemaMapDoc() const;
+    bsoncxx::document::value generateExtraOptionsDoc() const;
+
+private:
+    std::unique_ptr<EncryptionOptions> _encryptionOpts;
+    std::string _uri;
+};
```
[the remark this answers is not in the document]
erwee: Based on the discussion below, I've created an EncryptionManager class that will handle all the global initialization (setting up key vaults, data keys, etc). Then, when each pool sets up its encryption options, it'll just ask for those settings from the EncryptionManager.

src/gennylib/src/PoolManager.cpp
```diff
@@ -38,6 +39,12 @@ auto createPool(const std::string& name,
         poolFactory.setOptions(genny::v1::PoolFactory::kAccessOption, *accessOpts);
     }
 
+    auto encryptOpts = context["Clients"][name]["EncryptionOptions"].maybe<EncryptionOptions>();
+    if (encryptOpts) {
+        poolFactory.setEncryptionContext(
+            encryptionManager.createEncryptionContext(mongoUri, *encryptOpts));
```
aggrand, line 45: Am I correct in understanding that the mongoUri used here doesn't have all the options set that PoolFactory would have set? Do we want it to, in case some of the options spliced in there are needed for the connection, and we could re-path some of this so that the URI doesn't get used until a call inside PoolFactory?

src/gennylib/src/PoolManager.cpp
```diff
@@ -55,6 +62,10 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
     // ...but no need to keep the lock open past this.
     // Two threads trying access client("foo",0) at the same
     // time will subsequently block on the unique_lock.
+
+    if (!_encryptionManager) {
+        _encryptionManager = std::make_unique<EncryptionManager>(context, _dryRun);
+    }
```
aggrand, line 68: (nit) I think the above comment about releasing the lock should be after this encryption manager call

aggrand (changes requested):
Am I correct in understanding that the compile failures are due to the toolchain not being updated yet? If so, we'll want to merge the updates to the toolchain and update genny here to point at it, to get a happy commit queue. (That's my main requested change!)

Otherwise, I can't comment as much on the FLE doc-struct internals, but the integration with Genny at large and the poolmanager is looking nice!

src/gennylib/src/PoolManager.cpp
erwee: fixed

src/gennylib/src/PoolManager.cpp
erwee: I changed it to use the output of `poolFactory.makeUri()` instead.

src/gennylib/src/encryption.cpp
```diff
@@ -453,16 +453,42 @@ class EncryptionManager::EncryptionManagerImpl {
...
-    const auto& collsSequence = yaml["EncryptedCollections"];
+    const auto& encryptionNode = yaml["Encryption"];
+    if (!encryptionNode) {
+        return;
+    }
+
+    auto useShlibOpt = encryptionNode["UseCryptSharedLib"].maybe<bool>();
+    if (useShlibOpt) {
+        _useCryptSharedLib = *useShlibOpt;
+    }
```
aggrand, line 474: `_useCryptSharedLib = encryptionNode["UseCryptSharedLib"].maybe<bool>().value_or(false);` is a bit more idiomatic in Genny for setting defaults

src/gennylib/src/encryption.cpp
```diff
@@ -453,16 +453,42 @@ class EncryptionManager::EncryptionManagerImpl {
...
+    if (_useCryptSharedLib) {
+        auto shlibPathOpt = encryptionNode["CryptSharedLibPath"].maybe<std::string>();
+        if (!shlibPathOpt || shlibPathOpt->empty()) {
```
aggrand, line 477: Might wanna do `boost::filesystem::exists` as a basic validation too

src/workloads/docs/CrudActorEncrypted.yml
```diff
@@ -65,25 +65,28 @@ Description: |
...
-- Database: *encrypted_db
-  Collection: "medical"
-  EncryptionType: "fle"
-  FLEEncryptedFields:
-    patient_id: { type: "string", algorithm: "deterministic" }
-    insurance: { type: "string", algorithm: "random" }
-    conditions: { type: "array", algorithm: "random" }
+Encryption:
+  UseCryptSharedLib: true
+  CryptSharedLibPath: /data/workdir/mongocrypt/lib/mongo_crypt_v1.so
```
aggrand, line 70: Worth noting that hardcoding this in workload yamls seems a bit risky for breaking things for local users, unless they contort their local filesystem to match.

I wonder if it makes sense to encourage using [overrides](https://github.com/mongodb/genny/blob/master/docs/using.md#org22b7a0f) for the path, so only the caller has to know where the file is. Might be a DSI change worth doing either soon or in the future

src/workloads/docs/CrudActorEncrypted.yml
```diff
@@ -0,0 +1,212 @@
+SchemaVersion: 2018-07-01
+Owner: "@10gen/server-security"
+Description: |
```
aggrand, line 3: Since this is a core genny feature change, not just a workload-specific one, I'd recommend putting some docs for this in the [using](https://github.com/mongodb/genny/blob/master/docs/using.md) docs, and linking to this example workload.

aggrand (approved):
Nice, I think my main concerns are alleviated! I left a few more notes, but it looks good!

I'm happy to merge as long as the sys-perf patch is green, PR tests are green, and the perf implications are understood (which seemed like a yes based on our last convo?)

src/gennylib/src/encryption.cpp
erwee: libmongocrypt already throws a reasonable error message when it fails to `dlopen` this file name on client pool creation, so I opted not to do a redundant check here.

src/workloads/docs/CrudActorEncrypted.yml
erwee: Good point. I will do that as a follow-up ticket.

src/gennylib/src/encryption.cpp
erwee: much better. thanks!

src/workloads/docs/CrudActorEncrypted.yml
erwee: I commented these out to reduce the risk of breakage. I will explore how these can be overridden in sys-perf via the feature you mentioned.

erwee:
patch running CrudActorEncrypted workload with shared library: https://spruce.mongodb.com/version/631ba47b3e8e864cc3145df3/tasks

aggrand (approved):
Nice, the genny integration looks good! I'm comfortable merging if the perf implications for the toolchain update seem good for existing workloads!

jimoleary (approved):
LGTM since the  patch looks good.

PR #728, opened 2022-08-19
SERVER-26442 Add inclusion projection test to ProjectSort

TedTuckman:
I know changing the names of the test is unfortunate for burn in, but figure its better in the long run.


jimoleary (approved):
I'm good with changing names  there isn't a large population of historical data in this case.

Can you update the documentation as to what expected difference  in performance should be between the inclusive and exclusive versions

PR #730, opened 2022-08-23
Create an actor to run external script

allenwux:


src/cast_core/include/cast_core/actors/ExternalScriptRunner.hpp
```diff
@@ -0,0 +1,133 @@
...
+    // the `genny list-actors` command. Typically this should be the same as the
+    // class name.
+    //
+    static std::string_view defaultName() {
+        return "ExternalScriptRunner";
+    }
+
+private:
+
+    //
```
rtimmons, line 94: Remove template text

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include <iostream>
+#include <string>
+#include <cstdlib>
+#include <boost/filesystem.hpp>
+#include <cstdio>
+#include <stdexcept>
+#include <array>
+#include <stdio.h>
+#include <unistd.h>
```
rtimmons, line 44: ensure these are all needed
and run `clang-format` which will (I think) organize the imports.

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    // Record data retruned by the script
+    metrics::Operation operation;
+
+    // Record data of the script operation itself
+    metrics::Operation scriptOperation;
+
+    PhaseConfig(PhaseContext& phaseContext, ActorId id)
+        : script{phaseContext["Script"].to<std::string>()},
+          mongoServerURI{phaseContext["MongoServerURI"].maybe<std::string>().value_or("--nodb")},
```
rtimmons, line 61: Let's add a TODO to pipe through teh mongo URI from DSI (or we pass it in already to the genny binary - maybe access the actorContext and see if we can get the URI from the pool or the upper-level Node/yaml or something?)

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+          operation{phaseContext.operation("DefaultMetricsName", id)},
+          scriptOperation{phaseContext.namedOperation("ExternalScript", id)} {}
```
rtimmons, line 63: Let's just document what these metrics objects represent - here and/or in the docs yaml.

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * @brief Execute external script
+ * 
+ * @param cmd 
+ * @return std::string 
+ */
+std::string ExternalScriptRunner::exec(const char* cmd) {
+    std::array<char, 128> buffer;
+    std::string result;
+    std::unique_ptr<FILE, decltype(&pclose)> pipe(popen(cmd, "r"), pclose);
```
rtimmons, line 75: Nice!

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+            programCommand = "mongosh " + config->mongoServerURI + " --quiet --eval \"" + config->script + "\"";
+        }
+        else if (_command == "sh") {
+            programCommand = "sh -c \"" + config->script + "\"";
+        }
+        else {
+            throw std::runtime_error("Script type " + _command + " is not supported.");
+        }
+        
+        BOOST_LOG_TRIVIAL(info) << "ExternalScriptRunner running "
```
rtimmons, line 103: This is probably `debug` level

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+                                << programCommand;
+
+        for (const auto&& _ : config) {
+
+            // Execute the script and read result from stdout
+            auto ctx = config->scriptOperation.start();
+            const char* programCmdPtr = &*programCommand.begin();
+            std::string result = exec(programCmdPtr);
```
rtimmons, line 111: I think you can use `static_cast<char*>(std::string)` - if you wanna leave it as-is no biggie but let's add a comment that this is to convert a std::string to a char*, because it looks kinda funny at first glance

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+            ctx.success();
```
rtimmons, line 112: 1. if the output from the script isn't numeric, it's probably `ctx.fail()`?  Or add comment that non-numeric output is ignored.
2. what if program has a non-zero exit code?

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+            try {
+
+                // If the result from stdout can be parsed as integer, report the operation metrics
+                // Otherwise, ignore the result
+                int num = std::stoi(result);
+                config->operation.report(metrics::clock::now(),
+                                              std::chrono::milliseconds{num}); 
+            }
+            catch(std::exception& err) {
+                BOOST_LOG_TRIVIAL(info) << err.what();
```
rtimmons, line 123: This is an expected situation, so maybe `debug` level and/or consider having the command itself be a prefix of the `what` - but only maybe - can lead to log spam if not judicious.

src/gennylib/include/gennylib/context.hpp
```diff
@@ -544,6 +544,16 @@ class PhaseContext final : public v1::HasNode {
             internal);
     }
 
+    auto namedOperation(const std::string& metricsName, ActorId id, bool internal = false) const {
```
rtimmons, line 547: Let's mark this as experimental in a comment or at least write a good doc about why this was necessary; I think this API is covering up a shortcoming in the original context APIs.

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+    }
+}
+
+ExternalScriptRunner::ExternalScriptRunner(genny::ActorContext& context)
+    // These are the attributes for the actor.
+    : Actor{context},
+      _command{std::move(context.get("Command").maybe<std::string>().value_or(""))},
```
rtimmons, line 132: This command might make more sense in the `PhaseConfig` so you can access `config->command` (`config` is your `PhaseConfig`). See also other comment about constructing the command in there as well.

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -0,0 +1,138 @@
...
+void ExternalScriptRunner::run() {
+    // This section gets executed before the phases. Setup is executed here, so maybe
+    // we could add some sanity checks before actually running the shell.
+
+    for (auto&& config : _loop) {
+
+        std::string programCommand = "";
+        
+        if (_command == "mongosh") {
+            programCommand = "mongosh " + config->mongoServerURI + " --quiet --eval \"" + config->script + "\"";
```
rtimmons, line 94: 1. Assert at construction time that there are no quotes or other shell-scary things (not sure of best strategy there)
2. You can compute `programCommand` in the Actor's constructor (or in the PhaseContext's constructor) to avoid having to do this same calculation for each loop, plus that

rtimmons (changes requested):
Nice, simple approach. A few edge-cases to consider but nothing that would stop us from giving this a go-ahead and seeing where we want to take it next.

Only real blockers for me are:

1. remove the template text (mostly comments)
2. run `clang-format` on the new files (we don't have a standard clang-format version unfortunately so to prevent spacing wars just run it on your new files and we'll handle the consistency issue at some other point)

Glad to see this going in. I'm optimistic we can get stdin/stdiout and more fine-grained flow-control and configuration and output-processing in as we need things to be more specific.

src/cast_core/src/ExternalScriptRunner.cpp
allenwux: If the output isn't numeric, we will write a message to debug instead of fail the operation.

src/gennylib/include/gennylib/context.hpp
allenwux: [TIG-4328](https://jira.mongodb.org/browse/TIG-4328)

src/cast_core/src/ExternalScriptRunner.cpp
allenwux: const_cast<char*>(config->programCommand.c_str())

src/cast_core/test/ExternalScriptRunnerTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace genny {
+namespace {
+using namespace genny::testing;
+namespace bson_stream = bsoncxx::builder::stream;
+
+//
+// ⚠️ There is a "known" failure that you should find and fix as a bit of
+// an exercise in reading and testing your Actor. ⚠️
+//
```
rtimmons, line 36: Kill this :)

src/gennylib/include/gennylib/context.hpp
```diff
@@ -544,6 +544,29 @@ class PhaseContext final : public v1::HasNode {
             internal);
     }
 
+    /**
+     * @brief [Experimental] Create a metrics::Operation with the specified name
+     * 
+     * The current operation() function will always return the same operation if MetricsName is specified. 
+     * According to the HelloWorld actor, it seems not the expected behavior
+     * But it is too risky to change the behavior without thoroughly exam the impact
+     * Created TIG-4328 to track this work
```
rtimmons, line 553: 👍

rtimmons (dismissed):
LGTM! This could be the start of something really cool in Genny-land.

rtimmons (approved):

PR #732, opened 2022-08-29
PERF-3187 - Extend Genny to add queryable encryption support

erwee:
This patch builds on top of this patch: https://github.com/mongodb/genny/pull/726 to add QE support.

### DIFF
--- a/src/gennylib/src/encryption.cpp
+++ b/src/gennylib/src/encryption.cpp
@@ -92,6 +92,23 @@ private:
     std::string _algorithm;
 };
 
+class QueryableEncryptedField : public EncryptedField {
+public:
+    inline static const std::string kParentNodeName = "QueryableEncryptedFields";
+
+    QueryableEncryptedField(const Node& yaml);
+
+    void appendEncryptInfo(bsoncxx::builder::basic::sub_document subdoc) const override;
+
+private:
+    struct Query {
+        Query(const Node& yaml);
+        std::string queryType;
+        std::optional<int> contention;
+    };
+    std::vector<Query> _queries;
+};
+
 template <EncryptionType enctype, class FieldType>
 class EncryptedCollection {
 public:
@@ -176,6 +193,16 @@ public:
     FLEEncryptedCollection(const Node& yaml)
         : EncryptedCollection<EncryptionType::FLE, FLEEncryptedField>(yaml) {}
     void appendSchema(bsoncxx::builder::basic::sub_document builder) const;
+    void createCollection(const mongocxx::client& client) const override;
+    void dropCollection(const mongocxx::client& client) const override;
+};
+
+class QueryableEncryptedCollection
+    : public EncryptedCollection<EncryptionType::QUERYABLE, QueryableEncryptedField> {
+public:
+    QueryableEncryptedCollection(const Node& yaml)
+        : EncryptedCollection<EncryptionType::QUERYABLE, QueryableEncryptedField>(yaml) {}
+    void appendEncryptedFieldConfig(bsoncxx::builder::basic::sub_document builder) const;
     void createCollection(const mongocxx::client& client) const override;
     void dropCollection(const mongocxx::client& client) const override;
 };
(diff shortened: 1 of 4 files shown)


erwee:
patch test for CrudActorEncrypted workload w/ QE: https://spruce.mongodb.com/version/631f74a1d6d80a6ca5980abc/tasks

src/gennylib/src/encryption.cpp
```diff
@@ -305,6 +332,55 @@ void FLEEncryptedField::appendEncryptInfo(sub_document subdoc) const {
     }
 }
 
+QueryableEncryptedField::Query::Query(const Node& yaml) {
+    queryType = yaml["queryType"].to<std::string>();
+    contention = yaml["contention"].maybe<int>();
```
jimoleary, line 337: Should the default be explicit here?

src/gennylib/src/encryption.cpp
```diff
@@ -305,6 +332,55 @@ void FLEEncryptedField::appendEncryptInfo(sub_document subdoc) const {
...
+    if (queriesNode.isSequence()) {
+        for (auto& [_, queryNode] : queriesNode) {
+            if (!queryNode.isMap()) {
+                throw InvalidConfigurationException(
+                    "Each value in the 'queries' array must be of map type");
+            }
+            _queries.emplace_back(queryNode);
+        }
+    } else if (queriesNode.isMap()) {
+        _queries.emplace_back(queriesNode);
```
jimoleary, line 355: There doesn't appear to be a valid test for this.

src/gennylib/src/encryption.cpp
```diff
@@ -357,6 +433,33 @@ void FLEEncryptedCollection::dropCollection(const mongocxx::client& client) cons
...
+void QueryableEncryptedCollection::createCollection(const mongocxx::client& client) const {
+    auto createOpts = make_document(
+        kvp("encryptedFields", [&](sub_document subdoc) { appendEncryptedFieldConfig(subdoc); }));
+    BOOST_LOG_TRIVIAL(debug) << "Generated encryptedFieldConfig: "
+                             << bsoncxx::to_json(createOpts.view());
+    client[_database].create_collection(_collection, createOpts.view());
+}
+
+void QueryableEncryptedCollection::dropCollection(const mongocxx::client& client) const {
+    for (auto& suffix : {".esc", ".ecc", ".ecoc"}) {
```
jimoleary, line 456: Shouldn't mongodb drop these automatically? It looks like it creates [them](https://www.mongodb.com/docs/upcoming/core/queryable-encryption/fundamentals/manage-collections/#metadata-collections)

jimoleary (commented):
It broadly looks good, I have a few questions

src/gennylib/src/encryption.cpp
erwee: The current CXX driver does not have full support for QE, and automatic dropping of the auxilliary collections is one of the QE features that it lacks at the moment.

src/gennylib/src/encryption.cpp
erwee: I do not want to hard-code an explicit contention default value here as it's probably going to change again in the near future.

src/gennylib/src/encryption.cpp
jimoleary: Can you create a ticket to remove this code later and add it as a comment?

src/gennylib/src/encryption.cpp
jimoleary, line 337: What value do we get when it's empty, 0?

erwee: `contention` is a std::optional, so if there's no value we don't include the "contention" field in the schema. Server will interpret a missing "contention" field as the implicit default value of 4.

src/gennylib/src/encryption.cpp
erwee: done. see PERF-3306

src/gennylib/src/encryption.cpp
erwee: I added a test for `EncryptionContext::generateEncryptedFieldsMapDoc()`, which should cover the valid cases for this Node.

src/gennylib/src/encryption.cpp
erwee: I was wondering if there is a way to simply convert the yaml `Node` into JSON, and simply use `bsoncxx::from_json()` when appending it to the generated schema.  Since we're going to add more fields to the `query` node once QE range and substring search is implemented, it'll be nice to not have to parse and validate each field, only to simply serialize it back.  Also, this would allow for usage of extended JSON for complex types like decimal128. For instance, with "range" queryType, a queries node could look something like this:
```
queries: [ { queryType: "range", min: { $numberDecimal: "3.14" }, max: { $numberDecimal: "12.9" }, sparsity: 3 } ]
```

jimoleary: That might be something to try, I think the testlib code does something similar [here](https://github.com/mongodb/genny/blob/master/src/testlib/include/testlib/helpers.hpp#L42-L46).

My only concern would be that the syntax visible to the user must remain the same.

src/gennylib/src/encryption.cpp
jimoleary: :+1: Thanks.

jimoleary (dismissed):
LGTM from me as is, maybe look and see if the queries processed as json works.

jimoleary (approved):
still lgtm

aggrand (approved):

PR #734, opened 2022-09-13
PERF-3302 Add benchmarks for filters with complex expressions

laplab:


src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
...
+- Name: RunWideQuery
+  Type: CrudActor
+  Database: *Database
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 10
```
allenwux, line 60: does 10 times give us stable enough perf result?

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
```
allenwux, line 3: Can you add keywords as well?

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
...
+      PhaseConfig:
+        Repeat: 10
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {
+              $expr: {
+                $or: [
+                  {$eq: ["$a", 10]},
```
allenwux, line 68: Do we want to test dot path here as well?

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This workload stresses the query execution engine by running queries with complex logical
+  expressions that never match a document in the collection.
+
```
ghartnett, line 6: Did you fill the form for [new workload requests](https://wiki.corp.mongodb.com/display/STAR/New+Workload+Requests)?

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
...
+- Name: RunMixedQuery
+  Type: CrudActor
+  Database: *Database
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [4]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 10
```
ghartnett, line 146: Similar to @allenwux comment earlier 1 thread and 10 repetitions seems small. Did you schedule a number of patch-builds with identical configuration and git revision, then analyze the distribution of results? If so, did you see low result variability?

src/workloads/execution/FilterWithLogicalExpressions.yml
dstorch: +1

I don't know if it would go into this Genny workload or not, but I'd love to see two more test cases:

- A test with a deep field path expression, like "$a.b.c.d.e.f.g".
- A test that combines a wide logical expression with deep field paths, e.g. an $or with 10 clauses where each clause has a deep field path expression.

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
+SchemaVersion: 2018-07-01
```
dstorch, line 1: This only covers expressions, which makes sense. Is your intention to add workloads to cover complex `MatchExpression`s as a separate task? That would be fine with me, I just want to make sure I understand your plan.

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,162 @@
...
+                  {$eq: ["$a", 10]},
+                  {$eq: ["$b", 11]},
+                  {$eq: ["$c", 12]},
+                  {$eq: ["$d", 13]},
+                  {$eq: ["$a", 14]},
+                  {$eq: ["$b", 15]},
+                  {$eq: ["$c", 16]},
+                  {$eq: ["$d", 17]},
+                  {$eq: ["$a", 18]},
+                  {$eq: ["$b", 19]},
```
dstorch, line 77: I see that these only test equality. I _think_ that's ok, since it would be overkill to try to have complete test matrix where we have wide/deep logical expressions involving inequalities or any other leaf expression we can think of. Presumably other expressions would benefit from a similar optimization strategy and we will implement these optimizations in a general way that would apply to $lt, $gt, etc. But we just won't be able to reasonably add perf coverage for such cases explicitly. Does this match your understanding? Are there any particularly important expressions other than `$eq` that you think we should benchmark?

dstorch (changes requested):
I like it! A few minor thoughts/questions.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Done

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Number 10 was chosen to speedup the local debugging. This is the first time I hear about the requirement to perform distribution analysis through a number of Evergreen runs. It makes sense, but I would prefer to do this tuning when we settle on the types of queries this workload should include.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: I am not sure. Will run a couple of Evergreen patches to figure out these numbers once we settle on the queries this workload should include.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: I am not sure I added the correct keywords, but I tried to follow how other workloads did it.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Initially, I was thinking it will be a separate workload in a separate PR. But now I am thinking that we should just go ahead and add it here, since the workloads themselves are quite compact. Added find alternative for each of the queries.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Replied in Slack.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: I think that adding a wide query with deep fieldpaths is a good idea. I am not sure about the first case, though. 7-depth fieldpath traversal seems too synthetic. We can add a query with a lower number of fields (something like 3), but I think we already have queries like this in the `*Subpath*` microbenchmarks.

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -3,12 +3,30 @@ Owner: "@mongodb/query"
 Description: >
   This workload stresses the query execution engine by running queries with complex logical
   expressions that never match a document in the collection.
+  Each workload name consists of several parts: '{SyntaxType}{PredicateType}'.
+  'SyntaxType' can be:
+    - 'AggregationExpressions' means expressions which can be used inside $expr
+    - 'MatchExpressions' means operators of the find command language
```
dstorch, line 9:
```suggestion
    - 'AggregationExpression' means expressions which can be used inside $expr
    - 'MatchExpression' means operators of the find command language
```

This is a total nitpick, but if you make these names singular rather than plural the names used below will all roll off the tongue better. For example, `AggregationExpressionWidePredicate` sounds better to me than `AggregationExpressionsWidePredicate`.

src/workloads/execution/FilterWithLogicalExpressions.yml
dstorch: Sounds good to me! Resolving.

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -0,0 +1,375 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
```
dstorch, line 2: Should we rename this file `FilterWithComplexLogicalExpression.yml`? That might help us understand what this workload is about in the future, since it's specifically about deep/wide expressions. Another alternative is to call it `FilterWithDeepOrWideLogicalExpression.yml` but that's a bit wordier.

src/workloads/execution/FilterWithLogicalExpressions.yml
```diff
@@ -160,3 +336,40 @@ Actors:
...
+                {c: 16},
+                {d: 17},
+              ]
+            }
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - replica-all-feature-flags
```
dstorch, line 369: Ah, this isn't quite right. My recommendation would be to run it in four configurations, though I'm also curious what people from the perf team think:

- vanilla standalone
- standalone, all feature flags
- standalone, SBE
- standalone, classic

The first two will give us value going forward. The SBE and classic variants are necessary because we are currently focusing on this comparison for the SBE perf project.

src/workloads/execution/FilterWithLogicalExpressions.yml
dstorch: The Slack conversation suggests that we are on the same page. Do you want to copy any of the context here for posterity before resolving this thread?

src/workloads/execution/FilterWithLogicalExpressions.yml
dstorch: > 7-depth fieldpath traversal seems too synthetic.

These are all highly synthetic. If we have a performance problem which manifests specifically due to how we construct plans for deep paths, it seems worth at least knowing about. If we had data saying that customers almost never use field paths deeper than 3, then I could see an argument for omitting this. But since we don't actually know how deep field paths tend to go it seems worthwhile to have just one benchmark stressing this dimension.

Actually, it would be two additional benchmarks: One for a deep field path expression in the agg expression language and a corresponding test for the match expression language.

src/workloads/execution/FilterWithLogicalExpressions.yml
ghartnett: Thanks

src/workloads/execution/FilterWithLogicalExpressions.yml
ghartnett: lgtm

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: In the end, we decided to limit the benchmarks to the $eq expressions and derive some more optimizations for the type checker from analyzing common patterns in the stage builders.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Sounds good, done.

src/workloads/execution/FilterWithLogicalExpressions.yml
laplab: Sounds good. I have modified the "wide with deep fieldpaths" to use 7-depth fieldpaths and added a query with a single equality predicate with deep fieldpaths. I suggest to add other cases related to scalability along various axis in future pull requests.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -366,7 +430,10 @@ AutoRun:
 - When:
     mongodb_setup:
       $eq:
-      - replica-all-feature-flags
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - stadalone-sbe
```
dstorch, line 436:
```suggestion
      - linux-standalone-classic-query-engine
      - linux-standalone-sbe
```

You definitely have a typo in the last line, but I _think_ my edits are right in order to make this run in sys-perf. Unfortunately we cannot add new benchmarks to `performance` with Genny right now, only to `sys-perf` due to [BUILD-14593](https://jira.mongodb.org/browse/BUILD-14593). So these should run on the sys-perf standalone variants for SBE and classic.

I suggest that before merging this you run a sys-perf patch build in order to verify that if you run `schedule_patch_auto_tasks` and `schedule_variant_auto_tasks` that these benchmarks actually end up running in the expected build variants.

src/workloads/execution/FilterWithLogicalExpressions.yml
dstorch: I left a new comment about this, but I'll resolve this one to keep the comment threads cleaner.

dstorch (approved):
Thanks @laplab! LGTM with a request to fix the `AutoRun` section.

dstorch:
Oh, also to make this start running in our dashboard environment, you'll also have to name the task explicitly in a few config files:

- https://github.com/10gen/PerformanceReports/blob/main/scripts/perf_reports_sbe_perf.yml
- https://github.com/10gen/PerformanceReports/blob/main/scripts/tags.yml

This can be done as a followup change.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
laplab: Oh no, it is that "stadalone" typo catching up on me! But other than that, the current version seems to schedule all correct variants, so I plan to move forward with it.

laplab:
One last [Evergreen](https://spruce.mongodb.com/version/633572259ccd4e3024700726/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) to ensure that the AutoRun configuration was fixed.

dstorch (approved):

PR #737, opened 2022-09-27
PERF-3049 Add workload for Background TTL deletions

jordiolivares:
Adding @dalyd to see if this workload makes sense to test the impact of TTL deletions

### DIFF
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -1,7 +1,12 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/server-execution"
 Description: |
-  This workload tests the impact of background TTL deletions in a heavily modified collection
+  This workload tests the impact of background TTL deletions in a heavily modified collection. This
+  test does not quiesce between phases as we want the TTL deleter to be constantly running.
+
+  Previously this test was part of a legacy suite of workloads used to test called `insert_ttl.js`.
+  This is a port of the test in order to have better analytics with some minor changes to be
+  more correct.
 Keywords:
 - ttl
 - stress
@@ -13,7 +18,7 @@ Keywords:
 Clients:
   Default:
     QueryOptions:
-      maxPoolSize: 260
+      maxPoolSize: 256
 
 NumPhases: &NumPhases 20
 CollectionCount: &NumColls 1
@@ -134,6 +139,8 @@ Actors:
         - keys: {expireDate: 1}
           options: {expireAfterSeconds: 0}
 
+# We do not perform a quiesce between phases as it would allow the TTL deleter to catch up. We want
+# it to constantly be running for this workload.
 - ActorFromTemplate:
     TemplateName: WritesActorTemplate
     TemplateParameters:
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -1,0 +1,124 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
+  This workload tests the impact of background TTL deletions in a heavily modified collection
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 128
+
+Actors:
+- Name: Setup
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+    - Repeat: 1
+      Thread: 1
+      Database: admin
+      Operations:
+      - OperationName: AdminCommand
+        OperationCommand:
+          setParameter: 1
+          ttlMonitorSleepSecs: 5
+      - OperationName: AdminCommand
+        OperationCommand:
+          setParameter: 1
+          storageEngineConcurrentWriteTransactions: 32
+    - &Nop {Nop: true}
+    - *Nop
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 50000
+    BatchSize: 10000
+    Document: &generatedDoc
+      expireDate: {^Now: {}} # Instantly expire the document
+      y: &y {^RandomInt: {min: -1000, max: 1000}}
+      b: {^RandomString: {length: 15}}
+    Indexes:
+    - keys: {expireDate: 1}
+      options: {expireAfterSeconds: 0}
+  - *Nop
+
+- Name: Writes
+  Type: CrudActor
+  Database: *db
+  Threads: 96
+  Phases:
+  - *Nop
+  - *Nop
+  - Duration: 5 minutes
+    Collection: Collection0
+    Operations:
+    - OperationName: insertMany
+      OperationCommand:
+        Documents: # Insert 50 documents at a time
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+        - *generatedDoc
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - replica
+      - replica-all-feature-flags
+      - shard-lite
+      - shard-lite-all-feature-flags
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -20,7 +20,7 @@ Actors:
       - OperationName: AdminCommand
         OperationCommand:
           setParameter: 1
-          ttlMonitorSleepSecs: 5
+          ttlMonitorSleepSecs: 1
       - OperationName: AdminCommand
         OperationCommand:
           setParameter: 1
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -126,8 +126,8 @@ Actors:
         Database: *db
         Threads: 1
         CollectionCount: *NumColls
-        DocumentCount: 200000  # Add an initial set of documents. This will reduce noise in the next phases
-        BatchSize: 10000
+        DocumentCount: 200_000  # Add an initial set of documents. This will reduce noise in the next phases
+        BatchSize: 10_000
         Document: *generatedDoc
         Indexes:
         - keys: {expireDate: 1}
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -173,5 +173,3 @@ AutoRun:
       - standalone-all-feature-flags
       - replica
       - replica-all-feature-flags
-      - shard-lite
-      - shard-lite-all-feature-flags
--- a/src/workloads/execution/BackgroundTTLDeletions.yml
+++ b/src/workloads/execution/BackgroundTTLDeletions.yml
@@ -12,7 +12,7 @@ NumPhases: &NumPhases 20
 CollectionCount: &NumColls 1
 
 Document: &generatedDoc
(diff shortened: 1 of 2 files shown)


haleyConnelly:
@jordiolivares can you post a comment with a summary of the perf results please?

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
+  This workload tests the impact of background TTL deletions in a heavily modified collection
+
```
haleyConnelly, line 5: Brainstorming here --- Do we think it would be valuable to test the write throughput when there are writes on a secondary, non-ttl collection as well? 

The TTLMontior will be acquiring locks for the specific TTL collection. Since ticket contention is handled at the GlobalLock level, maybe we want to test write throughput on a non-TTL collection as well? 

I'm not sure if we would want to do that in this test, or another test completely/ if it is possible to test this reliably given the TTLMonitors sleeping nature.

haleyConnelly (commented):
Are we planning to replace the [legacy workload](http://sys-perf-workloads.s3-website-us-east-1.amazonaws.com/module-workloads_insert_ttl.html) with this?

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
...
+    TemplateName: WritesActorTemplate
+    TemplateParameters:
+      Name: Writes_32_Thread
+      Threads: 32
+      ActivePhases: [8]
+- ActorFromTemplate:
+    TemplateName: WritesActorTemplate
+    TemplateParameters:
+      Name: Writes_64_Thread
+      Threads: 64
```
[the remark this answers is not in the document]
jordiolivares: The uncontended case still has value as it measures the overhead of the TTL background deleter and provides datapoints for something we didn't really measure before. The purpose of this was to address @ykangpark comment regarding SERVER-63040:

> Given that after the change insert throughput for 16 and 32 threads are almost similar, would it be interesting to add additional thread levels, such as [2,4,8]?

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: I'm not sure it would matter that much. Both TTL and inserts acquire a MODE_IX lock on the collection so they will functionally do the same regardless of having a TTL or not. The only thing that would serve is to observe the overhead of modifying the same collection too much in WiredTiger.

jordiolivares:
@haleyConnelly A simple perf run of this workload between baseline and having TTL marked as low priority seems to show a remarkable improvement: 60% latency reduction in 50th percentile, no regression in tail latency nor throughput with a 3 node Replica Set when there are 128 writing threads.

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
```
dalyd, line 3: Could you add a keywords list also? See https://github.com/mongodb/genny/blob/73acba4a6866587e78f32d7b0261e9a1ca6f04c0/docs/using.md#L153


 For example https://github.com/mongodb/genny/blob/73acba4a6866587e78f32d7b0261e9a1ca6f04c0/src/workloads/scale/BigUpdate.yml#L9

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
...
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            ttlMonitorSleepSecs: 1
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            storageEngineConcurrentWriteTransactions: 32
+        SleepAfter: 60 seconds  # Wait 60 seconds so that the new ttl sleep config is active.
```
dalyd, line 109: Would a quiesce stage be sufficient? A minute seems generous. Alternatively is there something we can check?

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
...
+        Indexes:
+        - keys: {expireDate: 1}
+          options: {expireAfterSeconds: 0}
+
+- ActorFromTemplate:
+    TemplateName: WritesActorTemplate
+    TemplateParameters:
+      Name: Writes_1_Thread
+      Threads: 1
+      ActivePhases: [3]
```
dalyd, line 135: I assume we actively do not want to quiesce between phases here, because that would allow the ttl process to catch up and keep working. Is that correct? If it is, it's worth adding a comment somewhere in this file to that effect.

src/workloads/execution/BackgroundTTLDeletions.yml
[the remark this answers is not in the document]
dalyd: I'd encourage testing with a lot, but reducing the total number of configs before committing and running in regression.

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,187 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
+  This workload tests the impact of background TTL deletions in a heavily modified collection
```
dalyd, line 4: Could you add in a reference to the legacy workload and how this relates? It looks pretty close to a straight port, with only minor differences.

src/workloads/execution/BackgroundTTLDeletions.yml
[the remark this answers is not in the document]
jordiolivares: Maybe I can reduce the number to [1, 8, 32, 128]?

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: Done in the latest commit

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: You're right in that we want the TTL deleter to be working constantly for this test. I've added a comment explaining this behaviour.

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: I'm not sure the TTL config change would get picked up by the quiesce stage. Internally the TTL performs a [sleep](https://github.com/mongodb/mongo/blob/c0953d92e0a07dd130a0f4a4ac26a9fc4d7ef7f7/src/mongo/db/ttl.cpp#L271) for the ttl sleep config amount of seconds, so it will initially sleep for 60 seconds (default value) before changing to the new config.

jordiolivares: As an FYI, I've created [SERVER-70055](https://jira.mongodb.org/browse/SERVER-70055) to fix this eventually.

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,177 @@
...
+          OperationCommand:
+            setParameter: 1
+            ttlMonitorSleepSecs: 1
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            storageEngineConcurrentWriteTransactions: 32
+        SleepAfter: 60 seconds  # Wait 60 seconds so that the new ttl sleep config is active.
+
+- Name: InsertData
```
ykangpark, line 123: What is the reason for having this loading phase before the actual insert phases below?

jordiolivares: Sets up the TTL index mostly

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,177 @@
...
+      ActivePhases: [10]
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - replica
+      - replica-all-feature-flags
+      - shard-lite
```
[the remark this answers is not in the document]
jordiolivares: I sort of added it in automatically. On second thought I don't think testing it on a sharded cluster will provide meaningful data so I'll remove it

src/workloads/execution/BackgroundTTLDeletions.yml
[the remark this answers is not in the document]
jordiolivares: I reduced it to the threads given above

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,175 @@
...
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Duration: 3 minutes
+          CollectionCount: *NumColls
+          Operations:
+          - OperationName: insertMany
+            OperationCommand:
+              Documents:  # Insert 50 documents at a time
```
louiswilliams, line 48: 😳

src/workloads/execution/BackgroundTTLDeletions.yml
dalyd: Thanks.

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -0,0 +1,177 @@
...
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 256
+
+NumPhases: &NumPhases 20
+CollectionCount: &NumColls 1
+
+Document: &generatedDoc
```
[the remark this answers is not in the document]
dalyd: I'm curious where this part goes. I suspect the 1024 was fairly arbitrary at the time, to make sure the documents weren't too small. 

For reference, original workload added in https://jira.mongodb.org/browse/PERF-219. The 1024 may have been copied from the insert_capped workload. 

Would larger or smaller docs be more stressful for TTL deletion?

dalyd (approved):
Thanks for the changes. I'm interested to see where the document size discussion goes, but you should have better insight into the correct thing than me for that one.

haleyConnelly:
Update: We've had some discussion offline about the inherent noise for this workload. We are concerned that in between phases, whether the TTL Monitor catches up to the documents it needs to delete may have an impact on performance. 

Until we come to a decision, we may want to 
(1) hold off on publishing this workload
(2) only publish the non-contentious workload that recreates the legacy performance workload

ykangpark:
> Update: We've had some discussion offline about the inherent noise for this workload. We are concerned that in between phases, whether the TTL Monitor catches up to the documents it needs to delete may have an impact on performance.
> 
> Until we come to a decision, we may want to (1) hold off on publishing this workload (2) only publish the non-contentious workload that recreates the legacy performance workload

Just a question, the legacy workload drops and recreates the collection and ttl index for each thread level. Why can't we do the same here? Or does that not matter? Alternatively, you may make a template workload and split into different workloads for each thread level that will start in the same conditions.

jordiolivares:
@ykangpark The reason why we don't want to do that is that it introduces more noise to what we want to measure. What we want is "How does having concurrent TTL deletes impact new inserts". If we were to recreate the collection every time we risk introducing noise with a "ramp-up" of the collection before we have constant TTL deletions.

haleyConnelly:
I vote we create a ticket to signal wake TTLMonitor which wakes the monitor if it is sleeping.

In the meanwhile, do we think the noise here is acceptable for the standard workload (one ported from legacy)?

I'm okay holding off on the contentious workload until we get the command and can reduce noise. 

What do we think?

dalyd:
> @ykangpark The reason why we don't want to do that is that it introduces more noise to what we want to measure. What we want is "How does having concurrent TTL deletes impact new inserts". If we were to recreate the collection every time we risk introducing noise with a "ramp-up" of the collection before we have constant TTL deletions.

Another possibility would be to split each test into two parts, a warmup and a steady state portion, and document that we only care about the steady state portion.

haleyConnelly:
@dalyd since this repository is public, feel free to reach out to Jordi or me offline for patch builds. We discovered [SERVER-70391](https://jira.mongodb.org/browse/SERVER-70391) is likely the culprit for some of the noise. 

I'd say we should hold off on publishing this workload until SERVER-70391 is committed.

dalyd:
> SERVER-70391

Oh, the test would show the change with SERVER-70391, yes? I'd push to turn on the test first so we can see the improvement in the trend graphs. That sounds super cool.

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -13,20 +13,18 @@ Keywords:
...
-CollectionCount: &NumColls 1
+NumPhases: &NumPhases 6
 
 Document: &generatedDoc
-  expireDate: {^Now: {}}  # Instantly expire the document
-  y: &y {^RandomInt: {min: -1000, max: 1000}}
-  b: {^RandomString: {length: 15}}
+  expireDate: {^Date: "1970-01-01"}  # Instantly expire the document
+  y: 999
+  b: "Arbitrary string with 45 different characters"
```
dalyd, line 27: Does it matter that this string likely compresses very well since it's the same string over and over again?

src/workloads/execution/BackgroundTTLDeletions.yml
```diff
@@ -132,7 +126,7 @@ Actors:
         Database: *db
         Threads: 1
         CollectionCount: *NumColls
-        DocumentCount: 50000
+        DocumentCount: 200000  # Add an initial set of documents. This will reduce noise in the next phases
```
dalyd, line 129: Consider using 200_000 for readability. It should just work, although I am also failing to find an example right now. Scientific notation works also 2e5.

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: It's just filler for the document so that there are more bytes to delete. It doesn't really matter that much.

src/workloads/execution/BackgroundTTLDeletions.yml
jordiolivares: Didn't know you could do that in YAML!

jordiolivares:
Going to push this workload before SERVER-70391 so that we have some data collection going

haleyConnelly (approved):

PR #739, opened 2022-09-28
SERVER-58712: Add UpdateWithSecondaryIndexes workload

Fefer-Ivan:
Add UpdateWithSecondaryIndexes workload:
```
  Updates a large range of documents in the collection.
  Multiple secondary indexes are present.
  Update performed with and without a hint.
```
Fixed a bug that prevented update option from actually affecting updateMany operations.

### DIFF
--- a/src/phases/execution/UpdateWithSecondaryIndexes.yml
+++ b/src/phases/execution/UpdateWithSecondaryIndexes.yml
@@ -1,0 +1,116 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: >
+  Runs a workload that updates a large range of documents.
+
+Keywords:
+- RunCommand
+- Loader
+- LoggingActor
+- CrudActor
+- insert
+- update
+- latency
+
+GlobalDefaults:
+  # Configurable parameters.
+  DocumentCount: &DocumentCount {^Parameter: {Name: "DocumentCount", Default: 50_000}}
+  # Insert same document for both default and with secondary indexes to remove one variable.
+  InsertDocument: &DocumentToInsert
+    a: "foo"
+    b: &rs {^FastRandomString: {length: 6, alphabet: "0123456789"}}
+    c: *rs
+
+# Constants.
+WriteRate: &WriteRate 1 per 500 microseconds  # 2000/second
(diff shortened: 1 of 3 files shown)


rtimmons (approved):
Thanks for fixing the bug!

src/phases/execution/UpdateWithSecondaryIndexes.yml
```diff
@@ -0,0 +1,116 @@
...
+    Operation:
+      OperationName: updateMany
+      OperationMetricsName: UpdateManyWithoutHint
+      OperationCommand:
+        Filter: { a: "foo" }
+        Update: { $set: { a: "bar" } }
+  - *Nop
+  - *Nop
+
+# Phase 4: Issue a large update without a hint.
```
mtandrei, line 96: [nit] This should say "with a hint" right?

Also, I'm not sure that having an update with a hint is necessary for this workload. In the original test, the hint was there to show a "baseline" for update performance and that there was an issue with updating in the presence of multiple secondary indexes. In this performance test, it should be enough to have 'Phase 3' above as this will track the performance of updating in the presence of multiple secondary indexes.

src/workloads/execution/UpdateWithSecondaryIndexes.yml
```diff
@@ -0,0 +1,34 @@
...
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: 3_600_000  # = 1 hour
+
+LoadConfig:
+  Path: "../../phases/execution/UpdateWithSecondaryIndexes.yml"
+
+# Uncomment AutoRun to enable this test in evergreen patches
```
mtandrei, line 26: [nit] delete this comment?

mtandrei (changes requested):
Looks good, just one comment about the phase with a hint

src/phases/execution/UpdateWithSecondaryIndexes.yml
Fefer-Ivan: I want to add an execution with a hint as a sanity check. If we ever see two metrics diverge - it is a reason to be suspicious.

mtandrei (approved):
LGTM; thanks Ivan!

Fefer-Ivan:
evergreen merge

PR #740, opened 2022-09-30
PERF-3379 Make additional workloads run in SBE and classic build vari…

dstorch:
…ants

Update the `AutoRun` sections of a few Genny workloads such that they run in the SBE/classic sys-perf build variants. We missed these workloads on our first pass because they aren't obviously query-related. However, I did an analysis which found that measurements associated with these tasks have historically had change points tied to SBE. This suggests that there is some value in including these benchmarks in our system to compare SBE performance against the classic baseline.

I'm open to suggestions from the performance team about what should or shouldn't be included. See [PERF-3379](https://jira.mongodb.org/browse/PERF-3379) and the [linked spreadsheet](https://docs.google.com/spreadsheets/d/17s3ep6fb6b00MHjSU2W9ai5sxq8YmiRNpaM16tq_SxY/edit) for a description of how I performed my audit.

There are a few workloads from my audit that Mihai and I considered including but I had second thoughts about. Please let me know whether you think we should try to include any of the following:

- big_update_10k
- mixed_workloads_genny_stress

Also note that we agreed to include `snapshot_reads`, but this needs to be done by modifying `system_perf.yml` in the mongodb/mongo repo, so I'll file a separate ticket for this.

I kicked off a sys-perf patch build in order to validate that the tests run correctly on the new build variants: https://spruce.mongodb.com/version/63373d43d6d80a2daf48627c/tasks

### DIFF
--- a/src/workloads/scale/BigUpdate.yml
+++ b/src/workloads/scale/BigUpdate.yml
@@ -129,8 +129,10 @@ AutoRun:
       $eq:
       - atlas
       - replica
-      - replica-noflowcontrol
       - replica-1dayhistory-15gbwtcache
       - replica-all-feature-flags
+      - replica-noflowcontrol
       - single-replica
       - standalone
+      - standalone-classic-query-engine
+      - standalone-sbe
--- a/src/workloads/scale/LargeScaleLongLived.yml
+++ b/src/workloads/scale/LargeScaleLongLived.yml
@@ -90,7 +90,9 @@ AutoRun:
       - atlas
       - replica
       - replica-all-feature-flags
+      - shard
       - single-replica
       - single-replica-15gbwtcache
-      - shard
       - standalone
+      - standalone-classic-query-engine
+      - standalone-sbe
--- a/src/workloads/scale/MajorityReads10KThreads.yml
+++ b/src/workloads/scale/MajorityReads10KThreads.yml
@@ -22,3 +22,5 @@ AutoRun:
       $eq:
       - replica
       - replica-all-feature-flags
+      - single-replica-classic-query-engine
+      - single-replica-sbe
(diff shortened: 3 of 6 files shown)


dalyd (approved):
I assume that all of these are adding standalone-classic-query-engine and standalone-sbe and follow-your spreadsheet. Also assuming any other changes are from alphabetizing lists. Those are all good to me, and otherwise the change looks right. 
LGTM.

src/workloads/scale/BigUpdate.yml
```diff
@@ -129,8 +129,10 @@ AutoRun:
       $eq:
       - atlas
       - replica
-      - replica-noflowcontrol
       - replica-1dayhistory-15gbwtcache
       - replica-all-feature-flags
+      - replica-noflowcontrol
       - single-replica
       - standalone
+      - standalone-classic-query-engine
```
mtandrei, line 137: Is there a reason that we want "BigUpdate" running on the standalone variants and not the 1 node repl set variants? looking at https://evergreen.mongodb.com/task/sys_perf_linux_1_node_replSet_big_update_5cbe7f207d1abe694708bf729e4808a4a5413e60_22_09_28_21_01_35, this seems possible, but I'm not sure if this is deliberate.

src/workloads/scale/LargeScaleLongLived.yml
```diff
@@ -90,7 +90,9 @@ AutoRun:
       - atlas
       - replica
       - replica-all-feature-flags
+      - shard
       - single-replica
       - single-replica-15gbwtcache
-      - shard
       - standalone
+      - standalone-classic-query-engine
```
mtandrei, line 97: I was looking at your sys-perf patch build to confirm that this BV was running LargeScaleLongLived, and saw that it didn't run the standalone classic variant. I re-configured the patch build to do so, and will double check that this work (for future reference, is it ok to reconfigure someone else's patch build?)

mtandrei (approved):
LGTM with one question about whether certain tasks should run on both standalone and 1 node repl set or not. 

Also, it looks like your patch build wasn't running the classic standalone variant. Please verify that the tasks that this PR enables to run on the classic standalone variant (all except MajorityReads10k, ScanWithLongLived and LLTMixed) do run on this variant before merging.

PR #741, opened 2022-10-03
PERF-3275: Add change event application workload for mongosync

brad-devlugt:
This adds a new workload for testing mongosync's change event application phase. It also adds some changes to the external script actor:

1. Scripts that exit with a non-zero exit code will now fail the test run
2. Scripts are written to a temporary file rather than being appended to "sh -c" to reduce the amount of escape characters required to write a script and make it easier to compare strings with spaces in them in shell scripts


src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -0,0 +1,95 @@
...
+
+PollForCEA:
+  Repeat: 1
+  Command: sh
+  MetricsName: PollForCEA
+  Script: |
+    set -eou pipefail
+
+    getState() {
+      curl -s -S http://10.2.0.160:27182/api/v1/progress | jq '.progress.info' --raw-output
```
brad-devlugt, line 39: I think it would be neat if we added a way to pass in environment variables in the future, so we can pass down this data from DSI. I wanted to keep this simple for now though, as this might change a lot during the project

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -0,0 +1,95 @@
...
+    done
+
+DrainWrites:
+  Repeat: 1
+  Command: sh
+  MetricsName: DrainWrites
+  Script: |
+    set -eou pipefail
+
+    getLag () {
```
brad-devlugt, line 57: A lot of these scripts end up being pretty similar. I think it could be neat if we could come up with a way to add functions that can be re-used for external script actors in the future.

brad-devlugt:
This goes with https://github.com/10gen/dsi/pull/1099 and https://github.com/mongodb/genny/pull/741, and I put up a patch to test out the new test in https://spruce.mongodb.com/version/633b35830305b9222135cbb4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -66,29 +67,72 @@ struct ExternalScriptRunner::PhaseConfig {
...
-        BOOST_LOG_TRIVIAL(debug) << "ExternalScriptRunner running " << config->programCommand;
-
         for (const auto&& _ : config) {
+            TempScriptFile file{config->script};
+            std::stringstream fullInvocation;
+            fullInvocation << config->invocation << " "<< file.name() << " 2>&1";
+            std::string invocation = fullInvocation.str();
 
             // Execute the script and read result from stdout
             auto ctx = config->scriptOperation.start();
```
dalyd, line 134: Not your change, but it might be worth a comment to say that line 134 starts the timer on the metrics, but does not actually start the script. I got confused by that.

src/workloads/c2c/ChangeEventApplication.yml
```diff
@@ -0,0 +1,63 @@
...
+  Phases:
+  - *nop
+  - *nop
+  - Repeat: 1
+    BatchSize: 100
+    Threads: 1
+    DocumentCount: 1000000
+    Database: db
+    CollectionCount: 1
+    Document:
```
dalyd, line 57: Does the document shape matter at all in this workload?

dalyd: Or number of collections?

src/workloads/c2c/ChangeEventApplication.yml
```diff
@@ -0,0 +1,63 @@
...
+      a: {^RandomInt: {min: 0, max: 1000000}}
+      b: {^RandomString: {length: 8}}
+      c: {^RandomString: {length: 20}}
+  - *nop
+  - *nop
+  - *nop
```
dalyd, line 63: Should we add a comment at the bottom saying there is no auto-run, and currently the workload only is being run manually?

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -0,0 +1,96 @@
...
+
+  For curl commands we pass -s and -S so that an errors are still reported (-S)
+  but download progress stats are not logged to stderr
+
+StartMongosync:
+  Repeat: 1
+  Command: sh
+  MetricsName: StartMongosync
+  Script: |
+    curl -s -S -X POST http://10.2.0.160:27182/api/v1/start \
```
dalyd, line 28: I wish there was a way to pipe in the IP address. That looks really brittle, and it means we can't use this workload as is unless we use our infrastructure provisioning and ensure we have the correct IP. That said, I'm not sure how we would get it in. There is some ability to pass in variables to Genny, but likely painful and ugly.

src/workloads/c2c/MongosyncScripts.yml
dalyd: Oh cool I was just thinking about the same thing above.

src/workloads/c2c/MongosyncScripts.yml
dalyd: Alternatively you are already rsyncing the replicator node. You can upload a bunch of arbitrary scripts and call them. You don't need to dynamically generate each of them. I would encourage you to not do that in shell script though. Node or python would be better.

dalyd (commented):
Mostly questions and mild suggestions. Looking good.

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: Yeah I agree. I think this is going to get more complicated with adding multiple C2C urls also, so I have started thinking about how to handle that.

One idea I had is a top-level config in genny like:
```
EnvironmentVariables:
   key1: a
   key2: b
```

Then we could pass it in from DSI via genny's overrides.yml, and for the script invocation append all the key/value pairs under environment variables as `key1="a" key2="b" sh temporary_path.sh`.

What are your thoughts?

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: hmm that's a good point, I suppose we could just have a python script that lives alongside the workload, and give the external script actor a `ScriptPath` argument which could be provided in place of `Script`. Is that what you were thinking?

dalyd: yes, something like that.  It's all thinking out loud though.

src/workloads/c2c/MongosyncScripts.yml
dalyd: Oh that could work. I would just encourage you to not get too clever in bash. It doesn't end well :)

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: Sounds good! What are your thoughts on this proposal?

1. I'll change up some of the minor suggestions (e.g. cleanup, docs, etc.)
2. Merge this change without adding the EnvironmentVariables/ScriptPath/switch to python part
3. In the ticket to make this work with multiple mongosyncs, I'll implement 2) as I think it will be the most efficient path to get a correct, working solution

I think if we move forward before doing 2), that will let us shake out any surprises as we'll have this test running on a regular basis off of the waterfall over the next couple weeks

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -66,29 +67,72 @@ struct ExternalScriptRunner::PhaseConfig {
...
+    try {
+        while (fgets(buffer.data(), buffer.size(), pipe) != nullptr) {
+            result += buffer.data();
+            BOOST_LOG_TRIVIAL(info) << "Script output: " << buffer.data();
+        }
     }
+    catch (...) {
+        pclose(pipe);
+        throw;
+    }
```
aggrand, line 84: Would it be worth creating a RAII wrapper for this pipe?

src/cast_core/src/ExternalScriptRunner.cpp
```diff
@@ -45,12 +48,10 @@ struct ExternalScriptRunner::PhaseConfig {
...
           scriptOperation{phaseContext.namedOperation("ExternalScript", id)} {
 
-            // The script should not have double quotes
-            assert(script.find("\"") == std::string::npos);
             if (command == "mongosh") {
-                programCommand = "mongosh " + mongoServerURI + " --quiet --eval \"" + script + "\"";
+                invocation = "mongosh " + mongoServerURI + " --quiet --file";
             } else if (command == "sh") {
-                programCommand = "sh -c \"" + script + "\"";
+                invocation = "sh";
```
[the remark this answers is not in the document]
aggrand: I think it's added further down in the `fullInvocation << config->invocation << " "<< file.name() << " 2>&1";` line, if I'm reading correctly! I agree it's confusing though; should we combine the two lines?

aggrand (changes requested):
Can't comment much on the workloads themselves. Had a few nits on the actor, but push back if you disagree! This looks great!

src/workloads/c2c/MongosyncScripts.yml
dalyd: That general plan sounds good. I'm not sure which thing you are proposing with "2)", but I don't think that matters right now for your next steps.

src/cast_core/src/ExternalScriptRunner.cpp
brad-devlugt: It's a good thought. I think in this case, since I need the results from pclose to get the exitStatus, I'd need to track the state of whether or not the pipe is closed in the class and still call some sort of close method manually. My vote would be to keep the current implementation since it's relatively contained. Let me know if you disagree and I'm happy to add the new class.

src/cast_core/src/ExternalScriptRunner.cpp
brad-devlugt: Updated

src/workloads/c2c/ChangeEventApplication.yml
brad-devlugt: Updated

src/cast_core/src/ExternalScriptRunner.cpp
aggrand: Fair, sounds good. Can always add one later if it turns out to be needed

aggrand (approved):

src/cast_core/src/ExternalScriptRunner.cpp
dalyd: Thank you!

dalyd (approved):

brad-devlugt:
evergreen merge

brad-devlugt:
evergreen merge

PR #747, opened 2022-10-13
PERF-3420 Add workload for background index build impact

jordiolivares:
This workload is to test the impact of background index builds on a populated dataset while writes are occurring simultaneously.

### DIFF
--- a/src/workloads/execution/BackgroundIndexConstruction.yml
+++ b/src/workloads/execution/BackgroundIndexConstruction.yml
@@ -1,8 +1,14 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/server-execution"
 Description: |
-   This workload tests the performance of creating indexes on a collection large enough that the external
-   sorter must spill to disk. This can happen when the keys are larger than maxIndexBuildMemoryUsageMegabytes.
+  This workload tests the write performance impact of creating indexes on a very large collection.
+  It synthetically lowers the number of concurrent operations allowed to represent a saturated
+  server.
+
+Keywords:
+- stress
+- indexes
+- InsertRemove
 
 Clients:
   Default:
@@ -64,7 +70,7 @@ Actors:
   - *Nop
   - *Nop
   # Build an index on an integer field.
-  - Repeat: 10
+  - Repeat: 5
     Database: *db
     Operations:
     - OperationMetricsName: CreateIndexInt
@@ -81,7 +87,7 @@ Actors:
         dropIndexes: *coll
         index: random_int
   # Build an index on a string field.
-  - Repeat: 10
+  - Repeat: 5
     Database: *db
     Operations:
     - OperationMetricsName: CreateIndexString
@@ -104,11 +110,11 @@ Actors:
   Phases:
   - *Nop
   - *Nop
-  # Build an index on an integer field.
+  # Execute while building an index on an integer field.
   - Blocking: None
     Database: *db
     Collection: &otherColl Collection1  # Work on a different collection to avoid overloading the index target
-  # Build an index on a string field.
+  # Execute while building an index on a string field.
   - Blocking: None
     Database: *db
     Collection: *otherColl
--- a/src/workloads/execution/BackgroundIndexConstruction.yml
+++ b/src/workloads/execution/BackgroundIndexConstruction.yml
@@ -70,7 +70,7 @@ Actors:
   - *Nop
   - *Nop
   # Build an index on an integer field.
-  - Repeat: 5
+  - Repeat: 2
     Database: *db
     Operations:
     - OperationMetricsName: CreateIndexInt
@@ -87,7 +87,7 @@ Actors:
         dropIndexes: *coll
         index: random_int
   # Build an index on a string field.
-  - Repeat: 5
+  - Repeat: 2
     Database: *db
     Operations:
     - OperationMetricsName: CreateIndexString
--- a/src/workloads/execution/BackgroundIndexConstruction.yml
+++ b/src/workloads/execution/BackgroundIndexConstruction.yml
@@ -13,7 +13,7 @@ Keywords:
 Clients:
   Default:
     QueryOptions:
-      maxPoolSize: 260
+      maxPoolSize: 512
       socketTimeoutMS: 3600000  # = 1 hour
 
 Actors:
@@ -132,6 +132,16 @@ Actors:
   - Blocking: None
     Database: *db
     Collection: *otherColl
+  - *Nop
+
+- Name: BackgroundSideWrites
+  Type: InsertRemove
+  Threads: 128
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
   - Blocking: None
     Database: *db
     Collection: *coll  # Work on the same collection to excercise the side-writes.
--- a/src/workloads/execution/BackgroundIndexConstruction.yml
+++ b/src/workloads/execution/BackgroundIndexConstruction.yml
@@ -20,7 +20,7 @@ Clients:
   Default:
     QueryOptions:
       maxPoolSize: 1024
-      socketTimeoutMS: 21600000  # = 6 hours
+      socketTimeoutMS: 7200000  # = 2 hours
 
 Actors:
 # Phase 1: Insert enough data to ensure than an index build on each field spills to disk with a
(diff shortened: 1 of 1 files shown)


src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+# memory limit of 100MB.
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10000000
```
haleyConnelly, line 30: Q: Can we also use the thousands separators here to make it more readable?

haleyConnelly (dismissed):

dalyd:
@jordiolivares I'll take a look at this. Could you fill out https://docs.google.com/forms/d/e/1FAIpQLSf1oh23Ddo1khjLZMqZ9DHbRdObnpeH20PSXTBuXVg-7mxxTA/viewform quickly to help me with background? Or point me to a previous conversation if I'm forgetting one (which I sometimes do). Either should hopefully only take a couple of moments and it helps me and my colleagues make sure we're checking the right things on the PRs. Thanks.

jordiolivares:
@dalyd I've submitted the form.

The last item I'm playing with is the side-writes aspect for an extra datapoint, but if it causes me too much trouble then feel free to review it up to https://github.com/mongodb/genny/pull/747/commits/1f3b710af3988520a885563ed153fa7071b96018 (before the add side-writes exercising commit).

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
+  This workload tests the write performance impact of creating indexes on a very large collection.
+  It synthetically lowers the number of concurrent operations allowed to represent a saturated
+  server.
```
dalyd, line 6: Could you add something saying we measure and care about the index creation time and the throughput and latency of the inserts and removes of the background latency? Possibly also mention that we may see trade-offs between index creation time and background workload performance.

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+# Phase 2: Lower the memory limit for spilling to disk so that it occurs more often.
+- Name: Setup
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    # Fsync to force a checkpoint and quiesce the system.
```
dalyd, line 49: We have a quiesce actor you could use instead of the fsync, although it's documentation isn't as reassuring as I would like: https://github.com/mongodb/genny/blob/master/src/workloads/docs/QuiesceActor.yml It does an fsync, but it does more than that also. 

https://github.com/mongodb/genny/blob/d4897cf792d3ea30c40c94c5e04ce715a918cd1f/src/gennylib/src/quiesce.cpp

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10000000
+    BatchSize: 1000
+    Document:
+      randomInt: {^RandomInt: {min: -10000000, max: 10000000}}
```
dalyd, line 33: Does it matter to this workload that these two indexes should be fairly distributed? That is, I expect there will be few keys with more than 1 entry. Would anything interesting come out from having an indexed field with limited values, or with hot values?

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+    Operations:
+    - OperationMetricsName: CreateIndexInt
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: &coll Collection0
+        indexes:
+        - key:
+            randomInt: 1
+          name: random_int
+    - OperationMetricsName: DropIndexInt
```
dalyd, line 86: This is interesting. I frankly don't remember how index builds work these days. Is there a chance of the background workload speeding up as more of the index is completed, and then falling back down with the index is dropped? 

I suspect there are phases to the index build, and their impact (each phase) on the background workload may be different. It's worth looking at FTDC data from your runs to see if there's any interesting periodicity with the index builds.

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+      OperationName: RunCommand
+      OperationCommand:
+        dropIndexes: *coll
+        index: random_int
+  # Build an index on a string field.
+  - Repeat: 1
```
dalyd, line 92: Do we definitely start this phase without the int index? I think that should be the case, but it also seems worth verifying.

src/workloads/execution/BackgroundIndexConstruction.yml
dalyd: Could you also describe up here that you are both

1. Building an index with a workload against a different collection
2. Building an index with a workload against the same collection

At different points in the workload?

dalyd (commented):
thanks for filling out the workload doc -- that really helped me. I've added some questions and suggestions.

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: Done, I'll proceed to also revert the addition of point 2 as it seems it times out with our depriorisation changes and it would make the test fail with the feature flag enabled.

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: Done

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: We do, the last command executed is a dropIndex for the int index

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: It shouldn't matter that much as indexes aren't affected by hot keys. In this case we are only interested in the background aspect of the index build occurring simultaneously with a write workload

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+    Database: &db test
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10_000_000
+    BatchSize: 1000
+    Document:
+      randomInt: {^RandomInt: {min: -10000000, max: 10000000}}
+      randomString: {^RandomString: {length: 16}}
+  - &Nop {Nop: true}
```
dalyd, line 41: Does anything happen in phase 2? I see the loader in phase 1, and quiesce and run command in phase 3. Was one of those two supposed to be in phase 2?

src/workloads/execution/BackgroundIndexConstruction.yml
```diff
@@ -0,0 +1,156 @@
...
+  - *Nop
+  - *Nop
+  - *Nop
+
+# Phase 2: Lower the memory limit for spilling to disk so that it occurs more often.
+- Name: Setup
```
dalyd, line 47: This setup happens in phase 3, while the comment says 2. Does anything happen in phase 2? I see the loader in phase 1, and quiesce and this in phase 3. Was one of those two supposed to be in phase 2?

dalyd (commented):
One nit/question. Looks good.

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: Nice catch, I probably mixed up the number of Nops, Quiesce is phase 3 and RunCommand phase 2

src/workloads/execution/BackgroundIndexConstruction.yml
jordiolivares: Fixed with the latest commit

dalyd (approved):

PR #749, opened 2022-10-14
PERF-3276: Create Collection copy specific short workload and add it to the mongosync waterfall

huan-Mongo:


src/workloads/c2c/CollectionCopy.yml
```diff
@@ -0,0 +1,75 @@
...
+
+Keywords:
+- c2c
+- replication
+- collection copy
+- cluster to cluster sync
+
+# These two values should match those are the top of MixPhases.yml
+dbname: &dbname mongosync_test
+DocumentCount: &NumDocs 100000
```
brad-devlugt, line 15: Is this enough data to get a stable number for collection copy speed?

src/workloads/c2c/CollectionCopy.yml
```diff
@@ -0,0 +1,75 @@
...
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    BatchSize: 100
+    Threads: 1
+    DocumentCount: *NumDocs
+    Database: *dbname
+    CollectionCount: *NumColls
+    Document: &doc
```
brad-devlugt, line 38: If this is the same document structure as the CEA workload, what are your thoughts on using `LoadConfig` here to re-use some of the code?

src/workloads/c2c/CollectionCopy.yml
huan-Mongo: Yea, I agree 100K docs might not be enough. I am changing it to 1M.

src/workloads/c2c/CollectionCopy.yml
huan-Mongo: Good idea!

huan-Mongo: I moved the command insertion phase to `MongosyncScripts.yml`

src/workloads/c2c/CollectionCopy.yml
brad-devlugt: Sounds good! Was there any difference in the cea or collection copy test results with 1M docs?

huan-Mongo: Yea, Collection copy seems 1000 times faster than CEA... I expected it would be faster, but wasn't aware it's that fast.

brad-devlugt: Oh I see - nice finds!

When comparing the 100k docs to the 1M docs version of the test, is there any noticeable difference in results or runtime of the test? I'm just wondering if it worth extra runtime for additional docs, or if 100k is enough.

huan-Mongo: I saw there is no big difference between these two datasets regarding the metrics and test execution time. So I think we can keep the 1M data set here to be more confident.

brad-devlugt: Sounds good!

huan-Mongo:
evergreen merge

PR #750, opened 2022-10-17
[PERF-3416] - Generate new LLT workload with smaller collections

MAhmadShah:
Work on this PR is to cover [PERF-3416](https://jira.mongodb.org/browse/PERF-3416)

We have created a new workload that is a branch of the LLTMixed workload. This workload shares the same configuration, except that it now uses fewer documents (6 million instead of 10 million). This was done in an effort to reduce variance between runs, which we were able to confirm. Results can be found in this [analysis](https://docs.google.com/document/d/1x-xUVNg9SN1LXOWJ7vKhU_LneKClFxSAMzobiFvwDrc/edit?usp=sharing)

What has changed:

- Updated LLTMixed workload to parameterize the document counts at the top level (the load config did not work without this change)
- Add in LLTMixedSmall workload, which imports LLTMixed using LoadConfig and then updates the document counts

Test results can be found on this [patch](https://spruce.mongodb.com/version/634860bb5623433b16917f03/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). We can see that LLTMixed still runs fine after its changes and the LLTMixedSmall runs as expected with the logs showing 6 million documents being inserted.


src/workloads/transactions/LLTMixed.yml
```diff
@@ -25,11 +25,14 @@ Keywords:
 - update
 - delete
 
+InitialNumDocs: &InitialNumDocs {^Parameter: {Name: InitialNumDocs, Default: 10000000}}
```
jimoleary, line 28: Can you add something like the following (assuming it is correct):

```
Parameters:
  - &InitialNumDocs {^Parameter: {Name: InitialNumDocs, Default: 10000000}}
  - &SecondaryNumDocs {^Parameter: {Name: SecondaryNumDocs, Default: 10000000}}
```

jimoleary (commented):
small suggestion (more as a convention).

PR #752, opened 2022-10-21
PERF-3470 Add range query on '_id' and 'a' to clustered collection workload

Fefer-Ivan:
Added range queires on '_id' and 'a' to test if multi planner can efficiently choose between clustered index scan and secondary index scan

### DIFF
--- a/src/phases/execution/ClusteredCollection.yml
+++ b/src/phases/execution/ClusteredCollection.yml
@@ -15,20 +15,20 @@ GlobalDefaults:
   MaxPhases: &MaxPhases 9
 
 ValueGenerators:
-  SmallId: &SmallId {^Join: {array: [
-    { ^FormatString: {"format": "%07.0f", "withArgs": [
-      {^RandomDouble: {
-        min: 1,
-        max: {^NumExpr: {withExpression: "(2 * docCount) / 10", andValues: {docCount: *docCount}}}}}]}},
-    *PaddingGenerator]}}
-  BigId: &BigId {^Join: {array: [
-    { ^FormatString: {"format": "%07.0f", "withArgs": [
-      {^RandomDouble: {
-        min: {^NumExpr: {withExpression: "(8 * docCount) / 10", andValues: {docCount: *docCount}}},
-        max: *docCount}}]}},
-    *PaddingGenerator]}}
-  SmallSecondaryKey: &SmallSecondaryKey {^Join: {array: ["1", {^FastRandomString: {length: 5, alphabet: "0123456789"}}]}}
-  BigSecondaryKey: &BigSecondaryKey {^Join: {array: ["8", {^FastRandomString: {length: 5, alphabet: "0123456789"}}]}}
+- &SmallId {^Join: {array: [
+  { ^FormatString: {"format": "%07.0f", "withArgs": [
+    {^RandomDouble: {
+      min: 1,
+      max: {^NumExpr: {withExpression: "(2 * docCount) / 10", andValues: {docCount: *docCount}}}}}]}},
+  *PaddingGenerator]}}
+- &BigId {^Join: {array: [
+  { ^FormatString: {"format": "%07.0f", "withArgs": [
+    {^RandomDouble: {
+      min: {^NumExpr: {withExpression: "(8 * docCount) / 10", andValues: {docCount: *docCount}}},
+      max: *docCount}}]}},
+  *PaddingGenerator]}}
+- &SmallSecondaryKey {^Join: {array: ["1", {^FastRandomString: {length: 5, alphabet: "0123456789"}}]}}
+- &BigSecondaryKey {^Join: {array: ["8", {^FastRandomString: {length: 5, alphabet: "0123456789"}}]}}
 
 Actors:
 # Phase 0: create collection
--- a/src/phases/execution/ClusteredCollection.yml
+++ b/src/phases/execution/ClusteredCollection.yml
@@ -13,6 +13,8 @@ GlobalDefaults:
   Namespace: &Namespace test.Collection0
   DocumentCount: &docCount 1000000
   MaxPhases: &MaxPhases 9
+
+ValueGenerators:
   SmallId: &SmallId {^Join: {array: [
     { ^FormatString: {"format": "%07.0f", "withArgs": [
       {^RandomDouble: {
--- a/src/phases/execution/ClusteredCollection.yml
+++ b/src/phases/execution/ClusteredCollection.yml
@@ -12,7 +12,7 @@ GlobalDefaults:
   Collection: &Collection Collection0
   Namespace: &Namespace test.Collection0
   DocumentCount: &docCount 1000000
-  MaxPhases: &MaxPhases 8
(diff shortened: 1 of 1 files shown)


src/phases/execution/ClusteredCollection.yml
```diff
@@ -136,15 +135,41 @@ Actors:
...
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {
+              _id: {
+                $gte: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^RandomInt: {min: 1, max: *docCount}}]}}, *PaddingGenerator]}}
+              },
+              a: {
+                $gte: {^Join: {array: [
+                  {^Choose: {from: ["49", "51"]}},
```
ykangpark, line 159: Is there any specific reason for this prefixes?

ykangpark (commented):
Might be a dumb question, but the description of SERVER-58276 indicates the improvement is for time-series collections.  The collection here is not a time-series one, so how is this related?

Fefer-Ivan:
Timeseries collections are clustered.
Improving time series collection performance is the motivation behind this change, but enabling clustered index scan in multiplanning is general improvement for all clustered collection, no need to limit it only to timeseries collections.

src/phases/execution/ClusteredCollection.yml
Fefer-Ivan: The idea is to have index scan that scans over roughly the half of all documents.
I also want _id predicate to be similar: so not random int 1 - docsCount, but maybe something like 49 * docsCount / 100 - 51 * docsCount / 100. But I don't know how to implement it yet.

Currently I am testing if this works and get back to this idea later.
Maybe I will just scrap this idea and make range on a also random. 

Multiplanner should be able to pick correct index to scan in any case.

ykangpark: Not sure if it will help, but you can do YAML arithmetic (preprocessed) with `^NumExpr`.

Fefer-Ivan: Thanks. That is exactly what I was looking for.

Currently I decided to just pick ranges at random. It makes one of the two indexes a clear winner during multiplanner, which should be fine. I was over-complicating stuff.

Maybe it is a good idea to add two separate phases: one will have _id index be better fit for the query and the other will have secondary index clearly better.

src/phases/execution/ClusteredCollection.yml
```diff
@@ -134,17 +133,46 @@ Actors:
...
+        - OperationName: find
+          OperationCommand:
+            Filter: {
+              _id: {^Choose: {
+                from: [
+                  {$gte: *RandomId},
+                  {$lte: *RandomId}
+                ]
+              }},
+              a: {^Choose: {
```
lriuui0x0, line 160: Am I understanding correctly that before your planned change, the query will make use of a single secondary index, and after your planned change, the query will use the index intersection of both indices?

Fefer-Ivan: No.

Planner will pick either clustered index with filter on 'a' or secondary index, fetch and filter on '_id'.

In general it picks one which encounters matching elements more often during trial stage.

I didn't check how we generate index intersection plans, but I bet it will be much harder than what I already did because clustered index scan is still a collection scan, so I don't if our index intersection even supports that.

Fefer-Ivan: I can look into it, but I think it should be a separate ticket to enable clustered index + secondary index intersection.

Fefer-Ivan:
According to slack macos build failure is a known issue: https://mongodb.slack.com/archives/C01VD0LQZED/p1659617050659319

src/phases/execution/ClusteredCollection.yml
```diff
@@ -134,17 +133,86 @@ Actors:
...
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {
+              _id: {^Choose: {
+                from: [
+                  {$gte: &BigId {^Join: {array: [
```
lriuui0x0, line 156: Minor: We can move these values up instead of cluttered in the middle of workload.

lriuui0x0 (approved):

src/phases/execution/ClusteredCollection.yml
```diff
@@ -134,17 +146,76 @@ Actors:
         Operations:
         - OperationName: find
           OperationCommand:
-            Filter: {a: {^FastRandomString: {length: 6, alphabet: "0123456789"}}}
+            Filter: {a: *RandomSecondaryKey}
 
+# Phase 6: Range lookup with both primary and secondary key. Primary key predicate is more selective.
+- Name: RangeLookupBySelectivePrimaryAndWideSecondaryIndexes
```
ykangpark, line 152: Personally, I would find it helpful to clarify what being more selective or wide means.

src/phases/execution/ClusteredCollection.yml
```diff
@@ -12,8 +12,21 @@ GlobalDefaults:
...
+        min: 1,
+        max: {^NumExpr: {withExpression: "(2 * docCount) / 10", andValues: {docCount: *docCount}}}}}]}},
+    *PaddingGenerator]}}
+  BigId: &BigId {^Join: {array: [
+    { ^FormatString: {"format": "%07.0f", "withArgs": [
+      {^RandomDouble: {
+        min: {^NumExpr: {withExpression: "(8 * docCount) / 10", andValues: {docCount: *docCount}}},
+        max: *docCount}}]}},
+    *PaddingGenerator]}}
+  SmallSecondaryKey: &SmallSecondaryKey {^Join: {array: ["1", {^FastRandomString: {length: 5, alphabet: "0123456789"}}]}}
```
ykangpark, line 28: What's the benefit of having a random boundary each time the perf test is run?

ykangpark: That is compared with the fixed boundaries for the _id field.

Fefer-Ivan: It is not fixed for the whole test, it is different boundary for each query.

I just used this section to define this large expressions to make phase definitions more readable.

src/phases/execution/ClusteredCollection.yml
Fefer-Ivan: I re-wrote the comments and stage names to be more readable.

src/phases/execution/ClusteredCollection.yml
Fefer-Ivan: Moved them from GlobalDefaults to ValueGenerators field for clarity

ykangpark (approved):

src/phases/execution/ClusteredCollection.yml
Fefer-Ivan: It shouldn't be, because I think there are enough queries made for statistics to stabilize
If we see this tests to be unstable, we can narrow the random ranges.

There still has to be some randomness, so we won't cache exact queries.

PR #753, opened 2022-10-21
PERF-3268: Automate multiple mongosync use-case

brad-devlugt:
This adds sharding support to the mongosync workloads and refactors the external script actor so that we can pass in environment variables. This allows us to pass in the mongosync connection strings so they aren't hardcoded. I also ported the mongosync scripts to python since they were becoming more complex and needed threading support.

### DIFF
--- a/src/workloads/c2c/ChangeEventApplication.yml
+++ b/src/workloads/c2c/ChangeEventApplication.yml
@@ -13,6 +13,9 @@ Keywords:
 - change event application
 - CEA
 
+EnvironmentVariables:
+  MongosyncConnectionUrls: http://localhost:27182
+
 Clients:
   Default:
     QueryOptions:
@@ -21,17 +24,52 @@ Clients:
 ScriptsPath: &scriptsPath ./MongosyncScripts.yml
 
 Actors:
+
+- Name: SetupSharding
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    # We also run this on replica sets and so we
+    # ignore if this command fails
+    ThrowOnFailure: false
+    Operations:
+    - OperationMetricsName: EnableShardingMetrics
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: db
+
+  - Repeat: 1
+    Database: admin
+    ThrowOnFailure: false
+    Operations:
+    - OperationMetricsName: ShardCollectionMetrics
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: db.Collection0
+        key:
+          _id: hashed
+  - &nop {Nop: true}
+  - *nop
+  - *nop
+  - *nop
+  - *nop
+  - *nop
+
 - Name: Mongosync
   Type: ExternalScriptRunner
   Threads: 1
   Phases:
+  - *nop
+  - *nop
   - LoadConfig:
       Path: *scriptsPath
       Key: StartMongosync
   - LoadConfig:
       Path: *scriptsPath
       Key: PollForCEA
-  - &nop {Nop: true}
+  - *nop
   - LoadConfig:
       Path: *scriptsPath
       Key: DrainWrites
@@ -48,10 +86,12 @@ Actors:
   Phases:
   - *nop
   - *nop
+  - *nop
+  - *nop
   - Repeat: 1
     BatchSize: 100
     Threads: 1
-    DocumentCount: 1000000
+    DocumentCount: 500000
     Database: db
 
     # Note the document shape and number of collections doesn't
@@ -60,7 +100,7 @@ Actors:
     # insert workload
     CollectionCount: 1
     Document:
-      a: {^RandomInt: {min: 0, max: 1000000}}
+      a: {^RandomInt: {min: 0, max: 500000}}
       b: {^RandomString: {length: 8}}
       c: {^RandomString: {length: 20}}
   - *nop
--- a/src/workloads/c2c/MongosyncScripts.yml
+++ b/src/workloads/c2c/MongosyncScripts.yml
@@ -12,85 +12,32 @@ Description: |
         Path: ./MongosyncScripts.yml
         Key: StartMongosync
 
-  Note that the polling scripts start with set -eou pipefail so that any
-  subcommand run during the script failing will cause the entire script to fail.
-  This way if mongosync crashes during the test, it will stop the test execution
-  immediately and genny will bubble up the error and fail the test
-
-  For curl commands we pass -s and -S so that an errors are still reported (-S)
-  but download progress stats are not logged to stderr
-
 StartMongosync:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: StartMongosync
-  Script: |
-    curl -s -S -X POST http://10.2.0.160:27182/api/v1/start \
-      -H 'Content-Type: application/json' \
-      -d '{"Source": "cluster0", "Destination": "cluster1"}'
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m start
 
 PollForCEA:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: PollForCEA
-  Script: |
-    set -eou pipefail
-
-    getState() {
-      curl -s -S http://10.2.0.160:27182/api/v1/progress | jq '.progress.info' --raw-output
-    }
-
-    state=$(getState)
-    while [ "$state" != "change event application" ]
-    do
-      echo "Polling for change event application, current state=$state"
-      sleep 1;
-      state=$(getState)
-    done
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m poll_for_cea
 
 DrainWrites:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: DrainWrites
-  Script: |
-    set -eou pipefail
-
-    getLag () {
-      curl -s -S http://10.2.0.160:27182/api/v1/progress | jq '.progress.lagTimeSeconds'
-    }
-
-    lagTime=$(getLag)
-    while [ $lagTime -gt 0 ];
-    do
-      echo "Waiting for writes to drain, currentLag=$lagTime"
-      sleep 1;
-      lagTime=$(getLag)
-    done
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m drain_writes
 
 Commit:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: Commit
-  Script: |
-    curl -s -S -X POST http://10.2.0.160:27182/api/v1/commit \
-      -H 'Content-Type: application/json' \
-      -d '{}'
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m commit
 
 WaitForCommit:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: WaitForCommit
-  Script: |
-    set -eou pipefail
-
-    getState() {
-      curl -s -S http://10.2.0.160:27182/api/v1/progress | jq '.progress.state' --raw-output
-    }
-
-    state=$(getState)
-    while [ $state != "COMMITTED" ]
-    do
-      echo "Waiting for commit to finish, state=$state"
-      sleep 1;
-      state=$(getState)
-    done
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m wait_for_commit
--- a/src/workloads/c2c/mongosync_helpers.py
+++ b/src/workloads/c2c/mongosync_helpers.py
@@ -1,0 +1,117 @@
+import argparse
+import os
+import time
+import requests
+import functools
+from concurrent.futures import ThreadPoolExecutor
+
+
+def poll_mongosync(connection_urls, predicate, key):
+    """
+    Wait for all mongosyncs to reach a certain state (e.g. predicate returns False)
+    based on a value returned by the /progress endpoint
+    """
+
+    def get_progress():
+        res = requests.get(f"{url}/api/v1/progress")
+        return res.json()["progress"][key]
+
+    for url in connection_urls:
+        info = get_progress()
+        while predicate(info):
+            time.sleep(1)
+            print(f"Polling {url} for {key}, current value = {info}", flush=True)
+            info = get_progress()
+
+
+def change_one_mongosync_state(route, body, url):
+    """
+    Change state of a given mongosync running at the provided url
+    """
+    resp = requests.post(
+        f"{url}{route}", json=body
+    )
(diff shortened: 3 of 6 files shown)


huan-Mongo (approved):
LGTM! Just some minor suggestions.

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -12,88 +12,35 @@ Description: |
...
-    }
-
-    state=$(getState)
-    while [ $state != "COMMITTED" ]
-    do
-      echo "Waiting for commit to finish, state=$state"
-      sleep 1;
-      state=$(getState)
-    done
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m wait_for_commit
```
aggrand, line 43: (nit) Given the args later on, I'm not sure if `ScriptPath` is appropriate for this. Should it just be `ScriptArgs` or so?

src/workloads/c2c/ChangeEventApplication.yml
```diff
@@ -13,6 +13,9 @@ Keywords:
 - change event application
 - CEA
 
+EnvironmentVariables:
+  MongosyncConnectionUrls: http://localhost:27182
+
```
aggrand, line 18: Instead of a global thing should this be part of the ExternalScriptRunner actor config? It really only seems relevant to that actor, and having something called "EnvironmentVariables" inside the workload context might make future actor authors think it's a general scheme for reading whatever environment variables genny can see

aggrand (changes requested):
Main concern was the location of environmentvariables, the rest is looking great!

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -12,88 +12,35 @@ Description: |
...
 StartMongosync:
   Repeat: 1
-  Command: sh
+  Command: python3
   MetricsName: StartMongosync
-  Script: |
-    curl -s -S -X POST http://10.2.0.160:27182/api/v1/start \
-      -H 'Content-Type: application/json' \
-      -d '{"Source": "cluster0", "Destination": "cluster1"}'
+  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m start
```
dalyd, line 19: Are these patches going to be brittle? It looks like this will only run if we are in WORKDIR, which assumes src/genny within it, but it won't work from within src/genny if you ran Genny direction with ./run-genny. 

Why not be match the behavior of LoadConfig and load relative to this file?

src/workloads/c2c/MongosyncScripts.yml
dalyd: Alternatively, should it be split into scriptpath and script args?

src/workloads/c2c/mongosync_helpers.py
```diff
@@ -0,0 +1,117 @@
+import argparse
```
dalyd, line 1: I like this movement to python. I wonder about how formal we want to be with the script? It's a one-off here. Should it be a module? That is likely overkill, but I wanted to say it out loud.

src/workloads/c2c/MongosyncScripts.yml
dalyd: If we are already passing in `-m start`, why not explicitly pass the connection url through the command line also, rather than the environmental variable? Or write out and read in a json file?

dalyd (commented):
I have some concern about the explicit path we are using for the script. Otherwise nits and questions.

src/workloads/c2c/ChangeEventApplication.yml
brad-devlugt: Is there a way to override an actor specific config with `overrides.yml`? I was using that to pass in the connection string here.

Alternatively, I could just invoke genny with the environment variables required (e.g. something like
`MongosyncConnectionUrls="A B C" ./run-genny workload ...`). I think the child processes should inherit the environment variables.

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: having a way to include things could be interesting - e.g. what if we could have something like `Include: script.py` that has helper functions in it, and then have the python code inline in the yml to invoke the methods how we'd like. That helps avoid having to write an argument parser for every new script

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: Had an idea - what if we put the python script in a top-level folder like `cast_python` beside `cast_core`, and add `cast_python` to the python path in `run-genny`?

Then we can do something like this:

```
StartMongosync:
  Repeat: 1
  Command: python3
  MetricsName: StartMongosync
  Script: |
    import mongosync_actor as actor
    actor.change_state("/api/v1/start",
        {"Source": "cluster0", "Destination": "cluster1"})
```

src/workloads/c2c/ChangeEventApplication.yml
```diff
@@ -13,6 +13,9 @@ Keywords:
 - change event application
 - CEA
 
+EnvironmentVariables:
+  MongosyncConnectionUrls: http://localhost:27182
```
huan-Mongo, line 17: Actually, shouldn't this be set inside `mongodb_setup.xxx`file? The same workload should be able to run inside both single or multiple Mongosync instances setup.

huan-Mongo (changes requested):

src/workloads/c2c/ChangeEventApplication.yml
huan-Mongo: ah, great. Thanks for the explanation!

huan-Mongo (approved):

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -114,3 +61,28 @@ InsertShortTestData:
...
+  Repeat: 1
+  Database: admin
+  # We also run this on replica sets and so we
+  # ignore if this command fails
+  ThrowOnFailure: false
+  Operations:
+  - OperationMetricsName: EnableShardingMetrics
+    OperationName: AdminCommand
+    OperationCommand:
+      enableSharding: db
```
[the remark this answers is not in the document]
brad-devlugt: Do you mean via the command line or just in the workload yml?

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: I took a stab at the `cast_python` idea and uploaded a new change list. Let me know what you think.

I also set it up so we can write unit tests for the python actors and so that they get linted.

src/workloads/c2c/MongosyncScripts.yml
```diff
@@ -16,31 +16,42 @@ StartMongosync:
   Repeat: 1
   Command: python3
   MetricsName: StartMongosync
-  ScriptPath: ./src/genny/src/workloads/c2c/mongosync_helpers.py -m start
+  Script: |
+    import mongosync_actor as actor
```
dalyd, line 20: I like this better. 

It's probably diminishing returns, but we could likely also just make this script an endpoint of the module.

src/workloads/c2c/MongosyncScripts.yml
brad-devlugt: Made some updates to this, now there's a folder for python actors and it gets added to genny's PYTHONPATH. The yaml here now just imports the appropriate python actor and calls the function it needs in line.

dalyd (commented):
I think I'm confused about something. Hopefully my questions and suggestions will show what I'm not seeing.

src/workloads/c2c/ChangeEventApplication.yml
brad-devlugt: The connection strings are configured in mongodb_setup here:
https://github.com/10gen/dsi/pull/1118/files#diff-870e9d10c6e89edfd67250b4a3dca1e54fcfb5e2853b2ecaf541c0373499b8a9R120

And then passed to genny in the test_control here:
https://github.com/10gen/dsi/pull/1118/files#diff-371f59723914a0fe0f3fcfafade32f3d0255d7fcce70b8f70648133163cbda28R11

dalyd:
I think I'm seeing the issue now -- there's no good way to pass something into Genny, so you are using environmental variables. 

That was the exact issue that https://jira.mongodb.org/browse/TIG-2915 was supposed to address. Some code went in for that ticket. Could you take a look and see if that solves the problem? To be fair, I don't know if it does.

src/workloads/c2c/ChangeEventApplication.yml
aggrand: The `overrides.yml` file can override any key/value pair in the Genny config. So in theory you could override the specific actor config, though you would have to specify the paths to it inside the override yml, meaning it would be brittle for changing actor names / numbers I believe.

I still think the name `EnvironmentVariables` would be confusing though. Maybe there could be a global actor shared space? Like renaming it to `ActorVars` with a subkey for `ExternalScriptRunner` or something?

cc @rtimmons since this relates to the slack convo about config schemes

brad-devlugt: The `ActorVars` with a subkey for `ExternalScriptRunner` sounds good to me. Then the fact that the actor implements access to the global actor space in terms of env variables is hidden and easier to change in the future & other actors could implement it differently. Cool if I proceed with that option?

rtimmons: Using env vars for conn strings can be dangerous if the conn strings have any auth in them (a proc's env vars aren't private). Plus we don't have precedent for plumbing with env vars, so this could end up rather confusing. I think we can do something better within DSI to expose configdict in a sane way.

src/workloads/c2c/MongosyncScripts.yml
rtimmons: Okay so maybe hot-take, but I think a script Actor would look like this:

```yaml
Actors:
- Type: Script
  Script: src/my-repo/foo-bar.sh # no args!
```

And then the API for the script is:

1. Script is called from work dir (`.`) with the `v1 setup` arg at constructor time. This can do venv or whatever, no need to muck with PYTHONPATH and risk contamination of Genny's python. (The `v1` is there to version this little protocol so the Actor and scripts can evolve over time.)
2. Script is called with `begin` at the start of the workload. It is sent `loop X` to stdin for each iteration with `X` as the phase number.
3. Script outputs metrics data to stdout and log output to stderr.

Here's an example bash script that shows how it could be used:

```bash
#!/usr/bin/env bash

set -eou pipefail

# Called at constructor time
if [[ "$1" == "setup" ]]; then
  setup-venv
  echo "Setup venv" > /dev/stderr
fi

# Called before any Phases start
if [[ "$1" == "begin" ]]; then
  # Cal
  while read -r line; do
    echo "{Metrics: Foo}"
  done
fi

if [[ "$1" == "end" ]]; then
  cleanup-venv
fi
```

Again, there are many other ways to do this and there are obvious limitations with this, but imho it's a nice balance between simplicity and functionality.

src/workloads/c2c/MongosyncScripts.yml
dalyd: FYI @rtimmons I asked for script paths to be relative to workload rather than work directory in this thread. The work directory is a DSI concept, and it seems bad if Genny can't be run locally outside of DSI.

src/workloads/c2c/ChangeEventApplication.yml
dalyd: Thanks. I see it now. If @huan-Mongo is good with this one we can close this thread.

src/workloads/c2c/MongosyncScripts.yml
rtimmons: Strong disagree - paths are notoriously confusing everywhere throughout our ecosystem. Plus having them relative to the workload makes it harder to have them cross-repo. Having all paths everywhere be strictly relative to working directory makes everything simple and obvious.

The "work dir" concept is really just a workspace concept. A convention for how to check out repos. I don't think it's necessarily a DSI concept. See related jira ticket [here](https://jira.mongodb.org/browse/EVG-18189).

(Strong opinion weakly held here - happy to have a further convo, but I think this is a super important concept to get right.)

dalyd: I definitely agree with `paths are notoriously confusing everywhere throughout our ecosystem.`. At the same time I would find it surprising if I couldn't checkout out genny and 
```
cd genny
./run-genny install
./run-genny workload myworkload
```
I think that would mess with people's built in assumptions.

rtimmons: > I think that would mess with people's built in assumptions.

That's what we gotta change 🥹 Multi-repo setups are a nightmare unless we have a consistent way to manage them and for repos to know where each other are. I think `run-genny` and `run-dsi` could guide users to this, and it becomes pretty natural. This is, after all, how DSI and Genny are run from CI, and local mirroring CI is goodness.

rtimmons: Another angle that might be more convincing: Imagine wanting the genny yaml and the script to live in separate repos. You'd need both repos to be checked out to run locally. How would you coordinate those checkouts otherwise? You could check it out inside the genny repo but then do you still use evergreen modules?

dalyd: Some more thoughts:

1. These scripts are not in separate repos. They are in the same repo. If we use relative paths for in the same repo (current problem), it doesn't matter where you call ./run-genny at all. This would also be consistent to how we handle load_config. 
2. I grant that cross repo scripts would be an entirely different ball of wax, but that's not what we have here. 
3. The original purpose of the work directory was to free the user from having to run in a pre-ordained place. We seem to be moving back to having to work in a pre-ordained location or else things will fail in non-helpful ways. 

I have some other review notes to send also -- let me finish my open review, and then we can talk about how we get Brad and this work over the line.

dalyd (commented):
First, while I have some discussion still on design, the implementation of the current design (use environmental variables) looks good to me. If we can figure out a way that avoids the environmental variables, I would prefer that, but I also don't want to hold this work up if it depends on infrastructure that does not exist. 

Separately, we have a path discussion :)

src/workloads/c2c/MongosyncScripts.yml
dalyd: For the Genny case here, do any of us imagine a use case in which cross repo scripts exists and are called from within a workload? 

If we do, I'd suggest we consider supporting relative and absolute paths with different names. If we can't, I say we focus on the one repo case.

brad-devlugt:
New patch to check the end-end tests are working correctly: https://spruce.mongodb.com/version/6376897057e85a3d10e1b2b6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

huan-Mongo (approved):
LGTM!

dalyd (approved):

brad-devlugt:
evergreen merge

brad-devlugt:
evergreen merge

PR #755, opened 2022-10-25
PERF-3399 Extend NLJ lookup workload

leozzx:
[sys-perf](https://spruce.mongodb.com/version/6358765557e85a4d674fd39d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


src/phases/execution/LookupCommands.yml
```diff
@@ -59,6 +59,16 @@ ForeignSmallDocTemplate:
   arr_key: {^Array: {of: *RandIntUpTo, number: 5}}
   data1: *Rand32BitInt
 
+# The document template for both local and foreign collections in NLJ Lookups. The doc contains
+# scalar, single element array, 8 elements array, simple object and nested object.
+LargeComplexDocTemplate:
+  int_key1: &RandInt0To1k {^RandomInt: {min: 0, max: 1000}}
+  int_key2: {^RandomInt: {min: 2000, max: 3000}}
```
mtandrei, line 66: Should this be from 10000 to 12000:
```
r4: Random.randInt(10000) + 2000,
````
?

This might not matter however, since my understanding of r4 is that it's a value that's not meant to match any of the others (so as to stress the join logic, and not the logic to materialize documents).

Speaking of: in addition to the tests from https://docs.google.com/document/d/1XGMbzMrIrY1JX7uWOIbKjcdJMlQfu5cdcxtnIXVJ3gY/edit#heading=h.if6c26hl7ubr do you know if we have tests that test NLJ that produce matches? i.e. do you know if https://github.com/10gen/PerformanceReports/blob/main/scripts/tags.yml#L990-L994 produce a non-empty "as" array? This would be good to have in order to verify that the cost of materializing documents isn't a problem in SBE.

File: src/workloads/execution/lookup/LookupNLJ.yml [hunk 2]
```diff
@@ -0,0 +1,345 @@
...
+Owner: "@mongodb/query-execution"
+Description: |
+  This workload measures $lookup performance for scenarios that look up local scalar/array/dotted
+  key against foreign collection's scalar/array/dotted key, when the foreign collections have
+  multiple collection cardinalities such as 1000 / 10000. Note that 100k foreign collection
+  cardinality will make the testing duration too long, so we skip it.
+
+  While measuring performances, this workload collects numbers for either the SBE or the classic
+  engine $lookup implementations, depending on environments that it runs on. Both SBE and classic
+  $lookup implements will be nested loop join for this workload.
```
mtandrei, line 11: [nit] s/implements/implementations ?

src/phases/execution/LookupCommands.yml
```diff
@@ -59,6 +59,16 @@ ForeignSmallDocTemplate:
...
+  arr1: [*RandInt0To1k]
+  arr8: {^Array: {of: *RandInt0To1k, number: 8}}
+  obj1: {obj: *RandInt0To1k}
+  obj4: {obj3: {obj2: {obj1: {obj: *RandInt0To1k}}}}
```
mtandrei, line 70: [nit] recommend renaming 'obj' to something referencing that the value is an int. Here and above.

src/workloads/execution/lookup/LookupNLJ.yml
```diff
@@ -0,0 +1,345 @@
...
+  - *Nop
+  - *Nop
+
+- Name: LoadLocalDoc100k
+  Type: CrudActor
+  Database: *db
+  Threads: 10
+  Phases:
+  - *Nop
+  - Repeat: 10000
```
mtandrei, line 300: should this be 100k? or is this meant to be 10k?

src/workloads/execution/lookup/LookupNLJ.yml
```diff
@@ -0,0 +1,345 @@
...
+  - Phase: 1..3
+    Nop: true
+
+# The 'LoadForeign1K' / 'LoadForeign10K' / 'LoadLocalDoc100k' actors run in the 1st phase.
+- Name: LoadForeign1K
+  Type: CrudActor
+  Database: *db
+  Threads: 1
+  Phases:
+  - &Nop {Nop: true}
```
mtandrei, line 267: Instead of running 'insertOne' 1k or 10k times, could this use 'insertMany' or BulkWrite ops? Something like so: 
https://github.com/mongodb/genny/blob/master/src/workloads/docs/CrudActor.yml#L37-L42 or https://github.com/mongodb/genny/blob/master/src/cast_core/include/cast_core/actors/CrudActor.hpp#L60-L67 ?

src/workloads/execution/lookup/LookupNLJ.yml
```diff
@@ -0,0 +1,345 @@
...
+  - *Nop
+  - Repeat: 1
+  - *Nop
+
+# The 'NLJLookup' actor measures the NLJ lookup performances and runs in the 3rd phase.
+- Name: NLJLookup
+  Type: RunCommand
+  Database: *db
+  Phases:
+  - Phase: 0..2
```
mtandrei, line 327: [question] Am I correct that this is saying 'noop during phases 0 though 2', similar to how the setup above is a no op from phases 1 to 3?

src/phases/execution/LookupCommands.yml
leozzx: oh, you are elaborative , I didn't find this after reading it many times.
For your second question, `Foreign9500_IntTopField` query will for sure to produce non-empty result, other queries could have non-empty result depends on random number.

src/workloads/execution/lookup/LookupNLJ.yml
leozzx: There're 10 thread for this actor, each thread will generate 10k docs.

src/workloads/execution/lookup/LookupNLJ.yml
leozzx: I believe the insertMany and BulkWrite op only insert the listed documents once, not repeatedly one document many times. So I have to list all 10k documents in order to use one insertMany, or I can list 10 documents and repeat it 1k times, I don't think this is better than using insertOne.

src/workloads/execution/lookup/LookupNLJ.yml
leozzx: yes.

leozzx:
I have moved the yaml files out of `lookup` folder, evergreen doesn't recognize it.
[new sys-perf patch](https://spruce.mongodb.com/version/63596c533e8e86414ebe58ca/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

src/phases/execution/LookupCommands.yml
mtandrei: Ack; resolving.

mtandrei (approved):
LGTM; thanks Zixuan! It looks like the task times from this view are under an hour for each variant: https://spruce.mongodb.com/version/63596c533e8e86414ebe58ca/task-duration?duration=DESC

mtandrei:
One note before you merge: please make sure you update the PerformanceReports repository before merging like so: https://github.com/10gen/PerformanceReports/commit/8ca708b7dd65cd6b80d54786a6ecf7372efeeedb

PR #760, opened 2022-11-01
PERF-3464: Add SSB query workload against CSI.

ccstolley:
This adds the SSB queries that Colby adapted from the SQL equivalents (see https://docs.google.com/spreadsheets/d/1MEP1-mhUC9bHjvk9kIA6xiAHjQW42MdExOH59IJsJBM/edit#gid=1782429438 ).

The workload borrows from our existing CSI genny workload, by running the queries against a warm buffer cache first with CSI enabled, then without.


colbying (commented):
Queries themselves look good to me

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,595 @@
...
+          "customer_nation": "$_id.c_nation",
+          "part_brand": "$_id.p_brand"
+        }
+      },
+      {"$sort": {"order_year": 1, "customer_nation": 1, "part_brand": 1}}
+    ]
+  cursor: {}
+
+# Defines test operations.
+# Here we use explain with verbosity = execustionStats instead of using plain aggregate command
```
IrinaYatsenko, line 438: David Storch recently updated the CRUD actor to support aggregation (https://github.com/mongodb/genny/commit/bba39f66d82d02bfc9c064e8b3ec688e3a76f46a). Could it be used now instead of the explain-based workaround?

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,595 @@
...
+ColumnStoreIndexName: &ColumnStoreIndexName $**_columnstore
+
+Actors:
+- Name: Setup
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [*SetupPhase]
+      NopInPhasesUpTo: *MaxPhases
```
IrinaYatsenko, line 522: For my education: is the `NopInPhaseUpTo` field still required when `Active` is provided?

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,595 @@
...
+  Type: RunCommand
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [*ColumnstoreScanPhase]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *TestRepeatCount
+        Database: *db
+        Operations: *TestOperations
```
IrinaYatsenko, line 564: I presume this will run the operations in the order they are listed with no cleanup/reset of the actor between them, right? That is, the caching pattern will depend on the order of the queries. Whether that matters of not, I have no idea

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,595 @@
...
+  Type: RunCommand
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [*CollScanPhase]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *TestRepeatCount
+        Database: *db
+        Operations: *TestOperations
```
IrinaYatsenko, line 595: If there was rowstore fallback when running with CSI this might "warm up" the data for the collscan. Again, I have no idea whether that actually happens or matters.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: Genny crashes if you omit it, so I'd say yes.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: Resolved in 9d03739054cac7e76273af8b1f82a7d0005765e7

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: That's correct. My assumption is that with a data set larger than the size of RAM, a warm cache won't matter much.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: That's definitely true; some of these queries aren't able to use CSI yet. My as above, I think the size of the dataset make this inconsequential.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+  OperationCommand: *Q4_1Query
+- OperationMetricsName: Q4_2Query
+  OperationName: aggregate
+  OperationCommand: *Q4_2Query
+- OperationMetricsName: Q4_3Query
+  OperationName: aggregate
+  OperationCommand: *Q4_3Query
+
+# Defines how many times each test operation is executed.
+TestRepeatCount: &TestRepeatCount 1
```
jimoleary, line 451: Should you should run each test more than once? I found that there is a large degree of difference between the first and last run. 

Have you performed any analysis of how noisy this workload is likely to be? For workloads for more mature features we would like there to be limited noise, but I suspect that this may not be the case for this feature.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+          "order_year": "$_id.order_year",
+          "customer_nation": "$_id.c_nation",
+          "part_brand": "$_id.p_brand"
+        }
+      },
+      {"$sort": {"order_year": 1, "customer_nation": 1, "part_brand": 1}}
+    ]
+
+# Defines test operations.
+TestOperations: &TestOperations
```
jimoleary, line 418: I think you would probably be better off to run these queries separately in isolation as this would allow you to repeat them to both warm them up and get an average performance.

src/workloads/execution/SSBColumnStoreIndex.yml
jimoleary: As I noted earlier, I suspect you would be better off to run them separately.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  This workload compares the query performance of SSB (star schema benchmark) queries with columnar indexes to performances without them.
+
+  The workload consists of the following phases and actors:
+    1. Run SSB queries with columnstore indexes.
+    2. Drop columnstore indexes.
+    3. Run SSB queries without columnstore indexes.
```
jimoleary, line 9: So this workload is comparing column store indexes against no indexes (colscans). Is that a fair comparison? Should we also run the queries against regular indexes?

jimoleary (commented):
I think that the queries should be separated to allow them to be run more flexibly (and multiple times). 

The thing that mostly surprises me about this work is that CSI is being compared to the unindexed case.

FWIW, the following is what we came up with in the comparative project.

Queries:
* Q1
  * [mongo-q1-1.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q1-1.sh).
  * [mongo-q1-2.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q1-2.sh).
  * [mongo-q1-3.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q1-3.sh).
* Q2
  * [mongo-q2-1.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q2-1.sh).
  * [mongo-q2-2.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q2-2.sh).
  * [mongo-q2-3.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q2-3.sh).
* Q3 (I am less confident of these) 
  * [mongo-q3-1.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q3-1.sh).
  * [mongo-q3-2.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q3-2.sh).
  * [mongo-q3-3.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q3-3.sh).
  * [mongo-q3-4.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q3-4.sh).
* Q4 (again not confident)
  * [mongo-q4-1.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q4-1.sh).
  * [mongo-q4-2.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q4-2.sh).
  * [mongo-q4-3.sh](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/mongo-q4-3.sh).

The supporting [indexes](https://github.com/jimoleary/dsi/blob/PERF-3277-RDS/terraform/remote-scripts/ssb-mongo.sh#L110-L119) (although we didn't take Q3 / Q4 into account) and this is less relevant for you.


Backing [datasets](https://s3.console.aws.amazon.com/s3/buckets/comparative?prefix=ssb-tables/&region=us-east-1)

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: Yes, we want to see the speed-up (or penalty) compared to collection scans. In particular we want to evaluate CSI in contexts where the collection does not fit into memory.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+
+  The DSI test_control will drop caches inbetween each test to measure cold cache performance.
+
+Keywords:
+- columnstore
+- analytics
+- SSB
+
+Clients:
```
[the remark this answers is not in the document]
ccstolley: I think for the Q3 and Q4 queries, they can have an especially long runtime, particularly when doing collection scans that exceed the size of RAM, so I specified the longer timeout to accommodate those queries.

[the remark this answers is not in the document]
ccstolley: The comment there is misleading though, I should fix that.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+Collection: &Collection order_lineitem
+
+Q1_1Query: &Q1_1Query
+  Pipeline:
+    [
+      {
+        "$match": {
+          "order.o_orderdate": {
+            "$gte": ISODate("1992-01-01T00:00:00Z"),
+            "$lte": ISODate("1996-01-01T00:00:00Z")
```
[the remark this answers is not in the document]
ccstolley: I'm not sure how these ranges were chosen (@colbying would know) but I agree that it's odd to include the first second of 1996 here. I'll fix that.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+    ]
+
+Q1_2Query: &Q1_2Query
+  Pipeline:
+    [
+      {
+        "$match": {
+          "order.o_orderdate": {
+            "$gte": ISODate("1994-01-01T00:00:00.000Z"),
+            "$lte": ISODate("1994-02-01T00:00:00.000Z")
```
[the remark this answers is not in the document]
ccstolley: I agree. Thanks!

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+          revenue: { $sum: { $multiply: ['$l_extendedprice', '$l_discount'] } }
+        }
+      }
+    ]
+
+Q2_1Query: &Q2_1Query
+  Pipeline:
+    [
+      { $match: { 'part.p_brand': 'Brand#21' } },
+      {
```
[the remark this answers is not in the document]
ccstolley: The pipeline reorder probably wouldn't buy us much since we only use CSI and no traditional indexes.
As for further denormalizing, I can see why that would improve performance, but I believe we're trying to keep the schema as close to the original SSB benchmark as possible.

src/workloads/execution/SSBColumnStoreIndex.yml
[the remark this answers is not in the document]
ccstolley: addressed in 4d5a387c47f4243c17cb94a26c147569c42c2075

src/workloads/execution/SSBColumnStoreIndex.yml
IrinaYatsenko: What would it take in terms of the task's run-time to run the queries separately? If we run two patches with the queries ordered differently and see no significant difference, would that be sufficient "proof" of the expectation that warm cache doesn't matter much?

src/workloads/execution/SSBColumnStoreIndex.yml
IrinaYatsenko: while I tend to agree with you that the size of the dataset _should_ make this inconsequential, with performance our intuitions are often wrong... is there any way to confirm our expectations? I also asked on [on Slack](https://mongodb.slack.com/archives/C01VD0LQZED/p1668536753264619) whether we could observe actual disk activity for genny workloads.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+        }
+      }
+    ]
+
+Q2_2Query: &Q2_2Query
+  Pipeline:
+    [
+      {
+        $match: {
+          'part.p_brand': {
```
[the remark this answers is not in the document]
ccstolley: Right, I think for our purposes (evaluating CSI performance gains/regressions over time) this is probably ok.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+          "order_year": "$_id.order_year",
+          "customer_nation": "$_id.c_nation",
+          "supplier_nation": "$_id.s_nation"
+        }
+      },
+      {"$sort": {"order_year": 1, "revenue": -1}}
+    ]
+
+# Q4 queries are (currently) limited to 10000 documents because they take many hours to finish.
+Q4_1Query: &Q4_1Query
```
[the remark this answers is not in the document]
ccstolley: Right, the multiple lookups drag us down here, which is why we had to limit it to 10000 documents for now. Denormalizing to avoid them is an interesting idea though.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+        }
+      },
+      { $match: { 'customer.nation.region.r_name': 'AMERICA' } },
+      {
+        $group: {
+          _id: {
+            order_year: { $year: '$order.o_orderdate' },
+            c_nation: '$customer.nation.n_name'
+          },
+          revenue: {
```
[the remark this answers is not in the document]
ccstolley: Fixed in 02875ce3f9d5d5716e3e2ba361be98a8eb983b49

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+      },
+      {"$addFields": {"customer": {"$first": "$customer"}}},
+      {"$match": {"customer.nation.region.r_name": "AMERICA"}},
+      {
+        "$group": {
+          "_id": {
+            "order_year": {"$year": "$order.o_orderdate"},
+            "c_nation": "$customer.nation.n_name"
+          },
+          "revenue": {
```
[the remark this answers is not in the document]
ccstolley: Fixed in 02875ce3f9d5d5716e3e2ba361be98a8eb983b49

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+      {"$addFields": {"customer": {"$first": "$customer"}}},
+      {"$match": {"customer.nation.region.r_name": "AMERICA"}},
+      {
+        "$group": {
+          "_id": {
+            "order_year": {"$year": "$order.o_orderdate"},
+            "c_nation": "$customer.nation.n_name",
+            "p_brand": "$part.p_brand"
+          },
+          "revenue": {
```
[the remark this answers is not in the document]
ccstolley: Fixed in 02875ce3f9d5d5716e3e2ba361be98a8eb983b49

src/workloads/execution/SSBColumnStoreIndex.yml
IrinaYatsenko: Genny workloads collect measurements such as disk utilization (%) and reads/writes per second (mean, max, p99). The measurements are collected per phase, so if we split each query to run in its own phase (with or without explicit cleanup in between) at least we can observe some aspects of disk I/O behaviour? And I filed PERF-3535 to collect page faults which would be a better metric for our purposes.

src/workloads/execution/SSBColumnStoreIndex.yml
jimoleary: Hmm,  I think that it should be possible for genny to calculate MaxPhases automatically if is is not supplied or at least fail with a reasonable error message (rather than crashing). I've created [PERF-3536](https://jira.mongodb.org/browse/PERF-3536) to track this.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+      },
+      {
+        $group: {
+          _id: {
+            order_year: {
+              $year: '$order.o_orderdate'
+            },
+            brand: '$part.p_brand'
+          },
+          revenue: { $sum: { $multiply: ['$l_extendedprice', '$l_discount'] } }
```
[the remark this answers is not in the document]
ccstolley: Like above, not needed in yaml context.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: I've split out separate phases and also added quiesces in d59b8bd164c3fe074a913a942a52bb4282051e55. Thanks!

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: That is a good question. I'm running an evergreen build now that runs the tests 3 times to see how variable our numbers are: https://spruce.mongodb.com/version/637527a857e85a3432f91829/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+              'Brand#21',
+              'Brand#22',
+              'Brand#15',
+              'Brand#41',
+              'Brand#41'
+            ]
+          }
+        }
+      },
+      {
```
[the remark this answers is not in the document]
ccstolley: Yeah that makes sense for general performance comparisons, but I think for CSI probably not.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+Q2_3Query: &Q2_3Query
+  Pipeline:
+    [
+      {
+        $match: {
+          'part.p_brand': 'Brand#21',
+          'part.p_type': 'ECONOMY BRUSHED STEEL'
+        }
+      },
+      {
```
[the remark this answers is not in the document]
ccstolley: :+1: We'll stick to the normalized version here to stay as close to SSB as possible.

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,640 @@
...
+          "customer_nation": "$_id.c_nation",
+          "part_brand": "$_id.p_brand"
+        }
+      },
+      {"$sort": {"order_year": 1, "customer_nation": 1, "part_brand": 1}}
+    ]
+
+# Defines how many times each test operation is executed.
+TestRepeatCount: &TestRepeatCount 1
+MaxPhases: &MaxPhases 42
```
IrinaYatsenko, line 420: Is it possible to compute in yaml? I mean, have a variable for NumberOfTests, then set MaxPhases to `&NumberOfQueries * 2 +  2` (or however one would arrive to the number 42). A similar question for the elements of the `Active` arrays to make it easier in the future to add/drop a query and make it easier to understand how the phases are related to each other.

IrinaYatsenko (dismissed):
LGTM with suggestions (but I haven't looked at the actual queries for the tests)

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: This is a fair point. I'm going to have to reorganize the phases again because I noticed that now it is impossible to distinguish the CSI and non-CSI metrics (they use the same name--woops). I'll see if I can make the phase numbers more intuitive though.

ccstolley: See [edca461](https://github.com/mongodb/genny/pull/760/commits/edca461bdcf65f9eb2bbcbe7fa1306fc459b7016). It's more flat-footed about the phase numbers, which hopefully makes it easier to add queries in the future. It's still not ideal though.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: It seems like there's a fair amount of variability between the p99 and p50, so I'll run them 3 times instead of once. Addressed in a644af8c93905304760009ee145437dc211f6771 .

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,542 @@
...
+            },
+            brand: '$part.p_brand'
+          },
+          revenue: { $sum: { $multiply: ['$l_extendedprice', '$l_discount'] } }
+        }
+      }
+    ]
+
+Q3Query: &Q3Query
+  Pipeline:
```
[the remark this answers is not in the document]
ccstolley: :+1: we'll pass on the denormalization, but good to know.

src/workloads/execution/SSBColumnStoreIndex.yml
IrinaYatsenko: I wonder, whether having so many phases with a single actor in each is an indicator that Genny isn't the best tool for the testing we are trying to do here (this is not a critic of the patch! just thinking aloud)

IrinaYatsenko (approved):

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,791 @@
...
+  Threads: 1
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2, 4, 6, 8, 10, 12, 14, 16, 18, 20, 23, 25, 27, 29, 31, 33, 35, 37, 39, 41]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+# Run each SSB query in a separate phase with column store index.
```
jimoleary, line 473: Have you considered using an actor template? 

For larger workloads like this one, it will cut down on the size of the yml file, and more importantly it will mean less editing if you ever need to add / change something. For example in your case you would do something like:

```
# define before the actors.
ActorTemplates:
- TemplateName: SSB.Query
  Config:
    Name: {^Parameter: {Name: "Name", Default: "Q1_1"}}
    Type: CrudActor
    Phases:
      OnlyActiveInPhases:
        Active: {^Parameter: {Name: "Active", Default: [3]}}
        NopInPhasesUpTo: *MaxPhases
        PhaseConfig:
          Repeat: *TestRepeatCount
          Database: *db
          Collection: *Collection
          Operations:
          - OperationMetricsName: {^Parameter: {Name: "Name", Default: "Q1_1"}}
            OperationName: aggregate
            OperationCommand: {^Parameter: {Name: "OperationCommand", Default: *Q1_1Query}}

# skip
.
.
.


Actors:
# skip
.
.
.
.

- ActorFromTemplate:
    TemplateName: SSB.Query
    TemplateParameters:
      Name: Q1_1

- ActorFromTemplate:
    TemplateName: SSB.Query
    TemplateParameters:
      Name: Q1_2
      OperationCommand: *Q1_2Query
      Active: [5]

- ActorFromTemplate:
    TemplateName: SSB.Query
    TemplateParameters:
      Name: Q1_3
      OperationCommand: *Q1_3Query
      Active: [7]

- ActorFromTemplate:
    TemplateName: SSB.Query
    TemplateParameters:
      Name: Q2_1
      OperationCommand: *Q2_1Query
      Active: [9]

# etc/


# run same queries without column store index (Collection Scan)
- ActorFromTemplate:
    TemplateName: SSB.Query
    TemplateParameters:
      Name: Q1_1CollScan
      OperationCommand: *Q1_1Query
      Active: [24]

# etc.
```

Note: the metrics names will end up being the Template Parameter name but I don't see that as terribly significant.

You can view the processed workload with 
```
./run-genny evaluate src/workloads/execution/SSBColumnStoreIndex.yml
```

src/workloads/execution/SSBColumnStoreIndex.yml
```diff
@@ -0,0 +1,791 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: 1
+        Database: *db
+        Operations:
+        - OperationName: RunCommand
+          OperationCommand:
+            dropIndexes: *Collection
+            index: *ColumnStoreIndexName
```
jimoleary, line 640: [suggestion] Feel free to ignore, but if it were me, I would consider putting the DropColumnstoreIndex in phase 23 and changing the Quieces to all be in even phases. Then document that odd phases are setup and even phases are test related (except for the drop)  in the  the workload header. 

Operationally this shouldn't affect anything, it's only an extra fsync between CSI Q4_3 and the drop index.

src/workloads/execution/SSBColumnStoreIndex.yml
jimoleary: We would normally recommend running the task that you expect to deploy 3 to 5 times and then comparing the the results:

  
image


That is, compare the results from different execution numbers. Re-running the patch tasks means that you don't have to compile it every time.

jimoleary (commented):
I have left 2 suggestions, but the workload is looking good to me.

I'd like us to get a feel for how noisy the workload is likely to be before you deploy it.

If the noise level is high, we should inform the barons of this and determine whether there is work to be done before this workload is to be monitored by them.

src/workloads/execution/SSBColumnStoreIndex.yml
ccstolley: Makes sense, fixed in 88de40e2b8e3ed8b2d3a82106f435cd65d00ba3a

jimoleary:
>  How would you recommend we go about that? 
I don't see how the Perf Analyzer would help here as the workload hasn't been deployed.

Generally I would just execute the task 3 to 5 times in your [patch](https://spruce.mongodb.com/task/sys_perf_linux_standalone_all_feature_flags_ssb_column_store_comparison_patch_45524602443c5d913bf8957de1e57d6be0451a92_637806267742ae1a5767669d_22_11_18_22_25_48/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) and then compare the spread of the results.

ccstolley:
> Generally I would just execute the task 3 to 5 times 

Sounds good, I ran it 3 times, results under https://spruce.mongodb.com/task/sys_perf_linux_standalone_all_feature_flags_ssb_column_store_comparison_patch_45524602443c5d913bf8957de1e57d6be0451a92_637806267742ae1a5767669d_22_11_18_22_25_48/trend-charts?execution=1&sortBy=STATUS&sortDir=ASC but by and large the results seem to fall reasonably close together across executions. Perhaps that's aided by the fact that I'm running the queries 3 times already.

jimoleary:
It looks like our rubric is (max - min) / median < 10%. So for your patch results:

```
export tid=sys_perf_linux_standalone_all_feature_flags_ssb_column_store_comparison_patch_45524602443c5d913bf8957de1e57d6be0451a92_637806267742ae1a5767669d_22_11_18_22_25_48
(getCedar --headers ${tid}  -e -1 -s ' ' ) | egrep '(test_name|Q[1-3]_[1-3]\.Q.* OperationT)' | tee csi.txt | python -c 'import pandas as pd; print(pd.read_csv("/dev/stdin", delim_whitespace=True).sort_values("test_name").groupby("test_name")["result"].apply(lambda x: 1- abs(x.max() - x.min() / x.median())))'
test_name
Q1_1.Q1_1    0.244930
Q1_2.Q1_2    0.259578
Q1_3.Q1_3    0.255830
Q2_1.Q2_1    0.048016
Q2_2.Q2_2    0.030740
Q2_3.Q2_3    0.149006
Name: result, dtype: float64

(getCedar --headers ${tid}  -e -1 -s ' ' ) | egrep '(test_name|Q[1-3]_[1-3]\.No.*Q.* OperationT)' | tee csi.txt | python -c 'import pandas as pd; print(pd.read_csv("/dev/stdin", delim_whitespace=True).sort_values("test_name").groupby("test_name")["result"].apply(lambda x: 1- abs(x.max() - x.min() / x.median())))'
test_name
Q1_1.NoIndex.Q1_1.NoIndex    0.041186
Q1_2.NoIndex.Q1_2.NoIndex    0.066794
Q1_3.NoIndex.Q1_3.NoIndex    0.064207
Q2_1.NoIndex.Q2_1.NoIndex    0.038246
Q2_2.NoIndex.Q2_2.NoIndex    0.029400
Q2_3.NoIndex.Q2_3.NoIndex    0.072811
Name: result, dtype: float64
```

So the CSI indexes may be a bit noisy.

ccstolley:
@jimoleary I'm not sure I follow your numbers there. I took the raw json and came up with:
```
jq -c '[.[]|select(.info.test_name |match("Q1_1.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.17146776406035666,0.15924412123785764,0.17364125716270185]
 jq -c '[.[]|select(.info.test_name |match("Q1_2.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.17410481109627995,0.15916808149405773,0.1740442072286361]
jq -c '[.[]|select(.info.test_name |match("Q1_3.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.17394329448599757,0.15931177314003506,0.17352073572791948]
jq -c '[.[]|select(.info.test_name |match("Q2_1.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.04625560849252972,0.046016504586311624,0.04593547596809016]
jq -c '[.[]|select(.info.test_name |match("Q2_2.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.01551093003536492,0.015464475522312662,0.01522896752676491]
jq -c '[.[]|select(.info.test_name |match("Q2_3.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json
[0.13630168105406631,0.1342041692761922,0.13593112822836428]
```
Which are the same numbers in the web ui for `OperationThroughput`. Just eye-balling it, those numbers are very close, though the middle run seems lower across the board. Taking Q1_2 as a worst case, I see:
```
jq -c '[.[]|select(.info.test_name |match("Q1_2.Q"))| .rollups.stats |map(select(.name == "OperationThroughput"))| .[].val]' output.json | python -c 'import sys, statistics as stats; runs = eval(sys.stdin.read()); print((max(runs) - min(runs))/stats.median(runs))'
0.0858214693845015
```
which seems pretty reasonable, if 10% is the cut-off.

jimoleary (approved):
Oops, my calculation was not the one I sent you so you are correct.

ccstolley:
evergreen merge

ccstolley:
evergreen merge

PR #757, opened 2022-10-27
PERF-3476 add workload for $group queries that spill to disk

dstorch:
Proposed detailed commit message:

> Also extends the 'CrudActor' to support aggregate operations.

Hopefully reviewers are ok with lumping in the changes to the `CrudActor` here. The lack of support for "aggregate" in the `CrudActor` has led to the use of explain in genny tests in the past (e.g. for TPC-H; see [PERF-3251](https://jira.mongodb.org/browse/PERF-3251)), so it seemed useful to provide this functionality to support future agg perf tests.

I kicked off a sys-perf patch build here: https://spruce.mongodb.com/version/635ad15c5623435c5fb6d9a3/tasks

And a genny patch build here: https://spruce.mongodb.com/version/635ad2360305b91ce272e733/tasks

PR #761, opened 2022-11-02
PERF-3485 Extend 'CrudActor' to support the aggregate command

dstorch:
Proposed detailed commit message:

> Aggregate, like find, returns a cursor. We often want to measure the time taken to fully iterate the cursor, including the time to process the initial request as well as any subsequent getMore commands. Past workarounds obviated by this change include using manipulating the batchSize to ensure that the complete reply fits in a single batch, or using explain with "executionStats" verbosity.

This is depended on by [PERF-3476](https://jira.mongodb.org/browse/PERF-3476) (see https://github.com/mongodb/genny/pull/757).

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -699,6 +699,65 @@ private:
     metrics::Operation _operation;
 };
 
+struct AggregateOperation : public BaseOperation {
+    struct Pipeline {
+        Pipeline(const Node& node, PhaseContext& context, ActorId id) {
+            if (!node.isSequence()) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException("'Pipeline' must be an array"));
+            }
+            for (auto&& [_, stageNode] : node) {
+                DocumentGenerator stage = stageNode.to<DocumentGenerator>(context, id);
+                pipeline.append_stage(stage());
+            }
+        }
+
+        bsoncxx::document::value copyPipelineToDocument() const {
+            auto arrayView = pipeline.view_array();
+            bsoncxx::document::view docView{arrayView.data(), arrayView.length()};
+            return bsoncxx::document::value{docView};
+        }
+
+        mongocxx::pipeline pipeline;
+    };
+
+    AggregateOperation(const Node& opNode,
+                       bool onSession,
+                       mongocxx::collection collection,
+                       metrics::Operation operation,
+                       PhaseContext& context,
+                       ActorId id)
+        : BaseOperation(context, opNode),
+          _onSession{onSession},
+          _collection{std::move(collection)},
+          _operation{operation},
+          _pipeline{opNode["Pipeline"].to<Pipeline>(context, id)} {
+        if (opNode["Options"]) {
+            _options = opNode["Options"].to<mongocxx::options::aggregate>();
+        }
+    }
+
+    void run(mongocxx::client_session& session) override {
+        auto& pipeline = _pipeline.pipeline;
+        this->doBlock(_operation, [&](metrics::OperationContext& ctx) {
+            auto cursor = _onSession ? _collection.aggregate(session, pipeline, _options)
+                                     : _collection.aggregate(pipeline, _options);
+            for (auto&& doc : cursor) {
+                ctx.addDocuments(1);
+                ctx.addBytes(doc.length());
+            }
+            return _pipeline.copyPipelineToDocument();
+        });
+    }
+
+
+private:
+    bool _onSession;
+    mongocxx::collection _collection;
+    mongocxx::options::aggregate _options;
+    Pipeline _pipeline;
+    metrics::Operation _operation;
+};
+
 struct FindOneAndUpdateOperation : public BaseOperation {
     FindOneAndUpdateOperation(const Node& opNode,
                               bool onSession,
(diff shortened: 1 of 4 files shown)


src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -673,3 +673,30 @@ Tests:
...
+      - OperationName: bulkWrite
+        OperationCommand:
+          WriteOperations:
+            - WriteCommand: insertOne
+              Document: { a: "foo", b: 3 }
+            - WriteCommand: insertOne
+              Document: { a: "foo", b: 4 }
+      - OperationName: aggregate
+        OperationCommand:
+            Pipeline: [{$group: {_id: "$a", sum: {$sum: "$b"}}}, {$out: "test"}]
```
dstorch, line 688: Relying on `$out` here is maybe not ideal, but I couldn't figure out a better way to do it. For one thing, I didn't see any tests in this file for `OperationName: find`. The reason, I guess, is that genny throws away the results of the cursor, so there is no way to easily validate that the query is returning what we expect it to return. In the case of `$out`, the results of the query are persisted, so they can subsequently be validated using `OutcomeData`.

I considered testing this by trying to expose the metrics that are accounted for [here](https://github.com/mongodb/genny/pull/761/files#diff-d51310790eeb6ac904ed06a273fe737958403033c54d2a613a339a1cc61b8c3eR745-R746). I don't know the genny code too well, but it seemed to me like this wouldn't be very natural to implement. I'm also not sure if there are other genny self-tests that make assertions about metrics like these.

src/cast_core/src/CrudActor.cpp
```diff
@@ -699,6 +699,65 @@ struct FindOneOperation : public BaseOperation {
...
 
+struct AggregateOperation : public BaseOperation {
+    struct Pipeline {
+        Pipeline(const Node& node, PhaseContext& context, ActorId id) {
+            if (!node.isSequence()) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException("'Pipeline' must be an array"));
+            }
+            for (auto&& [_, stageNode] : node) {
+                DocumentGenerator stage = stageNode.to<DocumentGenerator>(context, id);
+                pipeline.append_stage(stage());
```
rtimmons, line 710: This is being evaluated at constructor-time versus for each iteration. Result is you'll have a random pipeline but each invocation will use the same random pipeline. Is that intended? If not, you can instead collect a vector of `DocumentGenerator`s (one for each stage) and then construct the pipeline by calling the generators in the `run` block.

src/cast_core/src/CrudActor.cpp
```diff
@@ -699,6 +699,65 @@ struct FindOneOperation : public BaseOperation {
...
+                       bool onSession,
+                       mongocxx::collection collection,
+                       metrics::Operation operation,
+                       PhaseContext& context,
+                       ActorId id)
+        : BaseOperation(context, opNode),
+          _onSession{onSession},
+          _collection{std::move(collection)},
+          _operation{operation},
+          _pipeline{opNode["Pipeline"].to<Pipeline>(context, id)} {
```
rtimmons, line 733: Nice!

src/cast_core/test/CrudActorYamlTests.yml
rtimmons: This is actually pretty clever. Consider putting a version of your comment in the code, though, because the `$out` was easy (for me) to miss / understand why it was there without your verbiage.

It doesn't seem worthwhile to change much testing infra to support an alternative and I can't think of anything better off-hand, so my vote is to leave this and potentially reconsider if there's additional use-cases in the future.

src/cast_core/src/CrudActor.cpp
```diff
@@ -699,6 +699,65 @@ struct FindOneOperation : public BaseOperation {
     metrics::Operation _operation;
 };
 
+struct AggregateOperation : public BaseOperation {
+    struct Pipeline {
+        Pipeline(const Node& node, PhaseContext& context, ActorId id) {
+            if (!node.isSequence()) {
+                BOOST_THROW_EXCEPTION(InvalidConfigurationException("'Pipeline' must be an array"));
```
rtimmons, line 706: Consider a test of this check.

rtimmons (approved):
Nice little change. Thanks for rolling up your sleeves here, and thanks for adding automated testing.

Approved assuming you want the doc-generator behavior. I'd be happy to take another look or provide more input/help if you decide you want runtime document-generation of the pipeline.

src/cast_core/src/CrudActor.cpp
dstorch: Oh, great catch! I didn't think that deeply about _when_ the generator was being called when I implemented this. You're right that it makes more sense to do in the `run()` method. Done.

src/cast_core/src/CrudActor.cpp
dstorch: 😄 

I thought the API for converting the YAML to typed constructs is pretty nifty.

src/cast_core/test/CrudActorYamlTests.yml
dstorch: > This is actually pretty clever. Consider putting a version of your comment in the code, though, because the $out was easy (for me) to miss / understand why it was there without your verbiage.

Done. Let me know what you think of the comment.

> It doesn't seem worthwhile to change much testing infra to support an alternative and I can't think of anything better off-hand, so my vote is to leave this and potentially reconsider if there's additional use-cases in the future.

Sounds good! That's pretty much where I landed as well.

dstorch (commented):
Thanks for the review @rtimmons! I pushed a new commit addressing your feedback. Please take another look when you have a moment.

rtimmons (approved):
Love it! Thanks for iterating on feedback. This turned out great.

PR #763, opened 2022-11-03
PERF-3487 Add artificial TPC-H queries for columnstore index

awagenmaker:


src/phases/tpch/denormalized/AvgAcctBal.yml
```diff
@@ -0,0 +1,44 @@
+SchemaVersion: 2018-07-01
```
awagenmaker, line 1: I could put all of these in an `artifical` subdirectory or somewhere else entirely if we want them separated from the standard queries. Wasn't sure where the best spot would be.

src/workloads/tpch/denormalized/10/AvgAcctBal.yml
```diff
@@ -0,0 +1,19 @@
+SchemaVersion: 2018-07-01
```
awagenmaker, line 1: I only added these queries for scale 10, but it would be easy to add to scale 1 as well if we wanted to run them on that in future (or if we want to do that now, I can add them in this PR).

starzia:
I know the term "artificial" was in the ticket, but I don't see why are we calling these "artificial" queries.  Is it just that these are new queries not in the offical TPCH benchmark suite?  If that's the reasoning, can we give them a more positive name, like "extra", "extended", "internal", or "columnstore"?

alyacb (commented):
I thought I'd have a look at this PR since I reviewed some of the accompanying dsi/server changes made by @banarun.

I do have a couple of larger (optional) comments:
1. Validation- it seems like we might want to have some for these queries. My suggestion would be to run the queries locally against the scale 10 data and get results. You could then add a validate.yml file similar to this [one](https://github.com/mongodb/genny/blob/master/src/workloads/tpch/denormalized/1/validate.yml) which uses the `AssertiveActor` to compare the output against an existing collection with the expected results from your local run. Alternatively, you could edit the existing [scale 10 validate.yml](https://github.com/mongodb/genny/blob/master/src/workloads/tpch/denormalized/10/validate.yml), which currently does nothing at all. Note that this second option would run this validation on all existing TPC-H tasks. This brings me to my second thought.
2. Is the plan to include these queries in the existing TPC-H scale 10 benchmark? It could eventually be very interesting/useful for the optimizer projects, since they seem simpler/ possibly like something we could actually use in the nearish future for performance testing. If so, I'd definitely edit the existing scale 10 validate.yml.

Let me know if you would like to discuss either of these ideas/ possible approaches outside this PR as well.

awagenmaker:
@starzia I changed the terminology to be 'internal' instead of artificial.

@alyacb I addressed your comments. Talking to the column store team about what we want to do about validation, I'll follow up on that.

starzia (approved):
I suggest we merge this as is and build on it later.

src/phases/tpch/denormalized/AvgAcctBal.yml
```diff
@@ -0,0 +1,44 @@
...
+Owner: "@mongodb/product-query"
+Description: |
+  Run an internal TPC-H query to get the average customer account balance against the denormalized
```
borosaurus, line 4: [note] IMO we should avoid benchmarking explains() because explain() enables all kinds of internal timers/metrics that actually have a fair amount of overhead (probably less than 10%, but still measurable)

Since these queries are all $groups, they return a fairly small result set, and we could get away with just using a large batch size.

Either way, it looks like the existing tests do use explain() and you were following the same model. Given this point is not specific to your new tests I see no reason to hold up the PR on this

borosaurus (approved):
lgtm

PR #764, opened 2022-11-04
PERF-3491 Add rooted $or filter tests with 50, 100, and 200 clauses

mtandrei:
Evergreen: https://evergreen.mongodb.com/patch/63653f4257e85a7e46fec987?redirect_spruce_users=true

The only question that I have is whether we are interested in adding even more variety in the test cases. Do we want to:
- Add some dotted paths into the mix?
- Have some rooted $and queries?

### DIFF
--- a/src/workloads/execution/FilterWithComplexLogicalExpression.yml
+++ b/src/workloads/execution/FilterWithComplexLogicalExpression.yml
@@ -32,7 +32,34 @@ GlobalDefaults:
   DocumentCount: &DocumentCount 1e6
   Repeat: &Repeat 50
   Threads: &Threads 1
-  MaxPhases: &MaxPhases 17
+  MaxPhases: &MaxPhases 23
+
+# Templates for executing a rooted $or over a configurable number of clauses for the aggregation
+# and match languages, respectively.
+AggregationOrWithManyChildClausesTemplate:
+  Filter:
+    {$expr: { $or:
+                { ^Array: { of: {$and: [
+                  {$eq:
+                     ["$a",
+                      { ^RandomInt:
+                          { min: 2, max: {^Parameter: {Name: "aMaxValue", Default: 1000}}}}]},
+                  {$eq:
+                     ["$b",
+                      { ^RandomInt:
+                          { min: 2, max: {^Parameter: {Name: "bMaxValue", Default: 1000}}}}]}]},
+                            number: {^Parameter: {Name: "NumberOfClauses", Default: 50}}}}}}
+
+MatchOrWithManyChildClausesTemplate:
+  Filter:
+    { $or:
+        { ^Array:
+            { of:
+                { a: { ^RandomInt:
+                         { min: 2, max: {^Parameter: {Name: "aMaxValue", Default: 1000}}}},
+                  b: { ^RandomInt:
+                         { min: 2, max: {^Parameter: {Name: "bMaxValue", Default: 1000}}}}},
+              number: {^Parameter: {Name: "NumberOfClauses", Default: 50}}}}}
 
 Actors:
 # Clear any pre-existing collection state.
@@ -516,6 +543,126 @@ Actors:
         - OperationName: find
           OperationCommand:
             Filter: {"n1.a.b.c.d.z": 0}
+
+- Name: AggregationExpressionFiftyClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [17]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: AggregationOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 50
+
+- Name: MatchExpressionFiftyClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [18]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: MatchOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 50
+
+- Name: AggregationExpressionOneHundredClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [19]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: AggregationOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 100
+
+- Name: MatchExpressionOneHundredClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [20]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: MatchOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 100
+
+- Name: AggregationExpressionTwoHundredClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [21]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: AggregationOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 200
+
+- Name: MatchExpressionTwoHundredClauseRootedOr
+  Type: CrudActor
+  Database: *Database
+  Threads: *Threads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [22]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: MatchOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 200
 
 AutoRun:
 - When:


src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -517,6 +544,126 @@ Actors:
...
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
+              Key: AggregationOrWithManyChildClausesTemplate
+              Parameters:
+                NumberOfClauses: 50
```
Adityav369, line 565: [nit/ optional] We could do without this field too, since the default is 50, but it might also be worthwhile to be explicit and keep it.

Adityav369 (commented):
This looks good! The number of clauses thing would also be helpful for the variadic expressions ticket I'll work on. Curious to see how this performs on SBE and classic! I only have a couple optional nits.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -517,6 +544,126 @@ Actors:
...
+      Active: [17]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            LoadConfig:
+              Path: "FilterWithComplexLogicalExpression.yml"
```
dstorch, line 562: [question] The self-reference here is a bit wacky, though I do think there is some value in keeping the template in this file instead of moving it into its own file. Is there no way to omit this parameter if the template is defined in the same file? If not, should we introduce something like this?

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,34 @@ GlobalDefaults:
   DocumentCount: &DocumentCount 1e6
   Repeat: &Repeat 50
   Threads: &Threads 1
-  MaxPhases: &MaxPhases 17
+  MaxPhases: &MaxPhases 23
```
dstorch, line 35: I actually suspect that this can be 22, not 23. I doubt it matters though.

Also, it's somewhat unfortunate that we need to keep maintaining this number... We don't necessarily have to do it as part of this patch, but I wonder if there is a way to write this workload as a single phase. The phase would be a single `CrudActor` that runs multiple find operations one after the other and measures each independently. I think this kind of thing is supported by the `CrudActor`, if I understand correctly. Theoretically this would get rid of some of the boilerplate in this patch as well.

dstorch (changes requested):
A few comments. Regarding this:

> The only question that I have is whether we are interested in adding even more variety in the test cases. Do we want to:
> * Add some dotted paths into the mix?
> * Have some rooted $and queries?

My opinion is that we shouldn't worry about dotted paths here. We can test separately for scaling along the dimension of deep dotted paths and scaling along the dimension of many clause $and/$or predicates. Anyhow, our anecdotal experience suggests that very large $and/$or predicates with lots of clauses are more likely in practice than very deep nested paths.

I do think there is some value in extending this path to deal with large $and predicates. How would this work exactly -- would we have some array field `arr` and have an $and predicate where each clause successfully matches the array? For example, the array could contain integers 0 to 199 and we could generate predicates up to 200 clauses where each clause matches one of the constituent integers.

mtandrei:
Thanks for taking a look Adi and Dave! I've leveraged ActorTemplates to reduce a lot of the boilterplate in this patch. 

Here's a new patch build: https://spruce.mongodb.com/version/636ad05fe3c3312e57aba757/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC 

PTAL

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+          Collection: *Collection
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {}}}
+
+Array: &Array
+  {^Cycle: {ofLength: 1, fromGenerator: {^Array: {number: 200, of: {^Inc: {start: 0, step: 1}}}}}}
+
+FiftyClauseRootedAnd: &FiftyClauseRootedAnd
```
mtandrei, line 117: I wasn't able to dedup these three queries; any suggestions on how to use anchors to deduplicate here would be appreciated.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+                { $or:
+                    { ^Array:
+                        { of:
+                            { a: { ^RandomInt:
+                                     { min: 2, max: { ^Parameter: { Name: "aMaxValue", Default: 1000 } } } },
+                              b: { ^RandomInt:
+                                     { min: 2, max: { ^Parameter: { Name: "bMaxValue", Default: 1000 } } } } },
+                          number: { ^Parameter: { Name: "NumberOfClauses", Default: 50 } } } } }
+
+- TemplateName: MatchAndWithManyChildClausesTemplate
```
mtandrei, line 96: Note that I did this only for the match language. I could do this for the agg expressions, but I'm not sure what this would look like (outside of doing a wide document style query, that is, {$and: [{$eq: ["$a", 1]}, {"$b", 1}, ...]})

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {}}}
+
+Array: &Array
+  {^Cycle: {ofLength: 1, fromGenerator: {^Array: {number: 200, of: {^Inc: {start: 0, step: 1}}}}}}
+
+FiftyClauseRootedAnd: &FiftyClauseRootedAnd
+  {$and:
+     {^Cycle: {ofLength: 1, fromGenerator:
```
dstorch, line 119: Ditto comment about why we need `^Cycle`.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: Aren't they identical but they have a different array length? Can't the array length just be treated as a parameter?

In fact, can the actual generator defining how to generate the query predicate be moved into the `MatchAndWithManyChildClausesTemplate`, and then have the template accept the array length parameter?

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -68,7 +169,8 @@ Actors:
         BatchSize: 1000
         Document:
           # NOTE: These documents will not produce a match for the filter below, forcing the query
-          # to scan the entire collection.
+          # to scan the entire collection. The only exception to this is the 'arr' field, which
+          # is intended to be used in rooted $and queries of the form {arr: 1, arr: 2, ... arr: n}
```
dstorch, line 173:
```suggestion
          # is intended to be used in rooted $and queries of the form {$and: [{arr: 1}, {arr: 2}, ... {arr: n}]}.
```

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: > Yeah, I'm not sure if if this is 0 indexed or not (i.e. is it the number of phases, or is it the highest phase number that we reach?)

I looked into this when we had a similar discussion on my genny patch for the $group spilling workloads. I'm 95% certain that it's the highest phase number we reach (zero-indexed) as opposed to the number of phases (one-indexed). That appears consistent with the latest code in this patch, so I think we can consider this topic resolved.

> Yeah; we could totally refactor these. I think it would look something like this. I can file a ticket and stick it in m4 if you would like.

I don't think it's worth messing with this particular workload, because it already has a different structure where each query has its own `CrudActor` declaration. So I wouldn't bother filing a ticket for m4. I doubt we'd actually do it. Going forward, however, one could argue that it would be better to write such workloads using a single `CrudActor` that executes multiple queries in sequence.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: Nice! This is definitely an improvement. Resolving the thread.

dstorch (changes requested):
Thanks @mtandrei! Looking good, but one more round of minor questions/comments about the latest changes.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
mtandrei: Oh dang 🤦  it occurs to me I could the same thing as the or expressions, lol. Done. 

Unlike the $or predicates which have random values, I think I was having trouble generating the fixed sequential values needed for $and.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+                           ["$a",
+                            {^RandomInt:
+                               {min: 2, max: { ^Parameter: { Name: "aMaxValue", Default: 1000}}}}]},
+                        {$eq:
+                           ["$b",
+                            {^RandomInt:
+                               {min: 2, max: {^Parameter: { Name: "bMaxValue", Default: 1000}}}}]}]},
+                                number: {^Parameter: { Name: "NumberOfClauses", Default: 50}}}}}}
+
+- TemplateName: MatchOrWithManyChildClausesTemplate
```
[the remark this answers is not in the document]
mtandrei: Indeed these do appear to work! Done and resolving

mtandrei (commented):
Thanks for taking another look Dave! Addressed the latest comments in the latest commit.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+        PhaseConfig:
+          Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
+          Collection: *Collection
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {}}}
+
+Array: &Array
+  {^Cycle: {ofLength: 1, fromGenerator: {^Array: {number: 200, of: {^Inc: {start: 0, step: 1}}}}}}
```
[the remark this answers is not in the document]
dstorch: Oh! Yes, that makes sense. I wonder if there is a different way that we are supposed to do this with Genny, but I guess if it works it works. Please do leave a comment when we use `^Cycle` this way in order to help future readers understand what's going on.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: Feel free to resolve this thread after reading my previous response.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: Latest changes here look good to me. Feel free to resolve this thread.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "ActivePhase", Default: -1}}]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
+          Collection: *Collection
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {}}}
```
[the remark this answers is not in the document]
dstorch: ☹️ 

We have another open thread about this topic, so I'll resolve this one.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
dstorch: We have another open thread above this topic, so I'll resolve this one. I do think adding comments to explain the use of `^Cycle` makes sense.

dstorch (approved):
LGTM with a couple more minor pieces of feedback and responses on ongoing threads.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,108 @@ GlobalDefaults:
...
   Threads: &Threads 1
-  MaxPhases: &MaxPhases 17
+  MaxPhases: &MaxPhases 25
+
+# Set of actor templates for executing a rooted $and/$or over a configurable number of clauses for
+# the aggregation and match languages, respectively.
+ActorTemplates:
+- TemplateName: AggregationOrWithManyChildClausesTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "AggregationOr"}}
```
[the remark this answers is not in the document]
mtandrei: https://jira.mongodb.org/browse/EVG-18242

src/workloads/execution/FilterWithComplexLogicalExpression.yml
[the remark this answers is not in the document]
mtandrei: Adding comment.

Adityav369 (approved):
LGTM. Thanks!

mtandrei (commented):
Thanks for the reviews!

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -53,58 +53,57 @@ ActorTemplates:
...
         Filter:
           {$and:
+             # Each time that this template is invoked, we wish to generate a query of the form:
+             #
+             # {$and: [{"arr": 0, "arr": 1, ... "arr": NumberOfClauses - 1}]}
+             #
+             # We achieve this using the "^Inc" generator. However, because ^Inc increments
+             # globally, we need to bind the generated array and keep using it to avoid generating
+             # values greater than/equal to NumberOfClauses. As such, we generate the array within
+             # a ^Cycle and keep repeating it to avoid this problem.
```
dstorch, line 106: [extremely optional]

```suggestion
             # a ^Cycle of length 1 and keep repeating it to avoid this problem.
```

src/workloads/execution/FilterWithComplexLogicalExpression.yml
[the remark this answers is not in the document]
dstorch: Thanks. Hopefully this is in the right state to be noticed by the appropriate Evergreen team! I'm not sure how their workflow usually works.

src/workloads/execution/FilterWithComplexLogicalExpression.yml
```diff
@@ -32,7 +32,83 @@ GlobalDefaults:
...
   Threads: &Threads 1
-  MaxPhases: &MaxPhases 17
+  MaxPhases: &MaxPhases 25
+
+# Set of actor templates for executing a rooted $and/$or over a configurable number of clauses for
+# the aggregation and match languages, respectively.
+ActorTemplates:
+- TemplateName: ManyChildClauseQueryTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "ManyChildClauseActor"}}
```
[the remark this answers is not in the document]
dstorch: [optional] You changed most of the places where this applies, but there are still a few spots where you should consider using -1 as the default for the `NumberOfClauses` parameter.

dstorch (approved):
LGTM

dstorch (approved):

PR #765, opened 2022-11-08
PERF-3488: Performance tests for variadic aggregate expressions

Adityav369:


src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: 1
+
+# Phase 3-N: Run variadic aggregate expression benchmarks
+- ExpressionAdd10:
```
mtandrei, line 157: [nit] prefer numbers to be written out (so, "Ten" instead of "10")

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionIfNull20
+      nopInPhasesUpTo: *MaxPhases
+      active: [10]
+      operationMetricsName: VariadicAggExpressionIfNull20
+      Pipeline: [{$project: {ifNull: { $ifNull: ["$arr1", "$arr2", "$arr3", "$arr4", "$arr5", "$arr6", "$arr7", "$arr8", "$arr9", "$arr10", "$arr11", "$arr12", "$arr13", "$arr14", "$arr15", "$arr16", "$arr17", "$arr18", "$arr19", "$arr20", 0]}}}]
+
+- ExpressionLet10:
```
mtandrei, line 238: I'm not sure that we want to test let? If we do though, we should make sure that the "in" expression is small and doesn't reference most of the bound fields. That way, we can measure the cost of binding variables.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionMultiply10
+      nopInPhasesUpTo: *MaxPhases
+      active: [13]
+      operationMetricsName: VariadicAggExpressionMultiply10
+      Pipeline: [{$project: {multiply: { $multiply: *IntFields10}}}]
+
+- ExpressionMultiply20:
```
mtandrei, line 268: we should also have coverage for the set expressions ($setUnion/equals/intersection)

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  This workload measures aggregation expression performance for aggregations that
+  can take in multiple expressions as arguments.
```
mtandrei, line 5:
```suggestion
  can take in a variable number of expressions as arguments.
```

mtandrei (changes requested):

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+          arr12: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr13: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr14: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr15: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr16: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr17: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr18: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr19: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          arr20: [{^Repeat: {count: 10, fromGenerator: {b: 0, c: 0}}}]
+          false1: false
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+      nopInPhasesUpTo: *MaxPhases
+      active: [8]
+      operationMetricsName: VariadicAggExpressionConcatArrays20
+      Pipeline: [{$project: {concatArrays: { $concatArrays: ["$arr1", "$arr2", "$arr3", "$arr4", "$arr5", "$arr6", "$arr7", "$arr8", "$arr9", "$arr10", "$arr11", "$arr12", "$arr13", "$arr14", "$arr15", "$arr16", "$arr17", "$arr18", "$arr19", "$arr20"]}}}]
+
+- ExpressionIfNull10:
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionIfNull10
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+
+- ExpressionConcat10:
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionConcat10
+      nopInPhasesUpTo: *MaxPhases
+      active: [5]
+      operationMetricsName: VariadicAggExpressionConcat10
+      Pipeline: [{$project: {concat: { $concat: ["$str1", "$str2", "$str3", "$str4", "$str5", "$str6", "$str7", "$str8", "$str9", "$str10"]}}}]
```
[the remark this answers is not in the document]
Adityav369: Done. I use 5 and 10 now.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,311 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document:
+          int1: {^RandomInt: {min: -100, max: 100}}
```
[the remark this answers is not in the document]
Adityav369: As per offline discussion, since we have only 10 fields, I chose not to use the ^cycle generator. Let me know if you think I should use it and reduce the number of fields we insert in a document.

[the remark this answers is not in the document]
mtandrei: There's two questions here:
- How many args should we have? 5 and 10 seem better, but I'm not sure whether something like 3 and 6 or 3 and 5 would be more realistic. I've pinged Chris and Katya on slack to see if they have any opinions; if they don't let's stick with with 5 and 10
- What should the document schema look like? The current schema feels a bit crazy, but it's probably for the best as a "realistic" use case wouldn't duplicate field arguments. However, one (very optional) suggestion: could you try and use the [object generator to cut down on some of the redundancy here?](https://github.com/mongodb/genny/blob/dcd3a0527cce7afbe1aaec90b3cb488672253b5e/src/workloads/docs/Generators.yml#L236) If not, it's no big deal, but it could make this more parseable.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,330 @@
...
+- ExpressionAddFive:
+  LoadConfig: &loadConfig
+    Path: "../../phases/execution/AggregateExpressions.yml"
+    Key: ExpressionTemplate
+    Parameters:
+      name: VariadicExpressionAddFive
+      nopInPhasesUpTo: *MaxPhases
+      active: [3]
+      operationMetricsName: VariadicAggExpressionAddFive
+      Pipeline: [{$project: {add: { $add: *FiveIntFields}}}]
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,330 @@
...
+
+- ExpressionIfNullFive:
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionIfNullFive
+      nopInPhasesUpTo: *MaxPhases
+      active: [9]
+      operationMetricsName: VariadicAggExpressionIfNullFive
+      Pipeline: [{$project: {ifNull: { $ifNull: ["$a1", "$a2", "$a3", "$a4", "$a5", 0]}}}]
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,330 @@
...
+
+- ExpressionSwitchTen:
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionSwitchTen
+      nopInPhasesUpTo: *MaxPhases
+      active: [16]
+      operationMetricsName: VariadicAggExpressionSwitchTen
+      Pipeline: [{$project: {switch: { $switch: { branches: [{case: {$gt: ["$int1", 100]}, then: "$int1"}, {case: {$gt: ["$int2", 100]}, then: "$int2"}, {case: {$gt: ["$int3", 100]}, then: "$int3"}, {case: {$gt: ["$int4", 100]}, then: "$int4"}, {case: {$gt: ["$int5", 100]}, then: "$int5"}, {case: {$gt: ["$int6", 100]}, then: "$int6"}, {case: {$gt: ["$int7", 100]}, then: "$int7"}, {case: {$gt: ["$int8", 100]}, then: "$int8"}, {case: {$gt: ["$int9", 100]}, then: "$int9"}, {case: {$gt: ["$int10", 100]}, then: "$int10"}], default: "$int1"}}}}]
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,685 @@
...
+  LoadConfig:
+    <<: *loadConfig
+    Parameters:
+      name: VariadicExpressionAddFifty
+      nopInPhasesUpTo: *MaxPhases
+      active: [5]
+      operationMetricsName: VariadicAggExpressionAddFifty
+      Pipeline: [{$project:
+                    {add:
+                       {$add:
```
mtandrei, line 147: discussing offline how we might simplify this further

mtandrei (changes requested):

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,685 @@
...
+                                  {^Array: {of: "$arr4", number: 5}},
+                                  {^Array: {of: "$arr5", number: 5}},
+                                  {^Array: {of: "$arr6", number: 5}},
+                                  {^Array: {of: "$arr7", number: 5}},
+                                  {^Array: {of: "$arr8", number: 5}},
+                                  {^Array: {of: "$arr9", number: 5}},
+                                  {^Array: {of: "$arr10", number: 5}}
+                                ]}}}}}]
+
+# The input fields do no exist so we execute every clause.
```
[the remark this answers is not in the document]
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: As per offline discussion, I made changes to the generator by adding a deterministic mode to Choose, and now we just control the parameter "number" when constructing an array.

rtimmons (changes requested):
Making `^Choose` have a `deterministic` option is an interesting concept I never would have thought of. I'm curious how useful it will be in the general sense. Glad to see it as an option.

A few suggestions to decrease chances of mis-configs and increase test-coverage to guard against future regressions. Please push back wherever you disagree. Only real block from my perspective is adding some mutual-exclusion for other options if `deterministic` is specified.

Note I didn't extensively review your workload yamls. I'll leave that to your team and patch-tests. In general I'd prefer a separate PR for the valgen vs the workload yaml so reverts (if needed) can be more surgical, but no biggie this time.

Thanks for pushing on this. Glad you were able to avoid a confusing double-`^Concat` situation.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1910,6 +1916,13 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
         // If not passed in, give each choice equal weight
         _weights.assign(_choices.size(), 1);
     }
+    if (node["deterministic"]) {
```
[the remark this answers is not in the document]
Adityav369: Done. If deterministic and weights are both specified, it fails with and InvalidSyntaxException. I checked for other configs, but looks like it is just these weights and deterministic.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1910,6 +1916,13 @@ ChooseGenerator::ChooseGenerator(const Node& node, GeneratorArgs generatorArgs)
...
+        auto val = node["deterministic"].maybe<int32_t>().value_or(-1);
+        val > 0 ? _deterministic = true : _deterministic = false;
```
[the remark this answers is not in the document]
Adityav369: Ah yes! Done.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -116,6 +116,12 @@ Tests:
       - {a: [ { "a": 1 }, 1, { "a": 1 } ] }
       - {a: [ 1, 1, { "a": 1 }, 1, "abc" ] }
 
+  - Name: ArrayChooseDeterministic
+    GivenTemplate:
+      a: {^Array: {of: {^Choose: {from: [1, 2, 3], deterministic: 1}}, number: 6}}
+    ThenReturns:
+      - {a: [ 1, 2, 3, 1, 2, 3 ] }
```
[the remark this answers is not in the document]
Adityav369: Done.

Adityav369 (commented):
Thanks Mihai and Ryan for going over my PR! I have implemented the suggested changes.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -516,13 +516,17 @@ class ZipfianInt64Generator : public Generator<int64_t> {
...
     Appendable& choose() {
+        if (_deterministic) {
+            ++_elemNumber;
+            return *_choices[_elemNumber % _choices.size()];
+        }
         // Pick a random number between 0 and sum(weights)
         // Pick value based on that.
         auto distribution = boost::random::discrete_distribution(_weights);
-        return (*_choices[distribution(_rng)]);
+        return (*_choices[distribution(_rng)]);
```
mtandrei, line 529: This looks like extra whitespace; please delete it (I'm not sure whether there's a cpp linter for genny or not)

src/workloads/execution/VariadicAggregateExpressions.yml
mtandrei: Nice work on this; resolving!

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,621 @@
...
+- sbe
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  DocumentCount: &DocumentCount 1e6
+  Repeat: &Repeat 10
+  Threads: &Threads 1
+  MaxPhases: &MaxPhases 32
+  TenIntFields: &TenIntFields ["$int1", "$int2", "$int3", "$int4", "$int5",
```
mtandrei, line 26: can we use the new deterministic ^Choose to auto generate these? As in, something like ^Join together 'fieldname' and a choose array from 1 to 10? 

Similar comment for the document schema, but if you're going to do that, it might be worth leaving a comment describing the schema.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,621 @@
...
+      name: VariadicExpressionLetTen
+      nopInPhasesUpTo: *MaxPhases
+      active: [15]
+      operationMetricsName: VariadicAggExpressionLetTen
+      Pipeline: [{$project:
+                    {let:
+                       {$let:
+                          {vars:
+                             {
+                               a: "$int1", b: "$int2", c: "$int3", d: "$int4", e: "$int5",
```
mtandrei, line 325: Similar comment, can we automatically generate these fields instead of hard coding them? One idea is to bind the referenced fields by name, but then auto-generate random fields for the non referenced fields.

mtandrei (changes requested):
Nice work Adi! Almost L-G-T-M, just a couple more suggestions and this should be good to go.

I only skimmed the cpp changes; they look good, but I will defer to Ryan for the final sign offs.

src/value_generators/src/DocumentGenerator.cpp
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -561,8 +572,20 @@ class ChooseStringGenerator : public Generator<std::string> {
             // If not passed in, give each choice equal weight
             _weights.assign(_choices.size(), 1);
         }
+        if (node["deterministic"]) {
+            auto val = node["deterministic"].maybe<int32_t>().value_or(-1);
```
mtandrei, line 576: Would it make sense for 'deterministic' to be a boolean value?

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -116,6 +116,34 @@ Tests:
       - {a: [ { "a": 1 }, 1, { "a": 1 } ] }
       - {a: [ 1, 1, { "a": 1 }, 1, "abc" ] }
 
+  - Name: Generate Array with choose deterministic where number is greater than from
```
mtandrei, line 119:
```suggestion
  - Name: Generate Array with choose deterministic where 'number' is greater than size of 'from'
```

src/workloads/docs/Generators.yml
```diff
@@ -225,6 +229,10 @@ Actors:
                                              prid: { ^RandomDouble: { min: 0.0, max: 1000.0 } },
                                              data: { ^Join: { array: ["aaaaaaaaaa", {^FastRandomString: {length: {^RandomInt: {min: 0, max: 10}}}}]}}},
                                        number: 10}}
+            # You can choose elements deterministically. from is an array of values to pick from.
```
mtandrei, line 232: [nit] s/from/'from' (generally, prefer to have named arguments in single quotes). Here and elsewhere.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,614 @@
...
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  DocumentCount: &DocumentCount 1e6
+  Repeat: &Repeat 10
+  Threads: &Threads 1
+  MaxPhases: &MaxPhases 32
+  ListOfTenInts: &ListOfTenInts {^Choose: {from: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"], deterministic: 1}}
+  ListOfFiftyInts: &ListOfFiftyInts {^Choose: {from: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10",
```
mtandrei, line 27: I guess it's not possible to use ^Inc and get a string, but: the lists of 50 and 100 integers ought to be simplified further . Could this be done by joining two deterministic choose lists; something like ["", "1", "2", "3", "4"] and the list of 10, then ["", "1", "2", "3", "4",  "5", "6", "7", "8", "9"]  and the list of 10? This would make it so that we start at 0, but this is definitely worth it if we can avoid hard coding a list.

mtandrei (approved):
LGTM with some final comments. Great work Adi!

rtimmons (dismissed):
Nicely done!

src/value_generators/src/DocumentGenerator.cpp
Adityav369: Done.

src/value_generators/test/DocumentGeneratorTestCases.yml
Adityav369: Done.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -54,6 +39,9 @@ GlobalDefaults:
                                                  {case: {$gt: ["$int8", 100]}, then: "$int8"},
                                                  {case: {$gt: ["$int9", 100]}, then: "$int9"},
                                                  {case: {$gt: ["$int10", 100]}, then: "$int10"}]
+  IntGenerator: &IntGenerator {^Join: {array: [{ ^FormatString: {format: "%03d", withArgs: [{^Inc: {start: 1}}]}}]}}
```
mtandrei, line 42: one problem that I ran into with ^Inc is that for using it repeatedly, you will keep getting higher and higher values ([see comment here](https://github.com/mongodb/genny/blob/b41ef95d4870c8d375a0ae3f7a99b44bd02bb643/src/workloads/execution/FilterWithComplexLogicalExpression.yml#L98-L111)). Before merging, please verify that this produces the fields that you expect.

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,594 @@
...
+                                                 {case: {$gt: ["$int2", 100]}, then: "$int2"},
+                                                 {case: {$gt: ["$int3", 100]}, then: "$int3"},
+                                                 {case: {$gt: ["$int4", 100]}, then: "$int4"},
+                                                 {case: {$gt: ["$int5", 100]}, then: "$int5"},
+                                                 {case: {$gt: ["$int6", 100]}, then: "$int6"},
+                                                 {case: {$gt: ["$int7", 100]}, then: "$int7"},
+                                                 {case: {$gt: ["$int8", 100]}, then: "$int8"},
+                                                 {case: {$gt: ["$int9", 100]}, then: "$int9"},
+                                                 {case: {$gt: ["$int10", 100]}, then: "$int10"}]
+  IntGenerator: &IntGenerator {^Join: {array: [{ ^FormatString: {format: "%03d", withArgs: [{^Inc: {start: 1}}]}}]}}
```
mtandrei, line 42:
```suggestion
  IntGenerator: &IntGenerator {^Join: {array: [{^FormatString: {format: "%03d", withArgs: [{^Inc: {start: 1}}]}}]}}
```

src/workloads/execution/VariadicAggregateExpressions.yml
```diff
@@ -0,0 +1,594 @@
...
+      operationMetricsName: VariadicAggExpressionConcatArraysHundred
+      Pipeline: [{$project:
+                    {concatArrays:
+                       {$concatArrays:
+                          {^Array: {of: *GenArrFields, number: 100}}
+                       }}}]
+
+# For the following $ifNull test cases, the arguments do not exist in the
+# documents that we insert into our collection. This results in executing
+# every argument to $ifNull
```
mtandrei, line 261:
```suggestion
# every argument of $ifNull.
```

mtandrei (approved):
Still LGTM

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: I was running into that issue initially. I used ^FixedGeneratedValue to generate the list of ints, so it is generated only once.

src/workloads/execution/VariadicAggregateExpressions.yml
Adityav369: Done.

mtandrei (approved):
Still LGTM

PR #767, opened 2022-11-09
PERF-3516 Temporarily disable constant_fold_arithmetic task

mtandrei:


src/workloads/execution/ConstantFoldArithmetic.yml
```diff
@@ -106,10 +106,11 @@ Actors:
...
         Expression: {^Concat: {arrays: [*constArray, ["$x"], *constArray]}}
 
-AutoRun:
-- When:
-    mongodb_setup:
-      $eq:
-      - replica
-      - single-replica-classic-query-engine
-      - single-replica-sbe
+# TODO PERF-3517: Re-enable this workload.
```
mtandrei, line 109: You might be wondering: why are we disabling the task entirely? Well, there's two reasons:
- it appears that disabling this task only for SBE would block collection of perf data for the perf dashboard. So, we disable it for classic. 
- It also appears that this issue is reproing [on the 'replica' cluster](https://evergreen.mongodb.com/task/sys_perf_linux_3_node_replSet_constant_fold_arithmetic_924c1769d07b776695cf511b0fa87b23219128be_22_11_07_23_23_54). So, in the spirit of "fixing the BF", I've disabled it as well

Small problem: this leaves us with no coverage for arithmetic constant folding! We have a couple of options:
- figure out why the 3 node repl set is running SBE
- Run this on a guaranteed, non-SBE perf related variant that has the classic engine
- Leave as is because we want to reconsider the coverage anyways

Open to any of these options

mtandrei:
Companion PR: https://github.com/10gen/PerformanceReports/pull/35

src/workloads/execution/ConstantFoldArithmetic.yml
ksuarz: This is fine to me; let's leave it as it is and have a discussion on whether we want the coverage. But fixing the build first is important.

ksuarz (approved):
LGTM

mtandrei:
Thanks for the review Kyle!

PR #768, opened 2022-11-14
PERF-3431 First Graviton support for Genny

dalyd:
There are some special cases in this code that we will need to take care of. I would like to defer any heavy lifting right now, but am happy to do easy things. This goes with a hand crafted genny toolchain for arm. I'll share notes on that also. 

This works in my test environment. I haven't tested the download portions of it, but I will try that tomorrow. Please take a look.

### DIFF
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -213,6 +213,9 @@ class CuratorDownloader(Downloader):
         if self._linux_distro in ("amazon2", "rhel8", "rhel62"):
             self._curator_distro = "rhel70"
 
+        if self._linux_distro == "amazon2arm":
+            self._curator_distro = "arm"
+
     def _get_url(self):
         return (
             "https://s3.amazonaws.com/boxes.10gen.com/build/curator/"
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -218,7 +218,7 @@ class CuratorDownloader(Downloader):
         if self._os_family == "Darwin":
             self._curator_distro = "macos"
 
-        self._curator_distro = DISTRO_MAPPING.get(self._linux_distro, self._curator_distro)
+        self._curator_distro = CuratorDownloader.DISTRO_MAPPING.get(self._linux_distro, self._curator_distro)
 
     def _get_url(self):
         # Check if we need a special curator version for the distro. Otherwise use the default
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -225,7 +225,7 @@ class CuratorDownloader(Downloader):
     def _get_url(self):
         # Check if we need a special curator version for the distro. Otherwise use the default
         # CURATOR_VERSION
-        version = CuratorDownloader.SPECIAL_CURATOR_VERSION.get(
+        version = CuratorDownloader.SPECIAL_CURATOR_VERSIONS.get(
             self._curator_distro, CuratorDownloader.CURATOR_VERSION
         )
         return (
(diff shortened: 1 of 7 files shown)


src/lamplib/src/genny/tasks/compile.py
```diff
@@ -50,9 +50,9 @@ def cmake(
     # using both.
     cmake_prefix_paths = [
         os.path.join(
-            toolchain_info.toolchain_dir, f"installed/x64-{toolchain_info.triplet_os}-dynamic",
```
dalyd, line 53: These hard-coded x64 are lousy in the vcpkg code also. I made similar changes there.

src/lamplib/src/genny/curator.py
```diff
@@ -213,11 +213,18 @@ def __init__(
...
             self._curator_distro = "rhel70"
 
+        if self._linux_distro == "amazon2arm":
+            self._curator_distro = "arm"
+
     def _get_url(self):
+        # The pinned curator version was before arm support was added. Use a newer version for Arm
+        version = CuratorDownloader.CURATOR_VERSION
+        if self._curator_distro == "arm":
+            version = "965d53845fd1987ddbf04a937ff625f3c243dee3"
```
rtimmons, line 223: Can there be a `CuratorDownloader.ARM_CURATOR_VERSION` constant so the version numbers can be updated in one place?

Also consider making it a dict mapping curator distro to version.

src/lamplib/src/genny/toolchain.py
```diff
@@ -16,7 +16,7 @@
 # Define complex operations as private methods on the module to keep the
 # public Context object clean.
 def _create_compile_environment(
-    triplet_os: str, toolchain_dir: str, system_env: Optional[dict] = None
+    triplet_os: str, toolchain_dir: str, arch: str, system_env: Optional[dict] = None
```
rtimmons, line 19: pedantry: make `arch` and `triplet_os` into one tuple or at least consecutive args. Only worth it if your IDE supports that kind of automated refactor.

src/lamplib/src/genny/toolchain.py
```diff
@@ -172,6 +191,7 @@ def __init__(
         workspace_root: str,
         os_family: str,
         linux_distro: str,
+        arch: str,
```
rtimmons, line 194: another nit: can this be `triplet_arch` more pervasively?

rtimmons (dismissed):
Seems you accidentally killed `build/.gitignore` which will make certain tooling cranky.

And do we wanna add an amazonlinuxarm genny build-variant so we know this works and can guard against regressions when we upgrade the toolchain in the coming weeks?

Otherwise just a few nit suggestions. Assuming you tested it locally, these changes are straightforward.

src/lamplib/src/genny/toolchain.py
dalyd: I'd prefer to leave this one for now. I agree it would be better, but also increases the changes of something breaking and I have enough of those right now :)

src/lamplib/src/genny/toolchain.py
dalyd: Done

rtimmons (approved):

src/lamplib/src/genny/curator.py
```diff
@@ -213,11 +218,19 @@ def __init__(
         if self._linux_distro in ("amazon2", "rhel8", "rhel62"):
             self._curator_distro = "rhel70"
 
+        if self._linux_distro == "amazon2arm":
+            self._curator_distro = "arm"
+
     def _get_url(self):
+        # Check if we need a special curator version for the distro. Otherwise use the default
+        # CURATOR_VERSION
+        version = SPECIAL_CURATOR_VERSION.get(
```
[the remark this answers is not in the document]
dalyd: I partly did this. I moved the _curator_distro into a map, except fro the fact tat the first check is not on _linux_distro. all the other checks collapsed away. Hopefully nothing breaks.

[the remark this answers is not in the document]
dalyd: And I'm so bad at scoping these things. Wow.

rtimmons (approved):

dalyd:
evergreen merge

dalyd:
evergreen merge

dalyd:
evergreen merge

dalyd:
evergreen merge

PR #769, opened 2022-11-15
PERF-2891: Adding workload for measuring create / drop time for views.

jawwadasghar:
@dalyd I was hoping to get your thoughts on this workload related to [PERF-2891](https://jira.mongodb.org/browse/PERF-2891). 
Do you think this provides enough value to be added? 
It takes less than a minute to run, and would monitor the duration of time for views to be created and deleted (averages are around 400,000 ns). 
I'm not particularly attached to it, and am fine just documenting having looked at this. 

I previous tested a different [view create / drop workload](https://github.com/mongodb/genny/compare/master...PERF-2891-create_view#diff-425cec6331df008a5d898ac29ac1757e5a3711f6c89710b182ad69fd0b19e8e8R112) (64 threads with 10,000 repeated create / drop views each) for an hour alongside mixed CRUD actors, and the server did seem to queue up the commands, create / drop view latency increased cyclically but dropped off, and the server managed well. Based on the servers behavior, I didn't see the need to add this kind of workload

### DIFF
--- a/src/workloads/scale/CreateDropView.yml
+++ b/src/workloads/scale/CreateDropView.yml
@@ -68,4 +68,5 @@ AutoRun:
 - When:
     mongodb_setup:
       $eq:
-      - single-replica
+      - replica
+      - shard-lite
--- a/src/workloads/scale/CreateDropView.yml
+++ b/src/workloads/scale/CreateDropView.yml
@@ -1,0 +1,81 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
+Description: |
+  This workload loads a collection with documents and then repeatedly 
+  create and drops a view on that collection.
+Keywords:
+- view
+- create
+- drop
+
+dbname: &dbname viewDB
+DocumentCount: &NumDocs 100000
+CollectionCount: &NumColls 1
+
+
+Actors:
+- Name: Setup
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    BatchSize: 100
+    Threads: 1
+    DocumentCount: *NumDocs
+    Database: *dbname
+    CollectionCount: *NumColls
+    Document: &doc
+      id: {^RandomInt: {min: 0, max: *NumDocs}}
+      a: {^RandomInt: {min: 0, max: *NumDocs}}
+      c: &string {^RandomString: {length: 50}}  # Adjust this so the doc comes out as 100 B.
+    Indexes:
+    - keys: {id: 1}
+    - keys: {a: 1}
+  - Phase: 1..3
+    Nop: true
+
+- Name: QuiesceBetweenLevels
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - &nop {Nop: true}
+  - &quiesce
+    Repeat: 1
+    Database: admin
+    Operations:
+    # Fsync to force a checkpoint and quiesce the system.
+    - OperationMetricsName: FsyncCommand
+      OperationName: AdminCommand
+      OperationCommand:
+        fsync: 1
+  - *nop
+  - *quiesce
+
+#This actor repeatedly creates and then drops a view
+- Name: ViewCreateDrop
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Nop: true
+  - Nop: true 
+  - &toggleView1
+    Repeat: 100000
+    Database: *dbname
+    Operations:
+    - OperationMetricsName: CreateViewMetric
+      OperationIsQuiet: true
+      OperationName: RunCommand
+      OperationCommand:
+        {create: myView, viewOn: Collection0, pipeline: [ { $match: { id: { $lt: 10 } } } ]}
+    - OperationName: RunCommand
+      OperationMetricsName: DropViewMetric
+      OperationIsQuiet: true
+      OperationCommand:
+        drop: myView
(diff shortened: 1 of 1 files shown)


jimoleary (approved):

src/workloads/scale/CreateDropView.yml
```diff
@@ -0,0 +1,81 @@
...
+  - Repeat: 1
+    BatchSize: 100
+    Threads: 1
+    DocumentCount: *NumDocs
+    Database: *dbname
+    CollectionCount: *NumColls
+    Document: &doc
+      id: {^RandomInt: {min: 0, max: *NumDocs}}
+      a: {^RandomInt: {min: 0, max: *NumDocs}}
+      c: &string {^RandomString: {length: 50}}  # Adjust this so the doc comes out as 100 B.
```
dalyd, line 30: Is that comment correct? Are docs coming out do 100 B?

src/workloads/scale/CreateDropView.yml
```diff
@@ -0,0 +1,81 @@
...
+    Indexes:
+    - keys: {id: 1}
+    - keys: {a: 1}
+  - Phase: 1..3
+    Nop: true
+
+- Name: QuiesceBetweenLevels
```
dalyd, line 37: We have a quiesce actor for this. See: https://github.com/mongodb/genny/blob/48b3fac2e0c8c29b5f196e1263acc8d89e3762ed/src/workloads/docs/QuiesceActor.yml. Please use that instead of the explicit FsycnCommand.

src/workloads/scale/CreateDropView.yml
```diff
@@ -0,0 +1,81 @@
...
+  - *quiesce
+
+# This actor repeatedly creates and then drops a view
+- Name: ViewCreateDrop
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Nop: true
+  - Nop: true
+  - &toggleView1
```
dalyd, line 61: Do you use this anchor anywhere? If not, consider deleting the yaml anchor.

src/workloads/scale/CreateDropView.yml
```diff
@@ -0,0 +1,81 @@
...
+      OperationIsQuiet: true
+      OperationCommand:
+        drop: myView
+  - Nop: true
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - single-replica
```
dalyd, line 81: Is it interesting to run in other configurations? I'm wondering if there's any interactions with replication or sharding. Thoughts. @jimoleary ?

dalyd (approved):
Thanks Jawwad, a couple comments but looking good.

dalyd (changes requested):
Sorry -- text correct, but I meant to hit request changes for the queisce actor.

src/workloads/scale/CreateDropView.yml
jawwadasghar: I did some poking with [dataSize()](https://www.mongodb.com/docs/manual/reference/method/db.collection.dataSize/), and the comment is off. Each document comes out to **103** bytes. I think the size breaks down as follows: 
- 22 bytes to store _id as an objectId type 
- 23 bytes to store the ints, 8 bytes each for the 2 randInts up to 100K, 4 bytes for the 2 sets of “:” and “,”, and 3 bytes for the key character values “i”, “d”, and “a” 
- 58 bytes to store the string: 8 + length of 50 bytes

I got rid of the comment because I’m not particular about the document size. Originally I had kept this comment form where I got the loader form, MixedWorkloadsGenny.yml. Let me know if you think I should make any change to the comment in that file. 
I also got rid of the indexes. Creating the view doesn’t actually run the pipeline, so the indexes had no effect here.

src/workloads/scale/CreateDropView.yml
jawwadasghar: Oh nice - replaced!

src/workloads/scale/CreateDropView.yml
jawwadasghar: Woops - deleted the anchor.

src/workloads/scale/CreateDropView.yml
jimoleary: maybe add replica and shard-lite?

jimoleary (dismissed):

src/workloads/scale/CreateDropView.yml
dalyd: Awesome. Thanks.

dalyd (dismissed):

src/workloads/scale/CreateDropView.yml
jawwadasghar: FYI, Jim had I think a good suggestion of creating a ticket for replacing the instances where RunCommand fsyncs are used with the QuieseActor. Created [PERF-3545](https://jira.mongodb.org/browse/PERF-3545)

src/workloads/scale/CreateDropView.yml
jawwadasghar: Sure thing - I added replica, and shard-lite. I also removed single-replica as it seemed unnecessary if we're including a run on a full replica set.

jimoleary (approved):

jawwadasghar:
evergreen merge

jawwadasghar:
evergreen merge

PR #770, opened 2022-11-16
EVG-17806 Support projection argument for 'find' operations in CrudActor

mtandrei:
genny self tests/integration tests: https://spruce.mongodb.com/version/637514732a60ed2934cd74e9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
verify that changes to ExpressiveQueries work: https://spruce.mongodb.com/version/637514aa3e8e86331032403e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Some questions: 
- Should we add projection support the [other set of "findOne" operations](https://github.com/mongodb/genny/blob/9fab05fdaef3f58711e00344abdafa31a8fa2c85/src/cast_core/src/CrudActor.cpp#L1191-L1195)?
- Should we try and add more generic support for other find options (thinking of sort limit skip, but I'm sure there are others)


src/cast_core/src/CrudActor.cpp
```diff
@@ -637,6 +637,9 @@ struct FindOperation : public BaseOperation {
         if (opNode["Options"]) {
             _options = opNode["Options"].to<mongocxx::options::find>();
         }
+        if(opNode["Projection"]){
+            _options.projection(opNode["Projection"].to<DocumentGenerator>(context,id)());
```
dstorch, line 641: Hmmm. The fact that this is implemented this way and is not done alongside the rest of the conversion from `Node` to `mongocxx::options::find` is a little bit surprising to me:

https://github.com/mongodb/genny/blob/9fab05fdaef3f58711e00344abdafa31a8fa2c85/src/cast_core/include/cast_core/actors/OptionsConversion.hpp#L125-L154

I am hoping we can move this to the OptionsConversion.hpp so it lives alongside the related code.

Another related topic. This currently invokes the `DocumentGenerator` once on construction. That's probably not what we want, assuming that we even want to use the `DocumentGenerator` here in the first place. You'll notice that for the filter, the document generator is invoked in `run()`. The consequence is that each time the find command is run against the server, it can do so with a different generated filter. I would argue that projection should have the same behavior.

src/cast_core/src/CrudActor.cpp
```diff
@@ -637,6 +637,9 @@ struct FindOperation : public BaseOperation {
         if (opNode["Options"]) {
```
dstorch, line 637: (Not related to this line.)

The set of find command options that we support here seems to be quite arbitrary. At the risk of making you do more work, I think we should expand this patch so that it adds support for a useful subset of find command parameters. At the very least, I would hope that we support filter, projection, sort, skip, limit, and collation. There is a good argument for adding things like batchSize, readConcern, tailable, awaitData, and let to this list as well.

The same comment would apply to `FindOneOperation` class below.

src/cast_core/src/CrudActor.cpp
```diff
@@ -637,6 +637,9 @@ struct FindOperation : public BaseOperation {
...
             _options = opNode["Options"].to<mongocxx::options::find>();
         }
+        if(opNode["Projection"]){
```
dstorch, line 640: One could argue that given the structure of the underlying C++ driver API, the `Projection` should be nested inside `Options` rather than top-level. I suppose that having `Projection` as top-level more closely resembles the actual find command though, so I'm not totally sure what's best. I'd be curious to hear if @rtimmons has a strong opinion here.

src/workloads/execution/ExpressiveQueries.yml
```diff
@@ -129,28 +129,25 @@ Actors:
   - *Nop
   - *Nop
 
-# TODO: TIG-3050 Switch to CrudActor once projection argument is supported for find operation.
 - Name: RunQueries
-  Type: RunCommand
```
dstorch, line 134: Does the queries in this file actually require multiple batches? If there's no actual problem with this workload as originally implemented, I'd be inclined to just delete the TODO comment above and leave it as is. That way, we avoid any risk of a discontinuity in the performance data, but we've still done the work to improve the underlying genny infrastructure.

dstorch (changes requested):
Thanks Mihai, a couple comments!

mtandrei (commented):
new evg: https://spruce.mongodb.com/version/637e76a157e85a529952e072/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC 

Thanks for the review Dave! I've updated the patch in response to your comments. This patch has kind of blown up in size, so here's a new commit message:

EVG-17806 Add support for various find options for 'find' operations in CrudActor

This commit introduces support for a subset of find options for find operations, including 'projection',  'sort', 'skip', 'collation', 'tailable', and 'awaitData',

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
```diff
@@ -118,37 +125,277 @@ struct NodeConvert<mongocxx::options::count> {
...
-            auto c = node["Comment"].to<std::string>();
-            rhs.comment(std::move(c));
+        if (const auto& collation = node["Collation"]) {
+            rhs.collation(collation.to<bsoncxx::document::value>());
         }
-        if (node["Limit"]) {
-            auto limit = node["Limit"].to<int>();
-            rhs.limit(limit);
+        if (const auto& hint = node["Hint"]) {
+            rhs.hint(mongocxx::hint(std::move(hint.to<std::string>())));
```
dstorch, line 360: The fact that we only support string hints isn't ideal, but I guess it works ok for now. Perhaps we should at least leave a comment here stating that in the future this could be extended to work with hints expressed either as strings or documents.

src/cast_core/test/CrudActorYamlTestRunner.cpp
```diff
@@ -103,6 +103,67 @@ bool isNumeric(YAML::Node node) {
...
+                return kTailableAwait;
+            } else if (tailable) {
+                return kTailable;
+            } else {
+                // It is illegal to specify 'awaitData' without specifying 'tailable'.
+                REQUIRE(!awaitData);
+                return kNonTailable;
+            }
+        }();
+
```
dstorch, line 162: delete this line (consecutive lines of vertical whitespace)

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -716,3 +716,285 @@ Tests:
...
+    Operations:
+      - OperationName: findOne
+        OperationCommand:
+          Filter: { a: 1 }
+          Options:
+            Comment: &comment "Please show up in the logs"
+    ExpectAllEvents:
+      comment: *comment
+
+  - Description: Test that a 'findOne' command ignores the specified limit.
```
dstorch, line 921: Oh! That's a little wacky. I think we should make this an error with an explicit check in `CrudActor.cpp`.

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -716,3 +716,285 @@ Tests:
...
+    Operations:
+      - OperationName: findOne
+        OperationCommand:
+          Filter: { a: 1 }
+          Options:
+            Skip: 2
+    ExpectAllEvents:
+      skip: 2
+
+  - Description: Test that a 'findOne' command applies the specified batchSize.
```
dstorch, line 941: Also wacky. I'd argue that `batchSize` should be explicitly rejected when used with the findOne actor, implemented with a special check in `CrudActor.cpp`.

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -716,3 +716,285 @@ Tests:
...
+    Operations:
+      - OperationName: findOne
+        OperationCommand:
+          Filter: { a: 1 }
+          Options:
+            MaxTime: 1001 milliseconds
+    ExpectAllEvents:
+      maxTime: 1001
+
+  - Description: Test that a 'findOne' command can construct a non-tailable  cursor.
```
dstorch, line 961:
```suggestion
  - Description: Test that a 'findOne' command can construct a non-tailable cursor.
```

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -716,3 +716,285 @@ Tests:
...
     Error: |
       .*'Pipeline' must be an array.*
+
+  - Description: Test that a 'find' command applies the specified projection.
+    Operations:
+      - OperationName: find
+        OperationCommand:
+          Filter: {a: 100}
+          Options:
+            Projection: {_id: 0, b: 1}
```
dstorch, line 726: [optional] We don't have to do this if it's tricky, but is there an easy way to test that the `DocumentGenerator` is kicking in correctly for the `Projection` option? Maybe we could use a simple `^Inc` generator or the like in order to show that we can generate the first integer in a sequence?

dstorch (changes requested):
Thanks Mihai! Another round of comments. As I mentioned in person, there are a few where I want to coordinate with Ryan to make sure we are on the same page.

I'm really glad that you are taking on this work, seems like a great improvement!

src/cast_core/src/CrudActor.cpp
rtimmons: I have little opinion here. In terms of SEO, find with distinct options and projection is more discoverable and consistent with the shell. This is compelling enough for me. Perhaps we could detect `Projection` inside of `Options` and barf to be helpful?

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
```diff
@@ -17,6 +17,13 @@
 
 #include <string_view>
 
+#include <boost/throw_exception.hpp>
```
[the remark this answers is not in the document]
rtimmons: Thanks for helping to keep things tidy. We informally and infrequently use `clang-format` albeit without explicit configs or versions. Following the style-guide as best-effort as possible is all that we require.

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
```diff
@@ -118,37 +125,277 @@ struct NodeConvert<mongocxx::options::count> {
...
 
     static type convert(const Node& node) {
         type rhs{};
 
-        if (node["Hint"]) {
-            auto h = node["Hint"].to<std::string>();
-            auto hint = mongocxx::hint(std::move(h));
-            rhs.hint(mongocxx::hint(hint));
+        if (const auto& sort = node["Sort"]) {
+            rhs.sort(sort.to<bsoncxx::document::value>());
```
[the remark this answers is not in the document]
rtimmons: The conversions from yaml to C++ driver structs for options and the like has always bothered me. Ultimately it all just gets sent as bson; the type-system just gets in the way in this kind of library/framework code.

(Aside: One advantage of hand-crafted conversions is we can have intuitive error-messages and guard against mis-configurations (e.g. projections on empty finds).)

I'd _love_ to just use extended json. Perhaps there's 3p code that we could crib. But ultimately don't we still have to map to C++ driver structs since that's the driver's API?

I don't have a great suggestion for how to handle this. We haven't prioritized any foundational changes like this in Genny for a while, so I'm hesitant to suggest we ticket that work out. Perhaps add it to the [genny ideas doc](https://docs.google.com/document/d/1K0SlHoKkIACIXhkNP_KPHvb3IV6r53ReqsHqEYTnXDA/edit#heading=h.9sx1dtlnczfk) which we intend to 💅 once we ramp up on more genny investment in the (hopefully near) future.

src/cast_core/test/CrudActorYamlTestRunner.cpp
```diff
@@ -103,6 +103,67 @@ bool isNumeric(YAML::Node node) {
 // (Mostly fighting the compiler with yaml scalars versus the many numeric bson types.)
 // For now just assert a subset
 void requireEvent(ApmEvent& event, YAML::Node requirements) {
+    if (auto sort = requirements["sort"]) {
```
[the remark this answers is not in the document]
rtimmons: Unfortunately I don't have a better idea off-hand. The testing library in Genny has never received serious love, and I'd love more ideas there. I'll again punt to the [genny ideas doc](https://docs.google.com/document/d/1K0SlHoKkIACIXhkNP_KPHvb3IV6r53ReqsHqEYTnXDA/edit#) where we can flesh out better expect/actual invariant checks if you'd like.

src/cast_core/test/CrudActorYamlTestRunner.cpp
mtandrei: Done.

src/cast_core/test/CrudActorYamlTestRunner.cpp
```diff
@@ -103,6 +103,67 @@ bool isNumeric(YAML::Node node) {
...
+        REQUIRE(event.command["limit"].get_int64() == limit.as<int64_t>());
+    }
+    if (auto skip = requirements["skip"]) {
+        REQUIRE(event.command["skip"].get_int64() == skip.as<int64_t>());
+    }
+    if (auto batchSize = requirements["batchSize"]) {
+        REQUIRE(event.command["batchSize"].get_int32() == batchSize.as<int32_t>());
+    }
+    if (auto maxTime = requirements["maxTime"]) {
+        auto actualMaxTime = genny::TimeSpec(std::chrono::milliseconds{event.command["maxTimeMS"].get_int64()});
```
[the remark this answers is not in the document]
mtandrei: Done. I actually ran clang format and formatted this file (and others).

src/cast_core/test/CrudActorYamlTestRunner.cpp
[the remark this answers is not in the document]
mtandrei: I'm not sure that this necessarily matters for this patch. Like, all we're concerned about here is that the projection/sort/whatever you specify is actually passed through the command.

Unless you feel strongly about adding support for testing the actual results of the command, I'm going to leave this as is.

src/cast_core/src/CrudActor.cpp
```diff
@@ -634,13 +634,20 @@ struct FindOperation : public BaseOperation {
           _collection{std::move(collection)},
           _operation{operation},
           _filter{opNode["Filter"].to<DocumentGenerator>(context, id)} {
-        if (opNode["Options"]) {
-            _options = opNode["Options"].to<mongocxx::options::find>();
+        const auto& options = opNode["Options"];
+        if (options) {
```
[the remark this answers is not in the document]
mtandrei: Sure; done

src/cast_core/src/CrudActor.cpp
```diff
@@ -634,13 +634,20 @@ struct FindOperation : public BaseOperation {
...
+            if (options["Projection"]) {
+                _projection.emplace(options["Projection"].to<DocumentGenerator>(context, id));
+            }
         }
     }
 
     void run(mongocxx::client_session& session) override {
         auto filter = _filter();
+        if(_projection){
+            _options.projection((*_projection)());
```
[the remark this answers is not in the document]
mtandrei: Sure; done.

src/cast_core/src/CrudActor.cpp
```diff
@@ -658,6 +665,7 @@ struct FindOperation : public BaseOperation {
     mongocxx::collection _collection;
     mongocxx::options::find _options;
     DocumentGenerator _filter;
+    std::optional<DocumentGenerator> _projection =  std::nullopt;
```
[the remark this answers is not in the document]
mtandrei: Done.

src/cast_core/src/CrudActor.cpp
mtandrei: I would prefer to leave this as is if that's alright, as I don't want to keep going back and forth between the two. This matches the CXX driver spec, so unless anyone feels strongly I'm inclined to leave it as is.

mtandrei (commented):
Thanks for taking a look Ryan and Dave! I've responded to your comments in the latest commit.

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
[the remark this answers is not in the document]
mtandrei: On this note: now that the genny test code isn't being shared anymore, I'm going to close  https://jira.mongodb.org/browse/EVG-18364 as gone away unless there's other refactoring that can be done

src/cast_core/src/CrudActor.cpp
dstorch: Thanks for filing those tickets! Looks like all topics under this thread have been resolved now, clicking the "resolve" button.

src/cast_core/src/CrudActor.cpp
dstorch: Sounds like we're all fine with it as currently implemented. Resolving the thread.

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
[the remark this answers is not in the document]
dstorch: I'm resolving this thread, but I opened a new one with one more nitpick about how the includes are structured.

src/cast_core/include/cast_core/actors/OptionsConversion.hpp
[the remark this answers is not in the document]
dstorch: Woo! I'm happy that this was easy to do. I'll leave the thread open in case @rtimmons has anything to add, but I'm happy with its resolution.

dstorch (approved):
LGTM!

src/cast_core/test/CrudActorYamlTestRunner.cpp
```diff
@@ -103,6 +103,67 @@ bool isNumeric(YAML::Node node) {
...
+
+    if (auto cursorType = requirements["cursorType"]) {
+        const auto kNonTailable = "non_tailable";
+        const auto kTailable = "tailable";
+        const auto kTailableAwait = "tailable_await";
+        const auto cursorTypeString = cursorType.as<std::string>();
+
+        auto getBoolValue = [&](const std::string& paramName) {
+            auto val = event.command[paramName];
+            return val ? val.get_bool().value : false;
```
rtimmons, line 146:
```suggestion
            return val && val.get_bool().value;
```

rtimmons (dismissed):
Pro-level changes here, my friend. Thanks for really digging into this.

One stupidly minor suggestion that's probably not even worth test cycles to change.

Since this has _potential_ implications on pretty much every workload, please:

1. Do a quick patch-build of some genny workloads and compare them with using the perf analyzer to ensure this doesn't cause a regression. I can help with this or please reach out in `#performance-testing-users`. Should only take about 10 minutes after the tasks complete.
2. After merged, please take a look at the sys-perf and microbenchmarks waterfalls for the day and rollback or hotfix any fallout.

I don't suspect either of these to bite you or cause further delay in getting this to production, but it's due-diligence for deep changes.

src/cast_core/test/CrudActorYamlTestRunner.cpp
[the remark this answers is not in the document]
mtandrei: Going to go ahead and resolve this thread.

src/cast_core/test/CrudActorYamlTestRunner.cpp
mtandrei: Good call; done

mtandrei:
Patch build is here: https://spruce.mongodb.com/version/638a2da4d1fe071a90793102/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

(I only picked a couple of variants)

mtandrei:
I did a perf analysis on the patch build above and apparently a lot of the aggregate expression workloads regressed significantly (30% increase in 50th percentile latency). This is very surprising as this PR doesn't touch the aggregate operation for the CrudActor. I will try and repro these regressions locally and if they do repro, I'll dig into why. If they don't, I will go ahead and merge and monitor the waterfall as Ryan suggested.

mtandrei:
After doing some local testing/analysis, it doesn't look like this PR made any real difference on the aggregate expression workloads: I ran the aggregate_expressions task with versions of genny compiled with and without my changes and there's no real difference in operation throughput (I used perf_results_summary.py to compare the results). For posterity, I'll post some partial results from my local runs below: 


```
WITH MY CHANGES: 
 226 ExpressionAdd.AggExpressionAdd summary:
 227         timers.dur (measured in nanoseconds, displayed in milliseconds):
 228                 count     : 10
 229                 average   : 14.8
 230                 median    : 14.1
 231                 mode      : 15.9
 232                 stddev    : 0.4
 233                 [min, max]: [14.5, 15.9]
 234                 histogram:
 235                         [14,15): ****** (6)
 236                         [15,15): **     (2)
 237                         [15,15):        (0)
 238                         [15,15): *      (1)
 239                         [15,15):        (0)
 240                         [15,15):        (0)
 241                         [15,16):        (0)
 242                         [16,16):        (0)
 243                         [16,16):        (0)
 244                         [16,16]: *      (1)
 245         throughput:
 246                 ops       : 10.0
 247                 seconds   : 0.147772396
 248                 ops per second: 67.6716
 
 1126 ExpressionDegreesToRadians.AggExpressionDegreesToRadians summary:
1127         timers.dur (measured in nanoseconds, displayed in milliseconds):
1128                 count     : 10
1129                 average   : 14.9
1130                 median    : 14.2
1131                 mode      : 16.2
1132                 stddev    : 0.5
1133                 [min, max]: [14.6, 16.2]
1134                 histogram:
1135                         [15,15): ********       (8)
1136                         [15,15): *      (1)
1137                         [15,15):        (0)
1138                         [15,15):        (0)
1139                         [15,15):        (0)
1140                         [15,16):        (0)
1141                         [16,16):        (0)
1142                         [16,16):        (0)
1143                         [16,16):        (0)
1144                         [16,16]: *      (1)
1145         throughput:
1146                 ops       : 10.0
1147                 seconds   : 0.148613
1148                 ops per second: 67.2889


WITHOUT MY CHANGES: 
  76 ExpressionAdd.AggExpressionAdd summary:
  77         timers.dur (measured in nanoseconds, displayed in milliseconds):
  78                 count     : 10
  79                 average   : 14.4
  80                 median    : 13.8
  81                 mode      : 15.5
  82                 stddev    : 0.4
  83                 [min, max]: [14.1, 15.5]
  84                 histogram:
  85                         [14,14): ****   (4)
  86                         [14,14): **     (2)
  87                         [14,15): ***    (3)
  88                         [15,15):        (0)
  89                         [15,15):        (0)
  90                         [15,15):        (0)
  91                         [15,15):        (0)
  92                         [15,15):        (0)
  93                         [15,15):        (0)
  94                         [15,15]: *      (1)
  95         throughput:
  96                 ops       : 10.0
  97                 seconds   : 0.143977622
  98                 ops per second: 69.4552
  801 ExpressionDegreesToRadians.AggExpressionDegreesToRadians summary:
 802         timers.dur (measured in nanoseconds, displayed in milliseconds):
 803                 count     : 10
 804                 average   : 15.2
 805                 median    : 14.6
 806                 mode      : 16.7
 807                 stddev    : 0.6
 808                 [min, max]: [14.9, 16.7]
 809                 histogram:
 810                         [15,15): ****** (6)
 811                         [15,15): ***    (3)
 812                         [15,15):        (0)
 813                         [15,16):        (0)
 814                         [16,16):        (0)
 815                         [16,16):        (0)
 816                         [16,16):        (0)
 817                         [16,16):        (0)
 818                         [16,17):        (0)
 819                         [17,17]: *      (1)
 820         throughput:
 821                 ops       : 10.0
 822                 seconds   : 0.152041015
 823                 ops per second: 65.7717
 824
```

With all that said, I'm going to go ahead and merge and keep an eye on the waterfall over the next day or so (also, on my inbox to see if I get a perf BF :) ) 

Thanks again Ryan and Dave for taking a look at this one!

dstorch (approved):

PR #771, opened 2022-11-16
PERF-3364 Add Genny workloads to run QE performance tests

erwee:
This adds a subset of the tests performed in the PM-2759 QE performance project, but with reduced number of iterations so that it does not take more than a few hours to run on DSI.

There are two types of workloads:
1. the YCSB-like workloads that try to emulate a YCSB workload, but with queries on encrypted fields, as opposed to simply the '_id' field. These workloads are meant to test the performance of encrypted finds and update, for various contention factor settings and various number of encrypted fields.
2. the exponential compact workload, which tests the latency of a compaction operation relative to the size of the ECOC collection.

evg: https://spruce.mongodb.com/version/637557d3d1fe072d1a4f7ab8/tasks


rtimmons (approved):

erwee:
evergreen merge

PR #772, opened 2022-11-17
PERF-3461 Add workload to measure inserts on Best Buy dataset

cswanson310:
Accompanies other related PRs:
* https://github.com/10gen/dsi/pull/1137
* https://github.com/10gen/mongo/pull/8810


starzia (commented):
I am not expecting to be the final approver on this because I can't really review the genny c++ code, but otherwise it looks good to me.

src/cast_core/src/SamplingLoader.cpp
```diff
@@ -70,42 +70,67 @@ struct SamplingLoader::PhaseConfig {
...
+    samplePipeline.sample(config->sampleSize);
+    samplePipeline.project(make_document(kvp("_id", 0)));
+    const auto suffixPipe = pipeline_helpers::makePipeline(config->pipelineSuffixGenerator);
+    samplePipeline.append_stages(suffixPipe.view_array());
+
+    std::vector<bsoncxx::document::value> sampleDocs;
+    const int maxRetries = 3;
+    for (int nRetries = 0; nRetries < maxRetries; ++nRetries) {
+        try {
+            auto cursor =
```
IrinaYatsenko, line 84: If I'm reading the mongocxx code correctly, the cursor's interator is only of type `std::input_iterator_tag` which means the constructor using cursor.begin()/.end() will reallocate the vector O(log N) times (according to https://en.cppreference.com/w/cpp/container/vector/vector). The copy assignment in the line that follows will move from the temporary but it would still waste the allocation that was done when creating sampleDocs. If we want to be more perf-minded here, I'd suggest to reserve sampleDocs's capacity and then push_back into it in a loop from the cursor. Totally your call (for small samples none of this should matter and I'm thinking we should probably add a line to the actor's documentation discouraging usage of large samples, if anything, to reduce the probability of retries).

src/workloads/execution/BestBuyInserts.yml
```diff
@@ -0,0 +1,141 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  This workload measures bulk insert performance with the BestBuy dataset used in some other
```
IrinaYatsenko, line 4: rather than vaguely mentioning "some other workloads" I'd provide here the link to the dataset. I understand that the link can be found in one of the other two repos that you had to touch to make this workload, but... you know, the future readers would appreciate having it here as well :)

src/workloads/execution/BestBuyInserts.yml
```diff
@@ -0,0 +1,141 @@
...
+  Database: &db bestbuy
+
+  # Defines the collection name.
+  Collection: &Collection products
+
+  ColumnStoreIndexSpec: &ColumnStoreIndexSpec
+    key: {$**: "columnstore"}
+    name: "columnstore"
+
+  LoadThreads: &LoadThreads 8
```
IrinaYatsenko, line 49: Why 8? Would you mind adding a comment explaining the choice of the number?

src/workloads/execution/BestBuyInserts.yml
```diff
@@ -0,0 +1,141 @@
...
+
+  SamplingLoaderPhaseConfig: &SamplingLoaderPhaseConfig
+    Collection: *Collection
+    Database: *db
+    Repeat: 1
+    InsertBatchSize: 400
+    Batches: 20  # Per Thread
+    SampleSize: 4000
```
IrinaYatsenko, line 57: Do we need such a large sample size? Is BestBuy dataset highly heterogenious to require this for stability of the benchmarks?

src/workloads/execution/BestBuyInserts.yml
```diff
@@ -0,0 +1,141 @@
...
+
+Actors:
+# This insert load phase will run with the default dataset in place, which includes ~7 indexes
+# already.
+- Name: BaselineBestBuyIndexes
+  Type: SamplingLoader
+  Threads: *LoadThreads
+  Phases:
+  - *SamplingLoaderPhaseConfig
+  - {Nop: true}
```
IrinaYatsenko, line 67: I saw in other workloads syntax like:
```
    OnlyActiveInPhases:
      Active: [1]
      NopInPhasesUpTo: *MaxPhases
```
which I find more readable. Your choice, though.

src/workloads/execution/BestBuyInserts.yml
```diff
@@ -0,0 +1,141 @@
...
+    Operations:
+    - OperationMetricsName: BuildColumnStoreIndex
+      OperationName: RunCommand
+      OperationCommand:
+        createIndexes: *Collection
+        indexes:
+          - *ColumnStoreIndexSpec
+  - {Nop: true}
+
+- Name: ColumnStoreIndex
```
IrinaYatsenko, line 116: By now the dataset has gained 120K+ documents compared to the experiment with the default indexes. Do we care?

IrinaYatsenko (commented):
Looks mostly fine to me but filed a few questions I'd like to be clarified before signing off.

IrinaYatsenko:
The workload has been committed via https://github.com/mongodb/genny/pull/806

PR #773, opened 2022-11-18
PERF-3502 PipelineUpdate

SimonEismann:
This pull request provides two fixes to reduce the performance variability of this test:
* adds the `OperationIsQuiet: true` flag to reduce the size of the DSI artifacts from >5GB down to reasonable levels and stop the logging from accidentally rate-limiting the test for large documents
* adds a rate limit to the test which significantly reduces the observed variability between test runs

Evg: https://spruce.mongodb.com/version/637219811e2d173a1b6bab34/


src/workloads/execution/PipelineUpdate.yml
```diff
@@ -47,193 +47,244 @@ Actors:
...
 
 - Name: RunUpdates
   Type: RunCommand
   Threads: 16
   Phases:
   - *Nop
   - *Nop
-  - Repeat: 500    # Classic updates on small documents phase.
+  - *Nop
+  - Repeat: 300    # Classic updates on small documents phase.
```
jimoleary, line 73: Is the repeat change deliberate?

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -47,193 +47,244 @@ Actors:
...
+    GlobalRate: 2 per 100 milliseconds
     Database: *db
     Operations:
     - OperationMetricsName: ClassicUpdateAddFieldsSmallDoc
       OperationName: RunCommand
+      OperationIsQuiet: true
```
jimoleary, line 79: Is there another ticket to make true the default or could this change be implemented here. The change itself would be a oneliner [here](https://github.com/mongodb/genny/blob/c058d4243dcb4ce181eaaa422fafd8c99df606cb/src/cast_core/src/RunCommand.cpp#L92).

Although it could also cause a change in performance on some tasks.

jimoleary (commented):
Overall the changes look good to me.

I just have 2 questions.

src/workloads/execution/PipelineUpdate.yml
SimonEismann: I opened up a ticket around this a couple of days ago and Ryan seems to agree with the change:
https://jira.mongodb.org/browse/EVG-18233

src/workloads/execution/PipelineUpdate.yml
SimonEismann: It's deliberate in the sense that this keeps each of the four test phases at 15 seconds, similar to the duration before. But to be honest, that line or argument does not make a lot of sense :) Increasing this back to 500 might be a good idea, since it increases the test duration by only 40 seconds in total and might even slightly increase stability. I'll change it to 500.

src/workloads/execution/PipelineUpdate.yml
```diff
@@ -47,193 +47,244 @@ Actors:
...
+        Repeat: 1
 
 - Name: RunUpdates
   Type: RunCommand
   Threads: 16
   Phases:
   - *Nop
   - *Nop
+  - *Nop
   - Repeat: 500    # Classic updates on small documents phase.
```
jimoleary, line 73: [suggestion] Feel free to ignore to limit the changes, but:

```
  - Repeat: &RunCommandRepeat 500    # Classic updates on small documents phase.
Write
```
And then later:
```
  - Repeat: *RunCommandRepeat
```

If having all the same Repeat is desirable.

src/workloads/execution/PipelineUpdate.yml
jimoleary: Would it be worth doing  EVG-18233 first then? This PR would be much smaller as a result.

jimoleary (commented):
I'm happy to lgtm this now if you don't want to wait for EVG-18233.

src/workloads/execution/PipelineUpdate.yml
SimonEismann: Good idea, I've made the corresponding change.

jimoleary (approved):

PR #774, opened 2022-11-21
PERF-3572 Add genny test for writes without shard key on a replica set

jz1242:


src/workloads/sharding/WriteOneWithoutShardKeyReplicaSet.yml
```diff
@@ -0,0 +1,76 @@
...
+            Level: majority
+            TimeoutMillis: 5000
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - replica
+      - single-replica
+      - standalone
```
jz1242, line 76: Doing 3 node replica, 1 node, and standalone (which is a node that's not started as a replica set at all)

src/workloads/sharding/WriteOneWithoutShardKeyReplicaSet.yml
```diff
@@ -0,0 +1,76 @@
...
+- Name: CRUDOpsWithoutShardKey
+  Type: CrudActor
+  Threads: 1  # Use 1 thread to avoid updates throwing WriteConflict errors.
+  Phases:
+  - *Nop
+  - Repeat: *DocumentCount
+    Database: *Database
+    Collection: Collection0
+    Operations:
+    - OperationMetricsName: UpdateOneWithoutShardKey
```
sphanse99, line 35: There's no concept of a shardKey so truly this is just updates, deletes, and findAndModify testing, we should change the names of the operation metrics to reflect that.

That being said, if we already have this for replica sets, we might not need this genny workload. However, now that it's written, it might be easier to compare timing between this and the other two workloads since it's modifying the same number of documents per workload.

jz1242: removed and renamed file to remove notion of shard key. For the second comment I dont think our previously written tests are against replica sets (specifically) if thats what you meant? I wanted a test case that mirrored our other ones just performed on a replica set.

sphanse99: Not the test we wrote previously for this project, but just genny tests that already exist. But yeah, it's nice to have ones that directly mirror our other tests so I don't mind keeping this workload.

sphanse99 (approved):

jz1242:
evergreen merge

PR #777, opened 2022-11-22
PERF-3576 Add a script to analyze local CedarMetrics output

cswanson310:


src/workloads/contrib/analysis/test_result_summary.py
```diff
@@ -0,0 +1,360 @@
...
+        },
+        'second metric example': {...},
+        ...
+    }
+    """
+    results = {}
+    for (metric_name, diffed_readings) in metrics_of_interest.items():
+        if diffed_readings == []:
+            print(
+                "No measurements to summarize for %s metric for %s. Skipping" % (metric_name, actor_name))
```
ccstolley, line 119: `actor_name` not in scope here

ccstolley (approved):
I left 1 comment and I think you can drop `requirements.txt` entirely. This is really useful, thanks!

src/workloads/contrib/analysis/test_result_summary.py
cswanson310: Good catch! Fixed.

cswanson310:
evergreen merge

PR #780, opened 2022-11-29
PERF-3591 Make Object generator explicit whether to allow duplicated key names

IrinaYatsenko:
I've updated all existing uses of `^Object` to allow duplicate keys so that their semantics doesn't change. Have a couple questions about how to test this -- please see my comments in the files.


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1298,13 +1300,22 @@ class ObjectGenerator : public Generator<bsoncxx::document::value> {
           _keyGen{stringGenerator(node["havingKeys"], generatorArgs)},
           _valueGen{
               valueGenerator<false, UniqueAppendable>(node["andValues"], generatorArgs, parsers)},
-          _nTimesGen{intGenerator(extract(node, "withNEntries", "^Object"), generatorArgs)} {}
+          _nTimesGen{intGenerator(extract(node, "withNEntries", "^Object"), generatorArgs)},
+          _allowDuplicateKeys{node["allowDuplicateKeys"].to<bool>()} {}
```
IrinaYatsenko, line 1304: I've decided against providing a default as for the users of the Object generator it might be equally unexpected when the object would contain duplicated keys _or_ would have fewer keys than requested.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -177,6 +183,19 @@ Tests:
           xYYXCh/Ybl: { "$numberLong": "9" }
           EHcmXZX/d4: { "$numberLong": "10" }
 
+  - Name: Object Generator disallow duplicate keys
```
IrinaYatsenko, line 186: I'm not sure how to run the tests from this file. From the instructions in https://github.com/mongodb/genny/blob/b3e5f1775a65869fd4413857cbd19638e0e4b33f/docs/developing.md it's not obvious to me which kind of tests this one is. There is also no `src/resmokeconfig/genny_standalone.yml` file and when I tried `./run-genny resmoke-test --create-new-actor-test-suite` it failed with `No such file or directory: '../../src/genny/build/src/cast_core/cast_core_test'` message.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -177,6 +183,19 @@ Tests:
...
+      a:
+        {^Object: {
+          withNEntries: 3,
+          havingKeys: "c",
+          andValues: {^Inc: {}},
+          allowDuplicateKeys: false
+        }}
+    ThenReturns:
+      - a:
+          c: { "$numberLong": "1" }
```
IrinaYatsenko, line 197: YAML doesn't support duplicated mappings. I don't understand how these tests work and how to check that when the duplicates are allowed, the do make it through into the object. The test above with random strings happens to generate unique keys.

src/value_generators/src/DocumentGenerator.cpp
rtimmons, line 1304: The implementation is actually *skip* duplicate keys and I think that's actually much better behavior than retrying until we find a unique key.

With this in mind, consider making the syntax

```yaml
^Object:
  withNEntries: 10
  havingKeys: {^Foo}
  andValues: {^Bar}
  # allowDuplicateKeys: bool
  duplicatedKeys: skipped
```

This allows for additional duplicated keys behaviors to be added in the future.

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: Ya this is un-testable in the current yaml-based testing mechanism. I think it's worth adding a comment to the test that the test-case of allowing duplicate keys is not tested just so someone in the future knows to take extra caution if they mess with this part of the code.

src/value_generators/test/DocumentGeneratorTestCases.yml
rtimmons: Ya this is a known issue. [This jira comment](https://jira.mongodb.org/browse/EVG-18189?focusedCommentId=4971961&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4971961) should help.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -177,6 +183,19 @@ Tests:
...
           EHcmXZX/d4: { "$numberLong": "10" }
 
+  - Name: Object Generator disallow duplicate keys
+    GivenTemplate:
+      a:
+        {^Object: {
+          withNEntries: 3,
+          havingKeys: "c",
+          andValues: {^Inc: {}},
+          allowDuplicateKeys: false
```
rtimmons, line 193: This test-case wasn't obvious to me at first. I think add a comment here similar to what you had below. Even though the template has `withNEntries: 3`, the generated document will only have one entry because the key for each entry was specified as a constant, and `allowDuplicateKeys: false` prevented additional (duplicate) entries from being added. Something like that.

rtimmons:
Looks like the task is failing due to this line not being updated - https://github.com/mongodb/genny/blob/master/src/workloads/docs/Generators.yml#L245. Or thereabouts. The test output tells you the "path" to the node it was looking at as

```
GeneratorsSeeded.yml/Actors/0/Phases/0/Operations/0/OperationCommand/WriteOperations/1/Document/randomObject/^Object/allowDuplicateKeys
```

This is done after [GeneratorsSeeded.yml](https://github.com/mongodb/genny/blob/master/src/workloads/docs/GeneratorsSeeded.yml) is preprocessed and the `LoadConfig` applied.

rtimmons (dismissed):
Nothing blocking merge, just a couple things for you to consider. Thanks for rolling up your sleeves and addressing this.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1298,13 +1300,22 @@ class ObjectGenerator : public Generator<bsoncxx::document::value> {
...
+
+        std::unordered_set<std::string> usedKeys;
         for (int i = 0; i < times; ++i) {
-            _valueGen->append(_keyGen->evaluate(), builder);
+            auto key = _keyGen->evaluate();
+            if (_allowDuplicateKeys || usedKeys.insert(key).second) {
+                _valueGen->append(key, builder);
+            }
+        }
+        if (!_allowDuplicateKeys && usedKeys.size() < times) {
```
IrinaYatsenko, line 1317: This is probably too verbose: reduce severity or not log at all?

src/value_generators/src/DocumentGenerator.cpp
IrinaYatsenko: Do you mean that you'd like for the duplicated keys to have three possible behaviours:
1. skipped (equivalent in this patch to `allowDuplicateKeys: false`)
2. inserted (equivalent in this patch to `allowDuplicateKeys: true`)
3. retried (would require switching from bool to enum)

I can change the param from a bool to an enum but only add "skipped" and "inserted" values so that "retried" can be implemented in the future when/if it's actually needed. Sounds good?

src/value_generators/src/DocumentGenerator.cpp
rtimmons: I'm torn here. I think we can just leave as-is to see how commonly this actually comes up. This (most likely) points to incorrect configuration if happens often, so spamming the logs might be the best canary we can shove into this coal-mine.

src/value_generators/test/DocumentGeneratorTestCases.yml
IrinaYatsenko: I also attempted to add a test that might be able to observe at least some difference in the behaviour between the two cases. I gave up on running the tests locally, so if it fails in evergreen I'll only keep the test with "skip" setting.

src/value_generators/src/DocumentGenerator.cpp
rtimmons: For that case it might be easier to just have `{a: {^Choose}, b: {^Choose}, c:{^Choose}}` since you know you want an object with exactly 3 keys (a,b,c).

rtimmons (approved):

src/value_generators/src/DocumentGenerator.cpp
IrinaYatsenko: I actually wanted and object which sometimes misses one or two of the keys

PR #782, opened 2022-11-30
EVG-18409: Updates to get genny to work with new toolchain on m1

johndaniels:
This involved upgrading cpp files to use the new version of catch that is part of our toolchain. This directly depends on the catch cmake package rather than using an intermediary library.

Additionally, there are some changes to handling triples and where we install stuff as part of our setup scripts

This also rebuilds the poplar libs from grpc.


rtimmons (dismissed):
Cool! Nice if it works. I'd just patch-test this against DSI and select a few genny/auto tasks to smoke-test this. Only real blocker for me is using a waterfall toolchain build, but I guess that can be a fast-follow change if needed for timing/orchestration reasons.

opsnlops (approved):
There's too many changes for me to review each one, but I did a lot of testing on this during the project, so I'm okay with saying it's good.

PR #785, opened 2022-12-01
PERF-1811 Port sys-perf secondary_reads.js workload to Genny

SimonEismann:
This patch adds a genny version of the benchrun workload secondary_reads.js

Summary of performance differences:
- Latency/Throughput for the reads on the secondary are almost identical
- Latency/Throughput differs due to genny using the C-driver under the hood, but genny's behavior should be more realistic
- The genny version actually shows some replication lag at 32 reader threads, which shows that the test is more stressful in this scenario

Evergreen patch: https://spruce.mongodb.com/version/6387464b56234346595dba1d


src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
+SchemaVersion: 2018-07-01
+Owner: "@TODO" # what should go here? us or the owners of secondary_reads?
```
dalyd, line 2: Let's add someone from replication to this PR. I think it would be great of replication team owned this. 

@ldennis @lankas could you recommend someone from your teams to review this workload? As described, Simon is porting over the legacy workload of the same name.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+DocumentSize: &document_size 459 # Target document size of 500 and Object.bsonsize({_id: ObjectId(), x: 0, str: "x"}) + 1 = 41
+ReadLimit: &read_limit 100
+TestDurationInSecs: &test_duration 180 seconds
+Db: &db test
+WriterThreads: &writer_threads 16
+NumPhases: &max_phase 13
+
+Clients:
+  Secondary:
+    URI: mongodb://md1:27017
```
dalyd, line 40: How does this part work? Does this run out of the box or will we need to pass it in somehow from DSI?

dalyd: Do we need a separate default client entry for the inserts?

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+          Repeat: 1
+          Database: *db
+          Threads: 1
+          CollectionCount: 1
+          DocumentCount: *initial_nb_docs
+          BatchSize: *initial_nb_docs
+          Document: 
+            x: {^Inc: {}}
+            #s: {^Repeat: {count: *document_size, fromGenerator: {^Choose: {from: ["x"]}}}} --> why does this not work?
+            s: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
dalyd, line 62: We might just want {^RandomString: {length: *document_size}} instead of the constant string. 

To follow how you started, Repeat generates an array. You could feed it to ^Join to get a string. Also, the {^Choose: {from: ["x"]}} is equivalent to ["x"].  See https://github.com/mongodb/genny/blob/e0b70ac49cea140f0a9cda56cc21e0647fba9b53/src/workloads/docs/Generators.yml#L119

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+              drop: Collection0
+
+Actors:
+- ActorFromTemplate:
+    TemplateName: LoadInitialDataTemplate
+    TemplateParameters:
+      Name: LoadInitialData
+      ActivePhases: [0, 5, 10]
+
+- ActorFromTemplate:
```
dalyd, line 171: Is this shorter and clearer for using a template? How long is it if you just flat instantiate the quiesce actor here? I'm suspicious of templates that are only used once.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+        Active: {^Parameter: {Name: "ActivePhases", Default: [-1]}}
+        NopInPhasesUpTo: *max_phase
+        PhaseConfig:
+          Repeat: 1
+          Database: *db
+          Threads: 1
+          CollectionCount: 1
+          DocumentCount: *initial_nb_docs
+          BatchSize: *initial_nb_docs
+          Document:
```
dalyd, line 59: Assuming this is the same as below, consider adding a yaml anchor here for the document, and reusing in the other template.

dalyd: Oh, there is a difference. Consider yaml anchor for parts that are reused. I need to think about x.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+        Active: {^Parameter: {Name: "ActivePhases", Default: [-1]}}
+        NopInPhasesUpTo: *max_phase
+        PhaseConfig:
+          Duration: *test_duration
+          Collection: Collection0
+          Operations:
+          - OperationName: insertOne
+            OperationCommand:
+              Document:
+                x: {^RandomInt: {min: *initial_nb_docs, max: 1000000000}}
```
dalyd, line 97: You could inc here also, but have a start value. Unfortunately inc behaves interesting with multiple threads (it's not a shared increment).

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+    Type: CrudActor
+    Threads: {^Parameter: {Name: "Threads", Default: -1}}
+    ClientName: Secondary
+    Database: *db
+    Phases:
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "ActivePhases", Default: [-1]}}
+        NopInPhasesUpTo: *max_phase
+        PhaseConfig:
+          Duration: *test_duration
```
dalyd, line 113: Should this maybe be phase_duration rather than test_duration? Open for argument there, but I might think of the whole run as the test, and that is going to be multiples of this value.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+          Collection: Collection0
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: { x: {$gte: {^RandomInt: {min: 0, max: *initial_nb_docs}}}} # this should be *initial_nb_docs - *read_limit but we don't seem to have and add/subtract generator and it doesn't really impact performance
```
dalyd, line 118: Correct. We don't have shared state here. We could make it larger, but it might return nothing some of the time. 
The question in my mind is if wt cache effects matter, and I suspect he answer is no for this workload, but maybe the replication team could chime in here as well.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,252 @@
...
+        PhaseConfig:
+          Repeat: 1
+          Collection: Collection0
+          Operations:
+          - OperationName: insertOne
+            OperationCommand:
+              Document: {s: 1}
+              OperationOptions:
+                WriteConcern:
+                  Level: 3
```
dalyd, line 144: This deserves a comment at top that this has to run with 3 nodes. I wonder if we want w:majority instead, as that will be more forgiving, and it will give the majority lag time.

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: I initially started with {^RandomString: {length: *document_size}} and switched to the constant string as this was giving me closer disk I/O usage compared to the benchrun workload. But in general, a random string seems more realistic so I'll switch this over to the random string.

Thanks for the explanation about Repeat, that makes a lot of sense. The inserted document always just had str: x and I could not figure out why.

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Good point, while working with the test I've also noticed that templates aren't that great for readability so removing them where they are not necessary makes sense.

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Good to know that inc is limited to single-thread actors. I went with the random IDs to keep it in line with the original test, but it seems like I might have dodged a bullet by doing this.

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Good point, phase_duration is closer to the truth than test_duration.

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: The original test ensures that the reads never go outside the initial documents:

https://github.com/10gen/workloads/blob/083f9d81c593c07a69d762b84b06dc4b55b628e5/workloads/secondary_reads.js#L160

I wasn't able to do the same thing dynamically using genny as I couldn't find any way to add/subtract two values via a generator. Maybe the simples workaround here would be to add the result of the subtraction as a static parameter at the top?

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: This needs to be Writeconcern == NumberOfNodes since we only have the additional reads on one secondary so with w:majority this would not correctly measure the replication lag of this node under the additional stress of the reads.

I'll add a comment at the top that this needs to be run in a 3-node replica set. The smooth solution would be to dynamically refer to the size of the cluster here, then it could also run in a 2-node replica set, do you know of any way to do this?

SimonEismann:
Added a new version that includes the first round of feedback and (hopefully) resolves lint errors

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Note: waiting on adding the comment to see if there is a better solution

src/workloads/execution/SecondaryReadsGenny.yml
dalyd: Let's get a comment in here then. All black magic gets commented :)

src/workloads/execution/SecondaryReadsGenny.yml
dalyd: Inc will work with multi-threaded actors, but each thread will increment independently. Hopefully this comment makes sense: https://github.com/mongodb/genny/blob/3ff552c1414e5be4fe7e5c813e136cb1547b1999/src/workloads/docs/Generators.yml#L104-L108

Basically, you can make each thread increment in its own range.

src/workloads/execution/SecondaryReadsGenny.yml
dalyd: I do not know a good way to do that in Genny. In the queisce logic I think we explicitly query for how many nodes there are in the replica set, and use that value. 

There is a real `w:majority` write concern, but unfortunately no `w:all` write concern (I wish there was). 
I suppose we could make the write concern an parameter to the workload, and then we could override it to set it. I know that is possible, but it would take me some work to do it myself. I think we have a recent example though passing in a parameter.

dalyd (approved):
Changes LGTM. I would still love to get input from someone on the query team on it. I think that will help it be a better test. I'm going to tag Jim on one comment that he might have thoughts also.

src/workloads/execution/SecondaryReadsGenny.yml
dalyd: @jimoleary do you have thoughts on the above to do a `w:all` write and not explicitly know the replica set size in this file?

jimoleary: I don't know of any magic or cleverness to get around this. **At the end of the day, we are still hamstrung by the inability to pass parameters into a genny workload file from dsi.** 

The only way of stopping them being run on the wrong mongo setup is to ensure the value is set correctly for the mongodb_setup autorun. 

Making the value a parameter with the correct default is probably a good start anyway.

```
              OperationOptions: {^Parameter: {Name: "OperationOptions", Default: {WriteConcern:{Level: 3, Journal: true}}}}
```

I think that [LargeRange.yml](https://github.com/10gen/PrivateWorkloads/blob/production/src/workloads/customer_representative/coinbase/balance/LargeRange.yml) calling [Large.ym](https://github.com/10gen/PrivateWorkloads/blob/production/src/workloads/customer_representative/coinbase/balance/Large.yml) is the closed to what you may be alluding to. 

But storing these actors in the phases directory (like [this](https://github.com/mongodb/genny/blob/9354506965654f4fd23ba61d1c7ccb1163b145d3/src/phases/scale/DesignDocWorkloadPhases.yml) and loading them like [this](https://github.com/mongodb/genny/blob/0c6262b06e8f543d69de8ac271359dbb7c82fd20/src/workloads/scale/UpdateMillionDocsInTxn.yml#L14-L29) might be better.

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
+SchemaVersion: 2018-07-01
+Owner: "@TODO"  # what should go here? us or the owners of secondary_reads?
```
XueruiFa, line 2: Seems like the owners of the workload, so feel free to add Repl if we will be the future owners!

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
...
+ ### Test
+
+ On an empty replica set, the workload performs the following actions:
+ - Setup stage:
+    - Inserted *initial_nb_docs* documents into both the primary and the
+      secondaries. This should be done before any writes and reads.
+ - Actual test stage:
+    - *nWriters* background writers doing infinite writes on the primary which
+      will replicate these writes to secondaries.
+    - 1/16/32 readers doing queries on a sepcified seconary. Each query has
```
XueruiFa, line 21: [nit] "specified secondary"

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
...
+DocumentSize: &document_size 459  # Target document size of 500 and Object.bsonsize({_id: ObjectId(), x: 0, str: "x"}) + 1 = 41
+ReadLimit: &read_limit 100
+MaxIdForFinds: &max_id_for_finds 9900
+TestDurationInSecs: &phase_duration 180 seconds
+Db: &db test
+WriterThreads: &writer_threads 16
+NumPhases: &max_phase 13
+NumberOfNodes: &number_of_nodes 3
+
+# This client allows the ReadsOnSecondary actor to directrly connect to the secondary node.
```
XueruiFa, line 40: [nit] "directly"

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
...
+- TemplateName: ReadsOnSecondaryTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: ReadsOnSecondary}}
+    Type: CrudActor
+    Threads: {^Parameter: {Name: "Threads", Default: -1}}
+    ClientName: Secondary
+    Database: *db
+    Phases:
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "ActivePhases", Default: [-1]}}
```
XueruiFa, line 76: Just for my own knowledge, where can I find the list of phases for this workload?

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
...
+ On an empty replica set, the workload performs the following actions:
+ - Setup stage:
+    - Inserted *initial_nb_docs* documents into both the primary and the
+      secondaries. This should be done before any writes and reads.
+ - Actual test stage:
+    - *nWriters* background writers doing infinite writes on the primary which
+      will replicate these writes to secondaries.
+    - 1/16/32 readers doing queries on a sepcified seconary. Each query has
+      *batch_size* numbers of documents and only reads documents inserted in the
+      setup stage in order to have more constant behavior all the time.
```
XueruiFa, line 23: What constant behavior are we looking for here? Is this performance of the secondary read, so that we guarantee that the document we're trying to read already exists on the secondary?

src/workloads/execution/SecondaryReadsGenny.yml
```diff
@@ -0,0 +1,231 @@
...
+      Active: [0, 5, 10]
+      NopInPhasesUpTo: *max_phase
+      PhaseConfig:
+        Repeat: 1
+        Database: *db
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *initial_nb_docs
+        BatchSize: *initial_nb_docs
+        Document:
```
XueruiFa, line 129: I'm not the most familiar here so apologies if this is done elsewhere, should these commands be done with writeConcern equal to the number of nodes, so that we guarantee they land on all secondaries before reading?

XueruiFa (changes requested):
Thanks for doing this work!

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Added repl as owners, but feel free to reach out if there are ever any issues questions around the genny stuff :)

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Fixed

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Fixed, maybe it's time to install a spellchecker in my VSCode :)

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: From lines 46-111, the workload defines a number of templates. Those templates are then used in the `ActorFromTemplate` block on lines 161-216, where the phase is passed as an argument. So the ReadsOnSecondary actor is running in phases 2, 7, and 12.

As for a general overview of the phases in this test:
```
0: LoadInitialData
1: Quiesce (fsync, etc)
2: WritesOnPrimary (16 Threads) + ReadsOnSecondary (1 Thread)
3: ReplicationLag (measure replication lag via write with w:3)
4: CleanUp (drop collection)
5: LoadInitialData
6: Quiesce
7: WritesOnPrimary (16 Threads) + ReadsOnSecondary (16 Threads)
8: ReplicationLag
9: CleanUp
10: LoadInitialData
11: Quiesce
12: WritesOnPrimary (16 Threads) + ReadsOnSecondary (32 Threads)
13: Replication lag
```

src/workloads/execution/SecondaryReadsGenny.yml
SimonEismann: Good catch, it's important to ensure that the documents are propagated to all nodes before we start the test. We currently do this via the Quiesce phase after loading the initial data, but wc:3 would also work.

SimonEismann:
@XueruiFa I've incorporated your feedback, let me know if this lgtms for you.

src/workloads/execution/SecondaryReadsGenny.yml
jimoleary: This list of phases has value, can we add it as a comment?

**Also, what does a ActivePhases value of [-1] actually do? I hope it throws an exception but I suspect it does nothing.**

src/workloads/execution/SecondaryReadsGenny.yml
jimoleary: I don't think Quiesce does that, it just flushes all the local data.

Wouldn't we would either have to use w:3 for all or a single extra w:3 after the LoadInitialData phases. Option 1 would be the easiest assuming it doesn't add too much overhead to the test runtime.

dalyd: The quiesce does do a "w:all". It figures out how many nodes there are and uses that in the write concern among other things. https://github.com/mongodb/genny/blob/master/src/gennylib/src/quiesce.cpp#L68-L80

SimonEismann: TIL, I thought that was just a fysnc, which would not have been sufficient. But seems we can keep this part as is :)

jimoleary: I think I must have known that in the past but had forgotten so good to know!

jimoleary (dismissed):
it looks like there are other outstanding comments but I'm mostly lgtm.

dalyd (dismissed):

jimoleary (approved):

PR #787, opened 2022-12-02
PERF-3056 Add workloads for fragmented inserts in presence of columnstore index

IrinaYatsenko:
There are two workloads, one with flat and one with nested data schema. The nested workload doesn't do wildcard index because creating the index takes too long (the time is proportional to the number values per document and when using arrays it explodes). However, for CSI we don't expect this workload to deviate too much from the flat schema, that is, the relative slowdown compared to the no-index case should be similar for both workloads.


src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,271 @@
...
+  and in presence of either columnstore or wildcard indexes. It uses an artificial data set with
+  a wide overall schema and narrow individual objects to model fragmented access to CSI, which
+  clusters entries by path.
+
+Keywords:
+- columnstore
+- wildcard
+- insert
+
+AutoRun:
```
cswanson310, line 14: I'm pretty unsure how this stuff works. This section means we don't have any new task associated with this workload, is that right? How will it show up in evergreen - both in your patch builds and once it's in the mainline? Do we need to worry about a 6.1 and 6.2 branch, or are those already disabled?

We have this open TODO, so I've been kinda ignoring this for the time being: https://jira.mongodb.org/browse/PERF-3094

src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,271 @@
...
+      NopInPhasesUpTo: *maxPhases
+      PhaseConfig:
+        # Cannot have more threads than the actor itself.
+        Threads: 1
+        Repeat: 1
+        Database: *db
+        CollectionCount: 1
+        DocumentCount: *docCount
+        BatchSize: 1000
+        Document:
```
cswanson310, line 125: If we have this document template - do we need to use the sampling loader? I would think we could just save this 'Document' as an alias and use the 'Loader' throughout? For some of these we want to exactly repeat the same document, for which we could hard-code one example.

src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,271 @@
...
+      NopInPhasesUpTo: *maxPhases
+      PhaseConfig:
+        Repeat: 1
+        Thread: 1
+        Database: admin
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            internalQueryFrameworkControl: "trySbeEngine"
```
cswanson310, line 108: Oh also, I don't think you need this command since you're not doing any queries from the CSI.

IrinaYatsenko: Good to know. I was wondering about it :)

src/workloads/query/CsiFragmentedInsertsFlat.yml
IrinaYatsenko: I considered using a hard-coded example and decided against it because it would prevent easy experimentation with different object widths or field distributions. The Loader could be used for inserting docs with randomized schema but I believe it would be easier to understand and maintain the workload if the same actor was used for the inserts we are trying to measure. The sampling loader, inho, fits quite well the scenarios here.

src/workloads/query/CsiFragmentedInsertsNested.yml
```diff
@@ -0,0 +1,245 @@
...
+      NopInPhasesUpTo: *maxPhases
+      PhaseConfig:
+        # Cannot have more threads than the actor itself.
+        Threads: 1
+        Repeat: 1
+        Database: *db
+        CollectionCount: 1
+        DocumentCount: *docCount
+        BatchSize: 1000
+        Document:
```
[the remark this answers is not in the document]
IrinaYatsenko: Answered in the other thread.

IrinaYatsenko:
The numbers from Evergreen for "cold" phase were all over the place so I removed the tests as too noisy and kicked off another 5 patch runs.

src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,258 @@
...
+          #     x271: NumberInt(9917),
+          #     x6305: NumberInt(11),
+          #     x8: NumberInt(1022),
+          #     <7 more fields like this>
+          #   }
+          # }
+          root: {^Object: {
+            withNEntries: *objectWidth,
+
+            # Do we want to use a non-uniform distribution for the fiedls, modelling that some are
```
starzia, line 125: spelling of "fields"

src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,258 @@
...
+
+Actors:
+# This workload is completely linear with one actor per phase. Having phases when nothing happens
+# is mildly wastful, but I'm tired of having to adjust the numbers, so assuming "sparse" numbering
+# scheme that is more stable.
+# Phases:
+# 0 setup (no index), 10 create CSI, 20 replace CSI with wildcard index
+# 1,                  11,            21                                  - quiese
+# 3,                  13,            23                                  - insert tail
+# 4,                  14,            24                                  - insert warm
```
starzia, line 94: Why do you call the first insert phase the "tail"?  I find this confusing.  I would have assumed this was just a "cold" or "warming" phase.

starzia (approved):
The results in evergreen look good.  I left some comments about comments and naming.

src/workloads/query/CsiFragmentedInsertsFlat.yml
IrinaYatsenko: The goal here was to observe the warm up behaviour and expecting to see a long tail. The test has been dropped per the discussion in https://mongodb.slack.com/archives/C03UE6J2AQ7/p1670952227018319?thread_ts=1670901746.133279&cid=C03UE6J2AQ7

src/workloads/query/CsiFragmentedInsertsFlat.yml
IrinaYatsenko: updated the comment

IrinaYatsenko:
https://spruce.mongodb.com/version/639a3eb63e8e865a8258050b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
https://spruce.mongodb.com/version/639a3eea9ccd4e4e6d5fd5ae/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
https://spruce.mongodb.com/version/639a3f0ea4cf472ac6028297/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

IrinaYatsenko:
Analyzed the results from the three runs above in https://docs.google.com/spreadsheets/d/1cBq78eHa6FlVNwEO0lX-5Vb4w-eggP3IqCU-HoNrMjE/edit#gid=665653627. The noise looks low.

src/workloads/query/CsiFragmentedInsertsNested.yml
```diff
@@ -0,0 +1,167 @@
...
+GlobalDefaults:
+  MaxPhases: &maxPhases 5
+  Database: &db csiFragmentedInsertsNested
+  Collection: &coll Collection0
+  DocumentCount: &docCount 1e6
+  SchemaWidth: &schemaWidth 10000
+  ObjectWidth: &objectWidth 5  # with two nested paths, get 10 paths_per_object
+  SampleSize: &sampleSize 10
+
+  Document: &document
```
dalyd, line 44: You can (don't have to) use LoadConfig for the common features between these two files. https://github.com/mongodb/genny/blob/3681cd089bb340eef3ed4a3643074d6a4171bc4a/docs/using.md#orga6d35c7

For example, I wouldn't be surprised if your two InsertFromSample templates are identical. You could write it once and load it into each file.

src/workloads/query/CsiFragmentedInsertsFlat.yml
```diff
@@ -0,0 +1,183 @@
...
+  and in presence of the columnstore index. It uses an artificial data set with a wide overall
+  schema and narrow individual objects to model fragmented access to CSI, which clusters entries by
+  path. The data size is relatively small (1e6 documents yield ~175MB data size and ~105MB storage
+  size).
+  We would like to be able to correlate the results of this workload with the similar one that uses
+  nested data (CsiFragmentedInsertsNested.yml). Please make sure to update both when making changes.
+
+Keywords:
+- columnstore
+- wildcard
```
dalyd, line 14: Are you actually testing against wildcard here? I don't see it.

src/workloads/query/CsiFragmentedInsertsFlat.yml
dalyd: This looks right to me. The test is in a patch build.

dalyd (commented):
This all looks pretty cool. And thank you for including the metadata and useful comments.

src/workloads/query/CsiFragmentedInsertsFlat.yml
IrinaYatsenko: oops, at the start we did test here against wildcard as well but since then the workload got, em, simplified :)

src/workloads/query/CsiFragmentedInsertsFlat.yml
IrinaYatsenko: Thank you for the confirmation! I'll keep it as is.

src/workloads/query/CsiFragmentedInsertsNested.yml
IrinaYatsenko: I'd like to keep the file for Flat workload as readable as possible on its own. Would it be OK, rather than loading some common file into both workload, to load into Nested from Flat?

dalyd: Yeah, that should be fine.

dalyd: Also, only if it makes sense to you.

IrinaYatsenko: Actually, the two workloads differ only in three parameters: object width, db name and the document generator. Is it possible to have a single file that describes the test but takes these three parameters to create two separate tasks?

dalyd: I'm not sure what the best path is. I could get it to two files, with one of the files being very small. I've asked in slack and tagged you on the thread.

rtimmons: Genny has no option for a single YAML file to produce multiple tasks.

This came up in a couple other conversations recently, so I think it's worth considering as a new syntax thing.

I think this fits with AutoRun pretty nicely:

```yaml
Actors:
- Type: {^Parameter: {Name: ActorType}}
AutoRun:
- When:
    mongodb_setup:
      $eq: replica
  ThenRun:  # Doesn't work, just a sketch.
  - DSIParameters:
      mongodb_setup: replica-delay-mixed
    GennyParameters:
      ActorType: Delayed
  - DSIParameters:
      mongodb_setup: replica-delay-mixed
    GennyParameters:
      ActorType: NotDelayed
```

This would be surprisingly low effort to add, but tbh I don't super love the syntax. And Genny's yaml syntax is already reaching critical mass of understandability, so I want to tread carefully here. (E.g. I completely forgot that `ActorTemplates` was a thing.)

ANYWAY. The end state here is subjective unless we invest in more syntax.

My $0.02 is to opt for a single parameterized workload yaml file that has these three parameters defined. Then two yaml files that just provide the parameters. Class and instance style kinda.

IrinaYatsenko: Thanks! Would you mind pointing me to an existing example of such parameterized workloads (ideally, that are AutoRun)?

starzia (approved):
I'm giving the missing official approval here, after reviewing the feedback from several experts on this PR.

src/workloads/query/CsiFragmentedInsertsNested.yml
IrinaYatsenko: I'm going to merge as is. I'll be working on similar workloads for update/delete soon and can refactor this one at the same time, after I learn more about using genny _well_ (it doesn't come naturally, btw :P)

PR #795, opened 2022-12-14
PERF-3659 Loosen assertions on encrypted field schemas in Genny's encryption_test.cpp

erwee:
This rewrites a couple of test cases in encryption_test.cpp that had too strict assertions on the layout of the generated encrypted field schema documents, which resulted in test failure when ran on M1 Macs.


jimoleary:
@dalyd Adding you to this PR

dalyd (approved):
I assume this is just bson comparion problem? 

The change looks fine to me. I feel like there's probably a better way to do document comparison here, but I'm not finding it yet.

erwee:
> I assume this is just bson comparion problem?
> 
> The change looks fine to me. I feel like there's probably a better way to do document comparison here, but I'm not finding it yet.

Yes, this is just a BSON comparison issue.

PR #798, opened 2022-12-15
PERF-3635 Share a sample across different SamplingLoader threads

cswanson310:


src/value_generators/include/value_generators/PipelineGenerator.hpp
```diff
@@ -29,16 +29,30 @@ namespace genny {
 struct PipelineGenerator {
     PipelineGenerator() = default;
 
-    PipelineGenerator(const Node& node, PhaseContext& context, ActorId id) {
-        if (!node.isSequence()) {
-            BOOST_THROW_EXCEPTION(InvalidConfigurationException("'Pipeline' must be an array"));
+    PipelineGenerator(const Node& node, ActorContext& context) {
+        assertIsArray(node);
+        for (auto&& [_, stageNode] : node) {
+            // The '0' here is a lie, but we don't necessarily have an actor ID yet in this
```
cswanson310, line 35: Oops the '0' is stale (we start at 1), but this is the part of the patch I'm not sure about. Should I make ActorId optional in the DocumentGenerator class? And then assert that it's present in the `{^ActorId` generator?

cswanson310: @rtimmons looking for your input here specifically.

IrinaYatsenko (approved):
Looks good to me but please get a sign off from someone more familiar with genny's arch

src/cast_core/include/cast_core/actors/SamplingLoader.hpp
```diff
@@ -36,21 +81,27 @@ namespace genny::actor {
 class SamplingLoader : public Actor {
 
 public:
-    explicit SamplingLoader(ActorContext& context);
+    SamplingLoader(ActorContext& context,
+                   std::string dbName,
+                   std::string collectionName,
+                   std::shared_ptr<DeferredSample> deferredSamplingLoader);
```
rtimmons, line 87: Is it "better" to use `ShareableState`?

https://github.com/mongodb/genny/blob/bfaff11cabfa9e9f362458f597943259e754f1f1/src/cast_core/include/cast_core/actors/MonotonicSingleLoader.hpp#L81-L82

It's slightly before my morning coffee. And what you have is fine. But either way, consider a comment here about why this is done versus `ShareableState` so future copy/pastas know the right tradeoffs to consider.

src/cast_core/include/cast_core/actors/SamplingLoader.hpp
```diff
@@ -36,21 +81,27 @@ namespace genny::actor {
...
 
+
 private:
     /** @private */
     struct PhaseConfig;
 
     metrics::Operation _totalBulkLoad;
     metrics::Operation _individualBulkLoad;
     mongocxx::pool::entry _client;
+    mongocxx::collection _collection;
```
rtimmons, line 103: Intuitively we wouldn't need both a client and a collection since a collection comes from its client. Consider a comment here explaining why both are necessary.

src/cast_core/src/SamplingLoader.cpp
```diff
@@ -48,67 +48,84 @@ using bsoncxx::builder::basic::make_document;
...
+std::vector<bsoncxx::document::value> DeferredSample::getSample() {
+    std::lock_guard<std::mutex> lock(_mutex);
+    if (_sampleDocs.empty()) {
+        _sampleDocs = gatherSample(lock);
+    }
+    return _sampleDocs;
+}
+
+std::vector<bsoncxx::document::value> DeferredSample::gatherSample(
+    const std::lock_guard<std::mutex>& lock) {
```
rtimmons, line 69: Out of curiosity, why is the lock guard passed in? It doesn't seem used here, so perhaps this is a C++-ism I'm not familiar with?

src/cast_core/src/SamplingLoader.cpp
```diff
@@ -138,15 +156,57 @@ void genny::actor::SamplingLoader::run() {
...
+      _deferredSample{std::move(deferredSample)},
       _loop{context, _client, SamplingLoader::id()} {}
 
+class SamplingLoaderProducer : public genny::ActorProducer {
+public:
+    SamplingLoaderProducer(const std::string_view& name) : ActorProducer(name) {}
+    genny::ActorVector produce(genny::ActorContext& context) {
+        if (context["Type"].to<std::string>() != "SamplingLoader") {
+            return {};
+        }  // namespace genny::actor
```
rtimmons, line 178: 💅

src/value_generators/include/value_generators/PipelineGenerator.hpp
rtimmons: This is a good question. But I'm confused about why we don't necessarily have the ActorId here when we do usually have it when constructing other generator types. Is there a usage pattern that we don't capture well?

src/cast_core/src/SamplingLoader.cpp
```diff
@@ -138,15 +156,57 @@ void genny::actor::SamplingLoader::run() {
     }
 }
 
-SamplingLoader::SamplingLoader(genny::ActorContext& context)
+SamplingLoader::SamplingLoader(genny::ActorContext& context,
+                               std::string dbName,
+                               std::string collectionName,
+                               std::shared_ptr<DeferredSample> deferredSample)
```
rtimmons, line 162: If you use `ShareableState` here, you don't need a custom Producer as well. Again, not sure of the implications.

rtimmons (dismissed):
Nothing that looks block-worthy. Nicely done overall. Thanks for making this surgical.

src/cast_core/include/cast_core/actors/SamplingLoader.hpp
cswanson310: Uhh I am not understanding the point of SharableState... Maybe you can help? Is this to make it easier so that I don't have to add the custom actor construction code?

src/cast_core/src/SamplingLoader.cpp
cswanson310: Oh yeah this was a quick and dirty carry-over of the "WithLock" concept in our server codebase: https://github.com/mongodb/mongo/blob/42667e3c762bd83d700696ec4635e41fb8471c0f/src/mongo/util/concurrency/with_lock.h#L40

I didn't carry over that type, but the intention was the same: make it impossible to call the function without holding the lock.

That being said - after typing this out. This function does not particularly _need_ to be called in the lock. The dangerous part is the assignment to _sampleDocs in the calling code.

src/cast_core/src/SamplingLoader.cpp
cswanson310: Oh of course now I am reading this after responding to your other comment :D I'll give that a go

src/value_generators/include/value_generators/PipelineGenerator.hpp
cswanson310: Well, conceptually we don't have an ActorId because we are trying to generate a single pipeline to be used to generate a single sample to be shared across all threads (which I am pretty sure all have different ActorIds).

We maybe could technically get an ActorId in here if I reworked a bunch of code to avoid even creating the PipelineGenerator until we are lazily loading the sample, but it would be a random ActorId - whichever one won the race.

src/cast_core/include/cast_core/actors/SamplingLoader.hpp
cswanson310: I think that a collection can't exist without it's client still being around in scope - so it's just convenience to have the collection member variable to save and use directly.

src/cast_core/src/SamplingLoader.cpp
cswanson310: Coming back again - it does still feel worthwhile to protect this, since we use the pipeline suffix generator which is not thread safe. I didn't change anything.

src/cast_core/src/SamplingLoader.cpp
cswanson310: Oops! Done.

src/cast_core/include/cast_core/actors/SamplingLoader.hpp
cswanson310: Ok after starting to try it I determined it would not actually work, since we want to be able to use one sample per phase, but SharableState only supports one sample per workload. Added a comment.

cswanson310:
evergreen patch

cswanson310:
Started a perf patch to see if this had any meaningful impact on the CSI workloads that use this: https://spruce.mongodb.com/version/63a47f78a4cf474d89a856e1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I don't expect it to really.

IrinaYatsenko (approved):

IrinaYatsenko:
Looked at the results of the above perf run for the csi workloads that use the sampling loader (csi_fragmented_inserts_flat and csi_fragmented_inserts_nested) and they are fine.

PR #806, opened 2023-01-12
PERF-3461 New workload for inserting into a large dataset in presence of CSI index

IrinaYatsenko:
Companion PRs:
https://github.com/10gen/dsi/pull/1137
https://github.com/10gen/mongo/pull/8810

Evergreen patches:
https://spruce.mongodb.com/version/63b870a02fbabe308124399b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC (1 run)
https://spruce.mongodb.com/version/63bdbfbb0305b968659ffb61/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC (3 reruns)
https://spruce.mongodb.com/version/63bf3fbd3627e06c3018361a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC (1 run)

### DIFF
--- a/src/workloads/query/CsiHeavyDiskUsage.yml
+++ b/src/workloads/query/CsiHeavyDiskUsage.yml
@@ -40,7 +40,7 @@ GlobalDefaults:
 
   LoadThreads: &loadThreads 8
   SampleSize: &sampleSize 1000
-  Batches: &batches 50 # Per Thread
+  Batches: &batches 50  # Per Thread
 
   ColumnStoreIndexSpec: &ColumnStoreIndexSpec
     key: {$**: "columnstore"}
@@ -53,7 +53,7 @@ GlobalDefaults:
     Collection: *coll
     Repeat: 1
     InsertBatchSize: 1000
-    Batches: 50 # Per Thread
+    Batches: 50  # Per Thread
 
 ActorTemplates:
 # We want to be able to compare results from the same test between the runs in presence of different
@@ -69,7 +69,7 @@ ActorTemplates:
     Threads: *loadThreads
     Phases:
       OnlyActiveInPhases:
-        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1024}}] # do not run by default
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1024}}]  # do not run by default
         NopInPhasesUpTo: *maxPhases
         PhaseConfig:
           InsertBatchSize: *sampleSize
--- a/src/workloads/query/CsiHeavyDiskUsage.yml
+++ b/src/workloads/query/CsiHeavyDiskUsage.yml
@@ -6,19 +6,18 @@ Description: |
   workload itself has no expectations as far as the size or schema of the data are concerned but it
   is only useful when being run against larger datasets to actually generate heavy disk IO.
 
-  The purpose of this workload is to compare the insert performance with columnar indexes to inserts
-  without them.
-   - "Default" indexes present - a set of indexes that are typically present on the target dataset,
-     those that would be useful for the typical query workloads. These indexes should be created
-     prior to running the workload.
-   - Only a column store index. A column store index isn't expected to realistically replace all
-     other indexes in production, but it could replace some and knowing it's relative performance
-     cost could help make that decision.
+  The purpose of this workload is to compare the insert performance in the following two situations:
+   - "Default" indexes are present - a set of indexes that are typically present on the target
+     dataset, those that would be useful for the typical query workloads. These indexes should be
+    created prior to running the workload.
+   - Only a column store index is present. A column store index isn't expected to realistically
+     replace all other indexes in production, but it could replace some and knowing its relative
+     performance cost could help make that decision.
 
   The workload consists of the following phases and actors:
-   - Measure bulk inserting with the default indexes (SamplingLoader)
+   - Measure bulk insertion with the default indexes (SamplingLoader)
    - Drop all non-_id indexes and build the column store index (RunCommand Actor)
-   - Measure bulk inserting again (SamplingLoader)
+   - Measure bulk insertion again (SamplingLoader)
 
 Keywords:
 - columnstore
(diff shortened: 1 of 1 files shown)


src/workloads/query/CsiHeavyDiskUsage.yml
```diff
@@ -0,0 +1,162 @@
...
+    name: "columnstore"
+
+  MaxPhases: &maxPhases 5
+
+  SamplingLoaderPhaseConfig: &SamplingLoaderPhaseConfig
+    Database: *db
+    Collection: *coll
+    Repeat: 1
+    InsertBatchSize: 1000
+    Batches: 50  # Per Thread
```
mdashti, line 56: This seems to be dead-code. I don't see it to be used/referenced anywhere.

src/workloads/query/CsiHeavyDiskUsage.yml
```diff
@@ -0,0 +1,162 @@
...
+    Pipeline: [{$set: {genny_added: true}}]
+    Threads: *loadThreads
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1024}}]  # do not run by default
+        NopInPhasesUpTo: *maxPhases
+        PhaseConfig:
+          InsertBatchSize: *sampleSize
+          Repeat: 1
+          Batches: *batches
```
mdashti, line 77: I looked into the `SamplingLoader` and apparently, it doesn't have a deterministic approach in adding more documents to a given collection. Also, this way of populating the collection hides the property of documents that are going to be indexed.

IMO, there are several dimensions that are interesting to see the performance of indexing over a collection (CSI vs non-CSI): 1) having big documents with too many columns 2) having a fixed schema vs. changing schema (for each different document), etc. IMO, right now, with the current perf test setting, it's not easy to see what's the property of source collection. Can you please consider using another actor-type, like a `LoaderPhase` (e.g., [here](https://github.com/mongodb/genny/blob/7bdf0e15a6cff84e7706a2b20398ae555f2ced90/src/workloads/scale/MatViewParamsGeneric.yml#L85-L93)) to make the document structure more clear in the experiment setup?

mdashti (changes requested):
@IrinaYatsenko Thanks for the PR. Please take a look at my comments and let me know if any of them is not clear.

src/workloads/query/CsiHeavyDiskUsage.yml
```diff
@@ -0,0 +1,162 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-execution"
+Description: |
+  This workload measures bulk insert performance against a collection named "coll" in a database
+  named "heavy_io". The collection and database must be set up before running the workload. The
+  workload itself has no expectations as far as the size or schema of the data are concerned but it
+  is only useful when being run against larger datasets to actually generate heavy disk IO.
+
+  The purpose of this workload is to compare the insert performance with columnar indexes to inserts
+  without them.
```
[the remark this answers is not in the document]
IrinaYatsenko: rephrased

mdashti (dismissed):
@IrinaYatsenko Thanks for applying the changes. LGTM.

dalyd:
@IrinaYatsenko the workload itself looks fine, but I'd like to broach an idea:

This workload is really running two experiments in order to compare them. One option would be to simplify the workload to just be the sampling loader, and make two different test_control files in DSI (one for stock data with default indexes, one with the column store index). 

This would do a couple of things. One, I think it makes it easier to compare, as now we are comparing two tasks to each other, and we have some tooling for that. It at least doesn't make it much harder to compare the two cases. SBE and other things are comparing task to task today, although across build variants. Two, it makes it a fairer test. You made sure to remove the added documents before running the second test, but I suspect there is still a difference in the underlying btrees and wiredTiger state due to adding and removing a lot of documents. Three: it makes it easier to extend if you discover there is something else you want to try, or a change in how you create a columnstore index. 

I'm double checking the feasibility of this idea to avoid any gotchas, but I think it would work. What do you think?

IrinaYatsenko:
@dalyd Thank you for reviewing!
1. I don't remember all of the rationales for choosing the approach of putting non-CSI and CSI experiments into the same test but we do this in other CSI-related workloads as well and have a way to compare the results (@starzia would have more context on this)
2. At least one of the reasons to go this route was the huge size of the dataset. If I'm reading the logs correctly, it takes 15min to set it up. Initially we wanted to run even more separate experiments against the same dataset so setting it up once seemed prudent.
3. The fairness, as far as WT state is concerned, is a valid point. We believe it doesn't matter here because inserts into the collection's rowstore should all be localized and we don't reuse indexes across the experiments so their internal state isn't a problem. We also only care about Latency50Percentile measurement, which should make the rowstore part of the insert to be even less sensitive to a potential internal state priming.
4. Personally, I like having the logic of dropping/creating the indexes as part of the genny's workload rather than in DSI as it makes the intent of the experiments more clear.

All in all, we'd like to keep this workload as is. Do you think that would be acceptable?

dalyd:
> @dalyd Thank you for reviewing!
> 
> 1. I don't remember all of the rationales for choosing the approach of putting non-CSI and CSI experiments into the same test but we do this in other CSI-related workloads as well and have a way to compare the results (@starzia would have more context on this)
> 2. At least one of the reasons to go this route was the huge size of the dataset. If I'm reading the logs correctly, it takes 15min to set it up. Initially we wanted to run even more separate experiments against the same dataset so setting it up once seemed prudent.
> 3. The fairness, as far as WT state is concerned, is a valid point. We believe it doesn't matter here because inserts into the collection's rowstore should all be localized and we don't reuse indexes across the experiments so their internal state isn't a problem. We also only care about Latency50Percentile measurement, which should make the rowstore part of the insert to be even less sensitive to a potential internal state priming.
> 4. Personally, I like having the logic of dropping/creating the indexes as part of the genny's workload rather than in DSI as it makes the intent of the experiments more clear.
> 
> All in all, we'd like to keep this workload as is. Do you think that would be acceptable?

Okay, that sounds fine then. Could you add some of points 2 and 3 to the comments (huge dataset setup time, etc)? Otherwise LGTM.

dalyd (approved):

dalyd (approved):

PR #808, opened 2023-01-16
Add compare operators support for AutoRun When 

Fefer-Ivan:


rtimmons (approved):
This is great. Thanks for adding comprehensive testing and updating the existing workloads to take advantage of it.

Please keep an eye on the sys-perf waterfall (and for the last couple branch versions too) for a couple days after this merges to ensure nothing goes purple. Lemme know if you need a hand with this part.

PR #809, opened 2023-01-18
PERF-3494 Consolidate MultiDelete workloads

haleyConnelly:


haleyConnelly:
Each workload runs with mass deletions on a set of 1KB documents then 10KB documents. 

Reduced to 4 workloads:
- 1 workload batched deletes, no secondary indexes
- 1 workload doc by doc deletes, no secondary indexes
- 1 workload batched deletes with secondary indexes
- 1 workload doc by doc deletes with secondary indexes

**alternative**
Have documents of varying sizes in the same collection, rather than running multi deletes on 2 distinct set of document sizes. 

My thinking was that having distinct sets of document sizes for each delete upholds the nature of the existing tests without having 16 files.

Once we decide what route to go, I'll file another ticket and add TODOs for deleting the existing workloads in about a week or so.

src/phases/execution/MixedMultiDeletes.yml
```diff
@@ -0,0 +1,304 @@
...
+          Database: *Database
+          Collection: *MassDeletionsColl
+          Operations:
+          - OperationName: bulkWrite
+            OperationCommand:
+              WriteOperations:
+              - WriteCommand: deleteMany
+                Filter:
+                  *DeleteQuery
+                OperationOptions:
```
haleyConnelly, line 147: This can be removed now that the 2022-11 variants are majoiryt by default

src/phases/execution/MixedMultiDeletes.yml
```diff
@@ -0,0 +1,304 @@
...
+  LoaderThreads: &LoaderThreads 4
+  # Delete by primary index always.
+  DeleteQuery: &DeleteQuery
+    {_id: {$gte: {^ObjectId: "000000000000000000000000"}}}
+  # Only used for generating the Document padding.
+  RandomStringLength: &RandomStringLength 6
+
+  #### Document Size Specifications ####
+
+  # __________  1KB Specifications __________
```
haleyConnelly, line 52: Taken from parameters in [MultiDeletes<>_1KB.yml](https://github.com/mongodb/genny/blob/6d4f0888d97bf5a753460db13e5974cfffaeae17/src/workloads/execution/MultiDeletesBatched_1KB.yml#L25-L30)

src/phases/execution/MixedMultiDeletes.yml
```diff
@@ -0,0 +1,304 @@
...
+    b: *rs_1KB
+    c: *rs_1KB
+    d: *rs_1KB
+    e: *rs_1KB
+    f: *rs_1KB
+    g: *rs_1KB  # up to field_count fields
+    padding: {^FastRandomString: {length: *DocumentPaddingSizeBytes_1KB}}
+
+  # __________  10MB Specifcations __________
+  TargetDocumentSizeBytes_10MB: &TargetDocumentSizeBytes_10MB 10_485_760  # 10MB
```
haleyConnelly, line 72: Parameters taken from [MultiDeletes<>_10MB.yml](https://github.com/mongodb/genny/blob/6d4f0888d97bf5a753460db13e5974cfffaeae17/src/workloads/execution/MultiDeletesBatched_10MB.yml#L27-L31)

haleyConnelly (commented):
**For the first pass, I'm looking for opinions on:**
(1) Think the amount of work done in each workload (2 rounds of 2 mass deletes, each round with specifically sized documents)
(2) Using 1KB then 10MB as document sizes

If the we think (1) and (2) are reasonable, then I'll put up another patch with the TODOs that say the existing files can be removed in a new PERF ticket after a week or so 
** I wanted to hold off on this until we decide on a direction, since some of the configuration files may need to stay vs go pending the implementation

henrikedin (commented):
Workloads look good but a bit difficult to understand what changes you've made. Is the idea to run these new ones in parallel with the old ones and then remove the old workloads after a while?

I see the 1KB and 10MB batching. We removed the others? Can you give a high-level overview of what is different? I think it would also be great if @ykangpark could give this a pass.

src/phases/execution/MixedMultiDeletes.yml
```diff
@@ -0,0 +1,304 @@
...
+  DocumentCount_10MB: &DocumentCount_10MB 4_000
+  InsertBatchSize_10MB: &InsertBatchSize_10MB 1
+  DocumentPaddingSizeBytes_10MB: &DocumentPaddingSizeBytes_10MB
+    ^NumExpr:
+      withExpression: "target_size - id_length - (field_count * (field_length+value_length)) - padding_field"
+      andValues: {target_size: *TargetDocumentSizeBytes_10MB, id_length: 22, field_count: 7, field_length: 8, value_length: *RandomStringLength, padding_field: 14}
+  # Insert same document for both default and with secondary indexes to remove one variable.
+  DocumentToInsert_10MB: &DocumentToInsert_10MB
+    a: &rs_10MB {^FastRandomString: {length: *RandomStringLength, alphabet: "0123456789"}}
```
ykangpark, line 81: `rs_1KB` is the same as `rs_10MB`, would it make sense to use the same definition?

src/workloads/execution/MixedMultiDeletesBatched.yml
```diff
@@ -0,0 +1,37 @@
...
+- deleteMany
+- latency
+
+Clients:
+  Default:
+    QueryOptions:
+      # Allow for longer duration since the loading of the data may take a while.
+      socketTimeoutMS: 3_600_000  # = 1 hour
+
+LoadConfig:
```
ykangpark, line 23: Would it be possible to parametrise the `MixedMultiDeletes` file, and do a `LoadConfig` for each document size?
I don't remember exactly, but I think it might be possible.

https://github.com/mongodb/genny/blob/91716090002bc1a76155c350bdf9fd17531f97d9/src/lamplib/src/genny/tasks/preprocess.py#L235

ykangpark (commented):
Looking good, left some suggestions regarding YML format. Wondering, why are we removing the 16KB and 2MB variants? I added them at Josef's request, I believe he wanted to be able to run them on demand if necessary, which is why those were not on autorun. Will that no longer be necessary?

haleyConnelly:
> Workloads look good but a bit difficult to understand what changes you've made. Is the idea to run these new ones in parallel with the old ones and then remove the old workloads after a while?
> 
> I see the 1KB and 10MB batching. We removed the others? Can you give a high-level overview of what is different? I think it would also be great if @ykangpark could give this a pass.

My original intent was to reduce the number of documents testing batched multi deletes and reduce the expense of performance testing multi deletes.

I removed the others after wondering: How much value does each variation give? Is it worth the expense? 

Going to respond to Yujin's comments then circle back

haleyConnelly:
> why are we removing the 16KB and 2MB variants

I wasn't sure if it was acceptable to have such a large intensive workload and wondered how much we gained from testing all 4. I can re-add them

ykangpark:
Seems like I messed up in PERF-2901, and in the last commit I uncommented the autorun for 16KB and 2MB. Sorry for the confusion. Those should have been disabled, and were intended to be run only on demand.

Commit where I was removing them from autorun:
https://github.com/mongodb/genny/pull/660/commits/c070914a8052aa02b411205f8c65ebc44b9b40af

And then I mistakenly pushed uncommented autorun:
https://github.com/mongodb/genny/pull/660/commits/adf21fea8afa7732e2630957c73d9ba2384b13fd

Might still be worth to keep them for on-demand runs.

haleyConnelly:
> Seems like I messed up in PERF-2901, and in the last commit I uncommented the autorun for 16KB and 2MB. Sorry for the confusion. Those should have been disabled, and were intended to be run only on demand.

Commit where I was removing them from autorun:
https://github.com/mongodb/genny/commit/c070914a8052aa02b411205f8c65ebc44b9b40af

And then I mistakenly pushed uncommented autorun:
https://github.com/mongodb/genny/commit/adf21fea8afa7732e2630957c73d9ba2384b13fd

Might still be worth to keep them for on-demand runs.

Any suggestions on how to do that without leaving the legacy files laying around? The goal of this is to reduce the number of files targeting these performance workloads

ykangpark:
> Any suggestions on how to do that without leaving the legacy files laying around? The goal of this is to reduce the number of files targeting these performance workloads

Well, we are trying to reduce task count. We are already reducing by half. So maybe it would be fine to have an alternate set with 16KB and 2MB? As in `MixedMultiDeletesBatched.yml` and `MixedMultiDeletesBatchedExtra.yml`, etc.

Another idea to keep them with the new files, maybe we could have 16KB and 2MB actors defined in a separate file. And allow that file to be passed as a parameter to `MixedMultiDeletes.yml`. In there, a LoadConfig would load the additional actor definitions at the end of the "Actors:" field. By default it would load an empty file. If a developer wants to run the extra file sizes, it is only necessary to uncomment to parameter.

I don't have the context to say whether we actually need 16KB and 2MB at this point. If it is unnecessary, I don't mind removing them.

henrikedin:
> 

It is unclear how much value the additional 16KB and 2MB workloads give. One small and one large would provide adequate coverage to me. 

If we can leave them in as on-demand workloads without creating a mess in these yaml files I would be totally okay with that. But I think we should try to share as much of the definition as possible then so it doesn't become stale and they don't work the day someone tries them out.

haleyConnelly:
Sounds like it isn't completely necessary to test the alternate sizes. I'm not convinced we will get much benefit from having a separate file for the alternative sizes that isn't going to be run automatically. If someone wants to test delete behavior on the alternate sizes, they can run a patch with the sizes changed in the existing workload. 

We already will be going 1/2 files over our target 4 files as I mentioned above in my previous comment. 

I'd like to move forward with the decision to completely remove the alternate size performance coverage.

src/workloads/execution/MixedMultiDeletesDocByDocWithSecondaryIndexes.yml
```diff
@@ -0,0 +1,40 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: >
+  Deletes a range of documents using the DELETE query exec stage both in isolation and while
```
ykangpark, line 4: Would it be useful to describe that this is with secondary indexes?

ykangpark: And maybe why it matters to test with secondary indexes at all.

src/workloads/execution/MixedMultiDeletesBatched.yml
```diff
@@ -0,0 +1,37 @@
...
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - replica-all-feature-flags
+      - replica
+    branch_name:
+      $neq:
+      # 6.0 Does not have the 'batchUserMultiDeletes' server parameter.
+      - v6.0
```
ykangpark, line 37: Hey, thanks for fixing this! One note. Not sure why we didn't get any BFs for other branches, but wouldn't this cause these workloads to still run for branches under 6.0? Such as 5.0, 4.4, 4.2?

ykangpark (approved):
LGTM! Thanks for the cleanup! Only a doubt regarding the `branch_name` filter.

src/workloads/execution/MixedMultiDeletesBatched.yml
haleyConnelly: I believe it will only run our most recent major release, but I could be wrong.

src/workloads/execution/MixedMultiDeletesDocByDocWithSecondaryIndexes.yml
haleyConnelly: Done! albeit maybe a bit wordy

henrikedin (approved):
Thanks for taking this on Haley!

src/workloads/execution/MixedMultiDeletesBatched.yml
ykangpark: Not sure either, but the docs and the other workloads seem to be excluding every branch explicitly.

https://github.com/mongodb/genny/blob/c1536caf49eeddf49d31d0307e7f8a9cecf5034f/docs/using.md#L553

haleyConnelly: Hm, okay, I'll modify it for down to 4.2 since we no longer support 4.0

haleyConnelly: Thanks for being so attentive to detail!

ykangpark (approved):

PR #812, opened 2023-01-20
PERF-3744 fix cursor leak in locf.yml

dstorch:
This was caught in [BF-27122](https://jira.mongodb.org/browse/BF-27122) which has more details on how I diagnosed the bug. I wrote a descirption for PERF-3744 which has more context as well.

Among the fixes I considered, I think adding a limit is the simplest way to do it while keeping the workload logically the same. The queries could run considerably more slowly if we instead starting using the `CrudActor` to exhaust the cursor. And I didn't want to go the more complex route of teaching genny how to partially read from a cursor and then kill it.

### DIFF
--- a/src/workloads/query/locf.yml
+++ b/src/workloads/query/locf.yml
@@ -12,6 +12,11 @@ Description: |
   To learn more about partitions, please check out the docs here:
   https://docs.mongodb.com/manual/reference/operator/aggregation/setWindowFields/
 
+GlobalDefaults:
+  # The limit ensures that the entire result set fits in an entire batch. That way we can use
+  # the 'RunCommand' actor without abandoning open cursors.
+  Limit: &limit 100
+
 Actors:
 - Name: InsertData
   Type: Loader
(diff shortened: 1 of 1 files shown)


dstorch:
Kicked off a sys-perf patch build to make sure that the workload still runs as expected on Evergreen: https://spruce.mongodb.com/version/63cabe7f9ccd4e5f89a38835/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

madelinezec:
I'm really glad you were able to diagnose the issue. Can you explain to me why using $limit enforces destroying cursors?

dstorch:
Sure! The server will cleanup the cursor when it reaches EOF. If the query logically has more results, then the `ClientCursor` is left alive on the server waiting for the client to execute another "getMore" operation. (The cursor or its session may eventually time out due to inactivity, so the cursor "leak" is not a true memory leak.) As originally written, the workload ran an initial aggregation command with a batchSize of 100. The result was that the client obtained the first 100 results, but the server would leave its cursor open waiting for a potential getMore that would never arrive. If instead of specifying a batchSize of 100 you specify a _limit_ of 100, then the query logically should only return 100 documents in total. Therefore, the server will return the 100 documents in the initial batch and clean up the query executor without leaving a cursor open. Make sense?

madelinezec (approved):
That makes sense! Thank you for the explanation (so glad we did it on github PR and not slack so I know I can always find it) LGTM

dstorch:
Cool! Thanks for the prompt review Maddie!

PR #815, opened 2023-01-25
PERF-3773 Use CrudActor Find instead of RunCommand to avoid leaking cursors

Fefer-Ivan:


mtandrei (approved):
These changes look LGTM; thanks Ivan! A couple of notes before you merge:
- Please verify on a patch build that the CrudActor indeed closes the cursors
- One thing that might be worth following up on is that using the CrudActor instead may change the test name in performance results. If this is the case, then the[ tags file may need to be update to avoid discontinuity in the dashboard](https://github.com/10gen/PerformanceReports/blob/main/scripts/tags.yml)

PR #817, opened 2023-01-31
Fix Projection spec for ExpressiveQueries workload

Fefer-Ivan:


SimonEismann:
PR looks good; some housekeeping:
* Could you add a small description to the PR to document the rationale behind the changes?
* If this change is associated with a Jira ticket, could you add the shorthand label to the PR title?
* It's good to have a green patch associated with every PR, no matter how small. I've scheduled a [small patch](https://spruce.mongodb.com/version/63d902fed6d80a136f93bfff)

rtimmons (approved):

SimonEismann:
Patch looks good:

https://spruce.mongodb.com/version/63d902fed6d80a136f93bfff

There seem to be some performance improvements:
image

I guess the performance changes are expected since the projection reduces the amount of data retrieved. In case that's true, we should let the build barons know that we expect this change to cause some changepoints :)

SimonEismann (commented):
LGTM

PR #818, opened 2023-01-31
PERF-3788 Use multi-threading to speed up Loader in ExpressiveQueries workload

Fefer-Ivan:
There is a TODO to use multithreading to load documents faster in ExpressiveQueries workload. This now can be fixed using MultipleThreadsPerCollection option.

### DIFF
--- a/src/workloads/query/ExpressiveQueries.yml
+++ b/src/workloads/query/ExpressiveQueries.yml
@@ -168,7 +168,8 @@ Actors:
       OperationName: find
       OperationCommand:
         Filter: *filter
-        Projection: &projection {_id: 0, firstname: 1, lastname: 1, dob: 1}
+        Options:
+          Projection: &projection {_id: 0, firstname: 1, lastname: 1, dob: 1}
   - *Nop
   - *Nop
   - Repeat: 1000  # This query is very fast, so we need to run it more times than others.
@@ -180,7 +181,8 @@ Actors:
       OperationName: find
       OperationCommand:
         Filter: *filter
-        Projection: *projection
+        Options:
+          Projection: *projection
 
 AutoRun:
 - When:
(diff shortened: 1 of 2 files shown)


src/workloads/query/ExpressiveQueries.yml
```diff
@@ -19,15 +19,13 @@ Actors:
...
-  # TODO: TIG-2938 Use multiple threads to speedup insertion.
-  Threads: 1
+  Threads: 4
   Phases:
   - Repeat: 1
     Database: &DB test
-    Threads: 1
+    MultipleThreadsPerCollection: true
     CollectionCount: 1  # Collection name will be Collection0, this is not configurable.
-    DocumentCount: 1e6
```
mtandrei, line 30: [optional nit] I would revert this change as it's easier to read '6' than to count six zeros.

mtandrei (dismissed):
LGTM; thanks Ivan!

src/workloads/query/ExpressiveQueries.yml
Fefer-Ivan: I agree, but for when MultipleThreadsPerCollection is true, yaml parsing is forcing this number to be an integer because it is doing integer division of DocumentCount / ThreadCount. 

We can address it in C++ code, but to keep the PR smaller, I decided to just change it to integer.

Fefer-Ivan: The fix for it is actually pretty easy, but I would like to do it in a separate PR

Fefer-Ivan:
https://spruce.mongodb.com/version/63d948bd3e8e86193aeae1fa/tasks

src/workloads/query/ExpressiveQueries.yml
mtandrei: Sounds good; thanks for addressing this!

mtandrei (approved):
Still LGTM; thanks Ivan!

PR #819, opened 2023-01-31
Use IntergerSpec for DocumentCount across Loader actor

Fefer-Ivan:
This allows us to use 1eN notation instead of 10...00 which is difficult to read of the number of zeros is high


allenwux (dismissed):

jimoleary (commented):
Thanks for the change, it looks a good update.

Can we update the collection scanner example so that it is clear that this can be used? That is. update [this line](https://github.com/mongodb/genny/blob/3f58f08b9147dc0052542240e380b2f462fce5e9/src/workloads/docs/CollectionScanner.yml#L24) to the following:

```
    DocumentCount: 2e5 # Same as above but easier to read.
```

jimoleary (approved):
thanks, lgtm!

PR #821, opened 2023-01-31
EVG-18844: Properly handle "!!str" in DocumentGenerator.

johndaniels:
We should check for an actual "!!str" value tag. Based on testing I don't think the value of the tag is ever "!".


rtimmons (approved):

johndaniels:
evergreen merge

PR #824, opened 2023-02-06
PERF-3807 Exclude graviton-single-lite.2022-11 from running BackgroundIndexConstruction

jordiolivares:


haleyConnelly (dismissed):
LGTM % if. you add a comment saying it causes us to run out of memory

haleyConnelly (approved):

PR #829, opened 2023-02-16
PERF-3657 Add sinusoidal read/write workload for execution control

pvselvan:
The patch runs successfully on evergreen, but I'm still working on manually obtaining the intra-run data so we can see the distribution.

In the meantime, I wanted to put this patch up for review to see if we agree on the format.

### DIFF
--- a/src/workloads/execution/SinusoidalReadWrites.yml
+++ b/src/workloads/execution/SinusoidalReadWrites.yml
@@ -42,7 +42,7 @@ ActorTemplates:
         NopInPhasesUpTo: *NumPhases
         PhaseConfig:
           Duration: 3 minutes
-          SleepAfter: 1.5 minutes
+          SleepAfter: 90 seconds
           Repeat: 3
           CollectionCount: *NumColls
           Operations:
@@ -62,7 +62,7 @@ ActorTemplates:
         Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
         NopInPhasesUpTo: *NumPhases
         PhaseConfig:
-          SleepBefore: 1.5 minutes
+          SleepBefore: 90 seconds
           Duration: 3 minutes
           Repeat: 3
           CollectionCount: *NumColls
@@ -120,7 +120,3 @@ AutoRun:
       - v5.1
       - v5.2
       - v5.3
-      - v6.0
-      - v6.1
-      - v6.2
-      - v6.3
--- a/src/workloads/execution/SinusoidalReadWrites.yml
+++ b/src/workloads/execution/SinusoidalReadWrites.yml
@@ -28,7 +28,7 @@ Document: &doc
 Clients:
   Default:
     QueryOptions:
-      maxPoolSize: 1000
+      maxPoolSize: 1001
 
 ActorTemplates:
 - TemplateName: WritesActorWithSleepTemplate
--- a/src/workloads/execution/SinusoidalReadWrites.yml
+++ b/src/workloads/execution/SinusoidalReadWrites.yml
@@ -4,7 +4,7 @@ Description: |
   This workload attempts to create a sinusoidal distribution of reads/writes to test
   the behavior of execution control when the load on the system changes over time.
 
-  We do this with two actors, one for writes and one for reads and spawn 5000 client
+  We do this with two actors, one for writes and one for reads and spawn 500 client
   threads. Though the actors run concurrently, we use the sleep functionality
   to effectively stop work on one actor while the other reaches its peak. We then
   alternate for a couple repetitions.
@@ -28,7 +28,7 @@ Document: &doc
 Clients:
   Default:
     QueryOptions:
-      maxPoolSize: 5100
+      maxPoolSize: 55
 
 ActorTemplates:
 - TemplateName: WritesActorWithSleepTemplate
@@ -36,7 +36,7 @@ ActorTemplates:
     Name: {^Parameter: {Name: "Name", Default: WritesActorWithSleep}}
     Type: CrudActor
     Database: *dbname
-    Threads: {^Parameter: {Name: "Threads", Default: 1000}}
+    Threads: {^Parameter: {Name: "Threads", Default: 100}}
     Phases:
       OnlyActiveInPhases:
         Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
@@ -57,7 +57,7 @@ ActorTemplates:
     Name: {^Parameter: {Name: "Name", Default: ReadsActorWithSleep}}
     Type: CrudActor
     Database: *dbname
-    Threads: {^Parameter: {Name: "Threads", Default: 1000}}
+    Threads: {^Parameter: {Name: "Threads", Default: 100}}
     Phases:
       OnlyActiveInPhases:
         Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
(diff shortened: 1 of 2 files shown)


src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,123 @@
...
+    Type: CrudActor
+    Database: *dbname
+    Threads: {^Parameter: {Name: "Threads", Default: 1000}}
+    Phases:
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Duration: 3 minutes
+          SleepAfter: 90 seconds
```
pvselvan, line 46: Once we get confirmation via the intra-run data that the sleep does what we want, I think we can parameterize this option. This way, we can reuse this actor for subsequent workloads.

We can do this for SleepAfter, SleepBefore, and Repeat.

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,123 @@
...
+      - standalone
+      - standalone-all-feature-flags
+    branch_name:
+      $neq:
+      - v4.2
+      - v4.4
+      - v5.0
+      - v5.1
+      - v5.2
+      - v5.3
```
gregorynoma, line 123: Is there a reason we want to skip this on these older branches?

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,123 @@
...
+      Name: {^Parameter: {Name: "ReadingName", Default: ReadsActor_50_50}}
+      Threads: {^Parameter: {Name: "ReadingThreads", Default: 2500}}
+      ActivePhases: [1]
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
```
gregorynoma, line 115: Will we be running this on more types of configurations later?

gregorynoma (commented):
About how long does this workload take to run in total?

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: I confined this to standalones for now since we discussed how flow control might be an issue for creating the right overload circumstances. However, I think we can file a ticket to experiment with more configurations once we get all the workloads running.

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Not really, do you think it's worth running on older branches?

pvselvan:
> 

Currently, this task takes about 38 minutes to run.

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,123 @@
...
+
+  We do this with two actors, one for writes and one for reads and spawn 5000 client
+  threads. Though the actors run concurrently, we use the sleep functionality
+  to effectively stop work on one actor while the other reaches its peak. We then
+  alternate for a couple repetitions.
+
+  In this workload, we try to ensure that reads and writes do not overlap much.
+Keywords:
+- sinusoidal
+- execution control
```
gregorynoma, line 15: Is it ok that this keyword is actually two words? Does it handle that?

src/workloads/execution/SinusoidalReadWrites.yml
gregorynoma: Is there a precedent for what we've done for other workloads we've added somewhat recently? Since we won't be working on execution control on the older versions but these workloads still generally apply there

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,123 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: |
+  This workload attempts to create a sinusoidal distribution of reads/writes to test
+  the behavior of execution control when the load on the system changes over time.
+
+  We do this with two actors, one for writes and one for reads and spawn 5000 client
```
gregorynoma, line 7: This comment says 5000 threads but below it looks like the reader and writer each spawn 1000 threads?

src/workloads/execution/SinusoidalReadWrites.yml
gregorynoma: I feel like this workload would make sense on a 3-node replica set as well (and maybe even 1-node)? Even if it ends up not being super useful for execution control for the time being, that seems like an issue with tuning flow control more than anything else

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Yep, it can handle multiple spaces!

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: The 1000 is a default value if no value for "Threads" has been passed in. We pass in 2500 writer threads and 2500 reader threads when creating the write and read actors from this template.

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Sure, I've added the corresponding tags to let this run on replica sets.

src/workloads/execution/SinusoidalReadWrites.yml
gregorynoma: Ah I see, do we think those numbers might be too high or is that in line with other workloads? I just want to make sure we're not bottlenecked by the server's ability to handle a large number of connections

pvselvan: Changed this to 500 reader/writer threads and lowered the connection pool to 50.

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,117 @@
...
+CollectionCount: &NumColls 1
+
+NumPhases: &NumPhases 2
+Document: &doc
+  a: {^RandomInt: {min: 0, max: *NumDocs}}
+  b: {^RandomString: {length: 16}}
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 55
```
gregorynoma, line 31: Is this intentionally 55 as opposed to 50?

gregorynoma (dismissed):

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Yes, the phase duration is a little imperfect right now because it's achieved with sleeps. They may not be perfectly opposing sin waves, because it's possible that we simultaneously do a bit more than the 500 reads or writes. To account for this, I just increased the conn pool size a bit by a factor of 50.

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,117 @@
...
```
gregorynoma, line 31:
```suggestion
      maxPoolSize: 550
```

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -0,0 +1,117 @@
...
+CollectionCount: &NumColls 1
+
+NumPhases: &NumPhases 2
+Document: &doc
+  a: {^RandomInt: {min: 0, max: *NumDocs}}
+  b: {^RandomString: {length: 16}}
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 1001
```
gregorynoma, line 31: Our of curiosity, did you figure out why a connection pool size is even useful at all, if it has to be greater than the max number of connections?

pvselvan:
evergreen merge

PR #832, opened 2023-02-23
EVG-19030 Propagate FrequencyMapSingletonGenerator's DefaultRandom

spencerjackson:
I observed segfaults in Genny while running tests in evergreen which used `TakeRandomStringFromFrequencyMapSingleton`. After running Genny locally with tsan, I observed that data races were occurring between FrequencyMapGenerator's use of RNG, and other generators' RNG use.

Digging into `FrequencyMapSingletonGenerator`, I noticed that in multithreaded environments, the first thread to make their `FrequencyMapSingletonGenerator` would construct the underlying `FrequencyMapGenerator` in the global table. All uses of this `FrequencyMapGenerator` across all actors would be protected by a mutex. Unfortunately, the `FrequencyMapGenerator` was constructed using the first thread's `GeneratorArgs`, and so would capture a reference to its `DefaultRandom`. `DefaultRandom` is intended to belong to a single thread, and may not be used concurrently. While another actor used the `FrequencyMapGenerator`, it could be concurrently using the borrowed `DefaultRandom` if the initiating thread was using it using a different Generator. The mutex would not prevent the initiating thread's use of its own `DefaultRandom` in other contexts.

Evgn: https://spruce.mongodb.com/version/63f67e8e3066154590204ca6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
Before: https://spruce.mongodb.com/version/63f4e5787742ae64ad5ccf18/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
After: https://spruce.mongodb.com/version/63f67d42d1fe076217d80285/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -622,15 +622,19 @@ public:
         }
     }
 
-    std::string evaluate() override {
+    std::string evaluateWithRng(DefaultRandom& rng) {
         if (_map.size() == 1) {
             return _map.take(0);
         }
 
         // Pick a random number between 0 and _map.size() inclusive
         auto distribution = boost::random::uniform_int_distribution<size_t>(0, _map.size() - 1);
-        auto value = distribution(_rng);
+        auto value = distribution(rng);
         return _map.take(value);
+    }
+
+    std::string evaluate() override {
+      return evaluateWithRng(_rng);
     };
 
 private:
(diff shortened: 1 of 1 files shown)


markbenvenuto (approved):

rtimmons (approved):
Good catch! I wish we had a linter that could help with this since it's an easy foot-gun!

PR #834, opened 2023-02-27
PERF-3401 Add workload to test multi planner performance

Fefer-Ivan:


src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/server-execution"
+Description: >
+ Create collection with multiple indexes and run queries with different selectivity on different
+ indexes to test how efficiently multi planning can fix most selective index.
```
dstorch, line 5:
```suggestion
 indexes to test how efficiently multi planning can choose the most selective index.
```

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
+SchemaVersion: 2018-07-01
```
dstorch, line 1: Just a heads up: once this is done, it's worth incorporating into the SBE perf dashboard. It's a slightly annoying process right now, unfortunately. You have to add the task name here: https://github.com/10gen/PerformanceReports/blob/main/scripts/perf_reports_sbe_perf.yml

Then once the measurements show up as untagged in the dashboard, you have to add the necessary tags to this file: https://github.com/10gen/PerformanceReports/blob/main/scripts/tags.yml

Maybe one day if we something like our PerformanceReports system becomes a more official tool owned by dev prod / Evergreen there can be some work done to make this whole process of incorporating a new benchmark into the dashboard less painful.

@mtandrei I assume you would agree that these benchmarks should be incorporated into the dashboard?

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        planCacheClear: *Collection
+  - Blocking: None
+    Database: *Database
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        planCacheClear: *Collection
```
dstorch, line 70: Is there a way to factor out this boilerplate so that it doesn't have to be repeated three times?

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+        - key:
+            c: 1
+          name: c_1
+  - &Nop {Nop: true}
+  - Blocking: None
+    Database: *Database
+    Operations:
+    - OperationName: RunCommand
+      OperationCommand:
+        planCacheClear: *Collection
```
dstorch, line 64: It would be good to leave a comment here about why we're repeatedly dropping the plan cache. Not sure if it would be obvious or not to future readers, but it seems helpful to state explicitly that we are frequently dropping the plan cache in order to make multi-planning more likely.

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *docCount
+        BatchSize: 1000
+        Document:
+          a: *Key
+          b: *Key
+          c: *Key
+
+- Name: Lookup
```
dstorch, line 97: I think this could be confused with the $lookup MQL feature. Recommend renaming it something like `MultiPlanning`.

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+  Type: CrudActor
+  Database: *Database
+  Threads: 4
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 2000
```
dstorch, line 104: Just curious: How did you choose this constant?

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - replica
+      - replica-all-feature-flags
+    branch_name:
+      $gte: v7.0
```
dstorch, line 147: Why shouldn't we run this against older branches?

src/workloads/execution/MultiPlanning.yml
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - replica
+      - replica-all-feature-flags
```
dstorch, line 145: I think this list is wrong. It seems sufficient to run this test just against our standalone variants, not against our 1-node replica set variants. However, we definitely want this to run against the classic and SBE specific build variants. I believe you can use the same list as the recently added $group spill-to-disk tests:

https://github.com/mongodb/genny/blob/9ee62c040a2bc7609d566040f012140daf8d81b1/src/workloads/query/GroupSpillToDisk.yml#L132-L135

dstorch (changes requested):
Not sure if you shared this elsewhere, but do you have numbers for these benchmarks from Evergreen for both classic and SBE? Assuming that the approach which repeatedly flushes the cache in the background shows a clear advantage for query latency on SBE, then I'd be in favor of the proposed approach. I definitely like avoiding making the parameter which disables plan caching a runtime parameter for now.

@mtandrei I'm curious for your thoughts on the approach of making the workload keep clearing the plan cache in the background.

I also left a few very minor suggestions on the particulars.

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: I will do it after the merge.

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: Done

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: Done

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: Done

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: I was experimenting and looking for a constant that will produce stable result and will run fast enough, so I can iterate on solution.

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: Initially I added it because it used internalQueryDisablePlanCache parameter. Now I can remove it.

src/workloads/execution/MultiPlanning.yml
Fefer-Ivan: Done

src/workloads/execution/MultiPlanning.yml
dstorch: Sounds good. Leaving this comment thread open now in case Mihai wants to jump in.

src/workloads/execution/MultiPlanning.yml
dstorch: Sounds good! Resolving.

dstorch (approved):
LGTM

PR #836, opened 2023-03-03
PERF-3726 external phase coordinator.

jimoleary:
**What:**
The genny portion of the phase event handling. When run with './run-genny' directly, the named pipes are not created so genny does not emit any events. When run under DSI, the named pipes are available so genny writes phase to the out named pipe and waits for an ack from dsi before continuing.

The code addes start and end phase events. These could be trimmed to just a single event type.

The phase number was added as a parameter to the callback because there is already a write lock taken out, so it is not possible to read the phase number through the accessor. Adding the number to the start and stop seems reasonable to me.

**Patches:** 
[updated sys perf patch](https://spruce.mongodb.com/version/64049adb5623436fe78fec71/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
~~[sys perf](https://spruce.mongodb.com/version/6402097e0ae6061e081c36be/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).~~

**Other PRs:**
[DSI](https://github.com/10gen/dsi/pull/1227)

### DIFF
--- a/src/gennylib/src/Orchestrator.cpp
+++ b/src/gennylib/src/Orchestrator.cpp
@@ -83,7 +83,7 @@ PhaseNumber Orchestrator::awaitPhaseStart(bool block, int addTokens) {
 
     const auto currentPhase = this->_current;
     if (_currentTokens >= _requireTokens) {
-        for (auto&& cb : _prePhaseHooks) {
+        for (auto&& cb : _prePhaseStartHooks) {
             cb(this, currentPhase);
         }
         BOOST_LOG_TRIVIAL(info) << "Beginning phase " << currentPhase;
@@ -136,12 +136,13 @@ bool Orchestrator::awaitPhaseEnd(bool block, int removeTokens) {
     // compare with >= rather than ==.
 
     if (_currentTokens <= 0) {
-        const auto currentPhase = _current;
-        ++_current;
-        BOOST_LOG_TRIVIAL(info) << "Ended phase " << currentPhase;
-        for (auto&& cb : _postPhaseHooks) {
-            cb(this, currentPhase);
+        // Fire the phase stop callbacks before updating the current phase counter to indicate that
+        // the current phase is complete.
+        BOOST_LOG_TRIVIAL(info) << "Ended phase " << _current;
+        for (auto&& cb : _postPhaseStopHooks) {
+            cb(this, _current);
         }
+        ++_current;
         _phaseChange.notify_all();
         state = State::PhaseEnded;
     } else {
@@ -157,12 +158,12 @@ bool Orchestrator::awaitPhaseEnd(bool block, int removeTokens) {
 
 void Orchestrator::addPrePhaseStartHook(const OrchestratorCB& f) {
     writer lock{_mutex};
-    _prePhaseHooks.push_back(f);
+    _prePhaseStartHooks.push_back(f);
 }
 
 void Orchestrator::addPostPhaseStopHook(const OrchestratorCB& f) {
     writer lock{_mutex};
-    _postPhaseHooks.push_back(f);
+    _postPhaseStopHooks.push_back(f);
 }
 
 void Orchestrator::abort() {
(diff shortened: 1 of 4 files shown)


src/gennylib/src/Orchestrator.cpp
```diff
@@ -136,8 +136,12 @@ bool Orchestrator::awaitPhaseEnd(bool block, int removeTokens) {
     // compare with >= rather than ==.
 
     if (_currentTokens <= 0) {
+        const auto currentPhase = _current;
         ++_current;
-        BOOST_LOG_TRIVIAL(info) << "Ended phase " << (this->_current - 1);
+        BOOST_LOG_TRIVIAL(info) << "Ended phase " << currentPhase;
```
dalyd, line 141: This log message is going to change by 1. Is that intentional?

src/gennylib/src/Orchestrator.cpp
```diff
@@ -136,8 +136,12 @@ bool Orchestrator::awaitPhaseEnd(bool block, int removeTokens) {
...
+        for (auto&& cb : _postPhaseHooks) {
+            cb(this, currentPhase);
```
dalyd, line 143: Do you want the next phase number for the call back, or the current? could you comment that somewhere?

src/gennylib/src/context.cpp
```diff
@@ -205,4 +214,55 @@ bool PhaseContext::isNop() const {
...
+}
+void ExternalPhaseCoordinator::onPhaseStart(PhaseNumber phase){
+    std::stringstream event;
+    event << R"({ "message": ")" << "Beginning phase" << R"(","phase":)"
+          <<  phase << R"(,"request":)" <<  1 << "}" << std::endl;
+    _onPhase(event.str());
+}
+void ExternalPhaseCoordinator::onPhaseStop(PhaseNumber phase){
+    std::stringstream event;
+    event << R"({ "message": ")" << "Ended phase" << R"(","phase":)"
```
dalyd, line 263: I think you are going to be off 1 on the ended phase based on the above.

src/gennylib/src/context.cpp
```diff
@@ -205,4 +214,55 @@ bool PhaseContext::isNop() const {
...
+    if(connect(sd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
+        BOOST_LOG_TRIVIAL(warning) << "connect('" << ipaddress << "',"
+                                   << port << "') failed --> '"<< ::strerror(errno) << "'";
+        return -1;
+    }
+    return sd;
+}
+
+ExternalPhaseCoordinator::ExternalPhaseCoordinator(std::string host, int port) :
+      m_host(host), m_port(port), m_socket(connect(host, port)){
```
dalyd, line 237: Not to self, this is where the connect function is called and the ip connection is made.

src/gennylib/src/Orchestrator.cpp
dalyd: Looking below I think you want to call this line before incrementing current on line 140.

dalyd (changes requested):

jimoleary (commented):
I think the confusion here is due to the vector names. I'll change  them to be more clear and adjust the code to reinforce this (plus add some comments).

src/gennylib/src/context.cpp
dalyd: A comment would be good. It's clear, but I also read right past it and then went searching for it.

jimoleary: done.

src/gennylib/src/Orchestrator.cpp
dalyd: All good now. Thanks. I still don't think the old code works, but that's no longer relevant :)

dalyd (approved):
Thanks for the updates. Much clearer to me now.

jimoleary:
evergreen merge

PR #839, opened 2023-03-09
EVG-19032 compile and run genny with ASAN

lgfang:
Without this library,  "run-genny install -s asan ..." will fail because of "undefined reference to `dlsym'"


johndaniels (dismissed):

lgfang:
Hi @spencerjackson,

Thank you very much for the additional information. Yes, I encountered the same error after I updated evergreen.yml to verify the compilation.

If I don't link `libdl` but instead change the compiler to clang++ (see below), the resulted binaries appear to be able to run as expected (and report memory leaks).

```text
     if sanitizer == "asan":
-        return ["-DCMAKE_CXX_FLAGS=-pthread -fsanitize=address -O1 -fno-omit-frame-pointer -g -ldl"]
+        return ["-DCMAKE_CXX_COMPILER=/opt/mongodbtoolchain/v4/bin/clang++", "-DCMAKE_CXX_FLAGS=-pthread -fsanitize=address -O1 -fno-omit-frame-pointer -g"]
```

That said, I feel that it might be a surprise to users that specifying `-s asan` changes the compiler being used. An alternative approach might be printing a warning message and then quit if compiler is g++ & asan is enabled. This way, the user will be "forced" to read the warning message and choose a different compiler (namely clang++) explicitly. But this can be a bit inconvenient.

I also tried running binaries generated by g++ with `LD_PRELOAD` and `ASAN_OPTIONS=verify_asan_link_order=0 ./run-genny workload src/workloads/docs/HelloWorld.yml` respectively.

```text
LD_PRELOAD=/opt/mongodbtoolchain/revisions/11316f1e7b36f08dcdd2ad0640af18f9287876f4/stow/gcc-v4.zdM/lib64/libasan.so ./run-genny workload src/workloads/docs/HelloWorld.yml
ASAN_OPTIONS=verify_asan_link_order=0 ./run-genny workload src/workloads/docs/HelloWorld.yml
```

The output of the later seems to suggest that it works. But again I'm not sure if we should use this approach because `ASAN_OPTIONS=verify_asan_link_order=0` basically means disabling the check.

I'll continue investigating this issue and keep you updated. In the meantime, your advice is highly welcome.

Regards,

Lungang

src/lamplib/src/genny/tasks/compile.py
```diff
@@ -10,12 +10,17 @@
 SLOG = structlog.get_logger(__name__)
 
 
-def _sanitizer_flags(sanitizer: str = None):
+def _sanitizer_flags(sanitizer: str = None, genny_repo_root: str = None):
```
johndaniels, line 13: This function doesn't need to have a default parameter.

johndaniels: Make both of them not have defaults

src/lamplib/src/genny/tasks/compile.py
```diff
@@ -10,12 +10,17 @@
...
     if sanitizer is None:
         return []
 
     if sanitizer == "asan":
-        return ["-DCMAKE_CXX_FLAGS=-pthread -fsanitize=address -O1 -fno-omit-frame-pointer -g"]
+        assert genny_repo_root is not None, "Code error: call _sanitizer_flags without genny_repo_root"
```
johndaniels, line 18: Skip assert, or put it at very top if you think its a risk.

johndaniels (changes requested):

johndaniels (approved):

lgfang:
evergreen merge

PR #841, opened 2023-03-14
PERF-3888 Add low-priority background operation to sinusoidal workload

pvselvan:
[sys perf evergreen patch](https://spruce.mongodb.com/version/640f6e0761837dd66a748a1d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/execution/SinusoidalReadWrites.yml
+++ b/src/workloads/execution/SinusoidalReadWrites.yml
@@ -87,9 +87,9 @@ ActorTemplates:
           Duration: 2 minutes
           CollectionCount: *NumColls
           Operations:
-          - OperationName: updateMany
+          - OperationName: updateOne
             OperationCommand:
-              Filter: {a: {$gt: 50000}}
+              Filter: {a: {^RandomInt: {min: 0, max: 100000}}}
               Update: {$set: {c: {^RandomInt: {min: 0, max: 100000}}}}
             OperationMetricsName: UpdateDocs
             ThrowOnFailure: false
--- a/src/workloads/execution/SinusoidalReadWrites.yml
+++ b/src/workloads/execution/SinusoidalReadWrites.yml
@@ -39,7 +39,7 @@ ExpiredDocument: &expiredDoc
 Clients:
   Default:
     QueryOptions:
-      maxPoolSize: 303
+      maxPoolSize: 403
 
 ActorTemplates:
 - TemplateName: WritesActorWithSleepTemplate
@@ -53,7 +53,8 @@ ActorTemplates:
         Active: {^Parameter: {Name: "ActivePhases", Default: [1]}}
         NopInPhasesUpTo: *NumPhases
         PhaseConfig:
-          Duration: 1 minute
+          SleepBefore: 10 seconds
+          Repeat: 3
           CollectionCount: *NumColls
           Operations:
           - OperationName: insertMany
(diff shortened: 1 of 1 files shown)


src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -75,37 +87,56 @@ ActorTemplates:
...
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *NumPhases
+      PhaseConfig:
+        # Add a large set of expired documents at start to make sure TLL monitor doesn't catch up in future phases.
```
gregorynoma, line 115: Did we confirm in the FTDC data that we were indeed always behind on TTL deletes?

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -75,37 +87,56 @@ ActorTemplates:
...
-        # Have 500 documents ready at start
         Repeat: 1
-        BatchSize: 100
+        Database: admin
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            ttlMonitorSleepSecs: 1
+        SleepAfter: 60 seconds  # Wait 60 seconds so that the new ttl sleep config is active.
```
gregorynoma, line 104: Do we still need this even after [SERVER-70055](https://jira.mongodb.org/browse/SERVER-70055)? I guess we'll still need it for old branches if we're running this workload there?

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -75,37 +87,56 @@ ActorTemplates:
...
       PhaseConfig:
-        # Have 500 documents ready at start
         Repeat: 1
-        BatchSize: 100
+        Database: admin
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            setParameter: 1
+            ttlMonitorSleepSecs: 1
```
gregorynoma, line 103: Does this mean we'll be running the TTL monitor every 1 second?

haleyConnelly: This means the TTL monitor will sleep 1 second after a pass.

A pass can continue indefinitely until there are no documents left to delete after the most recent subpass.

A subpass takes a snapshot of the TTLCollections, and continuously iterates over each index, deleting up to n-documents, until there are no more documents to delete on said index. A subpass is limited by a timer, to ensure that if new TTLIndexes are added, they won't be starved of deletes during the remainder of the pass.

haleyConnelly (commented):
Holding off until there is information about the FTDC data.

If the TTL deletes are being deprioritized, we should see a "concurrentTransactions write lowPriority queueLength" of greater than 0

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Yea, I think so. Since this is just the setup phase, it doesn't add a lot of overhead to the workload.

src/workloads/execution/SinusoidalReadWrites.yml
pvselvan: Based on the FTDC data, the TTL monitor is deleting fewer documents than it was deleting after the load phase, but @haleyConnelly is there another metric I should be using besides deletedDocuments to measure this?

pvselvan:
I'm not observing "queueLength" in the FTDC data, but I am seeing a lot of other "concurrentTransactions write lowPriority" metrics. Is queueLength nested in another metric?

src/workloads/execution/SinusoidalReadWrites.yml
```diff
@@ -63,49 +84,116 @@ ActorTemplates:
...
           CollectionCount: *NumColls
           Operations:
           - OperationName: find
             OperationCommand:
-              Filter: { a: { ^RandomInt: { min: 0, max: *NumDocs } } }
+              Filter: {a: {^RandomInt: {min: 0, max: 100000}}}
             OperationMetricsName: ReadDocs
             ThrowOnFailure: false
+
+- TemplateName: UpdatesActorWithSleepTemplate
```
pvselvan, line 97: Adding this update actor to update indexed documents definitely adds more stress to the system. Though the "eviction worker thread active" doesn't exceed 4, it is consistently at 4, which would imply that they're doing a lot of work? Let me know if you have other suggestions.

With this phase, we are able to see pretty decent improvements in average throughput and latency.

gregorynoma (approved):

src/workloads/execution/SinusoidalReadWrites.yml
haleyConnelly: Hm, in general there should be a queueLength of non-zero to show low priority operations are at least queueing. Compared to the load phase, aren't there more threads the TTL is competing with in phase 2?

haleyConnelly (commented):
Are we not seeing queueLength at all, or is it zero (and masked in t2 unless you select the option to "show all-zero charts")

PR #844, opened 2023-03-18
Perf-3460: Change MongosyncScript to support poll_for_commit_point action.

huan-Mongo:
Add command in Mongosync actor to keep polling until canCommit = true and lagTimeSeconds < 120

Patch run: https://spruce.mongodb.com/version/641b3d219ccd4e498aa2832d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=success

According DSI and Privateworkload change is 
https://github.com/10gen/dsi/pull/1244
https://github.com/10gen/PrivateWorkloads/pull/27


src/cast_core/include/cast_core/actors/ExternalScriptRunner.hpp
```diff
@@ -52,6 +52,7 @@ class ExternalScriptRunner : public Actor {
     struct PhaseConfig;
     PhaseLoop<PhaseConfig> _loop;
     std::string _command;
+    std::unordered_map<std::string, std::string> _environmentVariables;
```
dalyd, line 55: Is this new and/or needed? I don't see it referenced in the other changes.

dalyd (commented):
I'm not sure I completely understand this part of the code, but the change looks pretty small and similar to what already exists.

src/cast_core/include/cast_core/actors/ExternalScriptRunner.hpp
huan-Mongo: This is supposed to be from Brad's branch which I merge from. He might have removed it later. I am removing it from here.

PR #849, opened 2023-03-23
PERF-3950 Temporarily disable QE performance tests

shreyaskalyan:


SimonEismann:
This looks pretty good. 

Could we create a Jira ticket to reopen these and add a comment linking to this ticket above the autorun blocks. This would serve two purposes:
- documents why those autoruns are currently commented out
- keep a reminder around that we need to reactive them later

shreyaskalyan:
@SimonEismann this current ticket documents why the tests are currently disabled, and I have linked the new PERF ticket in a TODO. This ticket is also linked to the follow up PERF ticket. Let me know if you think this patch needs anything else.

SimonEismann (approved):
LGTM, thanks for adding the ticket

PR #851, opened 2023-03-29
PERF-3866 Workloads for $percentile accumulator, expression and window function

IrinaYatsenko:
Given we are not seeing any significant different between various distributions, I'm not convinced we need/should run this workload regularly in Evergreen but publishing as a PR for reference. Eventually, we'll settle here on what we really want to merge :)

[Patch w/ 1e7 docs](https://spruce.mongodb.com/version/643964540305b9bd36a61392/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[Patch 2 w/ 1e7 docs](https://spruce.mongodb.com/version/6439b06b61837d4a4aa6844d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[Patch 3 w/ 1e6 docs](https://spruce.mongodb.com/version/643d478a3e8e867d218f0101/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[Patch 4 w/ 1 e6 docs](https://spruce.mongodb.com/version/643d8e3b2a60edcb4efde6d7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


src/workloads/query/PercentilesSpillToDisk.yml
```diff
@@ -0,0 +1,158 @@
...
+        # We have 100k documents with each being between 10kB to 20kB. When divided into 20 partitions, it won't spill, but when divided into fewer partitions (like 2) it will spill.
+        DocumentCount: 100000
+        BatchSize: 1000
+        Document:
+          # We will use the _id field to create evenly distributed partition keys to test spilling.
+          _id: {^Inc: {start: 0}}
+          # This data field is what we calculate the percentiles of.
+          data: {^RandomDouble: {distribution: normal, mean: 50.0, sigma: 10.0}}
+          # This string pads the doc size out to around 10KB.
+          string: &string {^RandomString: {length: 10000}}
```
IrinaYatsenko, line 95: I don't understand the mechanics here: the "string" field isn't used in $setWindowFields. Why does it affect spilling?

src/workloads/query/PercentilesSpillToDisk.yml
```diff
@@ -0,0 +1,158 @@
...
+        Database: *db
+        Operations:
+        - OperationName: RunCommand
+          OperationCommand:
+            update: *coll
+            updates: [
+              {
+                q: {},
+                u: [
+                  {$set: {
```
IrinaYatsenko, line 115: I suspect it might be faster to create the dataset with these additional fields at the start, rather than adding them through and update operation.

src/workloads/query/PercentilesSpillToDisk.yml
```diff
@@ -0,0 +1,158 @@
...
+      Name: "percentile_sort_spill"
+      Partition: "$spill_unindexed"
+      Sort: {spill_unindexed: 1}
+
+# Just the $_internalSetWindowFields will spill. The partitions are too large to fit in memory, and the $sort is indexed.
+- ActorFromTemplate:
+    TemplateName: ComputePercentile
+    TemplateParameters:
+      OnlyActiveInPhase: 3
+      Name: "percentile_sort_spill"
```
IrinaYatsenko, line 147: Should use a different name?

src/workloads/query/PercentilesSpillToDisk.yml
galon1: When we partition the documents, entire documents are spilled to disk. So even though we don't use the `string` field in the `$setWindowFields` stage, the stage will add the entire document to a cache and then spill to disk if it runs out of memory. So the stage will still spill to disk.

I can change this to have more documents without the `string` field. I would just need to add around 1.25 million (I think) too have enough `double` values to spill.

IrinaYatsenko: The confusing bit for me was that I expected the $percentile window function to trigger spilling (because it accumulates values) but if I understand you correctly, the spilling might happen even for a window function that doesn't collect any data because the overall window functions infra causes the whole document to be kept in memory, and, therefore, might cause spilling. Right?

galon1: Yes I think so, the pulling in documents happens at a higher level (on the `setWindowFields` stage), so yes I think spilling could happen even if the window function doesn't do anything with data.

src/workloads/query/PercentilesSpillToDisk.yml
galon1: Done! I was concerned about having the partitions even, but that isn't necessary because they'll still be large enough to spill even if they aren't perfectly even.

galon1 (commented):
I marked this PR ready for review. I will run it on evergreen a few times to see if the tests are faulty. Additionally, we need to confirm that we want to keep all of these workloads.

[Patch](https://spruce.mongodb.com/version/64381a3f57e85adea3ee1824/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

src/workloads/query/PercentilesAgg_Distributions.yml
```diff
@@ -0,0 +1,226 @@
...
+
+Keywords:
+- group
+- percentile
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-all-feature-flags
```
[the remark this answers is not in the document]
galon1: We put down replica set just because we want to run the feature in both SBE and classic engine to compare the runtime when this feature eventually gets added to SBE, and we did not see an option in the form to have both options in a standalone. However, we don't need to run a replica set (the feature doesn't use any replication features). So is there a pair of build variants using standalone (one for SBE and one for classic)?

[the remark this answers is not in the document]
MAhmadShah: I see, thanks for the feedback I'll let the team know to update the form. These two variants should fit the bill for your use case:

- linux-standalone-classic-query-engine.2022-11

- linux-standalone-sbe.2022-11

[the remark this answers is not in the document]
galon1: Thank you! Those would work. Should we update our response to the form when these fields are added?

src/workloads/query/PercentilesAgg_Distributions.yml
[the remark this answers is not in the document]
galon1: Yes I agree we shouldn't keep all the different distributions, since it did take much longer than the other workloads to run. I think normal and mixed dataset would be good! I'll make sure to leave a comment explaining the other local variables.

IrinaYatsenko (commented):
BTW, because I created the review I can neither approve it nor request changes. So my "LGTM" will have to be without a green check mark :)

src/workloads/query/PercentilesAgg_Distributions.yml
```diff
@@ -0,0 +1,278 @@
...
+          Threads: 1
+          Repeat: 1
+          Database: *db
+          CollectionCount: 1
+          DocumentCount: *docCount
+          BatchSize: 1000
+          Document:
+            _id: {^Inc: {start: 0}}
+            data: {^Parameter: {Name: "Data", Default: 0}}
+            # We will use this field as our group key to test having ~500 groups (10,000 documents divided over 20 keys).
```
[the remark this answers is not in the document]
galon1: oh sorry! I mixed up the number of documents with this workload and expressions. Fixed it now!

src/workloads/query/PercentilesSpillToDisk.yml
```diff
@@ -0,0 +1,145 @@
...
+        CollectionCount: 1
+        # We have 100k documents with each being between 10kB to 20kB. When divided into 20 partitions,
+        # it won't spill, but when divided into fewer partitions (like 2) it will spill.
+        DocumentCount: 100000
+        BatchSize: 1000
+        Document:
+          _id: {^Inc: {start: 0}}
+          # This data field is what we calculate the percentiles of.
+          data: {^RandomDouble: {distribution: normal, mean: 50.0, sigma: 10.0}}
+          # This string pads the doc size out to around 10KB.
```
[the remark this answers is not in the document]
galon1: Yes I can add the comment. Based on my understanding spilling can happen in the window function (in this case if we have too many double values) or in the `setWindowFields` infrastructure when we are partitioning the data. However, there is not a difference in the spilling logic. In other words, both test are looking at the same spilling behavior and the same behavior for reading documents from disk. In one scenario we are spilling less but larger documents and the other we are spilling more but smaller documents. Because we are partitioning the data, part of the window function is iterating through the partition and this is what is spilling in this case.

However, we can change the test to have the window function itself spill the data.

src/workloads/query/PercentilesWindow.yml
```diff
@@ -0,0 +1,156 @@
...
+        PhaseConfig:
+          Repeat: 10
+          Database: *db
+          Collection: *coll
+          Operations:
+          - OperationName: aggregate
+            OperationCommand:
+              Pipeline: [
+                {$setWindowFields: {
+                  sortBy: { sort: 1 },
```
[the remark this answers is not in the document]
galon1: I can definitely do 1 and 2 but I'm a bit confused on 3. I'm not sure how to make sure document based windows and range based windows have around the same size windows. Document based windows have varying sizes (around 1000 for [-500, 500] and a different size for each document for [0, "unbounded"]). If I do ^Choose of a few differently bounded ^RandomDate then I can try and values that 1000 documents within each range. But to match the size of [0, "unbounded"] windows I can do something like ["current", "unbounded"], where the minimum value will have all the documents in the range and the maximum value will only have one document. Would that be close enough?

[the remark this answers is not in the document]
galon1: sortBy requires a specific field, and is required for bounded documents and range window. I added the same sort for all the different types of windows so we can just see the change in performance for the different types of window. Is there a different field that we can use for `sortBy` that would slow down the workload significantly less than this?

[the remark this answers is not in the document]
IrinaYatsenko: Bummer. I didn't know we cannot use $natural sort with window functions. And if we use anything else, I don't think we can avoid sorting (even if the field is presorted by construction). Oh well. 

For me, three aspects of perf testing are really important:
1. Is it clear what they test?
2. Is it clear how to interpret the results?
3. Are we wasting resources (CPUs/time/etc.) running them?

So, it's not required to be able to compare the results of document-based vs range-based windows, but the benchmarks should be authored in a way that doesn't temp our future selves (or other engineers) to do such a comparison, get surprised and waste time trying to understand what's going on.

And it's totally OK to test a regular data pattern as long as we understand/document that this is what we are covering. If our tools don't allow us to test more varied data, so be it, but at least we don't create any illusions about what's being tested.

Does this make sense? My apologies for getting into the philosophical weeds :)

src/workloads/query/PercentilesAgg_Distributions.yml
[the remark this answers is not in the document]
MAhmadShah: You should be okay to leave your submission as is.

src/workloads/query/PercentilesWindow.yml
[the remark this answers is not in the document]
galon1: That makes sense. I added a comment about how we do not intend to compare documents vs range windows. I think we should just separate the documents vs. range numbers, but set up a more realistic situation with range windows. I added a date field so we can have a date range and sort by date, which would make more sense.

src/workloads/query/PercentilesWindow.yml
```diff
@@ -0,0 +1,173 @@
...
+      # If the unit was larger than 'day' we would have nearly all the documents in every window,
+      # which would not test having a smaller bounded window.
+      Documents: { range: [-500, 500], unit: "day"}
+      OnlyActiveInPhase: 4
+
+- ActorFromTemplate:
+    TemplateName: ComputePercentile
+    TemplateParameters:
+      Name: "percentile_large_range_window"
+      # We want to test larger windows, so we increase the unit to 'month'.
```
IrinaYatsenko, line 163: the comment is somewhat confusing to me: we test larger windows by the virtue of them being unbounded, no? if I understand it correctly, the unit choice would affect the pattern of how many values are added/removed before recomputing the percentile (which is a great test to have, but could you please confirm that my understanding is correct here?)

IrinaYatsenko (commented):
LGTM with suggestions. Unfortunately, I cannot formally approve the PR because I created it initially...

src/workloads/query/PercentilesWindow.yml
galon1: yes the unit choice does affect the pattern of how many values are added/removed, but it is also required for the range windows that have a time range. So to test different patterns of adding/removing values we can have 2 tests. One we run a [0, "unbounded"] window with the 'year' unit to pull in as many document as possible and the other we can run on a tighter unit like 'day'.

src/workloads/query/PercentilesWindowSpillToDisk.yml
```diff
@@ -0,0 +1,149 @@
...
+          # and spills full documents, even though 'string' is not used in the $percentile window function, this
+          # field will increase the document size to trigger a spill.
+          string: &string {^RandomString: {length: 10000}}
+          # We will use the fields below to create approximately even partitions.
+          spill_indexed: {^Choose: {from: [1, 2]}}
+          spill_unindexed: {^Choose: {from: [1, 2]}}
+          not_spill_indexed: {^Choose: {
+            from: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]}}
+
+- Name: CreateIndexes
```
IrinaYatsenko, line 102: I've happened to see in BigUpdate.yml workload that it should be possible to create indexes as part of the loader rather than in a separate phase (which makes sense for us because we don't care to measure perf of index creation)

galon1: Done!

src/workloads/query/PercentilesWindow.yml
```diff
@@ -0,0 +1,188 @@
...
+    TemplateParameters:
+      Name: "percentile_range_use_t_digest"
+      Documents: { range: ["unbounded", 0], unit: "year"}
+      OnlyActiveInPhase: 7
+
+# Run the same test as above with a tighter 'unit', changing the frequency of adding/removing values.
+- ActorFromTemplate:
+    TemplateName: ComputePercentile
+    TemplateParameters:
+      Name: "percentile_range_use_t_digest_day"
```
IrinaYatsenko, line 186: nit: I'd name it `percentile_range_tdigest_day` because a) "use" doesn't add any information, b) I believe it would be more likely for people to search for "tdigest" rather than "t_digest"

src/workloads/query/PercentilesWindow.yml
```diff
@@ -0,0 +1,188 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
+  This tests $percentile window functions over various bounded window sizes. Bounded window
```
IrinaYatsenko, line 4: This comment might be confusing because we actually use window function with an accurate algorithm for computing over _unbounded_ removable windows. The list that follows also seems to imply that range windows aren't removable (because you differentiate "small removable windows" vs "small range windows"). I'd suggest "1. Small document-based removable windows" and "4. Small range-based removable windows"

IrinaYatsenko (commented):
LGTM with minor comments!

galon1 (approved):
Marking this as approved, since Irina created the PR and can't officially approve it.

galon1:
evergreen merge

PR #852, opened 2023-03-30
PERF-3967: Ensure tests that run against Atlas, also run against Atlas-like

miguelangelnieto:
Change AutoRun so the same tests that run on "atlas" also run on "atlas-like-replica.2022-10". Also changed those scheduled on the old "atlas-like-replica" that is not scheduled anymore so they now run on "atlas-like-replica.2022-10"


jawwadasghar (approved):
LGTM

PR #853, opened 2023-03-30
PERF-3460: add pause and resume Mongosync actor

huan-Mongo:


src/workloads/c2c/ChangeEventApplication.yml
```diff
@@ -55,7 +57,13 @@ Actors:
   - LoadConfig:
       Path: *scriptsPath
       Key: PollForCEA
+  - LoadConfig:
```
huan-Mongo, line 60: In this way, we pre-buffer the change events before CEA starts to overcome the genny slowness issue we observed before.

dalyd (approved):

PR #854, opened 2023-04-03
PERF-3918: Generate better data for arraytoobject and objecttoarray expression b…

projjal:
…enchmark


lriuui0x0 (approved):

lriuui0x0:
I think it's probably enough to not have the case of duplicate keys, since we will be testing adding to deduplication map anyway.

Also curious why the build is all red here?

lriuui0x0:
@projjal See [this comment](https://github.com/10gen/PerformanceReports/pull/82#discussion_r1156421397), can we update the test name here?

lriuui0x0 (approved):

PR #856, opened 2023-04-03
PERF-3988: Reduce slow query logging on mixed_workloads_genny_stress

miguelangelnieto:
As seen on PERF-3979, this workload is expected to cause slow queries. But on some systems, like M60-like the resulting log file could be up to 35GB in size, making evergreen to fail because of disk space issues or memory exhaustion. This change is to reduce the query logging on this particular workload.

### DIFF
--- a/src/workloads/scale/MixedWorkloadsGennyStress.yml
+++ b/src/workloads/scale/MixedWorkloadsGennyStress.yml
@@ -237,4 +237,5 @@ AutoRun:
     mongodb_setup:
       $eq:
       - atlas
+      - atlas-like-replica.2022-10
       - replica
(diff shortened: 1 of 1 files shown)


jimoleary (commented):
Change looks good, do we want a patch for paranoia?

miguelangelnieto:
@jimoleary I forgot to link it, it is [here](https://spruce.mongodb.com/version/642aaf3d61837d8a61a151a3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). Now M60-like does not fail.

jimoleary (approved):

PR #857, opened 2023-04-03
EVG-19133: Pre-warm connection for Loader.cpp.

johndaniels:
Since connections on clients in libmongoc are opened lazily, currently the connection time is being counted when we measure time to load initial data. This caused BF-27679 when we updated our infrastructure to have a valid certificate chain, likely due to OpenSSL doing more work to process opening a TLS connection. The recommended fix was to make sure our connections are warm and open before we measure database operation, since that steady state is what we care most about for user operations.

This change pings the database at the beginning of each phase to make sure that we already have an active connection before we time anything. This should bring timings down and reduce noise.

### DIFF
--- a/src/cast_core/src/Loader.cpp
+++ b/src/cast_core/src/Loader.cpp
@@ -126,6 +126,11 @@ struct Loader::PhaseConfig {
 
 void genny::actor::Loader::run() {
     for (auto&& config : _loop) {
+        if (!config.isNop()) {
+            // Warm up the connection to the database
+            auto ping = builder::stream::document{} << "ping" << 1 << builder::stream::finalize;
+            config->database.run_command(ping.view());
+        }
         for (auto&& _ : config) {
             for (uint i = config->collectionOffset;
                  i < config->collectionOffset + config->numCollections;


src/cast_core/src/Loader.cpp
```diff
@@ -126,6 +126,11 @@ struct Loader::PhaseConfig {
 
 void genny::actor::Loader::run() {
     for (auto&& config : _loop) {
+        if (!config.isNop()) {
+            // Warm up the connection to the database
+            auto ping = builder::stream::document{} << "ping" << 1 << builder::stream::finalize;
+            config->database.run_command(ping.view());
```
johndaniels, line 132: From Code review: This will throw an exception. Can we catch this easily?

rtimmons: This can throw an exception if the db server can't be reached for whatever reason. I don't think we'd want to catch or handle the exception in that case since the environment-under-test is proving unreliable if it can't ping. Probably better to error at the ping/warmups stage where the failure-condition ("db down?") is more obvious.

Apart from that--
This runs at the beginning of each Phase - is that what we want? Or perhaps it wants to run at the top of `run()` so it only runs at the beginning of the workload and can prevent other Actors from doing their thing until this is done.

Alternatively, perhaps this wants to be done in the Actor or maybe even pool constructor? Like pre-warm a pool in the PoolManager before giving them out? Connections are lazily loaded so we may end up re-implementing a pool construct in that extreme, so I'm not sure what's exactly ideal.

Won't block review since this code is perfectly valid, just want to throw a couple other ideas/considerations in the mix. Since this is limited to just the Loader Actor, it also seems reasonable to try this out here and then consider refactoring into PoolManager or whatever later.

johndaniels: You are right of course that probably we would want to rearchitect Genny to pre-warm the connection it passes to an actor, and probably relatedly also get rid of the connection pool entirely since everyone seems to hold onto a connection for the entire run of a test?

I'd be happy to not push this change as well but the feedback from Ger was that adding the extra noise from the TLS connection was a big problem.

thessem: I raised (aha) the exception comment in a meeting, the idea was that this code is ~probably going to be the one raising exceptions for connection issues, so it might be a good location to wrap the exception with some user friendly text (failed initial connection) rather than confuse them with (I handled exception during ping, or whatever it is). 

I am not fully convinced that getting rid of a connection pool is a great idea, we could use the pool to implement some kind of min_idle parameter to guarantee there are at least that many idle, but warm, connections sitting around ready for immediate consumption. Agree with the above that it’s useful to have a location for this kind of pre-warning logic and eventually take it away from the connection consumers. I think I’m just arguing for tomàto here (pool, factory, it’s in the eye of the user ☺️)

Mostly I’m just typing stuff to get into the swing of technical discussions again, I have no issues with the code as-is and I believe it solves the problem it is setting out to solve

thessem (approved):

src/cast_core/src/Loader.cpp
lgfang: I'd also prefer to get this in as-is.

This code change does not change any error/exception handling w/r connection error (due to the target mongodb cluster being unable to get online/response in time). It just excludes the time needed to establish the connection (up to 30 seconds) from performance metrics by running a command before the real test starts. 

Before the code change, an exception may be thrown when the test starts and the exception is caught and handled somewhere above  in the call stack. Now, if same connection error is encountered, the same exception will be thrown a few lines earlier from the same function.

thessem (approved):

PR #862, opened 2023-04-12
PERF-3974: Add workload for many different query shapes

naama-bareket:
I compared the time taken to run queries when run concurrently with queries with different shapes versus with queries of the same shape. 

Average time for queries to run when concurrent with queries of different shapes: `19134142652`

Actor `FindExists` performed around average with different shaped queries: `17295523476`
When run with queries of its same shape, it averaged at: `21371788523` 
(the difference was `4076265047`)

Actor `AggToLowerArr` performed slow with different shaped queries: `28186482902` 
When run with queries of its same shape, it averaged at: `41244399118`
(the difference was `13057916216`)

Actor `SimpleFind` performed quickly with different shaped queries: `13606118574`
When run with queries of its same shape, it averaged at `11110266494`
(the difference was `-2495852080`)

To summarize, it seems that for queries that performed average or slow when running concurrently with queries of different shapes, the performance when running with queries of the same shape was worse. For a query which ran fast when running concurrently with queries of different shapes, the performance when running with queries of the same shape was better.

Here is the link with the numbers: https://docs.google.com/spreadsheets/d/16GWOQIOfzhC_Z3-6uLRiT_RD8pU1VbA_3tOQNYjxU30/edit#gid=0


src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+# Different Query Shapes.
+- Name: SimpleFind
```
cswanson310, line 69: I think you can use some actor templating to make it more obvious what's going on here. Check out this example: https://github.com/mongodb/genny/blob/a603ba1a3b12c75f3c70509635ab6526daf4a5a4/src/workloads/query/CsiHeavyDiskUsage.yml#L54

You can also use fancy YAML features, like this: https://www.linode.com/docs/guides/yaml-anchors-aliases-overrides-extensions/

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+        Repeat: 1
+        Database: *Database
+        Collection: *Collection
+        Operations:
+        - OperationMetricsName: PipelineWithoutOptions
+          OperationName: aggregate
+          OperationCommand:
+            Pipeline: [{$match: {$expr: {$lt: [{$getField: {$literal: "$a"}}, 30]}}}]
+
+# Same Query Shape.
```
cswanson310, line 298: I think that you can just specify `Threads: 16` instead of repeating it - right?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: 1}
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-all-feature-flags
+    branch_name:
```
cswanson310, line 993: I think we support $gte now: https://github.com/mongodb/genny/blob/6d4f0888d97bf5a753460db13e5974cfffaeae17/src/workloads/execution/ClusteredCollection.yml#L27

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+    OnlyActiveInPhases:
+      Active: [3]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: 1}
```
cswanson310, line 83: Can you leave a comment here - we expect this to find ... 1e6 / 200 = 5000 documents? Or 0.5% of the collection?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: *Threads
+        Collection: *Collection
+        Operations:
+        - OperationName: drop
+
+- Name: InsertData
```
cswanson310, line 38: Does this support creating indexes? I think we probably want an index on {a: 1, b: 1} and {string: 1}?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+    OnlyActiveInPhases:
+      Active: [3]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: 1, b: 2}
```
cswanson310, line 99: Oh interesting so this will find even fewer documents - 0.005 * 0.005 * 1e6 = 2.5 documents?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,999 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {b: {$gt: 4}}
+
+- Name: FindInSingleArr
```
cswanson310, line 117: I think rather than defining such dramatically different query shapes, we can probably generate different shapes by just changing the path we query. We could generate data with lots of different field names as we do here: https://github.com/mongodb/genny/blob/6d4f0888d97bf5a753460db13e5974cfffaeae17/src/workloads/query/CsiFragmentedInsertsFlat.yml#L61

I think we could probably get away with adding a couple different ones that are like
```js
{a: 1, b: 2, f0: {$exists: false}}
{a: 1, b: 2, f1: {$exists: false}}
...
```
The field doesn't need to actually exist in the data for it to be considered a different shape.

cswanson310 (changes requested):
I'm trying to look into how many cores we expect to have in evergreen test environment. 

In the meantime, I have some comments below.

I think we are overall interested in measuring two things:
1) Performance with telemetry vs. without telemetry on these kinds of workloads. This we should get 'for free' via our telemetry build variants (as long as we make sure to include these tests in https://jira.mongodb.org/browse/SERVER-76083
2) Highlighting the scaling of the `Partitioned` telemetry store. I think the best way to do this would be to have one query shape - run that with a high thread count. Then take that query shape and add some do-nothing clauses to it like `{newField143: {$exists: false}}` - such that it is doing approximately the same amount of work, but there are now a lot more distinct shapes. So if we were running 32 threads all of 
```js
{a: 1, b: 2, newField0: {$exists: false}}
``` 
in the first phase, then that would compare to 32 different shapes like this, each running 1 thread:
```js
{a: 1, b: 2, newField0: {$exists: false}}
{a: 1, b: 2, newField1: {$exists: false}}
{a: 1, b: 2, newField2: {$exists: false}}
... // 32 times total
```

So that way when we compare the throughput - we know that the amount of query engine work is approximately the same, it's just the overhead of the 'partitioned' data structure that changed.

Although one thing that may be hard - if we have 32 different actors, then is it safe to total the throughput of all 32 and compare that to the throughput of the 1 which had 32 threads? I would think so? But the percentiles we couldn't really compare. So maybe we should look into whether we can use `{^RandomString: }` or something to have it be one actor that is oscillating between 32 shapes, and give that actor 32 threads. Does that make sense?

naama-bareket:
@cswanson310 I changed the workload to have 2 actors, one which oscillates between query shapes on 32 threads, and one that runs the same shape on 32 threads. It actually seems like oscillating shapes had worse performance, which is different from my initial analysis with the past workload. I was thinking maybe the extra time taken could be due to the building of the object / creating the random field name, but I ran the workload with the one shape query going through the same process and am still getting the same results. Here is the link to the new sheet with the results: https://docs.google.com/spreadsheets/d/1_cLQ5sZ3A-4ag05b46n1GZAKSTTzm_YzUqeBgSXFCOg/edit#gid=0

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,129 @@
...
+# 32 threads with distinct query shapes
+- ActorFromTemplate:
+    TemplateName: ComputeFindQuery
+    TemplateParameters:
+      Name: "AlternatingShapes"
+      Threads: 32
+      Filter:
+        {$and:
+           [{^Object: {
+             withNEntries: 3,
```
cswanson310, line 102: This is cool, but I'm curious if you could do
```
{$and: [
  {a1: 1},
  {a2: 1},
  {^Object: {...}}
]}
```

That way it would be more obvious that we're filtering primarily by "a1" and "a2" and the extra thing is the source of randomness to change the query shape.

Is that more or less the intention here? Or did you want some queries to not have "a1" or "a2" for some reason?

cswanson310: > @cswanson310 I changed the workload to have 2 actors, one which oscillates between query shapes on 32 threads, and one that runs the same shape on 32 threads. It actually seems like oscillating shapes had worse performance, which is different from my initial analysis with the past workload. I was thinking maybe the extra time taken could be due to the building of the object / creating the random field name, but I ran the workload with the one shape query going through the same process and am still getting the same results. Here is the link to the new sheet with the results: https://docs.google.com/spreadsheets/d/1_cLQ5sZ3A-4ag05b46n1GZAKSTTzm_YzUqeBgSXFCOg/edit#gid=0

This is quite interesting! I am just starting to look through, but it might be that the way you constructed the query here will sometimes skip the "a1" or "a2" predicates which are the indexable predicates?

Or is the `determenistic: true` part mean that it will always walk in order? If that's so, I might still recommend changing to the different $and for clarity, or at least adding a comment making this clear.

src/workloads/query/TelemetryQueryShapes.yml
cswanson310: Actually now that I say this - the workload doesn't actually depend on telemetry, right? Since we will be backporting telemetry, maybe we should just run this on all branches from the start? Do you see any reason why it wouldn't work?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,129 @@
...
+
+ActorTemplates:
+- TemplateName: ComputeFindQuery
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "ComputeFindQuery"}}
+    Type: CrudActor
+    Threads: {^Parameter: {Name: "Threads", Default: 1}}
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1024}}]
```
cswanson310, line 34: 1024??

cswanson310: oh is this just so it's not conflicting? A comment would be helpful - or is there a way to do this without a default and force template instantiators to specify it?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,129 @@
...
+
+KeyWords:
+- telemetry
+- query
+- shape
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 400
```
cswanson310, line 14: Do you know why this is here? Or just copied from another test?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,129 @@
...
+- ActorFromTemplate:
+    TemplateName: ComputeFindQuery
+    TemplateParameters:
+      Name: "OneShape"
+      Threads: 32
+      Filter: 
+        {$and:
+           [{a1: 1, a2: 1, x0: {$exists: false}}]
+        }
+      OnlyActiveInPhase: 4
```
cswanson310, line 120: > @cswanson310 I changed the workload to have 2 actors, one which oscillates between query shapes on 32 threads, and one that runs the same shape on 32 threads. It actually seems like oscillating shapes had worse performance, which is different from my initial analysis with the past workload. I was thinking maybe the extra time taken could be due to the building of the object / creating the random field name, but I ran the workload with the one shape query going through the same process and am still getting the same results. Here is the link to the new sheet with the results: https://docs.google.com/spreadsheets/d/1_cLQ5sZ3A-4ag05b46n1GZAKSTTzm_YzUqeBgSXFCOg/edit#gid=0

Another theory: this "OneShape" actor is going second - that might be an advantage because the first actor will pull the relevant data into the cache, making this one go slightly faster. Could you swap the order as an experiment? Or even just duplicate the "AlternatingShapes" in phase 5 as well as phase 3 to see if they have different performance?

cswanson310: If that turns out to be true - the next experiment might be to see if adding another quiesce stage between the two could help. I'm not really sure it is expected to help with flushing caches, it may just be about flushing to disk for insert/writing stability.

cswanson310 (commented):
Thanks for the update @naama-bareket ! This is much simpler and the results summary is helpful. The data is surprising, so I have some suggestions of follow-up experiments to try to dig deeper. You could also try to make a flame graph of the two phases, as described:
[here](https://github.com/10gen/performance-tooling-docs/blob/main/profiling_workflow/profiling_workflow.md) and also [here](https://github.com/10gen/performance-tooling-docs/blob/main/profiling_workflow/linux_perf.md)

I noticed that those pages make it easier to do when running with DSI or in evergreen. Running in evergreen might be an easy-ish option. But you can also do [this perf record step](https://github.com/10gen/performance-tooling-docs/blob/main/profiling_workflow/linux_perf.md#collecting-data)
```
perf record -F 99 -p $(pgrep mongod) -g -- sleep 30
```
step when you have a mongod running locally - as I expect you have been testing.

src/workloads/query/TelemetryQueryShapes.yml
cswanson310: Oh sorry - outdated comment but that was from the [line 53 above](https://github.com/mongodb/genny/pull/862/files/ddbb03b58a243493317b4cbeb9af13e37ab63141#diff-b92ce3540ab5fab9fec3758b7cbcdcf64578ffdc7afd232fb8234cf5599e0e83R53) which defined
```yml
a: &integer {^RandomInt: {min: -100, max: 100}}
```
I guess maybe it is actually either 199 if those bounds are exclusive or 201 if they are inclusive, since 0 would sit right in the middle.

cswanson310:
Also sorry, I didn't fully process this part:
> I ran the workload with the one shape query going through the same process and am still getting the same results

Is that what's reflected in the linked sheet? And the idea was that "One Shape" would use the `{^Object}` macro as well to generate the queries - to see whether that mattered?

naama-bareket:
> Also sorry, I didn't fully process this part:
> 
> > I ran the workload with the one shape query going through the same process and am still getting the same results
> 
> Is that what's reflected in the linked sheet? And the idea was that "One Shape" would use the `{^Object}` macro as well to generate the queries - to see whether that mattered?

Yes, I had the "One Shape" use the `{^Object}` macro to see if that was the reason the multiple shape took longer, but the results did not change (it was identical to one of the runs in which it didn't use the macro, so I didn't add it to the linked sheet)

src/workloads/query/TelemetryQueryShapes.yml
naama-bareket: Yeah I think that would be fine!

src/workloads/query/TelemetryQueryShapes.yml
naama-bareket: The `deterministic: true` part means that it will go in order, so we always had "a1" and "a2", but I agree it is cleaner to have them separately. I think for some reason I thought I couldn't do it that way, but I just tried again and it works!

src/workloads/query/TelemetryQueryShapes.yml
naama-bareket: I was having this error: `Error in parallel execution. First exception's what(): Failed to acquire an entry from the client pool.` And saw this was the fix someone else used for it on slack

src/workloads/query/TelemetryQueryShapes.yml
naama-bareket: Yes it was so it wouldn't conflict / by default wouldn't run in any of the phases, I fixed it to force specification

naama-bareket:
I ran the workload with telemetry and without telemetry to compare the relative overhead of telemetry between concurrent queries with the same shape vs alternate shapes. Here are the results: https://docs.google.com/spreadsheets/d/1aklsx4H0fmjPt_QXkZchTHSLszM_18P1Pt1RuQig8cc/edit#gid=0

I ran the workload both with the OneShape actor going first, the AlternatingShape actor going first, a Quiesce actor in between the two (no effect), and adding second AlternatingShape so that neither actor I looked at went first. It seems like there is an effect on the overhead of telemetry - when the OneShape actor went first, there was no difference in its performance without or without telemetry, but when it followed the AlternatingShape actor, the performance was faster without telemetry. However, there was no effect on the AlternatingShape if it went first or second.

Pretty consistently the % difference for the AlternatingShape running with and without telemetry was 6.1%, and for the OneShape actor when not going first was 8%. 

The OneShape actor also consistently performed better than the AlternatingShape with and without telemetry, and whether going first or second. @cswanson310

src/workloads/query/TelemetryQueryShapes.yml
cswanson310: Weird. Looks like many tests do this without any comment... 🤷🏼

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,133 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests performance of telemetry with different query shapes.
```
cswanson310, line 4: This isn't entirely accurate anymore. Could you update it to say something more about how it was designed to stress the telemetry collection - but doesn't actually enable telemetry. It can be run with or without telemetry for comparisons.

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,133 @@
...
+             duplicatedKeys: skip
+           }}]
+        }
+      OnlyActiveInPhase: 4
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-all-feature-flags
```
cswanson310, line 133: Oh interesting. In your patch build you might only find your tests run on certain variants. I'd be curious which it is. Relatedly, https://jira.mongodb.org/browse/SERVER-76083 is trying to get more genny tests to run on telemetry variants, which might mean this should change. I'm not sure yet...

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,133 @@
...
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: {unused: "please specify in which phases this actor should be active."}}}]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: *Repeat
+          Database: *Database
+          Collection: *Collection
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {a: 1, b: 2, x0: {$exists: false}}}}
```
cswanson310, line 43: [nitpick] this looks outdated for a: 1, b:2 - now it's a1 and a2, right?

src/workloads/query/TelemetryQueryShapes.yml
```diff
@@ -0,0 +1,133 @@
...
+    TemplateName: ComputeFindQuery
+    TemplateParameters:
+      Name: "OneShape"
+      Threads: 32
+      Filter: 
+        # Will find around 0.005 * 0.005 * 1e6 = 2.5 documents.
+        {$and:
+           [{a1: 1}, {a2: 1}, {^Object: {
+             withNEntries: 1,
+             havingKeys: "x0",
```
cswanson310, line 122: Can you add a comment here about how we could simplify this but want to leave it the same just to rule out the possibility of a perf impact there?

cswanson310 (approved):
I'm curious to see - could you [run a patch with profiling enabled](https://github.com/10gen/performance-tooling-docs/blob/aa35df9f7d2f0b64bac45f168f48f60a823c104f/profiling_workflow/profiling_workflow.md#how-to-trigger-profiling-for-an-evergreen-patch-run) and see if we can spot why the many different shapes are a little slower than the single shape?

I might have guessed the plan caching, but I think that with only one alternative index to use, we would not cache the plan. So there's something kinda odd going on. Not really telemetry related, but curious to see if a flame graph can identify it quickly.

src/workloads/query/TelemetryQueryShapes.yml
naama-bareket: Yeah, I saw there is a default limit of 100 for the `maxPoolSize` so I guess for certain workloads that max is reached and needs to be increased?

naama-bareket:
https://spruce.mongodb.com/version/644302d22a60ede63b141fb9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #867, opened 2023-04-19
PERF-3092 Adapt WouldChangeOwningShardBatchWrite workload to data size aware balancing

pierlauro:
:evergreen_tree: https://spruce.mongodb.com/version/64400f713627e0a930a96126/

Manually triggered a `moveRange` (atomic split+move) to distribute chunks in the way the test was originally expecting them.

### DIFF
--- a/src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
+++ b/src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
@@ -5,8 +5,7 @@ Description: |
   the shard key to trigger WouldChangeOwningShard errors.
 
   The workload consists of 3 phases:
-    1. Shard an empty collection (using ranged sharding) into two chunks and wait for the balancer
-    to distribute each chunk to its own shard.
+    1. Shard an empty collection (using ranged sharding) spreading 2 chunks across 2 shards.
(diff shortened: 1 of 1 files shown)


src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
```diff
@@ -34,20 +31,18 @@ Actors:
       OperationName: AdminCommand
       OperationCommand:
         enableSharding: *Database
+        primaryShard: "rs0"
```
jz1242, line 34: [q] Are the shardNames hard coded based on the number of shards? As in if we start a test cluster with N shards will the names be rs0, rs1, ... rsN ?

jz1242 (approved):
LGTM just 1 question

src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml
pierlauro: Yep, unfortunately there is not a pre-existing programmatic way to simply get the shard names in genny. [Here more details](https://mongodb.slack.com/archives/GSTBKTER5/p1681917907945459)

pierlauro:
evergreen merge

PR #868, opened 2023-04-20
PERF-4060: move runCommand to phase0

miguelangelnieto:
Moved the RunCommand actor to phase0 in order to eliminate the 50th latency regression.


jimoleary (approved):
lgtm

PR #873, opened 2023-04-23
EVG-17874: Update Genny with --no-activate flag.

johndaniels:
This allows us to create task generation tasks that don't immediately execute all the generated variants.

It's a bit awkward to pass through a flag that is phrased in the negative (as opposed to using an "activate" flag and variable), but in this case we need to because an explicit activate flag would have different behavior than the default of not specifying the "activate" parameter. If the "activate" parameter is present on a task, then that task is always activated, but if it is not present then it defaults to if the version is activated or through batchtime.

**Jira Ticket:** < Ticket Number >

**Whats Changed:**  
High level explanation of changes

**Patch testing results:**  

**Workload Submission form:**  

**Related PRs:**   

PLEASE READ AND REMOVE THE SECTION BELOW BEFORE SUBMITTING YOUR PR FOR REVIEW

1.  Once (1) you get approval from the STM or the Product Perf team for your change and (2) all your CI tests pass, you can initiate a merge by entering a comment on your PR with the text

    > evergreen merge

    This kicks your PR into the [evergreen commit queue][ecq]. Evergreen will automatically squash-merge your commit after checking that you have PR approval and that the CI tests all pass.

[ecq]: https://github.com/evergreen-ci/evergreen/wiki/Commit-Queue

### DIFF
--- a/src/lamplib/requirements.txt
+++ b/src/lamplib/requirements.txt
@@ -8,7 +8,7 @@ omegaconf==2.1.1
 PyYAML==5.1
 requests==2.21.0
 yamllint==1.15.0
-shrub.py==1.1.0
+shrub.py==3.0.5
 black==22.10.0
 setuptools==65.5.0
 pytest==7.2.0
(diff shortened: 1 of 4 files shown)


thessem:
It is going to take me a little bit of time to be able to review this one. There are a few things going on (unplumbing of `genny_repo_root` which doesn't seem to have been used anyway, modification of `exec_timeout_secs` in a test without any non-test code seemingly breaking? then again, CI is failing right now....) that are probably obvious with enough context

thessem (approved):

PR #876, opened 2023-04-25
PERF-3975: Add a workload that runs $telemetry w/ redaction during/under load

naama-bareket:
Here is the link to the sheets with results from running the workload a few times: https://docs.google.com/spreadsheets/d/1BAHrbqBk5vYsHdt44hg7iZjrh9GgxSlBiET4azy-X40/edit#gid=0
For the most part, it seems that running with redaction is slower than without. However there was one run in which this was not the case, and the percent differences seem to vary. I also noticed that some medians came out as negative... which I am not sure how its possible (all the min and max were positive).

----------------------------------------------------------------------------------------------------------------------------

**Jira Ticket:** PERF-3975

**Whats Changed:**  
There is a new workload which runs `$telemetry` with and without redaction. There is also a new BinData generator which returns a random `bsoncxx::types::b_binary` of the specified number of bytes.

**Patch testing results:**  
https://spruce.mongodb.com/version/6449813b2fbabe49a11e3e90/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

**Workload Submission form:**  

**Related PRs:**   
N/A


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1174,6 +1174,27 @@ class DateGenerator : public Generator<bsoncxx::types::b_date> {
...
+public:
+    BinDataGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _node{node} {}
+
+    bsoncxx::types::b_binary evaluate() override {
+        int64_t numBytes = _node["numBytes"].maybe<int64_t>().value_or(32);
+        uint8_t bytesArr[numBytes];
+        for (int i = 0; i < numBytes; i++) {
+            bytesArr[i] = rand();
+        }
```
naama-bareket, line 1188: Alternatively I can just hardcode a byte array like 
```
constexpr uint8_t localMasterKey[] = {
            0x77, 0x1f, 0x2d, 0x7d, 0x76, 0x74, 0x39, 0x08, 0x50, 0x0b, 0x61, 0x14, 0x3a, 0x07};
```
and return a Binary built from that, if that's better!

cswanson310: Hmm maybe have the generator always return the same value, since that's more what we want in our use-case, but you can generate that single value by calling a helper function 'genRandBinData()' which is basically your current 'evaluate()' method? That way someone could more easily extend/parameterize this to generate randomness in the future if they wanted.

What do you think?

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,158 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1000
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {string: {$in: ["hello", "hi", "bye"]}}
+
+# Telemetry with redaction.
```
cswanson310, line 137: Oh so it looks like here you're running the telemetry actor in its own phase?

I think the idea of the ticket was to see if we make everything else slower if we run it concurrently. So that would be something like
Phase 3: Run a bunch of concurrent finds on many threads - just maxing out load mimicing a stressed out cluster
Phase 4: Quiesce
Phase 5: Run all the same actors with all the same numbers of threads, but now add an actor running $telemetry also. How much slower does it get?

What might be interesting and hard about this though - should we repeat telemetry a lot in a loop? That's not really how we expect it to be used. We can start with a telemetry actor just repeating itself for the duration of the phase (you probably want to make it a duration in minutes rather than a "repeat" number? so they all run for the same amount of time?). It's worth checking if there's any way to get an actor to repeat N times but with some sort of sleep/delay between invokations? Like every 5 seconds say for a 1 minute trial workload?

cswanson310 (changes requested):
I am a little worried about what's going on with the median - that sounds like it's possibly a bug in my script again :\ 

You can run this on evergreen and get results from there - it might have a better/real median? That would help confirm a bug in my script. I hope to see if I can get to the bottom of it tomorrow though.

src/value_generators/src/DocumentGenerator.cpp
naama-bareket: Do you mean move what I currently have in evaluate to a `genRandBinData()` and separately have the generator return the same value? Or just have the `genRandBinData()` return a hardcoded value for now? I just worry that the name would be confusing if I am not actually returning a random BinData

cswanson310: Yes the first one! The generator's constructor can maybe call the random one once and store the result?

naama-bareket:
I updated the workload to run $telemetry concurrently with the find command! Put up a patch now since the medians are still giving funky negatives for the $telemetry actors: https://spruce.mongodb.com/version/644aec3a61837dd06f756d2f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Here is the table with the data from the first few times I ran the new workload: https://docs.google.com/spreadsheets/d/1xyAP8FJJoZKS6bHcCfkSo4n6OTCCR5WA-D426rKgyHA/edit#gid=0
It seems like there is no notable difference in the performance of the Find actors without telemetry vs. with redaction. Running everything together did make the $telemetry with redaction have a very big difference from the $telemetry without redaction though, at over 100%

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,193 @@
...
+  Threads: 1
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [4]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+- Name: FindCommandWithRedaction
```
cswanson310, line 129: Can we clarify what "with redaction" means in the name? I think you maybe just mean to say that it is expected to run in the same phase as the telemetry w/ redaction actor?

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,193 @@
...
+    OnlyActiveInPhases:
+      Active: [5]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {string: {$in: ["hello", "hi", "bye"]}}
```
cswanson310, line 143: So this will create just one query shape for us to consume in $telemetry, is that right? I think we should try to test with a bunch of different query shapes - which would probably be more realistic of the environment where we would worry about the performance of $telemetry. Could you borrow from the other workload you just did to make that happen?

I think also we'd probably want to try to vary the size of the query shapes if we could? Would it be possible to have the queries be small-ish (query on a couple fields, fits roughly on 1 or 2 lines if you have it all in json) to large-ish (hundreds of fields, ideally approaching like 1 to 10 KB in bson size?)

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,193 @@
...
+                        $telemetry: {
+                          redactIdentifiers: {^Parameter: {Name: "redactIdentifiers", Default: false}},
+                          redactionKey: *binDataGen
+                        }
+                      }
+                    ]
+              cursor: {}
+
+Actors:
+- Name: ClearCollection
```
cswanson310, line 54: I am now thinking - this will not clear the in-memory telemetry data, right? 

I don't know how we run things in evergreen in terms of a server restart or not, but it seems plausible that the $telemetry invocation here will have non-determenistic results based on what workloads ran before it. To ensure repeatable results, we'll probably want to clear it. I don't think there's any good way to do this other than setting the memory size limit to 0 and then back to what it was - which might be kinda challenging in this environment... I know there's a generic 'RunCommand' actor you could use to change the server parameter, but I don't think the YAML has a good way of remembering what the previous value of the parameter was - even if you ran getParameter to retrieve it. You could maybe skip the part about "back to what it was" since remembering the answer seems challenging to me on first thought. And just set it to be "1%" which we know is the default and seems reasonable. #performance-tooling-users might have some better ideas here.

cswanson310 (changes requested):

src/workloads/query/TelemetryWithRedaction.yml
cswanson310: Uhh you can test it out in the shell depending on the format:
```js
Object.bsonsize({x: 1})
16
```

naama-bareket: Okay thanks!

naama-bareket:
I updated the workload to set the telemetry store size and run varying query shapes and sizes. The results still align with what I previously found - no notable difference between the performance of running the find command on its own versus alongside telemetry, and a very large difference (over 100%) between running $telemetry with redaction vs without. 
https://docs.google.com/spreadsheets/d/1ejoITNdWJKsEUZZZtVSclprB9BenVF_j0cgLdSMJdKc/edit#gid=0

naama-bareket:
> I updated the workload to set the telemetry store size and run varying query shapes and sizes. The results still align with what I previously found - no notable difference between the performance of running the find command on its own versus alongside telemetry, and a very large difference (over 100%) between running $telemetry with redaction vs without. https://docs.google.com/spreadsheets/d/1ejoITNdWJKsEUZZZtVSclprB9BenVF_j0cgLdSMJdKc/edit#gid=0
> 
> running a patch with the new changes! https://spruce.mongodb.com/version/645be7b256234354a5da608a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

After seeing the naming conflict on evergreen, I updated the names and ran the workload on a fresh version of master and the results are significantly different. Running the Find actor concurrently with $telemetry with redaction has around a 4-5% difference in performance versus on its own, and running Find concurrently with $telemetry with redaction versus without also shows the same 4-5% difference. This means that running $telemetry with redaction does slow down the Find process. There is also an even larger percent difference between $telemetry with and without redaction, at around 194%. The results can be found at the bottom of the same sheet: https://docs.google.com/spreadsheets/d/1ejoITNdWJKsEUZZZtVSclprB9BenVF_j0cgLdSMJdKc/edit#gid=0

I ran another patch to confirm the naming issue is resolved! https://spruce.mongodb.com/version/645c16b03066154f43e709f0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

cswanson310:
I was shocked to learn that $telemetry without redaction was _slower_ than $telemetry with redaction, but then I remembered that we have this: https://github.com/10gen/mongo/blob/44d9b193ab7a2e86b7f63a276f0215d24c3c35d6/src/mongo/db/query/telemetry.h#L155-L158

That is actually slated to be removed. There was https://jira.mongodb.org/browse/SERVER-76526 but then Davis did it here: https://github.com/10gen/mongo/pull/12817/files#diff-09164370235e72f9b33739adfa2c4aa8e98a39bb01d1451a1dbe5d1ae94c3eceL154

That was the patch that was reverted though.. so it would be good to re-run the tests once that ticket gets merged again. I don't think we need to wait to merge this PR for that though, it would happen naturally. I'll just discount that comparison for now and try to remember later.

src/workloads/query/TelemetryWithRedaction.yml
cswanson310: Can you add some of this context as a comment near the 'setParameter' command? I don't think it will be obvious to future readers, and I also want to make it clear that 250MB was kinda arbitrary.

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,234 @@
...
+  Threads: 1
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [7]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+- Name: FindConcurrentlyWithTelemetryWithoutRedaction
```
cswanson310, line 197: Is this the same as the previous one? I think you could delete it and just have it active in multiple phases:
`Active: [4, 8]`

src/workloads/query/TelemetryWithRedaction.yml
```diff
@@ -0,0 +1,234 @@
...
+    TemplateParameters:
+      Name: "TelemetryAggregationWithoutRedaction"
+      applyHmacToIdentifiers: false
+      OnlyActiveInPhase: 8
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-all-feature-flags
```
cswanson310, line 233: Can you add regular 'standalone' here too?

cswanson310: Oh actually nvm - that won't work, right? Can't run $telemetry without the feature flag.

cswanson310: So we should explicitly compare the two you list here in the perf comparison report - the 'all-feature-flags' will have the rate limit set to 0 and then we can see the impact of collecting telemetry also.

cswanson310 (approved):
This looks good! I had a suggestion to simplify the YAML a bit, and then a request for a comment or two.

src/workloads/query/TelemetryWithRedaction.yml
naama-bareket: They are the same, but combining them into 1 actor would also combine the results right? I thought we wanted to look at how `$telemetry` with and without redaction affects the performance separately? If we don't care about that then I can move them into one actor with both phases active!

src/workloads/query/TelemetryWithRedaction.yml
naama-bareket: That sounds good! Should I add it to `scripts/perf_reports_query_telemetry_perf.yml` as part of this ticket?

src/workloads/query/TelemetryWithRedaction.yml
cswanson310: Oh that's a great point! I didn't realize/remember this. I think good to keep them separate then.

src/workloads/query/TelemetryWithRedaction.yml
cswanson310: Yes please! I can talk you through how to test it out. Let's merge this first though and give it a chance to run so we have some data to test out

naama-bareket:
evergreen merge

PR #880, opened 2023-05-04
PERF-4117 Exclude $percentile workloads from v6.1 and v6.3 in a correct way

IrinaYatsenko:


src/workloads/query/PercentilesAgg.yml
```diff
@@ -22,7 +22,7 @@ AutoRun:
       - standalone-sbe
     branch_name:
       $gte:
-      - v7.0
+        v7.0
```
jimoleary, line 25: [nit] For clarity, I would consider putting the equality matches on one line ie. ```$gte: v7.0``` but functionally your version is the same.

jimoleary (approved):
lgtm with an ignorable nit

jimoleary:
evergreen merge

jimoleary:
@IrinaYatsenko I've merged this PR as the previous commit causes system failures on all the sys-perf genny tasks. I've also raised EVG-19788 to have the commit queue catch this sort of problem going forward.

PR #881, opened 2023-05-04
Make the example URLs for toolchain build ID more newbie-friendly

PyGeek03:
When I was looking at this file to test my changes to [10gen/vcpkg/evergreen.yml](https://github.com/10gen/vcpkg/blob/master/evergreen.yml), I got confused for a while about how to get the toolchain build ID from the genny-toolchain Evergreen task because 
1) I'm using the new UI while the current example URL is for the old UI and 
2) My change is a patch so the build ID has "patch_" in front of it.

This change is to address these confusions.

### DIFF
--- a/src/lamplib/src/genny/toolchain.py
+++ b/src/lamplib/src/genny/toolchain.py
@@ -183,14 +183,16 @@ def toolchain_info(
 
 
 class ToolchainDownloader(Downloader):
-    # These build IDs are from the genny-toolchain Evergreen task.
+    # These build IDs are from the genny-toolchain Evergreen task (which pulls from 10gen/vcpkg repo)
     # Old UI: https://evergreen.mongodb.com/waterfall/genny-toolchain
     # New UI: https://spruce.mongodb.com/commits/genny-toolchain
     #
-    # Find a compile task (for any build variant) and modify the URL. 
-    # For example:
+    # Find a compile task (for any build variant) and get TOOLCHAIN_BUILD_ID from URL
+    # https://spruce.mongodb.com/task/genny_toolchain_<$build-variant>_t_compile_<$TOOLCHAIN_BUILD_ID>
+    # Example for merged PR build on archlinux variant:
     # https://spruce.mongodb.com/task/genny_toolchain_archlinux_t_compile_82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
     # =>                                                                  82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
+    # Example for patch build on macos_1100_arm64 variant:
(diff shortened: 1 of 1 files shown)


johndaniels (approved):

src/lamplib/src/genny/toolchain.py
```diff
@@ -184,13 +184,18 @@ def toolchain_info(
...
-    # =>                                  82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
+    # Old UI: https://evergreen.mongodb.com/waterfall/genny-toolchain
+    # New UI: https://spruce.mongodb.com/commits/genny-toolchain
+    #
+    # Find a compile task (for any build variant) and modify the URL. 
+    # For example:
+    # https://spruce.mongodb.com/task/genny_toolchain_archlinux_t_compile_82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
+    # =>                                                                  82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21
+    # https://spruce.mongodb.com/task/genny_toolchain_macos_1100_arm64_t_compile_patch_87457e6fec1d98f270c84d915f83bec53554ecee_6451d23fc9ec4441c9ce233d_23_05_03_03_17_20
+    # =>                                                                         patch_87457e6fec1d98f270c84d915f83bec53554ecee_6451d23fc9ec4441c9ce233d_23_05_03_03_17_20
```
thessem, line 195: [potentially bike-shedding so this is being left as a comment. Given that John has approved this, this feedback is not blocking!]

The original comment is describing how this function works normally. The added comment (I think) is introducing a new URL scheme (`patch_*`) without explaining that this is the URL scheme for a PR build or why using that would be useful. I am not sure this is adding clarity.

The URLs are appreciated. Could you please modify the comment to indicate that `82eb7c32ad09726f3ef0ddc8d7f24a18b03d9644_21_11_23_16_37_21` is the normal scheme, and `patch_87457e6fec1d98f270c84d915f83bec53554ecee_6451d23fc9ec4441c9ce233d_23_05_03_03_17_20` is what a patch build will look like? Perhaps having multiple forms of ID in the comment can be made clearer by introducing `https://spruce.mongodb.com/task/genny_toolchain_<$DISTRO>_t_compile_<$ID>` and giving examples of $DISTRO and $ID.

thessem (commented):
Just leaving a comment given the PR was reviewed while I was typing!

src/lamplib/src/genny/toolchain.py
PyGeek03: Yes I've added these details, thank you!

thessem (approved):

PyGeek03:
evergreen merge

PR #883, opened 2023-05-04
PERF-4119 Prevent ContentionTTLDeletions.yml from running on branches pre-7.0

pvselvan:
Since execution control and respective server parameters don't exist prior to 7.0, we should prevent this workload from running on older branches.


pvselvan:
evergreen merge

PR #886, opened 2023-05-05
PERF-4123 run match_filters in atlas and repl variants

ksuarz:
Currently, `match_filters` only runs in standalones. We care about these tests and it should run in more variants that actually reflect a release configuration.


jimoleary (approved):
lgtm

PR #891, opened 2023-05-11
PERF-4132 Change TimeSeriesSortCompound.yml to use bounded sort via hint

mattBoros:
This was supposed to use a bounded sort, not a blocking sort. See ticket description for more info

In the [design doc](https://docs.google.com/document/d/1d1artozDJUU1WmwHfcViV2BPIaDjns-xcylvPDuHiiI/edit#bookmark=id.edo8gitvzpap) you can see the third row of the example queries shows this query needs a hint to trigger a bounded sort.

[patch](https://spruce.mongodb.com/version/645d317f0ae606ff137a0cc2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/query/TimeSeriesSortCompound.yml
+++ b/src/workloads/query/TimeSeriesSortCompound.yml
@@ -88,8 +88,7 @@ Actors:
           {$skip: 1e10},
         ]
         cursor: {}
-        allowDiskUse: true
-
+        hint: {m: 1, t: 1}
   - *Nop
 
 AutoRun:


src/workloads/query/TimeSeriesSortCompound.yml
```diff
@@ -88,8 +88,7 @@ Actors:
           {$skip: 1e10},
         ]
         cursor: {}
-        allowDiskUse: true
-
+        hint: {m: 1, t: 1}
```
nzolnierzmdb, line 91: And we don't need to use disk for bounded sort?

nzolnierzmdb (approved):

src/workloads/query/TimeSeriesSortCompound.yml
mattBoros: It's possible to make bounded sort spill, but this test won't do it. BoundedSort is like a sorted merge, so the memory it uses is minimal, correlated with how many buckets it needs to have open. With the way the buckets are laid out I don't think it has enough buckets open to spill.

mattBoros:
evergreen merge

PR #892, opened 2023-05-12
PERF-4116 Add genny workloads to test latency and throughput of multi shard transactions

Fefer-Ivan:
**Jira Ticket:** PERF-4116

**Whats Changed:**  
Add genny workloads to test latency and throughput of multi shard transactions

**Patch testing results:**  
https://spruce.mongodb.com/version/645e4f2b850e6121c27d1d3e/tasks
https://spruce.mongodb.com/version/6461fbff9ccd4e2c6f1639e8/tasks

### DIFF
--- a/src/workloads/sharding/MultiShardTransactions.yml
+++ b/src/workloads/sharding/MultiShardTransactions.yml
@@ -6,7 +6,6 @@ GlobalDefaults:
   Database0: &Database0 databaseForMultiShardTestingIndex0
   Database1: &Database1 databaseForMultiShardTestingIndex1
 
-
   # Note that the exact document size may exceed ApproxDocumentSize because of field names and other
   # fields in the document.
   ApproxDocumentSize: &ApproxDocumentSize 2000  # = 2kB
@@ -14,8 +13,6 @@ GlobalDefaults:
   DocumentCount: &DocumentCount 100_000
   NumChunks: &NumChunks 240
 
-  ShardKeyValueMin: &ShardKeyValueMin 1
-  ShardKeyValueMax: &ShardKeyValueMax 24_000
   MaxPhase: &MaxPhase 6
 
 Actors:
@@ -39,12 +36,12 @@ Actors:
         - OperationName: AdminCommand
           OperationCommand:
             shardCollection: databaseForMultiShardTestingIndex0.Collection0
-            key: {shardKey: hashed}
+            key: {_id: hashed}
             numInitialChunks: *NumChunks
         - OperationName: AdminCommand
           OperationCommand:
             shardCollection: databaseForMultiShardTestingIndex1.Collection0
-            key: {shardKey: hashed}
+            key: {_id: hashed}
             numInitialChunks: *NumChunks
         - OperationName: AdminCommand
           OperationCommand:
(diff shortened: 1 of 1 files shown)


src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -0,0 +1,343 @@
...
+    - *Nop
+    - MetricsName: ShardedCollectionUpdate
+      Duration: 1 minute
+      Collection: Collection0
+      ThrowOnFailure: false  # Transactions are allowed to fail.
+      RecordFailure: true  # We still want to record the metrics even if the transaction is rolled back.
+      Operations:
+      - OperationName: withTransaction
+        OperationCommand:
+          OperationsInTransaction:
```
cheahuychou, line 97: [question] I noticed that we runs 11 update operations in each transaction in this workload. I have some questions:
1. Why don't we make each transaction just run the minimum number of operations it needs to do. For example, for a transaction to "targets two non-sharded collections on different primary shards", it just needs to do an update against `databaseForMultiShardTestingIndex0.Collection1` and an update against `databaseForMultiShardTestingIndex1.Collection1`. 
2. If we want to test coverage for large transaction, we could instead make each update touch multiple documents (e.g. similar to what this workload [does](https://github.com/mongodb/genny/blob/b84fde404e63763634497cacca5f662ed4db6a25/src/workloads/scale/UpdateMillionDocsInTxn.yml#L76)).

src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -0,0 +1,343 @@
...
+              Update: {$inc: {counter: 1}}
+          - OperationName: updateOne
+            OperationCommand:
+              Filter: *RandomIdFilter
+              Update: {$inc: {counter: 1}}
+          - OperationName: updateOne
+            OperationCommand:
+              Filter: *RandomIdFilter
+              Update: {$inc: {counter: 1}}
+    - MetricsName: ShardedAndNonShardedCollectionsUpdate
```
cheahuychou, line 142: Taken from the ticket description:
> Tests should include:
> 1. Single database transaction that targets two non-primary shards of a sharded collection.
> 2. Single database transaction that targets a non-primary shard of a sharded collection and a non-sharded collection on primary shard.
> 3. Multi-database transaction that targets two non-sharded collections on different primary shards.
> 4. Multi-database transaction that targets two non-primary shards of two sharded collections collections.

[question] Could you explain why we want to explicitly test the performance of a transaction that targets non-primary shards? What's the significance of not being the primary shard when it comes to change-stream?

Assuming that we still want to test the 4 categories above, I believe the current version only covers the third category correctly (under `DifferentDatabasesNonShardedCollectionsUpdate`). The reason is that with hashed sharding, an updater against a random collection can end up targeting any shard that owns data for the collection (i.e. primary or non-primary shard).

src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -0,0 +1,343 @@
...
+GlobalDefaults:
+  Database0: &Database0 databaseForMultiShardTestingIndex0
+  Database1: &Database1 databaseForMultiShardTestingIndex1
+
+  # Note that the exact document size may exceed ApproxDocumentSize because of field names and other
+  # fields in the document.
+  ApproxDocumentSize: &ApproxDocumentSize 2000  # = 2kB
+  ApproxDocumentSize50Pct: &ApproxDocumentSize50Pct 1000  # = 1kB
+  DocumentCount: &DocumentCount 100_000
+  NumChunks: &NumChunks 240
```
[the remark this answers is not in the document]
Fefer-Ivan: The main reason is copy-paste from another workload. 
I will check if distribution is equal and remove it if it is not needed.

src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -0,0 +1,343 @@
...
+    Threads: 4
+    BatchSize: 1000
+    DocumentCount: *DocumentCount
+    CollectionCount: 2
+    Database: *Database1
+    Document: *Document
+  - *Nop
+  - *Nop
+  - *Nop
+  - *Nop
```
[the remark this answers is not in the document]
Fefer-Ivan: The status of cluster is exactly how you described.

Primary shards are ensured by AdminCommand movePrimary on lines 46-53.

src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -0,0 +1,343 @@
...
+              Update: {$inc: {counter: 1}}
+          - OperationName: updateOne
+            OperationCommand:
+              Filter: *RandomIdFilter
+              Update: {$inc: {counter: 1}}
+    - MetricsName: ShardedAndNonShardedCollectionsUpdate
+      Duration: 1 minute
+      ThrowOnFailure: false  # Transactions are allowed to fail.
+      RecordFailure: true  # We still want to record the metrics even if the transaction is rolled back.
+      Collection: Collection0
```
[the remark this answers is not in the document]
Fefer-Ivan: It is a required field.

src/workloads/sharding/MultiShardTransactions.yml
Fefer-Ivan: 11 was a mistake. I was going for only 10.

I did it to make sure that most of the operations actually affect multiple shards on sharded collections.

The alternative would be to have not-hashed shard key and query documents from first half and last half of the _id range.
But I don't know how to ensure correct data distribution in that case.

src/workloads/sharding/MultiShardTransactions.yml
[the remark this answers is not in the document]
Fefer-Ivan: Yes, it works without big numInitialChunks. Removed it.
Thanks.

src/workloads/sharding/MultiShardTransactions.yml
cheahuychou: > The alternative would be to have not-hashed shard key and query documents from first half and last half of the _id range.
But I don't know how to ensure correct data distribution in that case.

One way is to specify some other field (not _id) as the shard key field, and do manually data distribution by specifying the range for that field (similar to [this](https://github.com/mongodb/genny/blob/48b3fac2e0c8c29b5f196e1263acc8d89e3762ed/src/workloads/sharding/ReshardCollection.yml#L95-L96)) and splitting and moving the chunks accordingly (similar to [this](https://github.com/mongodb/genny/blob/a061a91d04e07bd6da68336e6342ca8187809a6a/src/workloads/sharding/WouldChangeOwningShardBatchWrite.yml#L40-L51)).

src/workloads/sharding/MultiShardTransactions.yml
cheahuychou: I have read the design and the thread that you had with Andy, Max and Rachita in the sharding channel.  I see how you want to make the perf workload run transactions that target only non-primary shards, specifically because the current design intends to make the coordinator additionally send a writeEndOfTransaction command to the primary shard so you want to measure the performance impact of engaging another shard. I spoke to Max about this. We both think it isn't very important to make the transactions not have the primary shard as a participant since we just need to measure is the extra cost of going to the primary shard to write down the list of namespaces. Given this, we probably should just make the workload run any cross-shard transactions, i.e. not care about the participants (keep the current hashed sharding setup).

cheahuychou (approved):
LGTM

PR #895, opened 2023-05-15
PERF-3954 Re-enable QE performance tests

erwee:
test patch (using genny-toolchain with updated libmongocrypt): https://spruce.mongodb.com/version/64629292850e6173bd8bc3ca/tasks

### DIFF
--- a/src/lamplib/src/genny/toolchain.py
+++ b/src/lamplib/src/genny/toolchain.py
@@ -198,7 +198,7 @@ class ToolchainDownloader(Downloader):
     # If we were 💅 we could do the string logic here in python, but we're not that fancy.
     #
 
-    TOOLCHAIN_BUILD_ID = "patch_77d7b8df6042f53ef392097cde191b2bfef252cc_64628080562343a50196ef6f_23_05_15_18_57_15"
+    TOOLCHAIN_BUILD_ID = "c0f42baafd4845b1f77e0cc19875eba3660d64ba_23_05_17_14_21_03"
     TOOLCHAIN_GIT_HASH = TOOLCHAIN_BUILD_ID.split("_")[0]
 
     def __init__(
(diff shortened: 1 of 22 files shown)


src/lamplib/src/genny/toolchain.py
```diff
@@ -198,7 +198,7 @@ class ToolchainDownloader(Downloader):
     # If we were 💅 we could do the string logic here in python, but we're not that fancy.
     #
 
-    TOOLCHAIN_BUILD_ID = "77d7b8df6042f53ef392097cde191b2bfef252cc_23_05_05_02_21_38"
+    TOOLCHAIN_BUILD_ID = "patch_77d7b8df6042f53ef392097cde191b2bfef252cc_64628080562343a50196ef6f_23_05_15_18_57_15"
```
erwee, line 201: NOTE: this is only temporary, and will be replaced by a proper build ID once the vcpkg patch with the libmongocrypt update lands.

SimonEismann (commented):
LGTM, copying over your comments from the corresponding [vcpkg PR](https://github.com/10gen/vcpkg/pull/26) to give some context to anyone stumbling across this PR:

```
This patch updates the libmongocrypt to version 1.8.0, which implements version 2 of the queryable encryption protocol. This update is necessary in order to enable the QE performance tests to run against the server version 7.0+, where QEv2 must be used.

No updates to the C or CXX drivers are necessary since they will automatically start using the v2 protocol once this update is made. Therefore, there should be no risk of regression to current performance workloads that don't use encryption.
```

We can probably wait to merge this until we have a waterfall build of the updated toolchain, right?

erwee:
> We can probably wait to merge this until we have a waterfall build of the updated toolchain, right?

Yep, that's the plan.

SimonEismann (approved):
LGTM

PR #900, opened 2023-05-18
Add Evergreen build for Ubuntu 22.04

PyGeek03:
Now that Genny supports Ubuntu 22.04, and that Ubuntu 22.04 is the recommended distro for Evergreen workstations, we should add the Evergreen build for this distro as well. Given how we don't run Genny patches that often, this should not bump infra cost much.


johndaniels (approved):

johndaniels (approved):

PyGeek03:
evergreen merge

PR #903, opened 2023-05-19
EVG-18380 - Add instructions for running resmoke-test locally

PyGeek03:
Currently, the code to implement resmoke-test assumes that it will be run by Evergreen. Therefore, to run resmoke-test locally, some modifications must be made. This PR is to add the commands to make these modifications.

I've thought about creating a script to automatically run these commands, but the root cause of why we have to use these commands is fundamentally a bug in Genny ([EVG-18189](https://jira.mongodb.org/browse/EVG-18189)), so that script would just be papering over this bug.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -93,6 +93,9 @@ These commands assume that `git status` last output line is
 `nothing to commit, working tree clean`, and that the current working directory is
 `genny/` i.e. the root of the repo.
 
+Note: If you make any changes after having run `./run-genny install -d {your distro}`,
+please run that command again to compile your changes.
+
 #### MacOS:
 1. `mkdir -p data/db && sed -i '' '/mongos,/s!mongos,!mongos, "--dbpathPrefix", "data/db",!' src/lamplib/src/genny/tasks/run_tests.py`
 2. `sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/{YML file you want to use}`
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -100,7 +100,11 @@ These commands assume that `git status` last output line is
    # Example:
    sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/genny_sharded.yml
    ```
-3. If using genny_create_new_actor.yml: `./run-genny create-new-actor HelloWorldActor && sed -i '' '/REQUIRE(count == 101./s/101/100/' ./src/cast_core/test/HelloWorldActor_test.cpp`
+3. If using genny_create_new_actor.yml: `./run-genny create-new-actor HelloWorldActor`. Fix the bug in the generated code, which is left as an exercise to ensure the reader understand the code. 
+	<details>
+     <summary>But if you're short on time, click here for the cheat code</summary>
+     <pre><code>sed -i '' '/REQUIRE(count == 101./s/101/100/' ./src/cast_core/test/HelloWorldActor_test.cpp</code></pre>
+	</details>
 4. `./run-genny install -d not-linux`
 5. `./run-genny cmake-test` (Ignore the failed tests related to date & time, they're known to be flaky on MacOS ([EVG-19776](https://jira.mongodb.org/browse/EVG-19776)))
 6. `./run-genny resmoke-test --suites {full path to YML file}`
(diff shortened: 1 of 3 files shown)


docs/developing.md
```diff
@@ -77,19 +77,56 @@ The Actor tests use resmoke to set up a real MongoDB cluster and execute
...
+Currently, the code to implement resmoke-test assumes that it will be run by
+Evergreen. Therefore, to run resmoke-test locally, some modifications must be made.
+The commands to make these modifications on MacOS and on Linux are detailed below.
+
+These commands assume that `git status` last output line is
+`nothing to commit, working tree clean`, and that the current working directory is
+`genny/` i.e. the root of the repo.
+
+#### MacOS:
+1. `mkdir -p data/db && sed -i '' '/mongos,/s/mongos,/mongos, "--dbpathPrefix", "data\/db"/' src/lamplib/src/genny/tasks/run_tests.py`
```
lgfang, line 97: 1. there is an extra pair of single quotes (`''`) should be deleted?
2. I'd suggest, similar to the next line, using `sed -i '/xxx/s!from!to!' so no need of escaping slash(es) in path
3. the "to" does not contain an trailing ",". I'd suggest adding that so that it won't fail even if more options are appended in the future.

In summary I think this command should be like this
`mkdir -p data/db && sed -i '' '/mongos,/s!mongos,!mongos, "--dbpathPrefix", "data/db", !' src/lamplib/src/genny/tasks/run_tests.py`

docs/developing.md
```diff
@@ -77,19 +77,56 @@ The Actor tests use resmoke to set up a real MongoDB cluster and execute
...
+2. `sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/{YML file you want to use}`
+```sh
+# Example:
+sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/genny_sharded.yml
+```
+3. If using genny_create_new_actor.yml: `./run-genny create-new-actor HelloWorldActor && sed -i '' '/REQUIRE(count == 101./s/101/100/' ./src/cast_core/test/HelloWorldActor_test.cpp`
```
lgfang, line 103: I understand that it is supposed to be an exercise for users to find out this error. So I'd suggest remove line. Or, make it clear that you'll encounter an error, go fix it.

docs/developing.md
PyGeek03: MacOS's BSD implementation of `sed` is different from GNU/Linux's, hence the necessity for the extra pair of single quotes on MacOS. Took me a while to figure out why the Linux version of the command wasn't working on my Mac: https://stackoverflow.com/a/4247319

I'm a complete sed noob, so thank you for the suggested command! I've added it with the trailing space after the comma removed.

docs/developing.md
PyGeek03: Good idea, I've added a spoiler alert for this. Thanks!

docs/developing.md
```diff
@@ -77,19 +77,67 @@ The Actor tests use resmoke to set up a real MongoDB cluster and execute
...
+
+These commands assume that `git status` last output line is
+`nothing to commit, working tree clean`, and that the current working directory is
+`genny/` i.e. the root of the repo.
+
+Note: If you make any changes after having run `./run-genny install -d {your distro}`,
+please run that command again to compile your changes.
+
+#### MacOS:
+1. `mkdir -p data/db && sed -i '' '/mongos,/s!mongos,!mongos, "--dbpathPrefix", "data/db",!' src/lamplib/src/genny/tasks/run_tests.py`
```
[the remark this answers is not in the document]
PyGeek03: Yeah I know, that's why I put it under the section for MacOS.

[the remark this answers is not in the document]
PyGeek03: The commands without `''` doesn't work on MacOS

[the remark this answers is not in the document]
PyGeek03: (Cross-platform support is such a pain...)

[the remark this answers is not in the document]
lgfang: ah, my apologies. I should have seen the title "macos".

docs/developing.md
```diff
@@ -77,19 +77,67 @@ The Actor tests use resmoke to set up a real MongoDB cluster and execute
...
+2. `sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/{YML file you want to use}`
+   ```sh
+   # Example:
+   sed -i '' '/program_executable/s!:.*$!: ../../build/src/cast_core/cast_core_test!' src/resmokeconfig/genny_sharded.yml
+   ```
+3. If using genny_create_new_actor.yml: `./run-genny create-new-actor SelfTestActor`. Fix the bug in the generated code, which is left as an exercise to ensure the reader understand the code.
```
lgfang, line 106: This seems to suggest that for other tests there is no error in the generated code. But actually all have the same error.

On the other hand, according to my test  genny_create_new_actor.yml does require the actor named `SelfTestActor` while `genny_single_node_replset.yml` does not care it. 

Therefore, for simplicity, I'd suggest remove "If using genny_create_new_actor.yml:", just tell people to create the "SelfTestActor"

PyGeek03: My understanding is that only `genny_create_new_actor.yml` requires a new actor to be generated, the other YML files don't require any. I reckon it's important to document that fact in some way.

lgfang:
When create_new_actor is run, the screen output also gives (in-accurate) instructions about how to run genny smoke tests. I think we need to update it as well. At least, point the users to the updated doc.

https://github.com/mongodb/genny/blob/master/src/lamplib/src/genny/tasks/create-new-actor.sh#L622

PyGeek03:
> When create_new_actor is run, the screen output also gives (in-accurate) instructions about how to run genny smoke tests. I think we need to update it as well. At least, point the users to the updated doc.
> 
> [`master`/src/lamplib/src/genny/tasks/create-new-actor.sh#L622](https://github.com/mongodb/genny/blob/master/src/lamplib/src/genny/tasks/create-new-actor.sh?rgh-link-date=2023-05-19T05%3A50%3A15Z#L622)

Yup I've fixed these instructions as well

lgfang (approved):

PyGeek03:
evergreen merge

PR #907, opened 2023-05-23
EVG-19731 - Redact password in MongoDB URI logging

PyGeek03:
**Jira Ticket:** EVG-19731

**Whats Changed:** Redact password in MongoDB URI logging.

Patch Run:
https://spruce.mongodb.com/task/sys_perf_atlas_M60_atlas_repl_big_update_patch_cd1cfc52029ccfba3214f22b8dcb5b09848df5f3_646f29c49ccd4e17993b29c9_23_05_25_09_29_05/logs?execution=1&sortBy=STATUS&sortDir=ASC

Redacted log line outputted by Genny:
https://parsley.mongodb.com/evergreen/sys_perf_atlas_M60_atlas_repl_big_update_patch_cd1cfc52029ccfba3214f22b8dcb5b09848df5f3_646f29c49ccd4e17993b29c9_23_05_25_09_29_05/1/task?bookmarks=0,2488,4678&shareLine=2488

**Related PRs:** https://github.com/10gen/dsi/pull/1286

### DIFF
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -239,7 +239,12 @@ mongocxx::options::pool PoolFactory::makeOptions() const {
 
 std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
     auto uriStr = makeUri();
-    BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << uriStr << "'";
+
+    std::string password{*_config->get(OptionType::kAccessOption, "Password")};
+    _config->set(OptionType::kAccessOption, "Password", "[REDACTED]");
+    auto redactedUriStr = makeUri();
+    BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << redactedUriStr << "'";
+    _config->set(OptionType::kAccessOption, "Password", password);
 
     auto uri = mongocxx::uri{uriStr};
 
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -241,10 +241,9 @@ std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
     auto uriStr = makeUri();
 
     std::string password{*_config->get(OptionType::kAccessOption, "Password")};
-    _config->set(OptionType::kAccessOption, "Password", "[REDACTED]");
-    auto redactedUriStr = makeUri();
+
+    auto redactedUriStr = std::regex_replace(uriStr, std::regex(password), "[REDACTED]" );
     BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << redactedUriStr << "'";
-    _config->set(OptionType::kAccessOption, "Password", password);
 
     auto uri = mongocxx::uri{uriStr};
 
--- a/src/gennylib/test/PoolFactory_test.cpp
+++ b/src/gennylib/test/PoolFactory_test.cpp
@@ -237,16 +237,6 @@ TEST_CASE("PoolFactory behavior") {
 
         auto pool = factory.makePool();
         REQUIRE(pool);
-
-        // We should be able to change the value of an option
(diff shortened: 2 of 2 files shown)


src/gennylib/src/PoolFactory.cpp
```diff
@@ -239,7 +239,12 @@ mongocxx::options::pool PoolFactory::makeOptions() const {
 
 std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
     auto uriStr = makeUri();
-    BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << uriStr << "'";
+
+    std::string password{*_config->get(OptionType::kAccessOption, "Password")};
+    _config->set(OptionType::kAccessOption, "Password", "[REDACTED]");
+    auto redactedUriStr = makeUri();
```
thessem, line 245: Given that we have access to the password (`password`) wouldn't it be more efficient to substitute the string directly, rather than constructing the DSN twice?

src/gennylib/test/PoolFactory_test.cpp
```diff
@@ -237,6 +237,16 @@ TEST_CASE("PoolFactory behavior") {
 
         auto pool = factory.makePool();
         REQUIRE(pool);
+
+        // We should be able to change the value of an option
+        auto expectedRedactedUri = [&]() {
```
thessem, line 242: Given my other comment, don't update this test, I think this is a great way of testing it.

thessem (changes requested):

src/gennylib/src/PoolFactory.cpp
PyGeek03: I think either way, we would need 2 copies of the DSN (1 actual and 1 redacted), unless we want to mutate a single copy of DSN twice (from actual to redacted then back to actual). Also, an efficient mutation would require substituting the password with a string of the same length (hence leaking the length of the password to the log), unless we want to shuffle all the characters after the password forward and then backward.

thessem: Agreed, I just figured copying the string after generation was simpler (in terms of raw CPU and mental overhead) than generating it twice. 

In this case, I think we need to confirm that this function (`makePool`) does not ever run concurrently with any other function in this class on the same instantiation, given it is modifying `_config`. Trying to run `makePool` from two different threads, for example, would possibly generate a result where one of the pools is trying to use `Redacted` as its password.

I'd have to properly set up my IDE for this project (I am doing so right now) but I suspect that `_config` is actually a pointer to a global object, so modifying `Password` is modifying it for all users. This'd make it risky if two pools are being generated at the same time.

It's been a while since I've actually coded C++ so I might have the wrong end of the stick here, but copying a stack-local string and substituting it seems less prone-to-failure?

PyGeek03: Ah I see what you mean now. Yes creating the redacted copy from the actual DSN would be best then. I'll do this, thank you for the suggestion!

src/gennylib/test/PoolFactory_test.cpp
PyGeek03: As I'm gonna be creating the redacted DSN copy from the actual DSN instead of calling makeUri(), do you think I should keep my changes in this file or shouldn't?

thessem: Leave this how it is. I think testing using a different method from generating it originally is a great approach. Imagine if you're using `strings.sub(password, "REDACTED")` in the main function AND the test function, and for some reason `strings.sub` breaks and does nothing, the test wont pick it up.

What this test is currently stating is "the redacted DSN should look exactly as if we generated a DSN with `redacted` as a password, which is bang on.

PyGeek03: We've discussed this offline, and we agree that it's not worth setting up mocking just to test the log. The patch build shows that the code works anyway.

thessem (approved):
Thanks for this!

PyGeek03:
evergreen merge

PR #909, opened 2023-05-23
PERF-4152: Add new genny workloads for sharded $lookup where the local and f…

HanaPearlman:
…oreign data are co-located

**Jira Ticket:** [PERF-4152](https://jira.mongodb.org/browse/PERF-4152)

**Whats Changed:**  
This new workload exercises the behavior of $lookup when the local and foreign collection data is co-located. In the queries, for each local document on each shard, the $lookup subpipeline is targeted to the same shard. The queries include pipelines with let/pipeline syntax, pipelines with local/foreign field syntax, pipelines with different $limits, and pipelines run against sharded and unsharded colls.

**Workload Submission form:**  
Submitted! Submission [here](https://docs.google.com/forms/d/e/1FAIpQLSf1oh23Ddo1khjLZMqZ9DHbRdObnpeH20PSXTBuXVg-7mxxTA/viewform?edit2=2_ABaOnueF9oSrnorDeTQxFIjOEWZFXR73y7dNXcwjxf0PK8vkvg4N3Ix_631uJKF_WJcFLdI)

**Results and explanation**  
Results: [here](https://spruce.mongodb.com/task/sys_perf_linux_shard_lite.2022_11_lookup_colocated_data_patch_dc864de3212e83b2d2a7553b6d9a975c523305e6_646df918c9ec44795509dd32_23_05_24_11_49_21/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) (against master) and [here](https://spruce.mongodb.com/task/sys_perf_linux_shard_lite.2022_11_lookup_colocated_data_patch_dc864de3212e83b2d2a7553b6d9a975c523305e6_646dfa8561837d7089c11251_23_05_24_11_53_16/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) (against master with local read disabled). DSI modified in both cases to forceClassic 

New results with $limit [here](https://spruce.mongodb.com/task/sys_perf_linux_shard_lite.2022_11_lookup_colocated_data_patch_3aa51a4623b5869c6f57745ad1371aacf02c8d34_6483869c7742aef18d13c9e6_23_06_09_20_09_42/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC).

Config: unsharded (coll unsharded), sharded (coll sharded across multiple chunks/shards), sharded one chunk (coll sharded with a single chunk which exists on one shard)

local/foreign field syntax, sorted from lowest to highest latency

| Config | Batch size | Avg latency (sec)  |
|---|--|---|
| Unsharded | 1k | .125 |
| Sharded | 1k | .470 |
| Sharded one chunk | 1k | .478 |
| Unsharded | 5k | .550 |
| Unsharded | 15k | 1.60 |
| Sharded |  5k | 2.26 |
| Sharded one chunk |  5k | 2.32 |
| Sharded |  15k |  3.60 |
| Sharded one chunk |  15k | 6.92 |

let/pipeline syntax

| Config | Batch size | Avg latency (sec)  |
|---|--|---|
| Unsharded | 1k | .135 |
| Sharded one chunk | 1k | .554 |
| Sharded | 1k | .563 |
| Unsharded | 5k | .594 |
| Unsharded | 15k | 1.75 |
| Sharded one chunk |  5k | 2.70 |
| Sharded |  5k | 2.71 |
| Sharded |  15k |  4.18 |
| Sharded one chunk |  15k | 8.02 |

Things to note: 
- As expected from help ticket, sharded $lookup is slower than unsharded $lookup 
-  As mentioned in the help ticket, $lookup is sometimes faster when all the data is on a single shard. Now that I'm using $limit in the queries instead of batch size, the evergreen results no longer show this (though I do see it locally). Not sure why :/
- When I manually remove the local read path, unsharded $lookup regresses nearly to the "Sharded one chunk" numbers.

### DIFF
--- a/src/workloads/query/LookupColocatedData.yml
+++ b/src/workloads/query/LookupColocatedData.yml
@@ -1,0 +1,367 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This test exercises the behavior of $lookup when the local and foreign collection data is
+  co-located. In the queries below, for each local document on each shard, the $lookup subpipeline
+  is targeted to the same shard.
+  The workload consists of the following steps:
+    A. Creating empty collections, some unsharded and some sharded.
+    B. Populating collections with data.
+    C. Fsync.
+    D. Running $lookups. This includes pipelines using let/pipeline syntax, pipelines with
+       local/foreign field syntax, pipelines with different batch sizes, and pipelines run against
+       sharded and unsharded colls.
+GlobalDefaults:
+  UnshardedLocal: &UnshardedLocal UnshardedLocal
+  ShardedLocal: &ShardedLocal ShardedLocal
+  ShardedLocalAllOnOneShard: &ShardedLocalAllOnOneShard ShardedLocalAllOnOneShard
+  NumLocalDocs: &NumLocalDocs 15000
+  LocalShardKey: &LocalShardKey localField
+  BatchSize1k: &BatchSize1k 1000
+  BatchSize5k: &BatchSize5k 5000
+  BatchSize15k: &BatchSize15k 15000 
+  Database: &Database test
+  NumPhases: &NumPhases 23
+
+ActorTemplates:
+# Template for populating the sharded and unsharded collections. 
+- TemplateName: LoadDocuments
+  Config:
+    Name: LoadUnshardedInitialData
+    Type: CrudActor
+    Database: *Database
+    Threads: 1
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1}}]
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Repeat: *NumLocalDocs
+          Threads: 1
+          Collection: {^Parameter: {Name: "Coll", Default: "localColl"}}
+          Operations:
+          - OperationName: insertOne
+            OperationCommand:
+              Document:
+                *LocalShardKey: {^RandomString: {length: 10}}
+                str: {^RandomString: {length: 100}}
+                smallInt: {^RandomInt: {min: 1, max: 10}}
+                mediumInt: {^RandomInt: {min: 1, max: 100}}
+
+# Template for running a $lookup using local/foreign field syntax. Note that the local and foreign
+# collections are the same. This is how we can easily enforce that the local and foreign data
+# will always be co-located.
+- TemplateName: SelfLookupLocalFieldForeignField
+  Config:
+    Name: RunLookup
+    Type: RunCommand
+    Database: *Database
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1}}]
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Repeat: 10
+          Database: *Database
+          Operations:
+          - OperationMetricsName: {^Parameter: {Name: "MetricsName", Default: "Lookup"}}
+            OperationName: RunCommand
+            OperationCommand:
+              aggregate: {^Parameter: {Name: "Coll", Default: "localColl"}}
+              pipeline:
+                [
+                  {$lookup: {
+                    from: {^Parameter: {Name: "Coll", Default: "localColl"}},
+                    localField: "localField",
+                    foreignField: "localField",
+                    as: "joined"
+                  }},
+                ]
+              cursor: {batchSize: {^Parameter: {Name: "BatchSize", Default: 101}}}
+
+# Template for running a $lookup using let/pipeline (aka "expressive") syntax. For the reason
+# mentioned above, the local and foreign collections are the same.
+- TemplateName: SelfLookupLetPipeline
+  Config:
+    Name: RunLookup
+    Type: RunCommand
+    Database: *Database
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1}}]
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Repeat: 10
+          Database: *Database
+          Operations:
+          - OperationMetricsName: {^Parameter: {Name: "MetricsName", Default: "Lookup"}}
+            OperationName: RunCommand
+            OperationCommand:
+              aggregate: {^Parameter: {Name: "Coll", Default: "localColl"}}
+              pipeline:
+                [
+                  {$lookup: {
+                    from: {^Parameter: {Name: "Coll", Default: "localColl"}},
+                    let: {local_field: "localField"},
+                    pipeline: [
+                      {$match: 
+                        {
+                          $expr: {$eq: ['$localField','$$local_field']}
+                        }
+                      }
+                    ],
+                    as: "joined"
+                  }},
+                ]
+              cursor: {batchSize: {^Parameter: {Name: "BatchSize", Default: 101}}}
+
+Actors:
+- Name: CreateShardedCollections
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *NumPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: admin
+        Operations:
+        - OperationMetricsName: EnableSharding
+          OperationName: AdminCommand
+          OperationCommand:
+            enableSharding: *Database
+        # This collection will have data distributed across 6 chunks.
+        - OperationMetricsName: ShardLocalCollection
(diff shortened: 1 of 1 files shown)


SimonEismann:
We already have a couple of [variants that are used only by a specific test](https://github.com/10gen/performance-tooling-docs/blob/main/build_variants/build_variants.md#configurations-to-support-a-specific-test), so adding a new variant is definitely in the realm of possibilities.

That said, fewer variants are always better :) Would it maybe be possible to set the forceClassic parameter via the RunCommand actor at the start of the test?

SimonEismann:
Good point, hadn't really considered that the clusters are sharded. So it seems like there isn't really a way around adding a new variant if we want the unsharded $lookups baseline. If you think that this baseline adds significant value, then I would say let's go ahead with adding a new variant.

src/workloads/query/LookupColocatedData.yml
```diff
@@ -0,0 +1,368 @@
...
+- When:
+    branch_name:
+      $neq:
+      - v4.0
+      - v4.2
+      - v4.4
+      - v5.0
+    mongodb_setup:
+      $eq:
+      - shard
```
HanaPearlman, line 366: It may not be worth listing all of these configurations here, maybe we only need one or two? This first config is for the 3-shard cluster, while the last two are for 2-shard clusters. And as expected, $lookup is faster in the 3-shard case when the data is distributed across all 3 shards. Not really sure if this gives us useful intel. 

One thing to note is that if we introduce a sharded force-classic variant, we will need to pick which of these configs to mimic-- 3-shard or 2-shard? Feature flags on or off?

SimonEismann: I would probably schedule the test on the following variants:
* Linux Shard Lite Cluster 2022-11
* Linux 3-Shard Cluster 2022-11
* Linux Shard Lite Cluster Intel 2022-11
* And maybe Linux Shard Lite (all feature flags) 2022-11 if the behavior under test here could be impacted by feature flags

In case you decide to also add a new variant with forceClassic, I would probably stick to 1-2 variants.

src/workloads/query/LookupColocatedData.yml
```diff
@@ -0,0 +1,368 @@
...
+          OperationCommand:
+            createIndexes: *UnshardedLocal
+            indexes:
+            - key: {*LocalShardKey: 1}
+              name: localShardKey_1
+
+- ActorFromTemplate:
+    TemplateName: LoadDocuments
+    TemplateParameters:
+      OnlyActiveInPhase: 2
```
nzolnierzmdb, line 177: Can you elaborate on what these numbers do and why we need them?

src/workloads/query/LookupColocatedData.yml
```diff
@@ -0,0 +1,368 @@
...
+#
+# Unsharded
+#
+- ActorFromTemplate:
+    TemplateName: SelfLookupLocalFieldForeignField
+    TemplateParameters:
+      OnlyActiveInPhase: 6
+      MetricsName: UnshardedLocalFieldForeignField1k
+      Coll: *UnshardedLocal
+      BatchSize: *BatchSize1k
```
nzolnierzmdb, line 212: I don't have any concrete data to support this, but I feel like most customer workloads are not setting `batchSize` explicitly. Was it different in the HELP ticket that these came from? I'm wondering if we should also have a case for the default/unspecified batchSize

src/workloads/query/LookupColocatedData.yml
```diff
@@ -0,0 +1,368 @@
...
+AutoRun:
+- When:
+    branch_name:
+      $neq:
+      - v4.0
+      - v4.2
+      - v4.4
+      - v5.0
+    mongodb_setup:
+      $eq:
```
nzolnierzmdb, line 365: Presumably this is more like a $in?

src/workloads/query/LookupColocatedData.yml
HanaPearlman: When specifying an actor in Genny, you need to indicate what the actor should be doing in each phase. It may be performing some operations, which need to be specified, or it could be doing nothing (no-op). It's a common pattern that an actor is only active in exactly one phase, and is doing a no-op in every other phase. [Here](https://github.com/mongodb/genny/blob/939d9bdbdc8830a1a2af4b27237743ef07a4cb09/src/workloads/query/LookupSBEPushdown.yml#L362-L374) is an example of doing that with more verbose syntax. Here, I'm using `OnlyActiveInPhase` as a more concise syntax for the same thing. So for example, this actor is only performing actions in phase 2.

If you wanted to write workloads where reads/writes happen concurrently, you could specify the same phase number for multiple actors. In our case, each query should be executed in a phase where no other operations are running. So the `OnlyActiveInPhase` numbers should be increasing throughout the file. 

I suppose all the inserts could happen in the same phase. I had it this way for developing/debugging purposes.

src/workloads/query/LookupColocatedData.yml
HanaPearlman: See my response above. I'm pretty sure we need to specify a batchSize for the actors, so we could use the default. It would be fine with me to add that case, but it was interesting to me to see the change in perf characteristics as the batchSize (aka amount of data processed) increased.

src/workloads/query/LookupColocatedData.yml
HanaPearlman: Yes. This is just how it is done everywhere, see https://github.com/mongodb/genny/blob/master/docs/using.md#configuring-autorun

HanaPearlman (commented):
Some responses!

HanaPearlman (commented):
More responses, and some changes up. I'm now seeing a difference between local results and results on evergreen that I'd like to dig a bit more into.

nzolnierzmdb (approved):
lgtm though I'm curious on the change in perf behavior. Do you think adding the limit may have done this?

HanaPearlman:
Thanks Nick. 

>  I'm curious on the change in perf behavior. Do you think adding the limit may have done this?

I'm still trying to investigate this. At this point I'm pretty baffled. I do think adding the $limit may have done this-- I tested [string shard key](https://spruce.mongodb.com/task/sys_perf_linux_3_shard.2022_11_lookup_colocated_data_patch_3aa51a4623b5869c6f57745ad1371aacf02c8d34_64864325d6d80a91956558d8_23_06_11_21_59_33/trend-charts?execution=3&sortBy=STATUS&sortDir=ASC) vs [int shard key,](https://spruce.mongodb.com/task/sys_perf_linux_3_shard.2022_11_lookup_colocated_data_patch_3aa51a4623b5869c6f57745ad1371aacf02c8d34_6483869c7742aef18d13c9e6_23_06_09_20_09_42/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) and this didn't seem to restore the behavior I was seeing before.

As a recap, _locally_ I still see the same behavior trend, where $lookup all on one shard is significantly better at first, but does significantly worse as the data processed increases. But on evergreen it never does _significantly_ better; it's only slightly better, and then it does get significantly worse as the data processed increases.

HanaPearlman:
I'm going to merge these changes. In summary, we were able to show that sharded $lookup is slower than unsharded $lookup even when the data is perfectly co-located. We were not able to show as drastic a change when the data is all on one shard vs distributed across multiple shards. Still, these numbers make sense to me and seem expected. These workloads will be good for measuring the effectiveness of SERVER-77427.

My last step is to make sure I'm not seeing too much noisiness in my patch builds, and then I'll merge.

HanaPearlman:
I ran five identical patches and analyzed the noise from the measurements. I took average latency for each query (10 runs per test), and computed the max, min, and median across the 5 runs. Then I looked at `(max-min)/med`. For the most part, measurements are at or below 5%. There are a small number of measurements with a higher variance which is more like 10%. it should come as no surprise that these are some of the fastest queries (and interestingly, the only ones that go through SBE...). Compared to existing perf workloads, I think this is appropriate and good to merge. 

cc @SimonEismann I'm not sure if I need your sign-off here.

SimonEismann:
@HanaPearlman Thanks for doing this noise analysis!

Within the perf team, we usually use the coefficient of variation (cov) to describe the degree of noise in a test. As a rule of thumb, a cov of < 0.1 is considered good. I've added two rows to your sheet to calculate the cov, which is <0.1 for all of your queries. 

There are some queries that have significantly higher CoV values than the others, e.g., `RunLookup.UnshardedLocalFieldForeignField15k`, `RunLookup.UnshardedLocalFieldForeignField1k`, and `RunLookup.UnshardedLocalFieldForeignField5k`. Notably, these are the very short-running queries. Would it be possible to increase the number of repetitions for these queries? That said, the noise levels are already acceptable, so feel free to merge as is :)

HanaPearlman:
Thanks Simon for that explanation! I'll compute CoV next time I add a new workload. For now, I'm going to merge since it looks like the highest CoV is ~.06 which seems solid!

PR #923, opened 2023-06-08
EVG-18404 - Implement a new workload-level field: `IgnoredMetrics`

PyGeek03:
**Jira Ticket:** EVG-18404

**Whats Changed:**
Adding a workload-level field: `IgnoredMetrics`, which is the list of metric names to mark as ignored (by adding `__METRICS_IGNORED__` to the end of ignored metrics' names). This is (I think) the only reliable way to detect metrics that should be ignored, no matter what an Actor's metric naming logic is.

Example usage for [CrudActor.yml](https://github.com/mongodb/genny/blob/9c61ea42035c0ad3eb38d5080f019672bbc2efad/src/workloads/docs/CrudActor.yml):
```yaml
SchemaVersion: 2018-07-01
...
IgnoredMetrics:
- CrudActor.Crud
- CrudActor.drop
- CrudActor.InsertAndUpdate
- CrudActor.Phase3.2
- CrudActor.insertMany
- CrudActor.Phase3.OtherInsertAndUpdate

Actors:
...
```

**Patch testing results:** (still running)  
- [sys-perf](https://spruce.mongodb.com/version/6489e0cba4cf4752a1bb1302/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [sys-perf-atlas](https://spruce.mongodb.com/version/6489de660305b9215ad0abd7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/gennylib/src/context.cpp
+++ b/src/gennylib/src/context.cpp
@@ -78,6 +78,13 @@ WorkloadContext::WorkloadContext(const Node& node,
 
     _registry = genny::metrics::Registry(std::move(format), std::move(metricsPath));
 
+    auto& ignoredMetrics = (*this)["IgnoredMetrics"];
+    if (ignoredMetrics && ignoredMetrics.isSequence()) {
+        for (const auto& [k, metric] : (*this)["IgnoredMetrics"]) {
+            _registry.ignore(metric.to<std::string>());
+        }
+    }
+
     _seedGenerator.seed((*this)["RandomSeed"].maybe<long>().value_or(RNG_SEED_BASE));
 
     // Make a bunch of actor contexts
--- a/src/metrics/include/metrics/metrics.hpp
+++ b/src/metrics/include/metrics/metrics.hpp
@@ -275,6 +275,10 @@ public:
         return _pathPrefix;
     }
 
+    void ignore(const std::string& metricName) {
+        _ignoredMetrics.emplace(metricName);
+    }
+
 private:
     std::string createName(const std::string& actorName,
                            const std::string& opName,
@@ -292,7 +296,16 @@ private:
         if (phase) {
             str << '.' << *phase;
         }
-        return str.str();
+
+        const auto& metricName = str.str();
+        if (_ignoredMetrics.count(metricName) == 1) {
+            str << ".__METRICS_IGNORED__";
+            const auto& ignoredMetricName = str.str();
+            BOOST_LOG_TRIVIAL(info) << "Marked metric as ignored: " << ignoredMetricName;
+            return ignoredMetricName;
+        }
+        BOOST_LOG_TRIVIAL(info) << "Recording metric: " << metricName;
+        return metricName;
     }
 
     // Must be a ptr to keep the registry moveable.
@@ -300,6 +313,7 @@ private:
     std::unique_ptr<GrpcClient> _grpcClient;
     OperationsMap _ops;
     MetricsFormat _format;
+    std::unordered_set<std::string> _ignoredMetrics;
     boost::filesystem::path _pathPrefix;
     boost::filesystem::path _internalPathPrefix;
 };
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -519,6 +519,8 @@ public:
     /**
      * Convenience method for creating a metrics::Operation that's unique for this actor and thread.
      *
+     * If "MetricsIgnored" is specified for this actor, "__IGNORED__." will be prepended to metric name.
+     *
      * @param operationName the name of the operation being run.
      * @param id the id of this Actor.
      * @param internal whether this operation is Genny-internal.
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -597,8 +597,10 @@ public:
     /**
      * Convenience method for creating a metrics::Operation that's unique for this phase and thread.
      *
-     * If "MetricsName" is specified for a phase, it is used.
-     * Otherwise "[defaultMetricsName].[phaseNumber]" is used.
+     * If "MetricsIgnored" is specified for a phase, "__IGNORED__." will be prepended to metric name.
+     *
+     * If "MetricsName" is specified for a phase, it is used for metric name;
+     * otherwise, "[defaultMetricsName].[phaseNumber]" is used.
      *
      * @param defaultMetricName the default name of the metric if "MetricsName" is not specified
      *                          for a phase in the workload YAML.
@@ -607,6 +609,9 @@ public:
      */
     auto operation(const std::string& defaultMetricsName, ActorId id, bool internal = false) const {
         std::ostringstream stm;
+        if (this->_node["MetricsIgnored"].maybe<bool>()) {
+            stm << "__METRICS_IGNORED__.";
+        }
         if (auto metricsName = this->_node["MetricsName"].maybe<std::string>()) {
             stm << *metricsName;
         } else {
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -524,8 +524,13 @@ public:
      * @param internal whether this operation is Genny-internal.
      */
     auto operation(const std::string& operationName, ActorId id, bool internal = false) const {
+        std::ostringstream stm;
+        if (this->_node["MetricsIgnored"].maybe<bool>()) {
+            stm << "__METRICS_IGNORED__.";
+        }
+        stm << operationName;
         return this->_workload->_registry.operation(
-            this->_node["Name"].to<std::string>(), operationName, id, std::nullopt, internal);
+            this->_node["Name"].to<std::string>(), stm.str(), id, std::nullopt, internal);
     }
 
 private:
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -519,20 +519,13 @@ public:
     /**
      * Convenience method for creating a metrics::Operation that's unique for this actor and thread.
      *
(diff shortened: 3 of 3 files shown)


thessem (changes requested):
Clicking request changes on this because I'd like to understand this code a little better before it is merged.

- What are the performance implications of this change? Is `context.cpp` run inside any hot paths where we should be closely watching changes in algorithmic complexity? This code is adding a loop over ignored metrics (what if we have 100 ignored metrics, what if there are 10,000?). How complex is `emplace`? These aren't socratic questions, I genuinely do not have much experience here and would love to find out either way.

- Is info level logging the right choice here? What sort of number of metrics do are we dealing with? I'd like to avoid putting a big wall of text in the log files.

- I may have missed something big here, but it's not clear to me why renaming `__METRICS_IGNORED__` actually ignores the metric. I've taken a a look at https://github.com/10gen/dsi/pull/1310 and can't see a corresponding change

- What other possible repercussions does renaming a metric have? Is this definitely a safe thing to do with all the other consumers of metrics out of Genny?

PyGeek03:
> * What are the performance implications of this change? Is `context.cpp` run inside any hot paths where we should be closely watching changes in algorithmic complexity? This code is adding a loop over ignored metrics (what if we have 100 ignored metrics, what if there are 10,000?). How complex is `emplace`? These aren't socratic questions, I genuinely do not have much experience here and would love to find out either way.

From my understanding, the modified code path in `context.cpp` is run exactly once to create the workload's `WorkflowContext` based on the workload YAML file. A few lines below, there's [a loop for every single Actor  in the YAML file](https://github.com/mongodb/genny/blob/2d33cb0ddad170df7b20a9f19cb04fc031245aff/src/gennylib/src/context.cpp#L90-L93). This loop also uses `emplace_back`, std::vector's `emplace` version of `push_back()` (appending to array). Using the `emplace` version of insertion methods is a common pattern throughout the Genny C++ codebase, and for good reason: It is more efficient because it constructs the new object in-place inside the collection, instead of constructing outside then copying or moving the object into the collection. For more details: https://yasenh.github.io/post/cpp-diary-1-emplace_back

> * Is info level logging the right choice here? What sort of number of metrics do are we dealing with? I'd like to avoid putting a big wall of text in the log files.

My reasoning for doing this is observability: to make it easier for users to know the exact metrics names (for putting under `IgnoredMetrics:`) by just reading the log. The only alternative to this, from my understanding, is digging into the workload's output FTDC filenames (the FTDC filenames are basically `{metrics name}.ftdc`).

> * I may have missed something big here, but it's not clear to me why renaming `__METRICS_IGNORED__` actually ignores the metric. I've taken a a look at [EVG-18404 - Implement ignoring metrics using test_control YML file 10gen/dsi#1310](https://github.com/10gen/dsi/pull/1310) and can't see a corresponding change

[The initial commit in the DSI PR](https://github.com/10gen/dsi/pull/1310/commits/805f56c02c3d0ee6c591485c4ec654827d2e8b3a) would ignore the metrics with `__METRICS_IGNORED__` in their name, which is the counterpart to [the initial 3 commits in the Genny PR](https://github.com/mongodb/genny/pull/923/files/a1bf99de2364af1cc7829b0b3fb9c62232ec7a5b). 

I decided to tag a metric as ignored by renaming the metric, because 1) the FTDC files are still there should we need them & 2) workload generators other than Genny could also tag a metric as ignored by renaming the metric, and DSI analysis would filter them out just as well. Genny doesn't have an existing tagging mechanism, and its output is just time series data files, so implementing a tagging mechanism would involve either 1) renaming the file (which is the same as renaming the metric) or 2) adding some metadata to the file (which involves a bit more effort, so might be worth doing if we have other uses for tags).

Those 3 commits are based on the assumption that metric names are defined either at the Actor level or at the Phase level, and `MetricsIgnored: true` should be added at the corresponding level. However, I later discovered that Genny Actors actually could have arbitrary logic for metrics name. For example, [a CrudActor might define metrics name at a custom Operation level below the Phase level](https://github.com/mongodb/genny/blob/50f63afb1f1705dd635c1b72d52a1f67dbfb387a/src/cast_core/src/CrudActor.cpp#L1587-L1602), and this Operation level is unique to CrudActor.

I decided that the best way to deal with arbitrary logic for metrics name is by requiring the user to explicitly spell out the full names of metrics to ignore. I abandoned the above Genny-declare DSI-remove approach, and switched to a DSI-declare DSI-remove approach based on `test_control.*.yml` files under the `analysis` field, after having looked through all the DSI declaration interfaces.

However, it turns out that there were plans to remove all `test_control.*.yaml` files for Genny workloads, which I wasn't aware of. Quite a few Genny users have also voiced their preference to declare ignored metrics inside Genny itself (like in this recent Slack thread: https://mongodb.slack.com/archives/C01VD0LQZED/p1686288935508839)

This led me to adopt the spell-out-full-name mechanism for the Genny-declare DSI-remove approach, which is the current state of this Genny PR.

I planned to fix the DSI PR if this Genny PR is what Genny users want.

> * What other possible repercussions does renaming a metric have? Is this definitely a safe thing to do with all the other consumers of metrics out of Genny?

All the renamed metrics would be filtered out by DSI. I've just taken a look again at the [performance results data flow diagram](https://github.com/10gen/product-perf-experimentations/blob/master/users/david.daly/perf-architecture/results-flow.md), which shows that Genny sends "client_timeseries" to Cedar Report without going through DSI. I'll double check what "client_timeseries" is, but my understanding is that they're not metrics, but rather system measurements of the clients (like CPU, memory usage, etc.)

src/metrics/include/metrics/metrics.hpp
```diff
@@ -292,14 +296,24 @@ class RegistryT final {
         if (phase) {
             str << '.' << *phase;
         }
-        return str.str();
+
+        const auto& metricName = str.str();
+        if (_ignoredMetrics.count(metricName) == 1) {
+            str << ".__METRICS_IGNORED__";
+            const auto& ignoredMetricName = str.str();
+            BOOST_LOG_TRIVIAL(info) << "Marked metric as ignored: " << ignoredMetricName;
```
rtimmons, line 304: This is just operating as expected, I don't think this line would be useful.

src/metrics/include/metrics/metrics.hpp
```diff
@@ -292,14 +296,24 @@ class RegistryT final {
         if (phase) {
             str << '.' << *phase;
         }
-        return str.str();
+
+        const auto& metricName = str.str();
+        if (_ignoredMetrics.count(metricName) == 1) {
+            str << ".__METRICS_IGNORED__";
```
rtimmons, line 302: 1. But this doesn't actually cause us to ignore, it just appends a sentinel?
2. Would it make sense to lean into the existing `canary_` prefix logic instead  and treat ignored metrics as canaries? I don't know the full downstream implications of overloading canaries logic like that, but there might be rails there somewhere this could ride.
3. The user could do this today without this change by simply specifying `MetricsName: foo.__IGNORED__`. In essence this is yaml syntax sugar. As such, if we want to pursue this "ignore via suffix" strategy, I suggest we consider putting this logic into the existing [preprocessing python](https://github.com/mongodb/genny/blob/4e6e1c7da4d830f44fbf4a65c4d4e3bae2011a6a/src/lamplib/src/genny/tasks/preprocess.py). We've found syntax translations like this much easier to reason about, write, and test in Python and thus keep the C++-side of things kinda "dumb."

src/gennylib/src/context.cpp
```diff
@@ -78,6 +78,13 @@ WorkloadContext::WorkloadContext(const Node& node,
 
     _registry = genny::metrics::Registry(std::move(format), std::move(metricsPath));
 
+    auto& ignoredMetrics = (*this)["IgnoredMetrics"];
```
rtimmons, line 81: Consider using `getPlural("IgnoredMetric","IgnoredMetrics")` (info [here](https://github.com/mongodb/genny/blob/4e6e1c7da4d830f44fbf4a65c4d4e3bae2011a6a/src/gennylib/include/gennylib/Node.hpp#L501) to allow either config syntax. (I bring this up just to show off syntax features you might not be familiar with, I don't think it matters _that_ much in the long run.)

rtimmons (commented):
It's beyond my ability to push back against this if y'all are for it, but I think this is accumulating technical debt via string-hacking rather than solving the core issue. That sounds harsher than I intend, apologies. I get this is a simple solution that might be sufficient for what y'all want; and that's fine if you decide so.

I think we made a mistake in allowing metrics names to be defaulted and always collected even if not explicitly requested. If I could turn back time, I'd say we'd only report metrics if the name were specified in `MetricsName`. We could make that change and backfill, but it would be kinda painful.

What I would instead suggest is a "nop" metrics `Operation` implementation that simply discards all data rather than reporting it via gRPC. It would ideally still barf if not closed (indicating an unhandled exception).

(This might be some nontrivial surgery into the (admittedly twisty/turny) [`OperationT`](https://github.com/mongodb/genny/blob/4e6e1c7da4d830f44fbf4a65c4d4e3bae2011a6a/src/metrics/include/metrics/metrics.hpp#L223) internals to not create the streams and stuff. But I'd advocate that's super critical logic and investments in understanding there will pay dividends very quickly, especially as the team thinks of getting away from poplar/gRPC.)

Having a separate `IgnoredMetrics` is papering over the fact that we're needlessly collecting and downstream processing useless data, and it separates the thing from the metadata about the thing. Then allow the user to specify `MetricsName: Ignored` or something similar on the Actor/Operation yaml blocks for whose metrics they wish to ignore. Putting this right by the operation/config in the yaml makes it super obvious what applies to what.  If the workload later renames things an `IgnoredMetrics` list would become out of date, but the `MetricsName: Ignored` would be brought along for the ride.

My intent is to help give context and explain some potential alternatives and downsides. Normally I wouldn't comment (or care) in such detail, but given this is a nuanced area with new tech ownership, I wanted to provide some context or input that might not have been obvious if the goal is just to solve the single ticket. Happy to chat further, and of course I'm only one voice of many here.

PR #925, opened 2023-06-08
Batch miscellaneous fixes for Genny

PyGeek03:
**Whats Changed:**  
This is to batch some miscellaneous fixes into one PR:
- [context.hpp](https://github.com/mongodb/genny/pull/925/files#diff-91bfd6be8c6be7eea49589d45d788b78164899e2a666e0dc8780b4228da9d40a): 
    + Use explicit `this` pointer (instead of implicit class member attributes) throughout for consistency and readability.
    + Rename `PhaseContext`'s member attribute `ActorContext* _actor` to `ActorContext* _actorContext` to prevent confusion (as `ActorContext` and `Actor` are completely different: One `ActorContext` could map to multiple `Actor`'s). 
    + Code formatting done by linter.
- [metrics_test.cpp](https://github.com/mongodb/genny/pull/925/files#diff-c570f2a0bd0213787fa2a7e49f18122c107b5d4db0b7a2b97695ee528c4475ee):
    + Code formatting done by linter.
- [metrics.hpp](https://github.com/mongodb/genny/pull/925/commits/2514223c7a3819abd200fa23eac0e9b86551ca11#diff-87d703b99ca63cd730141f1a77a65d6a641b60cd56e559e64a1005190b8c2188):
    + Use explicit types to improve code searchability & comprehension (So that no one has to spend the same amount of time I did to comprehend what this section of the code actually does. Heck, it even confuses VSCode's type deduction...)

### DIFF
--- a/src/gennylib/include/gennylib/context.hpp
+++ b/src/gennylib/include/gennylib/context.hpp
@@ -128,7 +128,7 @@ public:
     ExternalPhaseCoordinator(ExternalPhaseCoordinator&&) = delete;
     void operator=(ExternalPhaseCoordinator&&) = delete;
 
-    ~ExternalPhaseCoordinator() {
+    ~ExternalPhaseCoordinator(){
         if (m_socket >= 0) {
             close(m_socket);
         };
@@ -487,7 +487,7 @@ public:
      *
      *   - Operation: Cleanup
      *     # inherits `Collection: links` from parent,
-     *     # and `Phase: 2` is derived based on index
+     *     # and `Phase: 3` is derived based on index
      * ```
      *
      * This would result in 3 `PhaseContext` structures. Keys are inherited from the
@@ -508,7 +508,7 @@ public:
     }
 
     /**
-     * @return a pool client from the "default" MongoDB connection-pool.
+     * @return a pool from the "default" MongoDB connection-pool.
      * @throws InvalidConfigurationException if no connections available.
      */
     template <class... Args>
@@ -562,9 +562,7 @@ private:
 class PhaseContext final : public v1::HasNode {
 public:
     PhaseContext(const Node& node, PhaseNumber phaseNumber, ActorContext& actorContext)
-        : v1::HasNode{node},
-          _actorContext{std::addressof(actorContext)},
-          _phaseNumber(phaseNumber) {}
+        : v1::HasNode{node}, _actor{std::addressof(actorContext)}, _phaseNumber(phaseNumber) {}
 
     // no copy or move
     PhaseContext(PhaseContext&) = delete;
@@ -573,7 +571,7 @@ public:
     void operator=(PhaseContext&&) = delete;
 
     DefaultRandom& rng(ActorId id) {
-        return this->_actorContext->rng(id);
+        return this->_actor->rng(id);
     }
 
     /**
@@ -585,15 +583,15 @@ public:
      * @return the parent workload context
      */
     WorkloadContext& workload() const {
-        return this->_actorContext->workload();
+        return _actor->workload();
     }
 
     ActorContext& actor() const {
-        return *this->_actorContext;
+        return *_actor;
     }
 
     SleepContext getSleepContext() const {
-        return SleepContext(this->_phaseNumber, this->actor().orchestrator());
+        return SleepContext(_phaseNumber, this->actor().orchestrator());
     }
 
     /**
@@ -612,24 +610,24 @@ public:
         if (auto metricsName = this->_node["MetricsName"].maybe<std::string>()) {
             stm << *metricsName;
         } else {
-            stm << defaultMetricsName << "." << this->_phaseNumber;
+            stm << defaultMetricsName << "." << _phaseNumber;
         }
 
         return this->workload()._registry.operation(
-            this->_actorContext->operator[]("Name").to<std::string>(),
+            this->_actor->operator[]("Name").to<std::string>(),
             stm.str(),
             id,
-            this->_phaseNumber,
+            _phaseNumber,
             internal);
     }
 
     /**
      * @brief [Experimental] Create a metrics::Operation with the specified name
      *
-     * PhaseContext::operation() always returns the same Operation if MetricsName is specified.
-     * But according to the HelloWorld actor, this behavior seems to be unexpected.
-     * However, it is too risky to change the behavior without thoroughly examining the impact.
-     * EVG-17788 tracks this work.
+     * The current operation() function will always return the same operation if MetricsName is specified.
+     * According to the HelloWorld actor, it seems not the expected behavior
+     * But it is too risky to change the behavior without thoroughly exam the impact
+     * Created TIG-4328 to track this work
      *
      * @param metricsName name of the metric
      * @param id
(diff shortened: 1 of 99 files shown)


src/gennylib/include/gennylib/context.hpp
```diff
@@ -610,24 +612,24 @@ class PhaseContext final : public v1::HasNode {
...
 
     /**
      * @brief [Experimental] Create a metrics::Operation with the specified name
      *
-     * The current operation() function will always return the same operation if MetricsName is specified.
-     * According to the HelloWorld actor, it seems not the expected behavior
-     * But it is too risky to change the behavior without thoroughly exam the impact
-     * Created TIG-4328 to track this work
+     * The current operation() function will always return the same operation if MetricsName is
+     * specified. According to the HelloWorld actor, it seems not the expected behavior But it is
```
thessem, line 630: This comment has not been improved by auto-formatting

src/gennylib/test/context_test.cpp
```diff
@@ -112,23 +114,30 @@ TEST_CASE("loads configuration okay") {
...
-SchemaVersion: 2018-07-01
-Clients:
-  Default:
-    URI: 'mongodb://localhost:27017'
-Actors:
-- Name: HelloWorld
-  Type: Nop
-  Count: 7
-        )",
+            SchemaVersion: 2018-07-01
```
thessem, line 117: This looks like a non-equivalent change to me. In YAML indentation matters. Is there some property of this string literal that makes these equivalent?

src/gennylib/test/context_test.cpp
```diff
@@ -112,23 +114,30 @@ TEST_CASE("loads configuration okay") {
...
+            )",
                                "");
 
         WorkloadContext w{yaml.root(), orchestrator, cast};
         auto& actors = w["Actors"];
     }
 
     SECTION("Invalid Schema Version") {
-        auto yaml = NodeSource("SchemaVersion: 2018-06-27\nClients: {Default: {URI: 'mongodb://localhost:27017'}}\nActors: []\n", "");
+        auto yaml = NodeSource(R"(
```
thessem, line 133: What is an R"" string? It looks like we've included indentation in this string where there wasn't any

thessem (changes requested):
There are a few changes I don't think are trivially simple in this. I don't immediately understand why adding indentation to YAML strings is making it more consistent.

This is in Go, but for example: https://go.dev/play/p/vQ3fj5n2ukq. Is there some property of C++ strings that avoids this?

Additionally, in C++ order of header importing _can sometimes_ be important. Changing it isn't what I would consider a pure linting/formatting change.

For reference: Formatting changes should result in equal ASTs before and after, linting changes should be immediately and trivially correct. Double-quotes to single-quotes is an example of both.

Generally I think it's okay to do header and import re-ordering when you're touching the file anyway for some non-trivial other change. Apologies if the header re-ordering here was extracted from some other already merged PR.

src/gennylib/include/gennylib/context.hpp
PyGeek03: I agree. I've fixed this comment.

src/gennylib/test/context_test.cpp
PyGeek03: It's a [raw string literal](https://en.cppreference.com/w/cpp/language/string_literal), which can be used to "avoid escaping of any character". Most other NodeSource initializations in this file use raw string literal instead of escaping the new lines, so I converted this one. The relative indentation levels between lines are preserved.

src/gennylib/test/context_test.cpp
PyGeek03: According to [YAML 1.2.2 spec](https://yaml.org/spec/1.2.2/#indentation-spaces): "Each node must be indented further than its parent node. All sibling nodes must use the exact same indentation level." So the absolute number of spaces doesn't matter, what matters is the relative number of spaces between YAML lines.

PyGeek03:
> I don't immediately understand why adding indentation to YAML strings is making it more consistent.

At first I was just annoyed that while most YAML strings in this file are indented, some YAML strings are not, so I fixed that small number of YAML strings. But then I realized that the the absolute number of spaces doesn't matter (explained below), so I decided to make them all indented nicely like [these lines](https://github.com/mongodb/genny/blob/a3bebe0144abbc04e42d690644ec31d8fa59729e/src/gennylib/test/context_test.cpp#L567-L584) in this same file.

> This is in Go, but for example: [go.dev/play/p/vQ3fj5n2ukq](https://go.dev/play/p/vQ3fj5n2ukq). Is there some property of C++ strings that avoids this?

According to [YAML 1.2.2 spec](https://yaml.org/spec/1.2.2/#indentation-spaces): "Each node must be indented further than its parent node. All sibling nodes must use the exact same indentation level." and "The amount of indentation is a [presentation detail](https://yaml.org/spec/1.2.2/#presenting-the-serialization-tree) and must not be used to convey [content](https://yaml.org/spec/1.2.2/#nodes) information." So the absolute number of spaces doesn't matter, what matters is the relative number of spaces between YAML lines.

> Additionally, in C++ order of header importing _can sometimes_ be important. Changing it isn't what I would consider a pure linting/formatting change.
> 
> For reference: Formatting changes should result in equal ASTs before and after, linting changes should be immediately and trivially correct. Double-quotes to single-quotes is an example of both.

The headers are sorted by [clang-format, via the C/C++ VSCode extension](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting). I don't think verifying this change manually is a good use of time, so I've reverted it.

src/gennylib/test/context_test.cpp
PyGeek03: It's a [raw string literal](https://en.cppreference.com/w/cpp/language/string_literal), which can be used to "avoid escaping of any character". Most other NodeSource initializations in this file use raw string literal instead of escaping the new lines, so I converted this one. The relative indentation levels between lines are preserved.

src/gennylib/test/context_test.cpp
thessem: I've been thinking about this, and I would really prefer it if we didn't do this. When I see indented strings like this in source code, it tends to be too jarring to the eye, and I think this is a valid concern, especially since this is purely a formatting, and therefore aesthetic, PR. Having dealt with too many bugs related to code formatting getting inside strings, my view of aesthetics here is based on that.

Turning single lines into multi-line strings is strictly worse. There have been cases where flow-based YAML strings have been converted into non-flow-based YAML, which changes the (fairly implicit) test coverage of the YAML parser here. These are not unambiguous improvements.

When it comes to formatting and linting changes, they should be AST equivalent to the code before them, and I think they should be unambiguously better.

If we already have multi-line YAML strings with code formatting in them, I would accept this change in those cases. But I don't think this is a super valuable use of time.

Could you please resubmit this PR without the non-string and variable name-related changes?

thessem (changes requested):

thessem (approved):
Thanks for this, sorry it took so long to get this tick!

PyGeek03:
evergreen merge

PR #930, opened 2023-06-13
PERF-4242 Add perf workload for multiple group stages on computed fields

Adityav369:
**Jira Ticket:** 
PERF-4242, PERF-4243, PERF-4244

**Whats Changed:**  
This change adds a perf load for a queries with group stages on computed fields on regular collections and time-series collections.

**Latest Patch testing results:**  
https://spruce.mongodb.com/version/649996b52a60edf99096a88f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

**Workload Submission form:**  
Submitted for this workload.


src/phases/query/MultipleGroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,84 @@
...
+      PhaseConfig:
+        Repeat: {^Parameter: {Name: "repeat", Default: 10}}
+        Database: *Database
+        Collection: {^Parameter: {Name: "collection", Default: Collection0}}
+        Operations:
+        - OperationMetricsName: {^Parameter: {Name: "operationsMetricsName", Default: MultipleGroupStagesOnComputedFieldsTemplate}}
+          OperationName: aggregate
+          OperationCommand:
+            Pipeline: [{$addFields:
+                          {"hourDiff": {$dateDiff: {"startDate": "$time", "endDate": "$$NOW", "unit": "hour"}}}},
```
jseyster, line 46: There's a risk that using `$$NOW` in the query will make its selectivity variable, depending on the timing of other phases of the test, which could make the measurements less stable.

I suggest loading `$time` with a fixed range of values and using a constant for `endDate` that will result in about 25% of documents getting filtered out.

My reasoning for 25% is that we should hopefully see some performance improvement by pushing down the `$match` and filtering out 25% of documents before they have to get converted to BSON. We don't want to lean on that _too_ heavily to get our performance improvement, though, because it won't be directly applicable to time series (which can use index bounds to skip most events that would be filtered out).

jseyster (commented):
Great work! I have one comment inline about the query.

src/phases/query/MultipleGroupStagesOnComputedFields.yml
Adityav369: Ah yes, that makes sense. Done!

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+                  },
+                  {$group:
+                    {"_id" : {
+                        $dateTrunc: {
+                          "date": "$time",
+                          "unit": "day"
+                        }
+                      },
+                      "count" : {$sum: 1}
+                    }},
```
jseyster, line 87: The second bracket (closing the `$group`) should be on its own line, and the subsequent `$sort` should be indented one level less.

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+                                  },
+                              }
+                          },
+                          {
+                              $sort : {
+                                  "_id" : -1
+                              }
+                          }]
+
+GroupStageAfterMatchOnDateComputation: &GroupStageAfterMatchOnDateComputation
```
jseyster, line 65: I would rewrite this name to be `GroupStateOnDateComputationAfterMatch` so that it's clear that the date computation is in the group and not the match filter.

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+                          "time": {
+                            $gte: "2022-01-03"
+                          }
+                          }
+                  },
+                  {$group:
+                    {"_id" : {
+                        $dateTrunc: {
+                          "date": "$time",
+                          "unit": "day"
```
jseyster, line 83: If you have time, I'd be curious to see how quantizing by minute would change our results. Having more groups might change which bottlenecks we see in each of the implementations. The characteristics might be different enough to justify two benchmarks.

jseyster (commented):
These looks great! There are a few things that I think would be interesting to try out, but other than that, there are no major changes.

src/phases/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,47 @@
...
+      NopInPhasesUpTo: {^Parameter: {Name: "nopInPhasesUpTo", Default: 5}}
+      PhaseConfig:
+        Repeat: {^Parameter: {Name: "repeat", Default: 10}}
+        Database: *Database
+        Collection: {^Parameter: {Name: "collection", Default: Collection0}}
+        Operations:
+        - OperationMetricsName: *Name
+          OperationName: aggregate
+          OperationCommand:
+            Pipeline: {^Parameter: {Name: "pipeline", Default: []}}
```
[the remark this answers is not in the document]
Adityav369: That's a good point. Parameter requires default to be set, so I changed it to an empty string. This way it would throw an error.

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+                                      $topN: {
+                                          "output": ["$_id.symbol", "$diffPercentage", "$close"],
+                                          "sortBy" : { "diffPercentage" : -1 },
+                                          "n": 3
+                                      }
+                                  },
+                                  "bottomN": {
+                                      $bottomN: {
+                                          "output": ["$_id.symbol", "$diffPercentage", "$close"],
+                                          "sortBy" : { "diffPercentage" : 1 },
```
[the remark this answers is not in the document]
Adityav369: Ah yes! Done.

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+    Parameters:
+      name: GroupStageAfterMatchOnDateComputation
+      active: {^Parameter: {Name: "active", Default: [4]}}
+      nopInPhasesUpTo: *NopInPhasesUpTo
+      pipeline: [{$match: {
+                          "symbol": "string050",
+                          "time": {
+                            $gte: "2022-01-03"
+                          }
+                          }
```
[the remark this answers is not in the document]
Adityav369: Done. I reflowed this query.

src/workloads/query/GroupStagesOnComputedFields.yml
Adityav369: Done.

src/workloads/query/GroupStagesOnComputedFields.yml
Adityav369: Yes! Done.

src/workloads/query/GroupStagesOnComputedFields.yml
```diff
@@ -0,0 +1,176 @@
...
+                                  "bottomN": {
+                                      $bottomN: {
+                                          "output": ["$_id.symbol", "$diffPercentage", "$close"],
+                                          "sortBy" : { "diffPercentage" : 1 },
+                                          "n": 3
+                                      }
+                                  },
+                              }
+                          },
+                          {
```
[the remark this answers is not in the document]
Adityav369: Done. I changed the $dateTrunc interval to minute, and now we get 2217 docs passed to the sort stage.

src/workloads/query/GroupStagesOnComputedFields.yml
Adityav369: I tried this. For the docs returned, quantizing by day returns 3 documents, and quantizing by minute returns 2937 documents. Perf results for this experiment are [here](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=5392c9c9-1c51-4bd2-ad09-9d6917120a0f&task_filter=group_stages_on_computed_fields&test_filter=GroupStageOnDateComputationAfterMatch.GroupStageOnDateComputationAfterMatch&test_filter=GroupStageAfterSortAndAddFields.GroupStageAfterSortAndAddFields&test_filter=MultipleGroupStagesOnComputedFields.MultipleGroupStagesOnComputedFields&measurement_filter=OperationThroughput&percent_filter=0%7C%7C100&z_filter=0%7C%7C10). Should I add this as a separate query or try seconds as well?

jseyster:
Thanks for trying those different quantization parameters, Adi! From the results so far (which don't show much of a difference), I think it makes sense to keep the quantization from the original queries that these tests are based on.

This LGTM for the Performance Team to review.

ghartnett (approved):

Adityav369:
evergreen merge

PR #932, opened 2023-06-20
EVG-20302 - Fix empty password redaction edge case, and add test to ensure PoolFactory supports empty password

PyGeek03:
**Jira Ticket:** EVG-20302

**Whats Changed:**  
Fix edge case when password is empty. For example, currently when running Genny locally, here's what I got in the log:

```
Constructing pool with MongoURI '[REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED][REDACTED]'
```

With this fix, here's what would be in the log:

```
Constructing pool with MongoURI 'mongodb://localhost:27017/?appName=Genny&maxPoolSize=1000'
```

This bug also affects Genny on Evergreen, for example [this log line](https://parsley.mongodb.com/evergreen/sys_perf_linux_microbenchmarks_standalone_arm.2023_01_genny_scale_InsertRemove_patch_b7fecb4445d3e54a601e4e4dd79b9d57d082117f_64a6cea3306615eb9c3cf612_23_07_06_14_25_36/3/task?bookmarks=0,2525&shareLine=2511).

The fix is to use a more obvious logic: constructs the redacted URI from source, instead of redacting constructed actual URI.

As these changes are outside metric recording path, there should be no impact at all to Genny workloads on Evergreen.

**Patch testing results:**  
[sys-perf, genny_scale_InsertRemove](https://spruce.mongodb.com/task/sys_perf_linux_microbenchmarks_standalone_arm.2023_01_genny_scale_InsertRemove_patch_2ef8927c2b6731fb68ed09dcbc80d0fd4c384ab0_64915414d6d80ad8606cbb43_23_06_20_07_26_00?execution=0)

**Related PRs:**
https://github.com/mongodb/genny/pull/907

### DIFF
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -240,17 +240,18 @@ mongocxx::options::pool PoolFactory::makeOptions() const {
 std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
     auto uriStr = makeUri();
 
+    std::string username{*_config->get(OptionType::kAccessOption, "Username")};
     std::string password{*_config->get(OptionType::kAccessOption, "Password")};
-
-    auto redactedUriStr = password == ""
+    auto credentials = username + ':' + password + '@';
+    auto redactedCredentials = username + ':' + "[REDACTED]" + '@';
+    auto credPosition = uriStr.find(credentials);
+    auto redactedUriStr = credPosition == std::string::npos
                         ? uriStr
-                        : std::regex_replace(uriStr, std::regex(password), "[REDACTED]" );
+                        : uriStr.replace(credPosition, credentials.length(), redactedCredentials);
     BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << redactedUriStr << "'";
 
     auto uri = mongocxx::uri{uriStr};
-
     auto poolOptions = makeOptions();
-
     return std::make_unique<mongocxx::pool>(uri, poolOptions);
 }
 
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -242,7 +242,9 @@ std::unique_ptr<mongocxx::pool> PoolFactory::makePool() const {
 
     std::string password{*_config->get(OptionType::kAccessOption, "Password")};
 
-    auto redactedUriStr = std::regex_replace(uriStr, std::regex(password), "[REDACTED]" );
+    auto redactedUriStr = password == ""
+                        ? uriStr
+                        : std::regex_replace(uriStr, std::regex(password), "[REDACTED]" );
     BOOST_LOG_TRIVIAL(info) << "Constructing pool with MongoURI '" << redactedUriStr << "'";
 
     auto uri = mongocxx::uri{uriStr};
--- a/src/gennylib/src/PoolFactory.cpp
+++ b/src/gennylib/src/PoolFactory.cpp
@@ -79,12 +79,14 @@ struct PoolFactory::Config {
         queryOptions["appName"] = "Genny";
     }
 
-    auto makeUri() const {
+    auto makeUri(bool redacted) const {
         std::ostringstream ss;
         size_t i;
 
         ss << accessOptions.at("Protocol");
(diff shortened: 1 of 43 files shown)


thessem (changes requested):
This looks good to me in principle. Would you mind pasting whatever error this caused into the PR request, just to be clear about the actual problem you're fixing? If other users have this problem, and they search for the error message, I'd like them to find something.

Is it possible to add a unit test for this exact case? That is to say, a unit test testing that the pool factory works with an empty password (not testing specifically that the log replace works correctly).

As far as I am aware I'd assume most regex parses would treat an empty string as matching nothing (this might be wrong) so I'm a little surprised this broke something. Would it be possible to make the regex slightly better (what if the password is "mongodb", the redaction would be come `[redacted]`://user:[redacted]@127....` right?), or use a non-regex string replace (which will be cheaper anyway)

PyGeek03:
> This looks good to me in principle. Would you mind pasting whatever error this caused into the PR request, just to be clear about the actual problem you're fixing? If other users have this problem, and they search for the error message, I'd like them to find something.

Ah yes I've added this to the PR description above.
 
> Is it possible to add a unit test for this exact case? That is to say, a unit test testing that the pool factory works with an empty password (not testing specifically that the log replace works correctly).

Yes I've done this. Thank you for this actually, this test helped me avoid a bug when switching from std::regex_replace to std::string::replace later.

> As far as I am aware I'd assume most regex parses would treat an empty string as matching nothing (this might be wrong) so I'm a little surprised this broke something. Would it be possible to make the regex slightly better (what if the password is "mongodb", the redaction would be come `[redacted]`://user:[redacted]@127....` right?), or use a non-regex string replace (which will be cheaper anyway)

That's a good point. I originally used std::regex_replace because I thought it would be more simple than std::string::replace, but turns out std::string::replace is not that much more complex. The code I write using std::string::replace is inspired by [the fixed YCSB redaction code for the same URL](https://github.com/mongodb-labs/YCSB/blob/8ad6e1fb5964654fef07c7045c1b199d380c2211/ycsb-mongodb/mongodb/src/main/java/com/yahoo/ycsb/db/MongoDbClient.java#L356-L359).

src/gennylib/src/PoolFactory.cpp
```diff
@@ -240,15 +240,18 @@ mongocxx::options::pool PoolFactory::makeOptions() const {
...
+    std::string username{*_config->get(OptionType::kAccessOption, "Username")};
     std::string password{*_config->get(OptionType::kAccessOption, "Password")};
-
-    auto redactedUriStr = std::regex_replace(uriStr, std::regex(password), "[REDACTED]" );
+    auto credentials = username + ':' + password + '@';
+    auto redactedCredentials = username + ':' + "[REDACTED]" + '@';
+    auto credPosition = uriStr.find(credentials);
+    auto redactedUriStr = credPosition == std::string::npos
+                        ? uriStr
+                        : uriStr.replace(credPosition, credentials.length(), redactedCredentials);
```
PyGeek03, line 250: @thessem Do you think it would be worth it to extract these lines into a helper method (like `std::string PoolFactory::getRedactedUri()`), which should be more testable, and add test for this method?

thessem:
I'm removing the review request from this one until it is out of draft :)

PyGeek03:
I decided that it's best to fix this using a more obvious logic: constructs the redacted URI from source, instead of redacting constructed actual URI.

lgfang (approved):

thessem (approved):

PyGeek03:
evergreen merge

PR #934, opened 2023-06-21
EVG-19909 Run cmake-test on macOS but skip flaky tests

lgfang:
**Jira Ticket:** [EVG-19909](https://jira.mongodb.org/browse/EVG-19909)  

**Whats Changed:**  

-   Split the test case "Actual Actor Example" to "Actual Actor Example" and  
    "Actual Actor Example Timing"
-   Modify 'run-genny cmake-test' to accept the option '&#x2013;regex-exclude'
-   Modify 'evergreen.yml' run cmake-test on macOS with '&#x2013;regex-exclude'

At the moment, the `f_cmake_test_ignore_flaky` function runs cmake-test with  
`--regex-exclude` to skip timing/random number related tests which are known to  
fail on macOS. In the future, should more such tests are discovered, please skip  
them similarly.  

**Patch testing results:**  
<https://spruce.mongodb.com/version/6492ab750ae6065cd41edb29/tasks>  
<https://spruce.mongodb.com/version/649a474457e85a45efc31d70/tasks>  
<https://spruce.mongodb.com/version/649a46c0d1fe074124687ff2/tasks>

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -377,7 +377,7 @@ functions:
       - -v
       - cmake-test
       - --regex-exclude
-      - "(sleepNonBlocking|Non-Blocking end|Actual Actor Example)"
+      - "(sleepNonBlocking|Non-Blocking end|Actual Actor Example Timing)"
 
   ##
   # Runs benchmarks via ctest.
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -116,7 +116,7 @@ buildvariants:
   run_on:
   - macos-1100
   tasks:
-  - name: tg_compile
+  - name: tg_compile_and_test_with_server_on_macos
 
 - name: macos-1100-arm64
   display_name: macOS Bug Sur arm64
@@ -124,7 +124,7 @@ buildvariants:
   run_on:
   - macos-1100-arm64
   tasks:
-  - name: tg_compile
+  - name: tg_compile_and_test_with_server_on_macos
 
 # - name: centos6-perf
 #   display_name: CentOS 6 for Performance
@@ -167,6 +167,10 @@ tasks:
 - name: t_cmake_test
   commands:
   - func: f_cmake_test
+
+- name: t_cmake_test_ignore_flaky
+  commands:
+  - func: f_cmake_test_ignore_flaky
 
 - name: t_benchmark_test
   commands:
@@ -269,6 +273,17 @@ task_groups:
   - t_integration_test_sharded
   - t_test_create_new_actor_script  # Must run this task last.
 
+- <<: *a_compile_and_test
+  name: tg_compile_and_test_with_server_on_macos
+  tasks:
+  - t_compile
+  - t_python_test
+  - t_cmake_test_ignore_flaky
+  - t_integration_test_single_node_replset
+  - t_integration_test_three_node_replset
+  - t_integration_test_sharded
+  - t_test_create_new_actor_script  # Must run this task last.
+
 - name: tg_compile
   max_hosts: 1
   tasks:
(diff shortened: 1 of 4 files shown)


thessem (approved):
Happy with this approach. In terms of API I like `regex-exclude` and the fact it doesn't scale well or have a lot of extra functionality is an indicator that "we shouldn't let the list get too long here". It passing on essentially the same cmake test parameter is nice.

I think this achieves the result of making an essentially "passing" build pass without lying to the user. MacOS is not a good platform for performance testing on _because_ those tests fail, we shouldn't hide it, but for every day development it doesn't matter.

PyGeek03:
1. From EVG-19909, a flaky test investigated is "value_generators_test:YAML". It seems like the regex doesn't have this test?
2. Have we investigated why we have to skip the "Actual Actor Example" tests?

PyGeek03 (changes requested):
1. From [EVG-19776](https://jira.mongodb.org/browse/EVG-19776), a flaky test investigated is "value_generators_test:YAML". It seems like the regex doesn't have this test?
2. Have we investigated why we have to skip the "Actual Actor Example" tests?

lgfang:
>  From [EVG-19776](https://jira.mongodb.org/browse/EVG-19776), a flaky test investigated is "value_generators_test:YAML". It seems like the regex doesn't have this test?

According to my test, this one never fails when run by evergreen.

> Have we investigated why we have to skip the "Actual Actor Example" test, and whether we should do that? If the issue is with a specific subtest within this test, should we split that subtest out to be its own test and ignore only that test?

Yes, this one fails due to timing similar to the sleepNonBlocking .

Given that the goal of this ticket is very if the "essential" build tests succeed, I think it is ok to skip such tests.

PyGeek03 (approved):
> > From [EVG-19776](https://jira.mongodb.org/browse/EVG-19776), a flaky test investigated is "value_generators_test:YAML". It seems like the regex doesn't have this test?
> 
> According to my test, this one never fails when run by evergreen.

It's interesting how our own Mac ARM laptops are also different from Evergreen's. But yeah I agree with you on this. Let's keep the test on Evergreen as a documentation of this fact.
 
> > Have we investigated why we have to skip the "Actual Actor Example" test, and whether we should do that? If the issue is with a specific subtest within this test, should we split that subtest out to be its own test and ignore only that test?
> 
> Yes, this one fails due to timing similar to the sleepNonBlocking .
> 
> Given that the goal of this ticket is very if the "essential" build tests succeed, I think it is ok to skip such tests.

I don't know if the "Actual Actor Example" test is essential or not, but I trust your judgement on this one.

Thank you for this PR, I really like your solution of using the regex-exclude feature in ctest to implement this!

PyGeek03 (approved):
This looks really awesome, thank you Lungang!

lgfang:
Thanks for Minh's suggestion, I've split the "Actual Actor Example" to "Actual Actor Example" and "Actual Actor Example Timing" and ignore the later only.

evergreen.yml
```diff
@@ -349,6 +364,21 @@ functions:
...
+  ##
+  f_cmake_test_ignore_flaky:
+  - command: subprocess.exec
+    params:
+      binary: ./src/genny/run-genny
+      args:
+      - -v
+      - cmake-test
+      - --regex-exclude
+      - "(sleepNonBlocking|Non-Blocking end|Actual Actor Example Timing)"
```
rtimmons, line 380: Consider a standard test name e.g. IgnoredOnMacOs so the tests can be marked ignored more explicitly and we don't accidentally break CI by renaming tests.

lgfang: Just to confirm I understand you correctly, you mean renaming "Acutual Actor Example Timing" to "Actual Actor Example IgnoredOnMacOS"?

rtimmons: Yes. That puts the test together with the specification to skip on macos. Again, just a suggestion.

lgfang: Good idea. It will also make the regex simpler.
Since I've already merged this PR, I'm creating a PR for this renaming.

PR #939, opened 2023-06-29
[PERF-2932] - Commit latency updates to include ReadConcern "Available"

MAhmadShah:
**Jira Ticket:** [PERF-2932](https://jira.mongodb.org/browse/PERF-2932)

**Whats Changed:**  
We have added in 4 new phases to the commit latency workload to cover testing of the "available" ReadConcern option. We are testing this with W:1, W:0, journaling and readMode as primary and secondary.

**Patch testing results:**  
A successful patch can be found here:
https://spruce.mongodb.com/version/649b294ad1fe07b787b40405/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/networking/CommitLatency.yml
+++ b/src/workloads/networking/CommitLatency.yml
@@ -76,9 +76,6 @@ Actors:
   - *DropCollection
   - *Load
   - *Nop
-  - *DropCollection
-  - *Load
-  - *Nop
 
 
 - Name: CommitLatency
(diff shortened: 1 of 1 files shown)


jawwadasghar (approved):
LGTM
I was kind of surprised the total duration of the test didn't seem to change much from what I see in the waterfall for older tasks, but the metrics do show up, so I guess we are good! 
Might be worth giving a heads up to build barons anyway.

MAhmadShah:
evergreen merge

PR #941, opened 2023-07-05
PERF-4369 parallel inserts for clustered collection workloads

cswanson310:
*Whats Changed:**  
Used 16 threads to load the data instead of 1.

**Patch testing results:**  
This patch includes some other unrelated changes and also comments out the later phases which I was doing while testing myself, but does prove that the insert phase can be parallelized and does work: https://spruce.mongodb.com/version/649f83385623435ea6cd12cb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/phases/execution/ClusteredCollection.yml
+++ b/src/phases/execution/ClusteredCollection.yml
@@ -12,6 +12,8 @@ GlobalDefaults:
   Collection: &Collection Collection0
   Namespace: &Namespace test.Collection0
   DocumentCount: &docCount 1000000
+  LoadingThreads: &LoadingThreads 16
+  DocsPerThread: &DocsPerThread 62500
   MaxPhases: &MaxPhases 9
 
 ValueGenerators:
@@ -73,10 +75,9 @@ Actors:
             key: {_id: "hashed"}
             numInitialChunks: 12
 
-# Phase 2: insert. TODO Make multithreaded once TIG-2938 is resolved.
 - Name: Insert
   Type: Loader
-  Threads: 1
+  Threads: *LoadingThreads
   Phases:
     OnlyActiveInPhases:
       Active: [2]
@@ -84,14 +85,14 @@ Actors:
       PhaseConfig:
         Repeat: 1
         Database: *Database
-        Threads: 1
+        MultipleThreadsPerCollection: true
         CollectionCount: 1
         DocumentCount: *docCount
         BatchSize: 1000
         Document:
           # Start from {_id: "0000001"} in order to be able to set the ^RandomInt max bound to *docCount
           # in the LookupByClusterKey phase. This works around TIG-3759.
-          _id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: 1}}]}}, *PaddingGenerator]}}
+          _id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: -62499, multiplier: *DocsPerThread}}]}}, *PaddingGenerator]}}
           a: &RandomSecondaryKey {^FastRandomString: {length: 6, alphabet: "0123456789"}}
           b: {^FastRandomString: {length: 1024}}
 


src/phases/execution/ClusteredCollection.yml
```diff
@@ -73,25 +75,24 @@ Actors:
...
-        Threads: 1
+        MultipleThreadsPerCollection: true
         CollectionCount: 1
         DocumentCount: *docCount
         BatchSize: 1000
         Document:
           # Start from {_id: "0000001"} in order to be able to set the ^RandomInt max bound to *docCount
           # in the LookupByClusterKey phase. This works around TIG-3759.
-          _id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: 1}}]}}, *PaddingGenerator]}}
+          _id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: -62499, multiplier: *DocsPerThread}}]}}, *PaddingGenerator]}}
```
naama-bareket, line 95: If I'm not mistaken since this is the 3rd phase, the multiplier will be multiplied by 3 right? So I think the start should account for that -187499? -187499 + 3 * 62500 = 1

josefahmad (commented):
I'd like clarification on why we're parallelising the data ingestion phase. Is the goal to reduce runtime, are we more interested in multi-threaded insert performance metrics? Have we looped in someone from the performance team to understand how this change is going to affect change point detection?

EDIT: I see we're probably addressing the TODO on the now-resolved TIG-2938. Still, I'd like to clarify if this is the right thing to do at this point.

cswanson310:
> I'd like clarification on why we're parallelising the data ingestion phase. Is the goal to reduce runtime, are we more interested in multi-threaded insert performance metrics? Have we looped in someone from the performance team to understand how this change is going to affect change point detection?
> 
> EDIT: I see we're probably addressing the TODO on the now-resolved TIG-2938. Still, I'd like to clarify if this is the right thing to do at this point.

Sorry for the long delay! This one fell of my radar.

I think the goal is primarily to reduce runtime. However, to the extent we are interested in metrics about insert performance, I can't imagine we are specifically interested in single-threaded performance. Maybe for debugging, but multi-threaded insert perf is the more interesting and probably more representative workload, right?

So while I am doing this to reduce the runtime, I am also doing so with the opinion that it makes it a better performance test overall.

src/phases/execution/ClusteredCollection.yml
cswanson310: You are absolutely right, thanks! Just confirmed after changing:
```
MongoDB Enterprise > db.Collection0.find({}).sort({_id: 1}).limit(1)
{ "_id" : "0000001", "a" : "841163", "b" : "osPFriKJjijnLNDnsHAELMqjKgDLnJiHMjAMtErqKjCuBNrjILjFlqkEAvOtgpDMqODqpiMrAGttsABIhtCMDmLkjqrirhlmLuLoFNjmlFDtGnrEBLKIJnjPCMrAqCNvFmopgBIHGGrklhDpPrvEgsCOrEAviKqHLKECqGmGLAACDAviHsFPoLLJFGHLjmgngAMkKErsmNFuMGsGHkoCHgiOhGvmttiopgBEBFLLDCIluAjIOCGlhAtkgpjKHIBJmHBKnImrnvMkrGNtnGjgLrjDruPvlCtqBrFnllJPHjMiCsLgDnjHHvohAvkKnqrBiuLPgivPOsJHgjpgnujKtiFKNmtEKiMCPMBMgLnDMqiKMngDMOtuigKtMkoKrHsKPkDhJMEgotCAlnEmjEruDsuivpAtJLlMCgrrvohCNGjCIEruiuGimrJNoCumrgGCOiLnqiJhokmulIAvvmEGsBoGPpNmErspikAlHskrJLImuINmhFgIPOhIgkKvAkCiNBgANMglhHmpAKDgBtptGGEjLjMjBJnjniGpGhrAHDvOGshHKHNFNBtLmNgtOAIPmtiIGOqpnGuvtCuGmhumtJiIhluFJhtjEtGGLBusHrLpiAmFOJACsJDrKuiJIjBnHCJmkOqDJluvPANCjlEhKqNGlrsjJDvMpgDtBiJDlAvpAHvOuoECmNHnFGmBghnODuphqmsvmoMFrgLDinmlDBAGCIBFMvrLFvpmDhlEHjLiKKFnnpqDtJvEANhKutHppAqmjBIiNvOrnJPuPALPutJCPKCAKkMoCpEEDMjJPgMqjolvqsGFMAsBvmjMjJphoALOMIoCjGJosoLvItiLAKlnmFCIFDHJmiFNlgPqtgooOpoCmPpsgAskgkJitJqAnDtopLJsKIhMnvvOkgjpGIMGpAPmoqJvtBNvnGmECPtInspOKorkqphHDnoqKOIggulKuqFlrriMAiBIGiiMBFEsDvrCiqhoHNJuLNtIDAgLADDOjrAKlDtFvhMpKMHJ" }
MongoDB Enterprise > db.Collection0.find({}).sort({_id: -1}).limit(1)
{ "_id" : "1000000", "a" : "545650", "b" : "CngnMCqnPkIrItOHNLlIhjMHOrnrmMuqqMHpKGBClAPlrhpGIPJGuAICuqDMqBMFvFPGHjBEiLhNFGEuokKnCmLivojiOjupgpLDkNCgtMiupOpKlqoIqhtDGrBrtogGmuGsBpsKEhBhgKvtCIhNhvChNtJLGPIrABnPvNEgLomEBhtkMiKtAJHKLujniHnpMtgOpBMDAqhLMHEHAKGMpuoIqvjjqGljEKElGDgEHLAKIHrgogpFtoCvrHDiMiDMKjLIoPinBJGgPvhGPKljMKLlCKvMuttOtJElHEAEnLhnDOpoumhEtPhPmPJkmIjHtBKPJkLIgIKPlkjvsvAMAumgtOuCiqCiluDKqtFjMFrlFkmunPmKGhkGnlEovBvmvNOOmFLtnnIvPPmgnOgkEnFkKogKvgPuliBKCAJMDFEqLsPJmrvGhpPKFqHiogmtLhnBIprmEOCCuErrnkMjConGCvOFMMNLvLOAKtCrjLDIBGOCkjhlhnKtGEtqOGDgCDuHMONtFOKvJMqMBtthLmKpDPhErBClCIuMFIKGpCAPiBLDlEIrPHgAjuJuODmElnkmHNKrjIJpqJNLKHGrCPmrBJsLBBvnhmMFkmvLEHKBKrOjvuNmjHGvjiFlEiMPoGIFvONkOMLljPHIgNnHgGrgJmgnsnototGFrOmHpoBouPNhpritntqtjCtKulHCCCBjLDEvkrhDMqioDFGOvDMLGhngqLIOKoEhkDqkDIMEglEFAGkBoiImNAELIILjDukKkHuErAJqoINDkjEgLqGglGEgJmCjpKGqGFEorDDrLtphKjNELoLEgjMgGAtvlOpIhPBkLGEJJpFpLltEvijoIhDOnqFMpMIiqgluuKjooIIolpEDDOGEANmMtiIknoJlkAkhIvmDtshFDDGhMBEEEiGEtgLMluuEGHokCmAFAqhqsLBPmuiDHuoHOMNPqMqosikqsvoCEFOhvMmMpGrNNNnGHAArCOluutFjJqsqnDoJrpMCJOHDqtkhgoJn" }
```

src/phases/execution/ClusteredCollection.yml
```diff
@@ -73,25 +75,34 @@ Actors:
...
+        MultipleThreadsPerCollection: true
         CollectionCount: 1
         DocumentCount: *docCount
         BatchSize: 1000
         Document:
           # Start from {_id: "0000001"} in order to be able to set the ^RandomInt max bound to *docCount
           # in the LookupByClusterKey phase. This works around TIG-3759.
-          _id: {^Join: {array: [{ ^FormatString: {"format": "%07d", "withArgs": [{^Inc: {start: 1}}]}}, *PaddingGenerator]}}
+          # The multiplier uses the actor ID so we start at -3*DocsPerThread + 1 to get to 1 as the
+          # first _id.
```
josefahmad, line 96: If I'm interpreting the below and [this thread](https://github.com/mongodb/genny/pull/941#discussion_r1253616420) correctly, the `multiplier` is an offset that is added to `start` and is computed as the `multiplier` value multiplied by the phase id, where the current phase is phase 3.

If the above is accurate, I suggest:
```suggestion
          # 'multiplier' is an offset to 'start', and the 'multiplier' value itself is multiplied by the phase id (3).  So we start at -3*DocsPerThread + 1 to get to 1 as the first _id.
```
And I also suggest refreshing the ["Phase" comments](https://github.com/mongodb/genny/pull/941/files#diff-7d05c7e5cc91b2536c056d4dde59005a6d5a1fe8320ca2715abe6d203c9ef678R36) that seem to be off by 1.

josefahmad (approved):
LGTM % suggesting an in-line doc improvement. Have we looped in someone from sys-perf so they're aware of a potential blip in the change point detection?

src/phases/execution/ClusteredCollection.yml
cswanson310: I can rephase the comment as you suggest but it's not quite right to say phase ID. It's by actor ID not phase ID. The idea is that each thread is a separate actor, so we need actor ID not phase ID to ensure they don't create duplicate ID values.

I think @naama-bareket mentioned 'phase 3' just to say that "there are at least 2 other actors that went before this", and often the setup phases are a single actor, so the two are sometimes interchangeable.

cswanson310:
Going to merge this now! I posted in #perf-build-baroning warning of the upcoming change.

cswanson310:
evergreen merge

PR #936, opened 2023-06-23
PERF-4207: Add IDHACK workload(s) to genny

naama-bareket:
Thanks for submitting a PR to the Genny repo. Please include the following fields (if relevant) prior to submitting your PR.

**Jira Ticket:** PERF-4207

**Whats Changed:**  
Added workload which should hit the IDHACK path with various _id types, sizes of documents, and threads

PR #942, opened 2023-07-05
PERF-4372 Improve throughput reporting of analysis tool

cswanson310:
**Whats Changed:**  
We previously export the results to csv which does not have a high enough precision timestamp to do the throughput calculation using the clock. Using the clock seemed to have much better results.

**Patch testing results:**  
N/A. Here's a sample of comparing the new math to the old math:
```
Processing expanded JSON results: 1131538it [00:06, 175983.35it/s]
LookupByClusterKey.Crud summary:
    timers.dur (measured in nanoseconds, displayed in milliseconds):
        count     : 1131538 
        average   : 0.8407  
        median    : 0.8115  
        mode      : 0.8065  
        stddev    : 0.5816  
        [min, max]: ['0.2907', '214.6']
        histogram:
            [  0, 15): ************************************************************...  (1131434)
            [ 15, 29): *************************************************    (49)
            [ 29, 43): ***************************  (27)
            [ 43, 57): **********   (10)
            [ 57, 72): *******  (7)
            [ 72, 86): **   (2)
            [ 86,100): **   (2)
            [100,115): **   (2)
            [115,129): *    (1)
            [129,143): **   (2)
            [143,157):  (0)
            [157,172):  (0)
            [172,186):  (0)
            [186,200):  (0)
            [200,215]: **   (2)
    throughput:
        ops       : 1131538.0
        seconds   : 951.252338361
        ops per second: 1189.5245
    throughput_new:
        ops       : 1131538.0
        seconds   : 59.999  
        ops per second: 18859.281

```
Notice that 'throughput_new' has 59.999 seconds, and this workload had 'Repeat: 1 minute', so that is looking great to me.

**Related PRs:**   
https://github.com/mongodb/genny/pull/941
https://github.com/mongodb/genny/pull/936


src/workloads/contrib/analysis/test_result_summary.py
```diff
@@ -77,9 +83,17 @@ def replace_suffix(string, suffixTarget, newSuffix):
     suffix_start = string.rfind(suffixTarget)
     return string[0:suffix_start] + newSuffix
 
+def animate(interrupt_event):
```
cswanson310, line 86: It turned out to be much slower to do it this way than to use the CSV format... but rather than fix that I just added a loading bar... It still doesn't take more than a handful of seconds which I think is acceptable. I was chatting with @naama-bareket and I do think we could do better by directly processing the '.ftdc' file, but that would be a more major re-write since it's a BSON-direct file format so we'd be messing with bits more and need to use some bson library.

wbuerger46: Yep, I think a handful of seconds is find given the complexity of working with .ftdc directly. And a loading bar will keep us patient :)

wbuerger46 (approved):
Nice! This script is so useful - thanks for the upkeep @cswanson310!

src/workloads/contrib/analysis/test_result_summary.py
naama-bareket: Agreed!

naama-bareket (approved):
LGTM! Yay for accurate throughput

cswanson310:
After getting some confusing patch failures ([example](https://spruce.mongodb.com/version/64a5d04bd1fe07d5c104378e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)), @rtimmons helped me figure out what to change and I updated the python handling here. Unfortunately the "test" in the name of the file was making it run the auto tests in a way that wasn't helpful, so while it's not ideal but I had to rename the script.

cswanson310:
Update: this PR may prevent us from having to rename it: https://github.com/mongodb/genny/pull/943

PR #943, opened 2023-07-06
EVG-20383 Explicitly enumerate pytest search paths

rtimmons:


thessem (approved):
Thanks for this. I think there's some tension here in that the `pyproject.toml` and the `requirements.txt` files are not in the same directory. This will likely be cleaned up in any project modernisation push though, and I think the `pyproject.toml` is in the corrext place (top-level).

I had a local version of this change, but I didn't have the second part of this (the change in tasks). This PR solves a problem I've been having. Thanks :)

PR #947, opened 2023-07-18
PERF-4371 More workloads on query stats

cswanson310:
Adds more variants into the mix for the query-stats comparisons

**Patch testing results:**  [🌲 link](https://spruce.mongodb.com/version/64b6fade61837d9ec98d9b4b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/execution/ClusteredCollection.yml
+++ b/src/workloads/execution/ClusteredCollection.yml
@@ -26,6 +26,8 @@ AutoRun:
       - replica
       - replica-all-feature-flags
       - replica-query-stats
+      - atlas-like-replica.2022-10
+      - atlas-like-replica-query-stats.2022-10
       - shard-lite-all-feature-flags
       - shard
       - shard-query-stats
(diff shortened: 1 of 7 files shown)


src/workloads/execution/ClusteredCollection.yml
```diff
@@ -21,8 +21,15 @@ AutoRun:
       $eq:
       - standalone
       - standalone-all-feature-flags
+      - standalone-query-stats
+      - standalone-query-stats-small-rate-limit
       - replica
       - replica-all-feature-flags
+      - replica-query-stats
+      - atlas-like-replica.2022-10
```
wbuerger46, line 29: Why are we adding non-query stats workloads too? (`shard` and `atlas-like-replica.2022-10`)

wbuerger46 (approved):
LGTM with one question!

naama-bareket (approved):
lgtm!

src/workloads/execution/ClusteredCollection.yml
cswanson310: Oh sorry I never responded to this! This is because we use the non-query-stats version as the baseline performance, then compute the diff to see how we compare. That's what's plotted in all our graphs on the dashboard.

PR #950, opened 2023-07-20
PERF-4422 Add genny workload for transaction with large number of namespaces

Fefer-Ivan:
**Jira Ticket:** PERF-4422

**Whats Changed:**  
Added new MultiShardTransactionWithManyNamespaces workload - that runs multi shard transaction across 50 namespaces.
Added QuiesceActor to MultiShardTransaction workload.

**Patch testing results:**  
https://spruce.mongodb.com/version/64b938223066154b1c206446/tasks


src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
```diff
@@ -0,0 +1,237 @@
...
+  UpdateOperationDatabase0: &UpdateOperationDatabase0
+    <<: *UpdateOperation
+    Database: *Database0
+
+  UpdateOperationDatabase1: &UpdateOperationDatabase1
+    <<: *UpdateOperation
+    Database: *Database1
+
+Actors:
+- Name: CreateShardedCollection
```
cheahuychou, line 32: Did you mean to shard the collections also?

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
```diff
@@ -0,0 +1,237 @@
...
+    Threads: 4
+    BatchSize: 1000
+    DocumentCount: *DocumentCount
+    CollectionCount: 25
+    Database: *Database1
+    Document: *Document
+  - *Nop
+  - *Nop
+
+- Name: QuiesceDatabase0
```
cheahuychou, line 82: [question] Could you explain the reason for doing this? I saw the header comment [here](https://github.com/mongodb/genny/blob/54e18a79d2836675e1a9f967c4cea2ca15f46991/src/cast_core/include/cast_core/actors/QuiesceActor.hpp#L31) but I'm still unclear on that.

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
Fefer-Ivan: No, I just need to make two different primary shards so the transactions are guaranteed to be multi-shard

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
Fefer-Ivan: Basically it does fsync for all cluster after we are finished writing data and before we start querying it to make sure the state of the cluster is more consistent between runs by the time we start executing queries.

Fefer-Ivan: The fsync command forces the mongod process to flush all pending writes to the storage layer

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
```diff
@@ -0,0 +1,237 @@
...
+  DocumentCount: &DocumentCount 100_000
+
+  MaxPhase: &MaxPhase 4
+
+  RandomIdFilter: &RandomIdFilter {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
+
+  UpdateOperation: &UpdateOperation
+    OperationName: updateOne
+    OperationCommand:
+      Filter: *RandomIdFilter
```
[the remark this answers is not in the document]
Fefer-Ivan: I double-checked that after the workload we affected all 50 collections and that in each collection random documents where affected.

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
cheahuychou: Then I think this should be renamed to `CreateDatabases` or something along that line since it doesn't create any (sharded) collections.

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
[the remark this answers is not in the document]
cheahuychou: I see. Thanks for verifying.

src/workloads/sharding/MultiShardTransactions.yml
```diff
@@ -74,6 +74,30 @@ Actors:
...
+      PhaseConfig:
+        Repeat: 1
+
+- Name: QuiesceDatabase1
+  Type: QuiesceActor
+  Threads: 1
+  Database: *Database1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [3]
```
cheahuychou, line 96: Should `QuiesceDatabase1` be done in a phase after `QuiesceDatabase0` like how is done in `src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml`?

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
```diff
@@ -0,0 +1,233 @@
...
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhase
+      PhaseConfig:
+        Repeat: 1
+        Database: admin
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            enableSharding: *Database0
```
cheahuychou, line 45: Since the goal is to measure the impact of writing the `"affectedNamespaces", should we make the total size of this field larger. Based on [this](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names) and [this](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Length-of-Database-Names), the max size for the namespace of an unsharded collection is 255 bytes and the max size of the database name is 63 bytes. I know that each transaction is already acting on 25 namespaces, but I think we might want to make increase the length of the namespaces for the test collections. Any thoughts?

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
Fefer-Ivan: Renamed the stage

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
Fefer-Ivan: I can make database name larger, but unfortunatelly, database loader have hard-coded collections names as Collection0, Collection1, ... CollectionN.

cheahuychou (approved):
LGTM

src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml
Fefer-Ivan: Made database name 63 chars long.

We can add another something like CollectionNamePrefix to Loader/MonotonicLoader class, so we can do larger collection. However, I feel like this should be a separate ticket.
Maybe instead we can increase the number of collections?

Currently we have 50 collections with CollectionXX (12 bytes) name and 63 bytes database name, so (12 + 63) * 50 = 3750 extra bytes per transaction.

How much extra space do you want to test? 4KB? 16KB?

cheahuychou: > How much extra space do you want to test? 4KB? 16KB?

I'm not sure.

> Currently we have 50 collections with CollectionXX (12 bytes) name and 63 bytes database name, so (12 + 63) * 50 = 3750 extra bytes per transaction.

This is probably fine. Maybe we should aim for the realistic worst case.

Fefer-Ivan: A googled to double-check: typical size of a block on a modern SSD and HDD is 4kb, so let's try to hit that. It is still an arbitrary constant, but I will raise the number of collections per database to 30, instead of 25.

PR #956, opened 2023-07-27
SERVER-78580 projection parsing workload

mattBoros:
**Whats Changed:**  
Added a workload for $project parsing performance.

**Patch testing results:**  
https://spruce.mongodb.com/version/64c2dcce3627e000ccaa3665/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

**Workload Submission form:**  
I've submitted this form.

**Related PRs:**   
https://github.com/10gen/mongo/pull/14542

### DIFF
--- a/src/workloads/query/ProjectParse.yml
+++ b/src/workloads/query/ProjectParse.yml
@@ -67,7 +67,7 @@ Actors:
         aggregate: *coll
         pipeline: [
           {$project : {^Object: {
-              withNEntries: 10000,
+              withNEntries: 20000,
               havingKeys: {^RandomString: {length: 10}},
               andValues: 1,
               duplicatedKeys: skip
--- a/src/workloads/query/ProjectParse.yml
+++ b/src/workloads/query/ProjectParse.yml
@@ -58,7 +58,7 @@ Actors:
             }}}
         ]
         cursor: {}
-  - Repeat: 100
+  - Repeat: 50
     Database: *db
     Operations:
     - OperationMetricsName: ProjectParseLarge
--- a/src/workloads/query/ProjectParse.yml
+++ b/src/workloads/query/ProjectParse.yml
@@ -1,0 +1,91 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Test $project parsing performance for small and large number of fields.
+
+GlobalDefaults:
+  dbname: &db test
+  coll: &coll Collection0
+  nop: &Nop {Nop: true}
+
+Actors:
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: *db
+    Collection: *coll
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10
+    BatchSize: 10
+    Document:
+      m: {^Inc: {start: 0}}
+  - *Nop
+  - *Nop
+  - *Nop
+
+- Name: Quiesce
+  Type: QuiesceActor
+  Threads: 1
+  Database: *db
+  Phases:
+  - *Nop
+  - Repeat: 1
+  - *Nop
+  - *Nop
+
+- Name: Queries
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10000
+    Database: *db
+    Operations:
+    - OperationMetricsName: ProjectParseSmall
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *coll
+        pipeline: [
+          {$project : {^Object: {
+              withNEntries: 10,
+              havingKeys: {^RandomString: {length: 10}},
+              andValues: 1,
+              duplicatedKeys: skip
+            }}}
+        ]
+        cursor: {}
+  - Repeat: 100
+    Database: *db
+    Operations:
+    - OperationMetricsName: ProjectParseLarge
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *coll
+        pipeline: [
+          {$project : {^Object: {
+              withNEntries: 10000,
+              havingKeys: {^RandomString: {length: 10}},
+              andValues: 1,
+              duplicatedKeys: skip
+            }}}
+        ]
+        cursor: {}
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+      - standalone-query-stats
+    branch_name:
+      $gte:
+      - v7.1
--- a/src/workloads/query/ProjectParse.yml
+++ b/src/workloads/query/ProjectParse.yml
@@ -68,7 +68,7 @@ Actors:
         pipeline: [
           {$project : {^Object: {
               withNEntries: 20000,
-              havingKeys: {^RandomString: {length: 10}},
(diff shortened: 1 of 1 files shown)


src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Test $project parsing performance for small and large number of fields.
```
dalyd, line 4: Could you add keywords in addition to the description? For example: https://github.com/mongodb/genny/blob/1dbcd0ba8b678721774787c31310fa4ec0a522fd/src/workloads/sharding/DistinctCommands.yml#L14

We will use those for general workload search.

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+        pipeline: [
+          {$project : {^Object: {
+              withNEntries: 10,
+              havingKeys: {^RandomString: {length: 10}},
+              andValues: 1,
+              duplicatedKeys: skip
+            }}}
+        ]
+        cursor: {}
+  - Repeat: 50
```
dalyd, line 61: Why 50? Would you prefer to run it for  a set period of time instead?

src/workloads/query/ProjectParse.yml
dalyd: Also, could you include a little more context. For example this text from your workload form is wonderful context for this workload 

> For large $projects it could take hours to parse previously. A good expected value would be milliseconds for each run.

Also, what was it about the $project that previously made it so slow? Feel free to link into relevant server tickets as well.

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: *db
+    Collection: *coll
+    Threads: 1
+    CollectionCount: 1
+    DocumentCount: 10
```
dalyd, line 21: You only need 10 small documents to show this issue? Wow. 
Going forward would a larger dataset be more interesting, or do we only care about this tiny size?

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+  - *Nop
+  - *Nop
+
+- Name: Queries
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - Repeat: 10000
```
dalyd, line 45: Do you want Repeat or Duration? If there's something interesting about the repeat, let's document that in the workload.

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: ProjectParseSmall
+      OperationName: RunCommand
```
dalyd, line 49: The crud actor supports aggregations now thanks to David Storch and I think it's preferred. https://github.com/mongodb/genny/commit/bba39f66d82d02bfc9c064e8b3ec688e3a76f46a

If there's a reason that we can't use the Crud Actor here, let's document that also.

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+    Database: *db
+    Operations:
+    - OperationMetricsName: ProjectParseLarge
+      OperationName: RunCommand
+      OperationCommand:
+        aggregate: *coll
+        pipeline: [
+          {$project : {^Object: {
+              withNEntries: 20000,
+              havingKeys: {^RandomString: {length: 10, alphabet: "abcdefghijklmnopqrstuvwxyz"}},
```
dalyd, line 71: You might want to use FastRandomString here. It's a lower quality number generator, but faster.

I have some concern that you are going to saturate the workload client generating the random data for this pipeline. You could compare the average latency to the throughput on this one to see if they match up. I think data generation time should not show up in the average latency, so you can check if OperationThroughput = 1/Average Latency.

src/workloads/query/ProjectParse.yml
```diff
@@ -0,0 +1,91 @@
...
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+      - standalone-query-stats
+    branch_name:
+      $gte:
+        v7.1
```
dalyd, line 91: Discussed in email. What happens if we run this on 7.0?

dalyd:
One more thought -- I would love to have the workload in and running before the server change (if it will run). I love being able to see a clear improvement in performance.

src/workloads/query/ProjectParse.yml
mattBoros: This came from an issue with parsing the query, so the document count is purposely small to emphasize the parsing rather than execution of the query.

src/workloads/query/ProjectParse.yml
mattBoros: It would be similar results as master. I can change this to >=7.0

src/workloads/query/ProjectParse.yml
mattBoros: Ah yea 50 was sort of arbitrarily chosen, I think setting a time would be better. I'll change this

src/workloads/query/ProjectParse.yml
mattBoros: Sounds good! I can add keywords and more details in the description.

src/workloads/query/ProjectParse.yml
mattBoros: Duration would be great actually.

src/workloads/query/ProjectParse.yml
mattBoros: CRUD would work fine I believe

src/workloads/query/ProjectParse.yml
mattBoros: What if we wrapped this with ^FixedGeneratedValue? I think that would be a good workaround.

mattBoros:
Sounds good! I'll make sure this goes in first.

mattBoros:
I think the only open question is whether using ^FixedGeneratedValue would work instead of ^FastRandomString.

src/workloads/query/ProjectParse.yml
dalyd: Where would you put the FixedGeneratedValue? Would you wrap the whole ^Object? If you did that the first execution still needs to generate all 20k entries and could still be slow. Would that be good enough? I don't think you can wrap the RandomString call itself, because then you would get 20k identical keys.

src/workloads/query/ProjectParse.yml
```diff
@@ -1,13 +1,21 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  Test $project parsing performance for small and large number of fields.
+  Test $project parsing performance for small and large number of fields. We previously had a quadratic
```
dalyd, line 4: Could you put a date in here to go with the "previously". Maybe a parenthetical "(before August 2023)"?

src/workloads/query/ProjectParse.yml
dalyd: Thank you

dalyd (approved):
Just a nit. Thanks!

src/workloads/query/ProjectParse.yml
mattBoros: I was thinking around the ^Object would be good. Maybe both would be beneficial. The first run would create 20k fields using ^FastRandomString and reuse that result for the rest of the runs.

mattBoros:
evergreen merge

PR #957, opened 2023-08-01
PERF-4478 Generator for an array of distinct values

wqian94:
**Jira Ticket:** PERF-4478

**Whats Changed:**  
Add the `DistinctArray` generator, which is essentially a set, but as an array with no duplicate values. Unlike `Array`, `DistinctArray` operates solely on homogeneous array values.

Also implements a refactor for the parsers maps to deduplicate maps and hedge against the risk of accidentally leaving a parser off of `allParsers`.

**Patch testing results:**  
If applicable, link a patch test showing code changes running successfully (TBD)

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -101,7 +101,6 @@ public:
     virtual ~Appendable() = default;
     virtual void append(const std::string& key, bsoncxx::builder::basic::document& builder) = 0;
     virtual void append(bsoncxx::builder::basic::array& builder) = 0;
-    virtual void appendDistinct(bsoncxx::builder::basic::array& builder, size_t nDistinct) = 0;
 
 protected:
     // Custom hasher to enable hashing for BSON types.
@@ -158,44 +157,6 @@ public:
     }
     void append(bsoncxx::builder::basic::array& builder) override {
         builder.append(this->evaluate());
-    }
-    void appendDistinct(bsoncxx::builder::basic::array& builder, size_t nDistinct) override {
-        std::unordered_set<T, SetHasher, SetKeyEq> distinctValues(2 * nDistinct);
-        // Choose a buffer length of max(nDistinct, 64).
-        const auto bufferLength = std::max<int64_t>(64, nDistinct);
-        // Keep track of collision offsets for next N runs.
-        std::deque<int8_t> offsets(bufferLength);
-        ssize_t collisions = 0;
-
-        /**
-         * Sliding window via cyclic buffer.
-         *
-         * This computes the number of insertion failures, i.e. repeated values, in the past
-         * `bufferLength` number of generated values. We achieve this by queueing up the
-         * inverse operation for N steps away. We set N to be >= 64 so that small counts are
-         * given more generous resources (very small N corresponds to higher variance). If
-         * all of the past N generated values are repeats, then we assume that we will not
-         * converge in a reasonable number of iterations.
-         */
-        while (distinctValues.size() < nDistinct) {
-            auto [itr, succeeded] = distinctValues.insert(this->evaluate());
-
-            // Iterate through cyclic buffer and update collisions counter and offsets.
-            collisions -= offsets.front();
-            offsets.pop_front();
-            offsets.push_back(!succeeded);  // Relies on bool translating into 1 or 0.
-
-            if (succeeded) {
-                builder.append(*itr);
-            } else if (++collisions >= bufferLength) {
-                // Rate of failure: 100% of the past N pulls.
-                std::stringstream msg;
-                msg << "All " << bufferLength << " most recently generated values are "
-                    << "duplicates. Assuming unique value convergence failure for node: "
-                    << _node;
-                BOOST_THROW_EXCEPTION(ValueGeneratorConvergenceTimeout(msg.str()));
-            }
-        }
     }
 };
 }  // namespace genny
@@ -323,9 +284,6 @@ std::unique_ptr<DocumentGenerator::Impl> documentGenerator(const Node& node,
 template <bool Verbatim>
 UniqueGenerator<bsoncxx::array::value> literalArrayGenerator(const Node& node,
                                                              GeneratorArgs generatorArgs);
-
-UniqueGenerator<bsoncxx::array::value> distinctArrayGenerator(const Node& node,
-                                                              GeneratorArgs generatorArgs);
 
 UniqueGenerator<bsoncxx::array::value> bsonArrayGenerator(const Node& node,
                                                           GeneratorArgs generatorArgs);
@@ -1452,13 +1410,48 @@ public:
           _node{node},
           _generatorArgs{generatorArgs},
           _valueGen{valueGenerator<false, UniqueAppendable>(node["of"], generatorArgs, parsers)},
-          _nTimesGen{intGenerator(extract(node, "number", "^Array"), generatorArgs)} {}
+          _nTimesGen{intGenerator(extract(node, "number", "^Array"), generatorArgs)},
+          _distinct {node["distinct"] ? node["distinct"].maybe<bool>().value_or(false) : false} {}
 
     bsoncxx::array::value evaluate() override {
         bsoncxx::builder::basic::array builder{};
         auto times = _nTimesGen->evaluate();
-        for (int i = 0; i < times; ++i) {
-            _valueGen->append(builder);
+        if (_distinct) {
+            // Choose initial bucket size of 2N to minimize hash collisions and resizes.
+            std::unordered_set<bsoncxx::array::value, SetHasher, SetKeyEq> distinctValues(2 * times);
+            int8_t consecutiveRepeats = 0;
+            while (distinctValues.size() < times) {
+                bsoncxx::builder::basic::array sbuilder{};
+                _valueGen->append(sbuilder);
+                auto [itr, succeeded] = distinctValues.insert(sbuilder.extract());
+
+                if (succeeded) {
+                    auto& element = *itr->view().begin();
+                    // Stolen shamelessly from the CXX driver library (bsoncxx/types/bson_value/view.cpp).
+                    switch (static_cast<int>(element.type())) {
+#define BSONCXX_ENUM(type, val)                                     \
+                        case val: {                                 \
+                            builder.append(element.get_##type());   \
+                            break;                                  \
+                        }
+#include <bsoncxx/enums/type.hpp>
+#undef BSONCXX_ENUM
+                    }
+                    consecutiveRepeats = 0;
+                } else if (++consecutiveRepeats > 100) {
+                    // Rate of failure: 100% of the past N pulls.
+                    std::stringstream msg;
+                    msg << "Repeatedly failed to find a new distinct value "
+                        << consecutiveRepeats << " times. Terminating distinct array value "
+                        << "generation because we are likely to hit an infinite loop. Node: "
+                        << _node;
+                    BOOST_THROW_EXCEPTION(ValueGeneratorConvergenceTimeout(msg.str()));
+                }
+            }
+        } else {
+            for (int i = 0; i < times; ++i) {
+                _valueGen->append(builder);
+            }
         }
         return builder.extract();
     }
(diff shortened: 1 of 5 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1674,121 +1728,161 @@ Out valueGenerator(const Node& node,
...
-         return std::make_unique<BinDataGenerator>(node, generatorArgs);
-     }},
-    {"^BinDataSensitive",
-     [](const Node& node, GeneratorArgs generatorArgs) {
-         // TODO: PERF-4467 Update this to bsoncxx::binary_sub_type::sensitive.
-         return std::make_unique<BinDataGenerator>(
-            node, generatorArgs, bsoncxx::binary_sub_type(0x8));
-     }},
-};
+const auto [allParsers, arrayParsers, dateParsers, doubleParsers, intParsers, stringParsers] = [] () {
```
dalyd, line 1731: What do we gain by the changes in this block?

dalyd (changes requested):
This is great. Thank you. 

Please do add some tests as Charlie asked for, and make sure add examples to https://github.com/mongodb/genny/blob/master/src/workloads/docs/Generators.yml or some other appropriate location. 

CC FYI: @rtimmons just because I think you will find this cool.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1841,6 +1935,36 @@ UniqueGenerator<bsoncxx::array::value> literalArrayGenerator(const Node& node,
...
+    } catch (const InvalidConversionException& e) {
+    }
+    try {
+        if (extractKnownParser(node["of"], generatorArgs, stringParsers)) {
+            // known parser type
+            return std::make_unique<DistinctArrayGenerator<std::string>>(node, generatorArgs);
+        }
+    } catch (const InvalidConversionException& e) {
+    }
+    BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax("Unsupported \"of\" for distinct array"));
```
dalyd, line 1965: Could we include the `of` node in the exception? Possibly even the whole node? Genny parsing error messages are generally awful, so any insight we can add is really helpful. This is a good start. Adding more information about the particular failing node would make it better.

src/value_generators/src/DocumentGenerator.cpp
wqian94: The main benefit is a deduplication of parser entries. It's really easy to only add the parser to just `arrayParsers`, for example, and forget to add it to `allParsers`. This makes the `allParsers` map inherit mappings from existing parser maps, to avoid the risk of accidentally forgetting to map a parser.

dalyd: That makes a lot of sense, and i know that error mode has occurred. Could you add a comment explaining this, and include instructions for what to do when adding a new parser--I think that would now be just add to one list, and it's then automatically included in the others appropriately.

wqian94: I spent some time debugging this yesterday, too, so I figured it'd be a good change.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1388,6 +1391,57 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+          _generatorArgs{generatorArgs},
+          _valueGen{getValueGen(node, generatorArgs)},
+          _nTimesGen{intGenerator(extract(node, "number", "^DistinctArray"), generatorArgs)} {}
+
+    bsoncxx::array::value evaluate() override {
+        bsoncxx::builder::basic::array builder{};
+        auto times = _nTimesGen->evaluate();
+        // Choose initial bucket size of 2N to minimize hash collisions and resizes.
+        std::unordered_set<T> distinctValues(2 * times);
+        while (distinctValues.size() < times) {
```
rtimmons, line 1410: This could infinite-loop on mis-configs. E.g. 

    {^DistinctArray: {of: {^RandomInt {start: 10, end: 10}}}, number: 10}

Hard to detect in the general case from config alone, but maybe stop the iteration and barf after like 100*N iterations or something?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1674,121 +1728,161 @@ Out valueGenerator(const Node& node,
...
+         }},
+        // There are other things of type Generator<int64_t>. Not sure if they should be here or not
+        {"^Inc",
+         [](const Node& node, GeneratorArgs generatorArgs) {
+             return std::make_unique<IncGenerator>(node, generatorArgs);
+         }},
+    };
+
+    // Set of parsers to look when we request an int parser
+    // see int64Generator
```
rtimmons, line 1834: string, not int I think?

rtimmons (commented):
nice! Couple things to consider - I approve if @dalyd approves :)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1388,6 +1393,130 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+            std::deque<int8_t> offsets(bufferLength);
+            ssize_t collisions = 0;
+
+            /**
+             * Sliding window via cyclic buffer.
+             *
+             * This computes the number of insertion failures, i.e. repeated values, in the past
+             * `bufferLength` number of generated values. We achieve this by queueing up the
+             * inverse operation for N steps away. We set N to be >= 64 so that small counts are
+             * given more generous resources (very small N corresponds to higher variance). If
```
cswanson310, line 1443: "N" is "times" right? Can we make them the same thing?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1388,6 +1393,130 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+            // Choose initial bucket size of 2N to minimize hash collisions and resizes.
+            std::unordered_set distinctValues({firstValue}, 2 * times, setHasher, setKeyEq);
+            // Choose a buffer length of max(times, 64).
+            const auto bufferLength = std::max<int64_t>(64, times);
+            // Keep track of collision offsets for next N runs.
+            std::deque<int8_t> offsets(bufferLength);
+            ssize_t collisions = 0;
+
+            /**
+             * Sliding window via cyclic buffer.
```
cswanson310, line 1438: uhhh... helper function/class with tests maybe? This is pretty complex to inline for this purpose... Can we just say "failure to insert 100 times in a row is fatal?" That's probably more than good enough for our purposes

cswanson310: I just worry that debugging this cyclic buffer may be harder than detecting that you have induced an infinite loop. The error message is also pretty... scary? None of those words are easy to parse/associate with "infinite loop"

dalyd (commented):
Thank you for adding the tests and documentations. Those look good. 
The C++ is getting over my head at this point, so going to need some help there.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -189,6 +189,35 @@ Tests:
             "data" : "aaaaaaaaaaoEtN" }}
         ]}
 
+  - Name: Array Generator creates an array with a repeated element
+    GivenTemplate:
+      a:
+        {^Array: {of: {^RandomInt: {min: 1, max: 4}}, number: 2}}
+    ThenReturns:
+      - {a: [{ "$numberLong": "1" }, { "$numberLong": "1" }]}
```
cswanson310, line 197: uhhh this test case doesn't make sense to me... how does it result in 2 ones deterministically?

cswanson310 (approved):
A couple more comments but this looks pretty good to me! I don't mind the type magic, but I do think the sliding window is overkill...

src/value_generators/test/DocumentGeneratorTestCases.yml
dalyd: Huh, I missed this. Conceptually it could do this, but why should it do it? 
If you want to force the functionality you can wrap a ^Repeat around the ^RandomInt. It would work for this case and the next.

dalyd: In fact, that would make a better test case as it would force duplicate remove fro the distinct generator. Unless I'm missing something, in which case we should add a comment about why this has to produce a repeated entry.

wqian94: I don't think `DistinctArray` supports `Repeat` yet because `Repeat` isn't a `Generator` class. I can work on changing this, but it might take a bit longer.

wqian94: It has been resolved, I think!

src/workloads/docs/Generators.yml
```diff
@@ -251,6 +251,22 @@ Actors:
             # The following example returns - repeatingSeq: [1, 2, 3, 1, 2]
             repeatingSeq: {^Array: {of: {^Choose: {from: [1, 2, 3], deterministic: true}}, number: 5}}
 
+            # You can also filter items out, similar to a set, by setting `distinct` = true.
+            # For example, you can generate N distinct random integers.
```
cswanson310, line 255: In this comment you claim to generate N distinct random integers, but then you use `Repeat`, which I think undermines that? err well I guess it doesn't but this is a weird example to give folks to copy/paste. Maybe just comment that it will de-duplicate and not display that via the repeat in the docs? You can say something like:
```
Without the 'distinct' argument here we would risk generating two of the same integers.
```

src/workloads/docs/Generators.yml
```diff
@@ -251,6 +251,22 @@ Actors:
...
+            distinctSeq10Ints: {^Array: {of: {^Repeat: {count: 10, fromGenerator: {
+              ^Inc: {start: 0, end: 1000}}}}, number: 10, distinct: true}}
+
+            # Just be careful, the generator will give up if the failure rate gets too high, such as
+            # with this next generator.
+            distinctSeqFailure: {^Array: {of: {^Repeat: {count: 10000, fromGenerator: {
```
cswanson310, line 261: repeat is ok in this example I think?

src/workloads/docs/Generators.yml
```diff
@@ -251,6 +251,22 @@ Actors:
...
+              ^Inc: {start: 0, end: 1000}}}}, number: 10, distinct: true}}
+
+            # A good rule of thumb is that you should be selecting less than 50% of your range.
+            # You're not limited to integers, though! You can also filter doubles, strings, even
+            # documents and arrays.
+            distinctSeq10Strings: {^Array: {of: {^Repeat: {count: 10, fromGenerator: {
```
cswanson310, line 267: I would also take the repeat out of here, since it inhibits the copy/paste-ability.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1369,13 +1409,49 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+          _nTimesGen{intGenerator(extract(node, "number", "^Array"), generatorArgs)},
+          _distinct {node["distinct"] ? node["distinct"].maybe<bool>().value_or(false) : false} {}
 
     bsoncxx::array::value evaluate() override {
         bsoncxx::builder::basic::array builder{};
         auto times = _nTimesGen->evaluate();
-        for (int i = 0; i < times; ++i) {
-            _valueGen->append(builder);
+        if (_distinct) {
+            // Choose initial bucket size of 2N to minimize hash collisions and resizes.
```
cswanson310, line 1419: Still somewhat confusing with the discrepancy between 'N' and 'times'... I actually don't really like 'times' since it kinda implies 'number of times to generate', which is not equivalent to 'number of output results'. Maybe replace both 'times' and 'N' with 'outputSize' or 'nOutput'?

cswanson310 (approved):
Looks great, thanks @wqian94 ! My remaining comments are all just to improve maintainability.

src/value_generators/src/DocumentGenerator.cpp
dalyd: I'd note that nTimes was existing, but also internal, so I'm all in favor of clearer names. The input is `number`, so Charlie's suggestions make sense to me.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1369,13 +1409,49 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
-          _nTimesGen{intGenerator(extract(node, "number", "^Array"), generatorArgs)} {}
+          _nTimesGen{intGenerator(extract(node, "number", "^Array"), generatorArgs)},
+          _distinct {node["distinct"] ? node["distinct"].maybe<bool>().value_or(false) : false} {}
 
     bsoncxx::array::value evaluate() override {
         bsoncxx::builder::basic::array builder{};
         auto times = _nTimesGen->evaluate();
-        for (int i = 0; i < times; ++i) {
-            _valueGen->append(builder);
+        if (_distinct) {
```
dalyd, line 1418: Could you add a lot of comments to this code? Both intention and implemenation. For example, call out that if _distinct is true, we need to make sure that all the elements in the array are distinct. 

There's a lot of C++ magic here. I'm fine with that, but would appreciate leaving a map behind so I can make sense of it when I or anyone else comes back to look at this code again.

src/value_generators/src/DocumentGenerator.cpp
dalyd, line 1419: Maybe additional comment here that we generate individual values and place them in a set. We dump them to JSON for a simple hashable element we can put into our set. If an element is already in the set, we don't use it, but instead try again.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1369,13 +1409,49 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+            // Choose initial bucket size of 2N to minimize hash collisions and resizes.
+            std::unordered_set<bsoncxx::array::value, SetHasher, SetKeyEq> distinctValues(2 * times);
+            int8_t consecutiveRepeats = 0;
+            const int8_t consecutiveRepeatsThreshold = 100;
```
dalyd, line 1422: Comment that there is a chance of an infinite loop, and we are tracking this value so we can fail in a useful way if we think we are in an infinite loop.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1369,13 +1409,49 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+
+            while (distinctValues.size() < times) {
+                bsoncxx::builder::basic::array sbuilder{};
```
dalyd, line 1425: Could you add a comment (if not covered above), that we are only using this array builder temporarily, so that we can get one arbitrary bson value?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1369,13 +1409,49 @@ class ArrayGenerator : public Generator<bsoncxx::array::value> {
...
+                _valueGen->append(sbuilder);
+                auto [itr, succeeded] = distinctValues.insert(sbuilder.extract());
+
+                if (succeeded) {
+                    auto& element = *itr->view().begin();
+                    // Stolen shamelessly from the CXX driver library (bsoncxx/types/bson_value/view.cpp).
+                    switch (static_cast<int>(element.type())) {
```
dalyd, line 1432: Is the \#define and \#include come from the shamelessly stolen part? Maybe just given an end line number or add a comment "end stolen shamelessly"

dalyd (approved):
The C++ is beyond me at this point, so I'm leaning on Charlie here. I did ask for a lot more comments in line so I can make sense of this code if I need to come back to it at some point.

PR #961, opened 2023-08-02
PERF-4225 add a workload to test resharding performance with indexes

YoungYang0820:

**Jira Ticket:** [PERF-4225](https://jira.mongodb.org/browse/PERF-4225)

**Whats Changed:**  
Add a workload for resharding with 10 indexes

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/64cbcb7ca4cf47bf54dbdcc6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). The patch hasn't run successfully due to lack of available instance and long sys-perf queue.

**Workload Submission form:**  
[Form](https://docs.google.com/forms/d/e/1FAIpQLSf1oh23Ddo1khjLZMqZ9DHbRdObnpeH20PSXTBuXVg-7mxxTA/viewform?edit2=2_ABaOnueG5Z-djqaIYwlIPrAzY5kbAC7mtqj-G6FpcajAukgf6ir-0hH56oObruNBoF3PxXI)

**Related PRs:**   
[dsi](https://github.com/10gen/dsi/pull/1369)
https://github.com/10gen/mongo/pull/14666

### DIFF
--- a/src/workloads/sharding/ReshardCollectionWithIndexes.yml
+++ b/src/workloads/sharding/ReshardCollectionWithIndexes.yml
@@ -182,6 +182,17 @@ Actors:
     Operations: *WriteOperations
     GlobalRate: *WriteRate
 
+- Name: LoggingActor
+  Type: LoggingActor
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1, 2]
+      NopInPhasesUpTo: 4
+      PhaseConfig:
+        LogEvery: 15 minutes
+        Blocking: None
+
 AutoRun:
 - When:
     mongodb_setup:
--- a/src/workloads/sharding/ReshardCollectionWithIndexes.yml
+++ b/src/workloads/sharding/ReshardCollectionWithIndexes.yml
@@ -15,8 +15,8 @@ Description: |
 
   The workload consists of 3 phases:
     1. Turning off balancer and make the sharded collection exists on only 1 shard.
-    4. Running read and write operations on the collection while it is being resharded to 2 shards.
-    5. Running read and write operations on the collection while it is being resharded to 3 shards.
+    2. Running read and write operations on the collection while it is being resharded to 2 shards.
+    3. Running read and write operations on the collection while it is being resharded to 3 shards.
 
   All documents are generated through genny' data loader where the integer and length of short 
   string fields are randomly generated. The fields are designed like to form different combinations 
@@ -196,7 +196,7 @@ Actors:
   Phases:
     OnlyActiveInPhases:
       Active: [1, 2]
-      NopInPhasesUpTo: 4
+      NopInPhasesUpTo: 2
       PhaseConfig:
         LogEvery: 15 minutes
         Blocking: None
--- a/src/workloads/sharding/ReshardCollectionWithIndexes.yml
+++ b/src/workloads/sharding/ReshardCollectionWithIndexes.yml
@@ -181,3 +181,12 @@ Actors:
     Collection: *Collection
     Operations: *WriteOperations
     GlobalRate: *WriteRate
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - resharding-3shard.2023-08
+    branch_name:
+      $gte:
+      - v7.1
--- a/src/workloads/sharding/ReshardCollectionWithIndexes.yml
+++ b/src/workloads/sharding/ReshardCollectionWithIndexes.yml
@@ -1,0 +1,175 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+Description: |
+  This test expects the target cluster is created using ebs snapshot with 1 billion 1KB records 
+  and has 3 shards, named shard-00, shard-01, shard-02. The collection has 10 indexes including 
+  _id index, single-key index and compound index. ReshardCollection should use same-key resharding 
+  and use shardDistribution parameter to reshard into 2 shards then 3 shards. There will be random 
+  CRUD operations during resharding but should run at a very low rate.
+
+  The workload consists of 3 phases:
+    1. Turning off balancer and make the sharded collection exists on only 1 shard.
+    4. Running read and write operations on the collection while it is being resharded to 2 shards.
+    5. Running read and write operations on the collection while it is being resharded to 3 shards.
+    
+
+  The inserted documents have the following form:
+
+      {
+        _id: 10,
+        shardKey: 20,
+        counter: 0,
+        num1: 1,
+        num2: 2,
+        str1: 'abc',
+        str2: 'xyz',
+        padding: 'random string of bytes ...'
+      }
+
+  The indexes are:
+      [
+        {_id: 1},
+        {shardKey: 1},
+        {counter: 1},
+        {num1: 1},
+        {num2: 1},
+        {str1: 1},
+        {str2: 1},
+        {shardKey: 1, counter:1},
+        {str1: 1, num1: 1},
+        {num2:1, str2: 1}
+      ]
+
+GlobalDefaults:
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+  # Collection0 is the default collection populated by the MonotonicSingleLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentCount: &DocumentCount 1_000_000_000  # for an approximate total of 1TB
+  NumChunks: &NumChunks 1                # start with only 1 shard
+  Timeout: &Timeout 360_000_000 # = 100 hours
+
+  ShardKeyValueMin: &ShardKeyValueMin 1
+  # We use 100x the number of chunks so that the $sample aggregation to choose the new split points
+  # has a wide range of distinct values to choose from. It could be even larger.
+  ShardKeyValueMax: &ShardKeyValueMax 1_000_000
+  ShardName0: &Shard0 shard-00
+  ShardName1: &Shard1 shard-01
+  ShardName2: &Shard2 shard-02
+
+  ReadRate: &ReadRate 100 per 1 second
+  WriteRate: &WriteRate 10 per 10 seconds  # Keep the writes slow to pass critical section.
+
+  ReadOperations: &ReadOperations
+  - OperationName: findOne
+    OperationCommand:
+      Filter: {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
+
+  WriteOperations: &WriteOperations
+  - OperationName: updateOne
+    OperationCommand:
+      Filter: {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
+      Update: {$inc: {counter: 1}}
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 202
+      socketTimeoutMS: *Timeout  # = 100 hours
+  # The reshardCollection command is expected to take a long to complete so we give the actor
+  # running it a much higher socket timeout.
+  ReshardCollection:
+    QueryOptions:
+      maxPoolSize: 1
+      socketTimeoutMS: *Timeout  # = 100 hours
+
+Actors:
+- Name: CreateShardedCollection
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationMetricsName: StopBalancer
+      OperationName: AdminCommand
+      OperationCommand:
+        balancerStop: 1
+        maxTimeMS: *Timeout
+    - OperationMetricsName: EnableSharding
+      OperationName: AdminCommand
+      OperationCommand:
+        enableSharding: *Database
+    - OperationMetricsName: ShardCollection
+      OperationName: AdminCommand
+      OperationCommand:
+        shardCollection: *Namespace
+        key: {shardKey: hashed}
+  - *Nop
+  - *Nop
+
+- Name: ReshardCollection
+  Type: AdminCommand
+  Threads: 1
+  ClientName: ReshardCollection
+  Phases:
+  - *Nop
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationMetricsName: ReshardCollectionToTwoShards
+      OperationName: AdminCommand
+      OperationCommand:
+        reshardCollection: *Namespace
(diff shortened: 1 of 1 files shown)


src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,175 @@
...
+  The workload consists of 3 phases:
+    1. Turning off balancer and make the sharded collection exists on only 1 shard.
+    4. Running read and write operations on the collection while it is being resharded to 2 shards.
+    5. Running read and write operations on the collection while it is being resharded to 3 shards.
+    
+
+  The inserted documents have the following form:
+
+      {
+        _id: 10,
```
mtrussotto, line 19: Can you make it more clear what the form is here?  As it is, it looks like the documents are all literally the same aside from the padding.  I'm guessing what you really mean is that _id is an integer, counter is an integer that's unique for all docs, num1 and num2 are (small?) integers, str1 and str2 are three characters strings, etc?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,175 @@
...
+
+Actors:
+- Name: CreateShardedCollection
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationMetricsName: StopBalancer
```
mtrussotto, line 97: A nit, but several of the other workloads do this and they use the name "DisableBalancer", so we should too just for consistency.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,175 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+Description: |
+  This test expects the target cluster is created using ebs snapshot with 1 billion 1KB records
```
mtrussotto, line 4: Can you make it clearer that the cluster starts with all records in the same shard?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,175 @@
...
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: admin
+    Operations:
+    - OperationMetricsName: StopBalancer
+      OperationName: AdminCommand
+      OperationCommand:
+        balancerStop: 1
+        maxTimeMS: *Timeout
```
mtrussotto, line 101: I think we don't want the 100-hour timeout for this one?

mtrussotto (commented):
I believe this also needs a perf team reviewer.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Done

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Done

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Done

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Sounds good.

mtrussotto (approved):

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+  # Collection0 is the default collection populated by the MonotonicSingleLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentCount: &DocumentCount 1_000_000_000  # for an approximate total of 1TB
+  NumChunks: &NumChunks 1                # start with only 1 shard
+  Timeout: &Timeout 72_000_000 # = 20 hours
+
+  ShardKeyValueMin: &ShardKeyValueMin 1
+  # We use 100x the number of chunks so that the $sample aggregation to choose the new split points
```
dalyd, line 57: Above you have `NumChunks` set to 1, but this comment says 100x the number of chunks before 1M. Are one of those settings wrong? The comment? My reading of it?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+  # has a wide range of distinct values to choose from. It could be even larger.
+  ShardKeyValueMax: &ShardKeyValueMax 1_000_000
+  ShardName0: &Shard0 shard-00
+  ShardName1: &Shard1 shard-01
+  ShardName2: &Shard2 shard-02
+
+  ReadRate: &ReadRate 100 per 1 second
```
dalyd, line 64: How were the read and write rate picked? Are they important? If not, why not?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+Description: |
+  This test expects the target cluster is created using ebs snapshot with 1 billion 1KB records 
+  and has 3 shards, named shard-00, shard-01, shard-02. The collection has 10 indexes including 
+  _id index, single-key index and compound index. The whole dataset is on 1 shard at the beginning.
+  ReshardCollection should use same-key resharding and use shardDistribution parameter to reshard 
+  into 2 shards then 3 shards. There will be random CRUD operations during resharding but should 
+  run at a very low rate.
```
dalyd, line 9: Is the important part that there is something running in the background, but what it is is not very important? Should you be driving conflicts in which crud operations target actively being moved data?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+  Nop: &Nop {Nop: true}
+
+  Database: &Database test
+  # Collection0 is the default collection populated by the MonotonicSingleLoader.
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+
+  DocumentCount: &DocumentCount 1_000_000_000  # for an approximate total of 1TB
+  NumChunks: &NumChunks 1                # start with only 1 shard
+  Timeout: &Timeout 72_000_000 # = 20 hours
```
dalyd, line 54: Could you add a comment about why 20 hours was chosen?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+  This test expects the target cluster is created using ebs snapshot with 1 billion 1KB records 
+  and has 3 shards, named shard-00, shard-01, shard-02. The collection has 10 indexes including 
+  _id index, single-key index and compound index. The whole dataset is on 1 shard at the beginning.
+  ReshardCollection should use same-key resharding and use shardDistribution parameter to reshard 
+  into 2 shards then 3 shards. There will be random CRUD operations during resharding but should 
+  run at a very low rate.
+
+  The workload consists of 3 phases:
+    1. Turning off balancer and make the sharded collection exists on only 1 shard.
+    4. Running read and write operations on the collection while it is being resharded to 2 shards.
```
dalyd, line 13: Do you want a pause or queisce between the first and second resharding?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
dalyd: You have the reads and writes rate limited separately and running from different threads. That's probably the right thing for this test so you can see the impact on reads and writes separately. That said, there's an argument that could be made for having threads do a mixture of reads and writes (probably more representative), and that may behave differently coming from a shared thread (eg slow writes may impact following queries). 

I wanted to bring up the point and make sure you had considered it. Assuming you have considered it, you can close this comment thread.

dalyd (commented):
Two small requests and a lot of questions. 
The workload does what it says it's going to do (nice work). The questions mostly are about making sure we are testing the right thing. 

Also, let's get a working patch build for this before turning it on.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/sharding"
+Description: |
```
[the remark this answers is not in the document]
YoungYang0820: done

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Added

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: The random CRUD operations here are to make sure that these operations work as normal and won't make resharding fail, but as the goal of this test is to see the performance improvements of resharding, I can also remove these operations.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Not necessary, we expect the `reshardCollection` command return when everything about this resharding is completed so we can start next round immediately.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+  The workload consists of 3 phases:
+    1. Turning off balancer and make the sharded collection exists on only 1 shard.
+    4. Running read and write operations on the collection while it is being resharded to 2 shards.
+    5. Running read and write operations on the collection while it is being resharded to 3 shards.
+    
+
+  The inserted documents have the following form:
+
+      {
+        _id: integer(default _id index),
```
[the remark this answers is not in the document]
YoungYang0820: All documents are generated through genny's data loader where the integer and length of short string fields are randomly generated. The main reasons I make it in this format are:

- We want every document is around 1KB
- We want to have multiple fields so we can form different combinations of indexes.

Regarding the type of those fields, I don't think that's important because we just want to see the performance improvements over the old way on a certain setup. The performance on different kinds of fields are not what we are looking to check here.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
```diff
@@ -0,0 +1,174 @@
...
+        shardKey: integer(for the shard key),
+        counter: integer(used for counting updates),
+        num1: integer(random generated between [1, 1000000000]),
+        num2: integer(random generated between [1, 1000000000]),
+        str1: string(length <= 20),
+        str2: string(length <= 20),
+        padding: string(random string of bytes ...)
+      }
+
+  The indexes are:
```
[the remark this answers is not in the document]
YoungYang0820: It comes from the design, but as I said above, it's just a setup to compare the performance before and after PM-2322.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: Done. It is picked from previous run.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: This is actually no longer needed in this test since the data is now coming from the snapshot. The idea is that because we need to reshard the collection into 20000 chunks, we need to make the shardKey range large enough to have enough cardinality.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
YoungYang0820: The read and write operations are not important here. We only want to have some CRUD operations here at a slow rate, so the rate is picked as some number I feel it's small.

Thanks for the idea to make this an argument, I didn't think about that because CRUD is not the main player in this test. I would leave it as it is now and make it tunable if we need it one day.

YoungYang0820 (commented):
Thanks for the detailed review! I'm trying to answer these but feel free to ask more if it's not clear.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
dalyd: Don't remove the background operations for me. I'm just pushing for clarity and making sure things are intentional. 

Do you expect any impact on the background load? It might be worth explicitly saying that the performance of the background crud operations is of secondary concern to the performance of the resharding operation in this test (assuming those words are true).

src/workloads/sharding/ReshardCollectionWithIndexes.yml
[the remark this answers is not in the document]
dalyd: Could you add some of the above to the comment string? Capturing that workload intent can be very valuable when we come back to the workload in the future.

src/workloads/sharding/ReshardCollectionWithIndexes.yml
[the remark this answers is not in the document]
dalyd: Could you say that in the comment? It could be to just say that we care that there are indexes, but the 10 selected were somewhat arbitrary?

src/workloads/sharding/ReshardCollectionWithIndexes.yml
dalyd: That sounds good. Maybe add another comment here saying the values were picked to be small.

dalyd (commented):
Looking good. Just some requests for a couple more comments. 
I think you might be adding in the logging actor as well. Should be good after those and a working patch build.

dalyd (approved):
Changes LGTM. Please do check that you get a reasonable patch build, and the performance looks like you expect.

YoungYang0820:
Green patch https://spruce.mongodb.com/version/64d924efc9ec4418435fa710/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #963, opened 2023-08-03
PERF-4449 Perf tests for query IDHACK using _id and regular int index

wqian94:
**Jira Ticket:** PERF-4449

**Whats Changed:**  
Add workloads for comparing queries with IDHACK on `_id` and without IDHACK on a regular int index.

**Patch testing results:**  
[Patch run](https://spruce.mongodb.com/version/64d6652ad1fe0789f12f6c98/tasks)

[Perf analysis](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=660131b0-8838-4eb1-8b60-99d15ffbc8b8&variant_filter=linux-standalone-query-stats&task_filter=ten_m_doc_collection__int_id&test_filter=IDHACK128Threads.find&test_filter=IDFieldPointProject128Threads.find&test_filter=IDHACK256Threads.find&test_filter=IDHACKPointProject128Threads.find&test_filter=IntFieldEq128Threads.find&test_filter=IntField100KIn128Threads.find&test_filter=IntFieldOr128Threads.find&test_filter=IntField10KIn128Threads.find&test_filter=IDHACK32Threads.find&measurement_filter=OperationThroughput&percent_filter=0%7C%7C100&z_filter=0%7C%7C10)

**Related PRs:**   
PERF-4478 [#957]


wqian94:
This is still pending #957, but here are the proposed additional tests.

We were also talking about merging the YAML tests vs using the base template -- do you have further opinions on this?

src/lamplib/src/genny/tasks/preprocess.py
wqian94, line 1: Not 100% sure these changes to the preprocessor will fly.

And even if they do, I can only imagine that they should be broken up into a separate PR.

wqian94:
The goal is to standardize the actors' names, but it's probably better to just go back to what's currently available in the preprocessor.

Templates are a genny preprocessor concept, so generators are generally not used. Since things like template names need to be available at the preprocessor level, it creates some restrictions on what can be used for some specific fields.

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -1,109 +1,224 @@
...
       OnlyActiveInPhase: 4
-
+      Threads: 128
 - ActorFromTemplate:
-    TemplateName: IDHACK
+    TemplateName: IDHACKEq
     TemplateParameters:
-      Name: IDHack256Threads
-      Threads: 256
+      Name: IDHACKEq0256Threads
```
cswanson310, line 129:
```suggestion
      Name: IDHACKEq256Threads
```

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -1,109 +1,224 @@
...
+    Key: QuiesceActor
+    Parameters:
+      MaxPhases: *MaxPhases
 - ActorFromTemplate:
-    TemplateName: IDHACK
+    TemplateName: IDHACKEq
     TemplateParameters:
-      Name: IDHack32Threads
-      Threads: 32
+      Name: IDHACKEq032Threads
```
cswanson310, line 117: Did you mean to add the 0 here? Also I think the "Eq" is not necessarily needed since only equalities are eligible for IDHACK. I worry it may be more confusing, since the name sort of implies there might be an IDHACKLt or IDHACKIn

cswanson310: Oh, I see that this is indeed a topic of confusion :) Try loading the dataset (it doesn't go away after your test run), and running some explains with these queries. I believe that you should only see the IDHACK plan stage in the explain output for the existing workloads - I don't know of another way to generate that plan stage aside from exact equality on _id.

cswanson310 (changes requested):

src/workloads/query/TenMDocCollection_IntId.yml
wqian94: Oh.

The `0` in `032` is intentional, so that when sorting on the actor name, the 32-thread runs will show up first. The `0`s in front of the other thread counts is unintentional.

wqian94: Is there any point in retaining the `id` field, then? Shouldn't we be just testing on `_id` anyway?

wqian94: Actually, I take that back, the eq test still needs to happen, right?

cswanson310: We want the `id` field so that we can compare a single point lookup between IDHACK and "regular" index scans. The non-_id point lookup is probably the cheapest query you could imagine otherwise. These other, more expensive queries are more interesting as "other representative queries" which will presumably make us feel better that our "worst case" performance is indeed worse than "average case" or something like that.

Actually after typing that out, there probably is a cheaper one: An equality query with a projection on only that field, so that we would not need to have a FETCH in the query plan - all the data comes from that one location in the index. Could we add that also?

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -30,6 +30,17 @@ ActorTemplates:
...
+  Config:
+    LoadConfig:
+      Path: TenMDocCollection_Base.tmpl
+      Key: ActorTemplateConfig
+      Parameters:
+        MaxPhases: *MaxPhases
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {_id: {^RandomInt: {min: 0, max: 9999999}}}
```
cswanson310: Was this supposed to be no underscore?

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -62,7 +73,22 @@ ActorTemplates:
...
+      Parameters:
+        MaxPhases: *MaxPhases
+        Operations:
+        - OperationName: aggregate
+          OperationCommand:
+            Pipeline:
+            - $match:
+                id: {^RandomInt: {min: 0, max: 9999999}}
+            - $project:
+                id: '$id'
```
cswanson310: I think you should use '1' or 'true' otherwise this won't be considered a "simple" projection: https://github.com/mongodb/mongo/blob/2ef28165da8a2deb91bc46757aec32c0f419ae81/src/mongo/db/query/projection.h#L143

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -51,7 +62,7 @@ ActorTemplates:
         Operations:
         - OperationName: find
           OperationCommand:
-            Filter: {_id: {$in: {^FixedGeneratedValue: {fromGenerator: {^Array: {distinct: true, number: 100000, of: {^RandomInt: {min: 0, max: 9999999}}}}}}}}
+            Filter: {_id: {$in: [{^RandomInt: {min: 0, max: 9999999}}]}}
```
cswanson310: I see you trimmed the number of in values. Is there a story here? Does this single $in get optimized into an equality? If yes, no need to keep it I don't think. If not, then can we add to it?

In either case, I think we would benefit from a medium-sized in list and a large size in list. I've seen some large ones, including in a recent help ticket where they are causing problems. I don't really know what large means exactly, but 10k seems good? medium is maybe 1k then?

cswanson310 (changes requested):
A couple rapid comments before I sign off for the day!

src/workloads/query/TenMDocCollection_IntId.yml
wqian94: Ah... yea, a length of 1 makes it an `$eq`. But a length of 2 keeps it as an `$in`.

I'm still confused about what the purpose of testing with an `$in` of 10K points is. This is a fundamentally different query than the IDHACK query.

wqian94: Although, funnily, `$eq` doesn't invoke `IDHACK`...

wqian94:
Pulling the conversation out of the code itself.

Here are some query plans that I have:

```
> test.find({_id: 1}).explain()["queryPlanner"]["winningPlan"]
{ "stage" : "IDHACK" }

> test.find({_id: {: 1}}).explain()["queryPlanner"]["winningPlan"]
{
        "queryPlan" : {
                "stage" : "FETCH",
                "planNodeId" : 2,
                "inputStage" : {
                        "stage" : "IXSCAN",
                        "planNodeId" : 1,
                        "keyPattern" : {
                                "_id" : 1
                        },
                        "indexName" : "_id_",
                        "isMultiKey" : false,
                        "isUnique" : true,
                        "isSparse" : false,
                        "isPartial" : false,
                        "indexVersion" : 2,
                        "direction" : "forward",
                        "indexBounds" : {
                                "_id" : [
                                        "[1.0, 1.0]"
                                ]
                        }
                }
        }
}
> test.find({id: 1}).explain()["queryPlanner"]["winningPlan"]
{
        "queryPlan" : {
                "stage" : "FETCH",
                "planNodeId" : 2,
                "inputStage" : {
                        "stage" : "IXSCAN",
                        "planNodeId" : 1,
                        "keyPattern" : {
                                "id" : 1
                        },
                        "indexName" : "id_1",
                        "isMultiKey" : false,
                        "multiKeyPaths" : {
                                "id" : [ ]
                        },
                        "isUnique" : false,
                        "isSparse" : false,
                        "isPartial" : false,
                        "indexVersion" : 2,
                        "direction" : "forward",
                        "indexBounds" : {
                                "id" : [
                                        "[1.0, 1.0]"
                                ]
                        }
                }
        }
}

> test.find({id: {: 1}}).explain()["queryPlanner"]["winningPlan"]
{
        "queryPlan" : {
                "stage" : "FETCH",
                "planNodeId" : 2,
                "inputStage" : {
                        "stage" : "IXSCAN",
                        "planNodeId" : 1,
                        "keyPattern" : {
                                "id" : 1
                        },
                        "indexName" : "id_1",
                        "isMultiKey" : false,
                        "multiKeyPaths" : {
                                "id" : [ ]
                        },
                        "isUnique" : false,
                        "isSparse" : false,
                        "isPartial" : false,
                        "indexVersion" : 2,
                        "direction" : "forward",
                        "indexBounds" : {
                                "id" : [
                                        "[1.0, 1.0]"
                                ]
                        }
                }
        }
}

> test.find({id: {: [1, 2]}}).explain()["queryPlanner"]["winningPlan"]
{
        "queryPlan" : {
                "stage" : "FETCH",
                "planNodeId" : 2,
                "inputStage" : {
                        "stage" : "IXSCAN",
                        "planNodeId" : 1,
                        "keyPattern" : {
                                "id" : 1                                                                                                                                                                                                   },
                        "indexName" : "id_1",
                        "isMultiKey" : false,
                        "multiKeyPaths" : {
                                "id" : [ ]
                        },
                        "isUnique" : false,
                        "isSparse" : false,
                        "isPartial" : false,
                        "indexVersion" : 2,
                        "direction" : "forward",
                        "indexBounds" : {
                                "id" : [
                                        "[1.0, 1.0]",
                                        "[2.0, 2.0]"
                                ]
                        }
                }
        }
}

> test.find({: [{id: 1}, {id: 2}]}).explain()["queryPlanner"]["winningPlan"]
{
        "queryPlan" : {
                "stage" : "FETCH",
                "planNodeId" : 2,
                "inputStage" : {
                        "stage" : "IXSCAN",
                        "planNodeId" : 1,
                        "keyPattern" : {
                                "id" : 1
                        },
                        "indexName" : "id_1",
                        "isMultiKey" : false,
                        "multiKeyPaths" : {
                                "id" : [ ]
                        },
                        "isUnique" : false,
                        "isSparse" : false,
                        "isPartial" : false,
                        "indexVersion" : 2,
                        "direction" : "forward",
                        "indexBounds" : {
                                "id" : [
                                        "[1.0, 1.0]",
                                        "[2.0, 2.0]"
                                ]
                        }
                }
        }
}
```

Also, the plans generated with `id` are identical to the ones generated with `_id`, except for the direct point query, which uses IDHACK.

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -1,109 +1,172 @@
...
-      Active: [2]
-      NopInPhasesUpTo: *MaxPhases
-      PhaseConfig:
-        Repeat: 1
+        - OperationName: find
+          OperationCommand:
+            Filter: {_id: {^RandomInt: {min: 0, max: 9999999}}}
+- TemplateName: IDFieldPointProject
+  Config:
+    LoadConfig:
```
cswanson310, line 35: I feel like I've lost the plot here a little bit... why did we split this file into two? Do you have a grand plan? Or did we give up on the grand plan and now we should have one file?

src/workloads/query/TenMDocCollection_IntId.yml
cswanson310: Oh okay one more interesting thing I noticed. If you run explain for this query it still does use the IDHACK query plan stage. But if you throw in an $eq like you do below - it disqualifies the IDHACK plan but it does then make you eligible for a 'PROJECTION_COVERED' plan, which will avoid fetching the whole document. [IDHACK will perform a fetch](https://github.com/10gen/mongo/blob/d0d74180a28b2c9eb9343dd0782f22ec81888eb0/src/mongo/db/exec/idhack.cpp#L123). So plausibly this query would get faster if it used $eq. Can we throw an $eq version in as well?

```js
winningPlan: {
      stage: 'PROJECTION_COVERED',
      transformBy: { _id: 1 },
      inputStage: {
        stage: 'IXSCAN',
        keyPattern: { _id: 1 },
        indexName: '_id_',
        isMultiKey: false,
        multiKeyPaths: { _id: [] },
        isUnique: true,
        isSparse: false,
```
vs: 
```js
  winningPlan: {
      stage: 'PROJECTION_SIMPLE',
      transformBy: { _id: 1 },
      inputStage: { stage: 'IDHACK' }
    },
```

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -1,109 +1,172 @@
...
+    Path: TenMDocCollection_Base.tmpl
+    Key: QuiesceActor
+    Parameters:
+      MaxPhases: *MaxPhases
 - ActorFromTemplate:
     TemplateName: IDHACK
     TemplateParameters:
-      Name: IDHack32Threads
-      Threads: 32
+      Name: IDHACK032Threads
```
cswanson310, line 124: can you please avoid changing these names as part of this commit? Can we throw a separate cleanup ticket in to do it? I think it is a good change, but it will make all our graphs break and I think we will stop caring about our graphs soon enough that we can/should just defer this.

cswanson310 (changes requested):

src/workloads/query/TenMDocCollection_IntId.yml
wqian94: Ah, the template file is helpful for simplifying some of the common syntax... but an argument could be made that it's no longer necessary.

src/workloads/query/TenMDocCollection_IntId.yml
wqian94: Don't think we'll need a cleanup ticket for this, since there aren't as many 32t tests now.

cswanson310 (approved):
LGTM! Thanks for all the work on this! Let's start getting some results! I have some other ideas for future experiments but I think this is what we wanted for now.

PR #960, opened 2023-08-02
EVG-20193: Warm up pool connection before returning

PyGeek03:
**Jira Ticket:** EVG-20193

**Whats Changed:**  
Warm up the pool connection before returning this connection to Actor. 

This code is only called [during setup](https://github.com/mongodb/genny/pull/960/files#diff-1c7f2006fe8a4fc6fdac7e96796e826197884da4111f3d850ac4127c31541d34R134-R135) in Actor constructors, and Actor constructors are only called by the `WorkloadContext` to create all Actors, before any of Actor is run. Therefore, this change should not negatively impact metric measurement. However, it is expected that this change might cause changepoints where throughput is improved (as the connection has been warmed up), but this change would stabilize metrics in the long run, because Actors no longer spend time waiting for the pool connection to warm up the first time they use the connection. Indeed, Perf Analyzer results below seem to indicate that this change would stabilize metrics.

**Patch testing results:**  
- [sys-perf on ARM](https://spruce.mongodb.com/version/64c9d84f3e8e86704a2163a3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (PERF-4166, connection_pool_stress)
[Perf Analyzer](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=a5a4b88d-e90d-4bba-a789-2c3bffbc68bc)
- [sys-perf on Intel](https://spruce.mongodb.com/version/64caff4532f417d5c6eac635/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (BF-28785, mixed_workloads_genny_stress, device_monitoring)
[Perf Analyzer](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=973a48df-1581-4ada-93da-8c9e94817ae5)
- [sys-perf on Intel](https://spruce.mongodb.com/version/64caff4532f417d5c6eac635/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (PERF-4166, majority_reads10_k_threads, majority_writes10_k_threads)
[Perf Analyzer](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=8a14fc25-5fa5-4649-95e2-f0d898742091)

PR #970, opened 2023-08-07
DEVPROD-2090: Pre-warm pool connection before returning (configurable)

PyGeek03:
**Jira Ticket:** [DEVPROD-2090](https://jira.mongodb.org/browse/DEVPROD-2090) (EVG-20193)

**Whats Changed:**  
Pre-warm the pool connection before returning this connection to Actor, by sending a ping to the admin database.

Pre-warming is needed to stabilize metrics, because an Actor's first operation can be very noisy.

You can disable pre-warming by setting `NoPreWarm: true` under `Clients` in the workload YAML file:

```yaml
...
Clients:
  $client_name:
    NoPreWarm: true
...
```

Each pool connection is created and pre-warmed when an Actor thread is created, instead of pre-warming the entire pool with a loop upon initial pool construction. This is both to minimize the number of pool connections created and pre-warmed (which is expensive to run) and for ease of implementation.

This pre-warming logic is only called as part of pool connection creation [during setup](https://github.com/mongodb/genny/pull/960/files#diff-1c7f2006fe8a4fc6fdac7e96796e826197884da4111f3d850ac4127c31541d34R134-R135) in Actor constructors, and Actor constructors are only called by the `WorkloadContext` to create all Actors in one go, before any of Actor is run. Therefore, the pre-warming is outside of any measurement code path. 

This change should not negatively impact metric measurement. It is expected to cause changepoints where latency and throughput are improved (as the connection has been warmed up), but this would stabilize metrics in the long run because Actors no longer spend time waiting for the pool connection to warm up the first time they use the connection.

I disable pre-warming for [ConnectionPoolStress.yml](https://github.com/mongodb/genny/pull/970/files#diff-aca636168207ba043737047ed72426735d8296db4bfd473c4e67a02966d07d4b), as [PERF-4166](https://jira.mongodb.org/browse/PERF-4166) mentions that "the overwhelm behavior seems to be the more interesting case". Please let me know if I should disable pre-warming for any other workloads before merging this PR in.

Discussion: Should there be a pre-warmed version of pre-warming-disabled workloads, to separate issues caused by the sheer number of ongoing connections from issues caused by cold connections?

**Patch testing results:**  
- [sys-perf, relevant workloads](https://spruce.mongodb.com/version/64d095b79ccd4e2eb2f28670). This patch includes workloads mentioned in BF-28502 & BF-28785, which led to this request: connection_pool_stress, majority_reads_10k_threads, majority_writes_10k_threads, device_monitoring & mixed_workloads_genny_stress.
[Perf Analyzer link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=c85c5c17-7064-48cc-a019-746f3e1bda3a)
- [sys-perf, includes both high value Genny workloads and relevant workloads](https://spruce.mongodb.com/version/65041dc0850e61cb9a0db0e4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).
[Perf Analyzer link](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=f0b2316a-6de2-4c33-a1a8-e8e6206432b4)

Note that although mixed_workloads_genny_stress's LatencyMin measurements regressed, all other latency and throughput measurements improved, which suggests that the latency distribution is tighter and less noisy.

**Related PRs:**   
The non-configurable previous iteration of this PR: https://github.com/mongodb/genny/pull/960


src/gennylib/src/PoolManager.cpp
```diff
@@ -53,9 +54,21 @@ auto createPool(const std::string& name,
...
-mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
-                                                     size_t instance,
-                                                     const Node& context) {
+mongocxx::pool::entry genny::v1::PoolManager::create_client(const std::string& name,
+                                                            size_t instance,
+                                                            const Node& context) {
+    auto pool_entry = this->_create_client(name, instance, context);
+    bool prewarm = context["Clients"][name]["PreWarm"].maybe<bool>().value_or(true);
+    if (prewarm) {
+        pool_entry->list_databases();
```
PyGeek03, line 63: Regarding which command to use for pre-warming:

1. Both [ping](https://www.mongodb.com/docs/manual/reference/command/ping/) and [hello](https://www.mongodb.com/docs/manual/reference/command/hello/) are issued against a specific database, not the server. I could try creating a database to do this, but that might mess with tests which test creating a database that happens to use the same database name as the pre-warming code.
2. The pre-warming is run within Actor constructors, so it is completely outside any measurement code path.

PyGeek03: As discussed in Slack with @thessem , I've decided to pre-warm by pinging the admin DB instead.

SimonEismann:
I'll focus on the performance implications of this change:
- To address your question, I think we currently don't need a prewarmed version of ConnectionPoolStress since the behavior is quite boring outside of the connection storm itself.
- The minimum latency of ConnectionPoolStress seems to be heavily impacted, even though there should be no change in behavior for this test, right?
image
- The FTDC data shows that for majority_reads10_k_threads we need ~20 minutes to establish the 10K connections, I assume this would be addressed by [CDRIVER-3514](https://jira.mongodb.org/browse/CDRIVER-3514) & [CDRIVER-4656](https://jira.mongodb.org/browse/CDRIVER-4656) which are scheduled for this quarter.
- The throughput data for the high-thread-count tests show significant improvements from your change, which is awesome.
- My main concern currently is if these changes have any unexpected impacts on the "normal" tests. Could you run some of the [high-value workloads](https://docs.google.com/document/d/1M5dSFM4gTA1048hIA9AnmlM0Q9LDb15onh61c5om1Tk/edit#heading=h.wdpdgdpcthgg) to see if there are any unexpected changes?

SimonEismann (changes requested):
I'll focus on the performance implications of this change:
- To address your question, I think we currently don't need a prewarmed version of ConnectionPoolStress since the behavior is quite boring outside of the connection storm itself.
- The minimum latency of ConnectionPoolStress seems to be heavily impacted, even though there should be no change in behavior for this test, right?
image
- The FTDC data shows that for majority_reads10_k_threads we need ~20 minutes to establish the 10K connections, I assume this would be addressed by [CDRIVER-3514](https://jira.mongodb.org/browse/CDRIVER-3514) & [CDRIVER-4656](https://jira.mongodb.org/browse/CDRIVER-4656) which are scheduled for this quarter.
- The throughput data for the high-thread-count tests show significant improvements from your change, which is awesome.
- My main concern currently is if these changes have any unexpected impacts on the "normal" tests. Could you run some of the [high-value workloads](https://docs.google.com/document/d/1M5dSFM4gTA1048hIA9AnmlM0Q9LDb15onh61c5om1Tk/edit#heading=h.wdpdgdpcthgg) to see if there are any unexpected changes?

src/workloads/issues/ConnectionPoolStress.yml
```diff
@@ -41,9 +41,11 @@ Constants:
 
 Clients:
   Default:
+    PreWarm: false
     QueryOptions:
       maxPoolSize: *GlobalMaxPoolSize
   PerThread:
+    PreWarm: false
```
thessem, line 48: I have a fairly strong preference for `NoPreWarm: true`. I feel that with bools you should try and keep the default option `false`

src/gennylib/src/PoolManager.cpp
```diff
@@ -53,9 +55,22 @@ auto createPool(const std::string& name,
...
-mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
-                                                     size_t instance,
-                                                     const Node& context) {
+mongocxx::pool::entry genny::v1::PoolManager::create_client(const std::string& name,
+                                                            size_t instance,
+                                                            const Node& context) {
+    auto pool_entry = this->_create_client(name, instance, context);
+    bool prewarm = context["Clients"][name]["PreWarm"].maybe<bool>().value_or(true);
+    if (prewarm && !_dryRun) {
+        auto ping = bsoncxx::builder::stream::document{} << "ping" << 1 << bsoncxx::builder::stream::finalize;
```
thessem, line 64: Upon reading this, it is not clear how this solves the issue. It seems to rely upon the particular connection dynamic of Genny.

To improve clarity, it would be useful to add a comment explaining why pre-warming has been implemented and why it is being done here instead of pre-warming the entire pool with a loop upon initial pool construction (which would be the naive assumption). The comment does not need to be here, but wherever it makes sense.

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -143,7 +143,9 @@ class PoolManager {
      *   the WorkloadContext used to look up the configurations
      * @return a connection from the pool or throw if none available
      */
-    mongocxx::pool::entry client(const std::string& name, size_t instance, const Node& context);
+    mongocxx::pool::entry create_client(const std::string& name,
```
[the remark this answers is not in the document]
thessem: Also, we've changed exactly what this function does and I think that is worth a documentation update

thessem (changes requested):
I would like to see testing for the added functionality. I would like to make sure that pre-warmed connections are actually pre-warmed.

Additonally: How does this functionality interact with `minPoolSize` and `maxPoolSize` URI parameters documented for the cxx driver? Does `get_client` start freezing up if we're above maxPool size? What affect would that have?

PyGeek03:
Hi @SimonEismann, thank you for your feedback, that was quick! 

Regarding ConnectionPoolStress, I believe it's because of the inherent high variance of this workload stemming from no pre-warming. The case you mentioned is for linux-shard-single.2022-11 variant, which I have re-run and [found no significant changes](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=d37a11c8-2816-4255-bbed-cd116d1948a9&task_filter=connection_pool_stress&percent_filter=0%7C%7C100&z_filter=0%7C%7C10).

Regarding testing this PR on high-value workloads, thank you for this great idea! I will run another patch for these workloads after addressing @thessem's feedback.

src/gennylib/include/gennylib/v1/PoolManager.hpp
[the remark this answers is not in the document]
PyGeek03: I've updated this function's comment

src/gennylib/src/PoolManager.cpp
```diff
@@ -53,9 +55,22 @@ auto createPool(const std::string& name,
...
+        pool_entry->database("admin").run_command(ping.view());
+    }
+    return pool_entry;
+}
+
+
+mongocxx::pool::entry genny::v1::PoolManager::_create_client(const std::string& name,
```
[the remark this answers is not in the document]
PyGeek03: I initially wrap `create_client` around `_create_client` because of a conditional block for apm that returns the pool connection in seperate conditional branches. But after moving the pre-warming logic to its own helper function, this wrapping is no longer necessary.

src/gennylib/src/PoolManager.cpp
```diff
@@ -53,9 +55,22 @@ auto createPool(const std::string& name,
...
 
 
-mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
-                                                     size_t instance,
-                                                     const Node& context) {
+mongocxx::pool::entry genny::v1::PoolManager::create_client(const std::string& name,
+                                                            size_t instance,
+                                                            const Node& context) {
+    auto pool_entry = this->_create_client(name, instance, context);
+    bool prewarm = context["Clients"][name]["PreWarm"].maybe<bool>().value_or(true);
```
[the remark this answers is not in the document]
PyGeek03: Mocking context is done a lot in Genny test code, and the workload context is used by `EncryptionManager` constructor and the `createPool` function, which are called by this function.

src/gennylib/src/PoolManager.cpp
PyGeek03: I've added more details for these questions in the PR description.

SimonEismann:
Hi @PyGeek03, I've checked in with the rest of the team about the expected performance impact below. The consensus was that this is an exciting, positive change. Make sure to give the perf build barons a heads-up when you are merging this. 

Throughput:
- Throughput increases for the high-thread-count workloads 
- Throughput increases for a few tests that run a single request (e.g. array_traversal_high_value/ClearCollection.Crud)
Average latency/LatencyP50:
- Average latency decreases for the high-thread-count workloads
- Average latency decreases for a few tests that run only a few request (e.g. array_traversal_high_value/ClearCollection.Crud)
LatencyP95:
- LatencyP95 decreases for the high-thread-count workloads
- LatencyP95 decreases for some tests, e.g., 15% on load_test
Summary:
- The changes seem to make the initial requests faster, some tests are sensitive to this if they run a low number of total requests.
- There are significant benefits for the high-thread-count workloads

SimonEismann (approved):
Hi @PyGeek03, I've checked in with the rest of the team about the expected performance impact below. The consensus was that this is an exciting, positive change. Make sure to give the perf build barons a heads-up when you are merging this. 

Throughput:
- Throughput increases for the high-thread-count workloads 
- Throughput increases for a few tests that run a single request (e.g. array_traversal_high_value/ClearCollection.Crud)
Average latency/LatencyP50:
- Average latency decreases for the high-thread-count workloads
- Average latency decreases for a few tests that run only a few request (e.g. array_traversal_high_value/ClearCollection.Crud)
LatencyP95:
- LatencyP95 decreases for the high-thread-count workloads
- LatencyP95 decreases for some tests, e.g., 15% on load_test
Summary:
- The changes seem to make the initial requests faster, some tests are sensitive to this if they run a low number of total requests.
- There are significant benefits for the high-thread-count workloads

PyGeek03:
@thessem

> I would like to see testing for the added functionality. I would like to make sure that pre-warmed connections are actually pre-warmed.

Good idea, I've added test for this via resmoke-test, thanks!

> How does this functionality interact with `minPoolSize` and `maxPoolSize` URI parameters documented for the cxx driver? Does `get_client` start freezing up if we're above maxPool size? What affect would that have?
> What happens if this errors? Have you moved the place where it throws an error further up the stack? What does this look like for a user? Is this more confusing or less confusing?

By default, when pool clients are created by the driver, they are not pre-warmed. If `minPoolSize` is N, then [according to the specs](https://github.com/mongodb/specifications/blob/a10e270087bc8ec661561be8cffd287bc6982d91/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst?plain=1#L526-L530), N pool clients would be created by the driver either when the pool is in "ready" state after its creation or (from my understanding) when `acquire()` is called(*). In either case, none of the clients would be pre-warmed by the driver.

(*): Unfortunately this is not the case for the CXX driver because of a [C driver bug](https://jira.mongodb.org/browse/CDRIVER-4663), but there's not much we can do on this front except for finding a way to convince the team that owns the C driver to fix this bug.

Before this change, if there's more Actor threads than `maxPoolSize`, then Actor construction would fail with an error like this:

```
...
[2023-09-26 00:46:07.655877] [0x00007f9812234840] [info]    ExternalPhaseCoordinator::ExternalPhaseCoordinator('',4400') --> -1
[2023-09-26 00:46:07.661067] [0x00007f9810fff640] [info]    Constructing pool with MongoURI 'mongodb://localhost:27017/?appName=Genny&maxPoolSize=4'
[2023-09-26 00:46:08.878039] [0x00007f9812234840] [error]   Caught boost::exception /home/ubuntu/coding/genny-workspace/src/genny/src/gennylib/include/gennylib/parallel.hpp(132): Throw in function throwIfExceptions
Dynamic exception type: boost::wrapexcept<genny::ExceptionBucket::ParallelException>
std::exception::what: Error in parallel execution. First exception's what(): Failed to acquire an entry from the client pool.
 0# 0x00000000004B6789 in /home/ubuntu/coding/genny-workspace/src/genny/dist/bin/genny_core
 1# 0x00000000004B12B1 in /home/ubuntu/coding/genny-workspace/src/genny/dist/bin/genny_core
 2# 0x00007F9812029D90 in /lib/x86_64-linux-gnu/libc.so.6
 3# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6
 4# 0x00000000004B10A5 in /home/ubuntu/coding/genny-workspace/src/genny/dist/bin/genny_core
...
```
After this change, this error happens at the same place as before.

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -398,6 +398,7 @@ def _parse_clients(self, clients):
         clients_dict = self._recursive_parse(clients)
         for _, client in clients_dict.items():
             client.setdefault("URI", self._default_uri)
+            client.setdefault("NoPreWarm", False)
```
lgfang, line 401: This is intentional that change default behaviour from no pre-warm to pre-warm?

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -129,7 +129,16 @@ class PoolManager {
     /**
      * Obtain a connection or throw if none available.
      *
-     * This is allowed to be called from multiple threads simultaneously.
+     * This connection will be pre-warmed by default, unless "NoPreWarm" is set to true, e.g.:
+     *
```
lgfang, line 133: nit: there are trailing space chars on this line and line 158, please delete them :)

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -139,11 +148,15 @@ class PoolManager {
...
-     * @param context
+     * @param workloadCtx
      *   the WorkloadContext used to look up the configurations
      * @return a connection from the pool or throw if none available
      */
-    mongocxx::pool::entry client(const std::string& name, size_t instance, const Node& context);
+    mongocxx::pool::entry createClient(const std::string& name,
+                                       size_t instance,
+                                       const Node& workloadCtx);
+
```
lgfang, line 158: nit: trailing space chars

src/gennylib/include/gennylib/v1/PoolManager.hpp
```diff
@@ -129,7 +129,16 @@ class PoolManager {
     /**
      * Obtain a connection or throw if none available.
      *
-     * This is allowed to be called from multiple threads simultaneously.
+     * This connection will be pre-warmed by default, unless "NoPreWarm" is set to true, e.g.:
+     * 
+     * ```yaml
+     * Clients:
+     *   Default:
+     *     NoPreWarm: true
```
lgfang, line 137: nit: IMO, "PreWarm: true/false" is more straightforward than "NoPreWarm: false/true"

src/gennylib/src/PoolManager.cpp
```diff
@@ -74,24 +74,34 @@ mongocxx::pool::entry genny::v1::PoolManager::client(const std::string& name,
 
     Pools& pools = lap.second;
 
+    auto& clientNode = workloadCtx["Clients"][name];
+    bool shouldPrewarm = !_dryRun & !clientNode["NoPreWarm"].maybe<bool>().value_or(false);
```
lgfang, line 78: nit: `&` ==> `&&` ? Although `&` gives correct result, we are doing an "logical and" and hence I'd prefer the logical operator `&&` instead of the bitwise operator `&`.

lgfang (approved):
lgtm except for some nits. Hence I'm approving

src/lamplib/src/genny/tasks/preprocess.py
PyGeek03: Yes. This is because it is preferable for the default value for booleans to be False, [as suggested by James](https://github.com/mongodb/genny/pull/970#discussion_r1315379736).

src/gennylib/include/gennylib/v1/PoolManager.hpp
PyGeek03: Done, thanks!

src/gennylib/include/gennylib/v1/PoolManager.hpp
PyGeek03: Done, thanks!

src/gennylib/src/PoolManager.cpp
PyGeek03: Nice catch, thanks!

PyGeek03:
evergreen merge

PR #974, opened 2023-08-07
EVG-17874: Generate all tasks

johndaniels:

**Jira Ticket:** [EVG-17874](https://jira.mongodb.org/browse/EVG-17874)

**Whats Changed:**  
This is a draft PR for switching to having a single task generation task for all the variants in Genny. The advantage of this is that generated tasks will be scheduled only when other variant tasks are scheduled, but will also always be available to run on any waterfall build, allowing individual tasks to be run. 

This change also makes it possible to generate tasks in a patch build without running all of them, in case there is a specific generated task that needs to be run.

See https://github.com/mongodb/mongo/blob/EVG-17874-taskgen-test/etc/system_perf_taskgen_test.yml#L455 for how this would be used in a sysperf test. That file is a modified sys-perf that doesn't actually run DSI ever in order to test scheduling

See https://spruce.mongodb.com/commits/sys-perf-taskgen-test?view=FAILED to view how scheduling works.

### DIFF
--- a/src/lamplib/src/genny/cli.py
+++ b/src/lamplib/src/genny/cli.py
@@ -585,13 +585,15 @@ def auto_tasks(ctx: click.Context, tasks: str):
     required=True,
     help="An evergreen project file",
 )
+@click.option("--no-activate", default=False, is_flag=True, help="Ensure that generated tasks don't activate")
 @click.pass_context
-def auto_tasks_all(ctx: click.Context, project_file: str):
+def auto_tasks_all(ctx: click.Context, project_file: str, no_activate: bool):
     from genny.tasks import auto_tasks_all
 
     auto_tasks_all.main(
         project_file=project_file,
         workspace_root=ctx.obj["WORKSPACE_ROOT"],
+        no_activate=no_activate,
     )
 
 
--- a/src/lamplib/src/genny/cli.py
+++ b/src/lamplib/src/genny/cli.py
@@ -583,7 +583,7 @@ def auto_tasks(ctx: click.Context, tasks: str):
 @click.option(
     "--project-file",
     required=True,
-    help="An evergreen project file",
+    help="An evergreen project file, such as system_perf.yml",
 )
 @click.option("--no-activate", default=False, is_flag=True, help="Ensure that generated tasks don't activate")
 @click.pass_context
--- a/src/lamplib/src/genny/cli.py
+++ b/src/lamplib/src/genny/cli.py
@@ -568,7 +568,29 @@ def auto_tasks(ctx: click.Context, tasks: str):
 
     auto_tasks.main(
         mode_name=tasks,
-        genny_repo_root=ctx.obj["GENNY_REPO_ROOT"],
+        workspace_root=ctx.obj["WORKSPACE_ROOT"],
+    )
+
+@cli.command(
+    name="auto-tasks-all",
+    help=(
+        "Determine which Genny workloads should be schedulable for a given evergreen yaml config. "
+        "Parses the config to find every build configuration."
+        "This is used by evergreen and allows new genny workloads to be created "
+        "without having to modify any repos outside of genny itself."
+    ),
+)
+@click.option(
+    "--project-file",
+    required=True,
+    help="An evergreen project file",
+)
+@click.pass_context
+def auto_tasks_all(ctx: click.Context, project_file: str):
+    from genny.tasks import auto_tasks_all
+
+    auto_tasks_all.main(
+        project_file=project_file,
         workspace_root=ctx.obj["WORKSPACE_ROOT"],
     )
 
--- a/src/lamplib/src/genny/tasks/auto_tasks.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks.py
@@ -70,11 +70,8 @@ class WorkloadLister:
     Separate from the Repo class for easier testing.
     """
 
-    def __init__(self, workspace_root: str, genny_repo_root: str, reader: YamlReader):
+    def __init__(self, workspace_root: str):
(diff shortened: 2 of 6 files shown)


src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -449,19 +425,23 @@ def __init__(self, lister: WorkloadLister, reader: YamlReader, workspace_root: s
...
-        return [
-            Workload(
-                workspace_root=self.workspace_root,
-                file_path=fpath,
-                is_modified=fpath in modified,
-                reader=self.reader,
-            )
-            for fpath in all_files
-        ]
+        if self._all_workloads is None:
```
dalyd, line 434: Should you move the first two lines inside the if? You don't need to do that work if `self._all_workloads` exist.

src/lamplib/src/genny/cli.py
```diff
@@ -568,10 +568,34 @@ def auto_tasks(ctx: click.Context, tasks: str):
...
+        "Determine which Genny workloads should be schedulable for a given evergreen yaml config. "
+        "Parses the config to find every build configuration."
+        "This is used by evergreen and allows new genny workloads to be created "
+        "without having to modify any repos outside of genny itself."
+    ),
+)
+@click.option(
+    "--project-file",
+    required=True,
+    help="An evergreen project file",
```
dalyd, line 586: Is this system-perf.yml by default? Maybe add "such as system_perf.yml".

src/lamplib/src/genny/tasks/auto_tasks_all.py
```diff
@@ -0,0 +1,60 @@
...
+def get_all_builds(global_expansions: dict[str, str], project_file_path: str) -> List[CurrentBuildInfo]:
+    """
+    Attempts to compute the expansions 
+    """
+    with open(project_file_path) as project_file:
+        project = yaml.safe_load(project_file)
+    variants = project["buildvariants"]
+    all_builds = []
+    for variant in variants:
+        if "expansions" in variant and "mongodb_setup" in variant["expansions"]:
```
dalyd, line 26: Based on https://github.com/mongodb/genny/blob/593b1f4eac70c901dabbd45bd207b846a76b7203/docs/using.md#configuring-autorun I think you need to check more than just mongodb_setup. I think you have to check all possible settings. 

Oh, maybe this is sufficient -- you are just getting the list of build variants here, and all the relevant build variants will have a mongodb_setup. NVM. Maybe worth a comment in the code. 

Also, please update that documentation when putting this in production.

src/lamplib/src/genny/tasks/auto_tasks_all.py
```diff
@@ -0,0 +1,60 @@
...
+        if len(build_tasks) > 0:
+            SLOG.info(f"Generating auto-tasks for variant: {build.variant}")
+            ConfigWriter.configure_variant_tasks(config, build_tasks, build.variant, activate=activate_param)
+        else:
+            SLOG.info(f"No auto-tasks for variant: {build.variant}")
+    return config
+
+def main(project_file: str, workspace_root: str, no_activate: bool) -> None:
+    reader = YamlReader()
+    global_expansions = reader.load(workspace_root, "expansions.yml")
```
dalyd, line 51: For production, an informative error message for no such file would be useful.

dalyd (commented):
I like this proposal. Can we run the generate all on every commit? 

I don't follow why we need the no-activate version. I thought (and it looks like) that because the generated tasks are in a different build variant they don't automatically get scheduled. Or do I have that backwards? Unless I'm missing something, I would prefer to have on generate task that runs on every commit. And maybe have something to preserve the auto-patch functionality for people working on new workloads.

src/lamplib/src/genny/tasks/auto_tasks_all.py
johndaniels: I wasn't sure if it would be a problem if someone made a generated task without a "When" block. In practice I think every task has a "when" on mongodb setup so maybe it wouldn't come up but this seemed moderately safer?

src/lamplib/src/genny/cli.py
johndaniels: Will add this

src/lamplib/src/genny/tasks/auto_tasks.py
johndaniels: Yes definitely

johndaniels:
> I don't follow why we need the no-activate version. I thought (and it looks like) that because the generated tasks are in a different build variant they don't automatically get scheduled. Or do I have that backwards? Unless I'm missing something, I would prefer to have on generate task that runs on every commit. And maybe have something to preserve the auto-patch functionality for people working on new workloads.

It was mainly a workaround for if you want to generate all tasks in a patch build. In that context, tasks seem to always be fun regardless of the variant scheduling. 

I think what I want to do ideally is have it generate with no_activate in the patch context but not in the waterfall context. That will allow it to run on every waterfall build and have appropriate scheduling, while making sure that in a patch context the standard behavior is not to run every single sys-perf task.

src/lamplib/src/genny/tasks/auto_tasks.py
johndaniels: Fixed

src/lamplib/src/genny/cli.py
johndaniels: Fixed

dalyd (approved):
LGTM modulo backwards compatiple patch build.

johndaniels:
evergreen merge

johndaniels:
evergreen merge

PR #975, opened 2023-08-07
PERF-4516 Implement Genny preprocessor key FormatString

wqian94:
**Jira Ticket:** PERF-4516

**Whats Changed:**  
Add `^FormatString` as a preprocessor key. It mimics the `^FormatString` value generator. If `^FormatString` fails at the preprocess step, it'll ignore the failure and let the value generator version of `^FormatString` handle it.

**Patch testing results:**  

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -26,7 +26,10 @@
         1.  [LoadConfig](#orga6d35c7)
         2.  [ActorTemplate](#orga45b1d8)
         3.  [OnlyActiveInPhases](#orgf9c328f)
-        4.  [Defaults and Overrides](#org22b7a0f)
+        4.  [FlattenOnce](#preprocessorflattenonce)
+        5.  [PreprocessorFormatString](#preprocessorformatstring)
+        6.  [NumExpr](#preprocessornumexpr)
+        7.  [Defaults and Overrides](#org22b7a0f)
     5.  [Connecting to the Server](#orgd6b0450)
         1.  [Connection Strings and Pools](#orgd2659db)
         2.  [Multiple Connection Strings](#orga591018)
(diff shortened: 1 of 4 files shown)


src/workloads/docs/HelloWorld.yml
```diff
@@ -33,6 +33,13 @@ Actors:
       ^NumExpr:
         withExpression: "messages / threads"
         andValues: {messages: *TotalMessageCount, threads: *Threads}
+  # To concatenate strings, use the ^FormatString preprocessor directive. It works similarly to the
+  # ^FormatString generator.
+  - Message:
+      ^FormatString:
```
dalyd, line 39: How do we distinguish the pre-processor version of this and the generator version of this? I'm worried that we are going to process all of them.

wqian94: Well, it _does_ process all of them, but any that fail will be left as a `^FormatString` (although the inner docs are still recursively processed). So basically, if it fails to do the formatting at preprocess time, it'll attempt to run it as a generator.

Happy to change it up to something like `^PreprocessorFormatString`, but my concern was primarily that this requires genny users to be fairly aware of what is preprocessed vs generator-able.

dalyd: After looking and thinking a while, I think I would like a different name for the preprocessed version. I think we should be explicit about what is continually regenerated at runtime versus what is evaluated once before running. 

I haven't been able to come up with any truly bad failure cases, but I could still imagine someone wondering why they can't use a  ^RandomString inside a ^FormatString when defining the actor name, but they could when creating a document within their workload. 

@rtimmons thoughts?

wqian94: I think that's valid. I couldn't fully decide on which behavior I wanted, so either is fine by me.

rtimmons: If we lean into lisp-isms, preprocessing is a macro. Most lisps have a thing that calls out macros in the syntax. I like `^PreprocessorFormatString` although I wonder if a better syntax like `^^` for preprocessor things would be more intuitive long-term.

wqian94: Created PERF-4543 to track this. Also realized that this probably belongs in the EVG queue...

dalyd (approved):
Change LGTM. 

Could you add documentation to https://github.com/mongodb/genny/blob/master/docs/using.md?plain=1 or wherever is most appropriate for this change and the FormatString? Either in this PR or another? Maybe here: https://github.com/mongodb/genny/blob/b80986958c232a7f430f110715b2cd730240fd59/docs/using.md?plain=1#L622

It you felt charitable you could cover ^NumExpr while you are there, but not required.

wqian94:
> Change LGTM.
> 
> Could you add documentation to https://github.com/mongodb/genny/blob/master/docs/using.md?plain=1 or wherever is most appropriate for this change and the FormatString? Either in this PR or another? Maybe here:
> 
> https://github.com/mongodb/genny/blob/b80986958c232a7f430f110715b2cd730240fd59/docs/using.md?plain=1#L622
> 
> It you felt charitable you could cover ^NumExpr while you are there, but not required.

How are the anchor IDs determined?

docs/using.md
```diff
@@ -26,7 +26,10 @@
         1.  [LoadConfig](#orga6d35c7)
         2.  [ActorTemplate](#orga45b1d8)
         3.  [OnlyActiveInPhases](#orgf9c328f)
-        4.  [Defaults and Overrides](#org22b7a0f)
+        4.  [FlattenOnce](#preprocessorflattenonce)
+        5.  [PreprocessorFormatString](#preprocessorformatstring)
```
wqian94, line 30: @dalyd @rtimmons double-checking, is there a set scheme for these anchor names?

PR #976, opened 2023-08-07
PERF-4517 Implement Genny preprocessor key FlattenArray

wqian94:
**Jira Ticket:** PERF-4517

**Whats Changed:**  
Implement the `^FlattenArray` key, which reduces the dimensionality of an array by 1. For example:
- `[1, 2, 3, 4, 5, 6, 7, 8, 9]` -> `[1, 2, 3, 4, 5, 6, 7, 8, 9]`
- `[[1, 2, 3], 4, 5, 6, [7, [8, 9]]]` -> `[1, 2, 3, 4, 5, 6, 7, [8, 9]]`
- `"123456789"` -> `[1, 2, 3, 4, 5, 6, 7, 8, 9]`

**Patch testing results:**  

### DIFF
--- a/src/lamplib/src/genny/tasks/preprocess.py
+++ b/src/lamplib/src/genny/tasks/preprocess.py
@@ -239,6 +239,8 @@ class _WorkloadParser(object):
             out = self._replace_param(value)
         elif key == "^NumExpr":
             out = self._replace_numexpr(value)
+        elif key == "^FlattenArray":
+            out = self._replace_flattenarr(value)
         elif key == "ActorTemplates":
             self._parse_templates(value)
         elif key == "ActorFromTemplate":
@@ -322,10 +324,23 @@ class _WorkloadParser(object):
             raise ParseException(msg)
         except Exception as e:
             msg = (
-                "Failure trying to parse '{OP_KEY}' node.\n"
+                f"Failure trying to parse '{OP_KEY}' node.\n"
                 f"Node source: {input}\n"
                 f"Exception: {e}\n"
             )
+            raise ParseException(msg)
+
+    def _replace_flattenarr(self, input):
+        OP_KEY = "^FlattenArray"
+
+        parsed_values = self._recursive_parse(input)
+
+        try:
+            return [value
+                    for parsed_value in parsed_values
+                    for value in (parsed_value if type(parsed_value) == list else [parsed_value])]
+        except TypeError as e:
+            msg = f"Invalid value type for '{OP_KEY}', which must be iterable: {input}"
             raise ParseException(msg)
 
     def _parse_templates(self, templates):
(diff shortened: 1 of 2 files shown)


wqian94:
Should examples be written into the docs somewhere?

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -322,12 +324,25 @@ def _replace_numexpr(self, input):
             raise ParseException(msg)
         except Exception as e:
             msg = (
-                "Failure trying to parse '{OP_KEY}' node.\n"
+                f"Failure trying to parse '{OP_KEY}' node.\n"
```
rtimmons, line 327: lol nice!

src/lamplib/src/tests/test_preprocess.py
```diff
@@ -628,3 +628,71 @@ def test_numexpr_with_dict_string_value_throws(self):
...
         self._assertParseException(yaml_input)
+
+    def test_flatten_array_valid_type_arrays(self):
+        yaml_input = """SchemaVersion: 2018-07-01
+Document:
+  ^FlattenArray:
+  - 1
+  - "2"
+  - [3, "4"]
+  - ^FlattenArray: [5, [6, "7", ["8", 9]]]
```
rtimmons, line 639: How about something like

```
- ^FlattenArray: [5, [6, "7", ["8", 9, [], [], [1], [[[1]], [False], [""]]]
```
(I lost track of my brackets there but that's kinda the point :)

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -322,12 +324,25 @@ def _replace_numexpr(self, input):
...
 
+    def _replace_flattenarr(self, input):
+        OP_KEY = "^FlattenArray"
+
+        parsed_values = self._recursive_parse(input)
+
+        try:
+            return [value
+                    for parsed_value in parsed_values
+                    for value in (parsed_value if type(parsed_value) == list else [parsed_value])]
```
rtimmons, line 341: I don't think this handles multiple levels of nesting, which I think you want.

rtimmons (changes requested):
Thanks for the PR. Please ensure it handles arbitrary depth. Or if that's not what you're after, perhaps rename to `^FlattenOneLevel` or something to make that clearer.

I'm fine once a more complicated/deep test-case is added, even if you decide to not handle deep/recursive-cases yet. That way if we want to change the spec/feature in the future we'll lock down the current behavior.

src/lamplib/src/genny/tasks/preprocess.py
wqian94: Ah, I can see how one might think that.

`^FlattenOnce` is closer to what I wanted to do. Automatically flattening all the subarrays seems like maybe an optional flag to toggle in the future.

rtimmons (approved):
Nice! Thanks for iterating.

PR #977, opened 2023-08-07
PERF-4520 Correct spelling of Evergreen variant for Big Sur arm64

wqian94:
**Jira Ticket:** PERF-4520

**Whats Changed:**  
Update name of Evergreen build variant from "Bug Sur arm64" to "Big Sur arm64"

### DIFF
--- a/evergreen.yml
+++ b/evergreen.yml
@@ -119,7 +119,7 @@ buildvariants:
   - name: tg_compile_and_test_with_server_on_macos
 
 - name: macos-1100-arm64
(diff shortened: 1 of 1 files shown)


thessem (approved):
Looks good here, so long as the tests pass.

I apologize to @PyGeek03, who submitted this pull request several months ago when I was new to this environment and did not fully understand the potential impact of such a change.

While there are filtering functions around that allow results to be filtered by display name, I do not believe that changing this will pose a much of an issue.

PR #978, opened 2023-08-08
PERF-4523: Have IDHACK workloads run for 5 minutes instead of 30 seconds

wbuerger46:
**Whats Changed:**  
When investigating PERF results from the IDHACK workloads, it often takes many days of repeated runs for the results to stabilize and remove significant noise. By having the workloads run for 5 minutes instead of 30 seconds, we'll be able to narrow down if noise is coming from the find commands, or the hardware, or something else. This hopefully will also help the results stabilize sooner.

**Patch testing results:**  
[Evergreen patch](https://spruce.mongodb.com/version/64d26292562343c2b650cc05/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


naama-bareket (approved):
LGTM!

PR #980, opened 2023-08-09
PERF-4522: Add more window functions in existing window function workloads

projjal:


projjal:
[evergreen build](https://spruce.mongodb.com/version/64d363e75623433efd46cc13/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[perf analyzer](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=f16d15cd-7687-4e77-b40a-30fca5d45fc2&percent_filter=0%7C%7C100&z_filter=0%7C%7C10)

src/workloads/query/SlidingWindows.yml
```diff
@@ -377,6 +377,674 @@ Actors:
...
+      nopInPhasesUpTo: *MaxPhases
+      Pipeline:
+          [{$setWindowFields: {
+            partitionBy: "$partitionKey",
+            sortBy: {time: 1},
+            output: {covarianceSamp: {
+              $covarianceSamp: "$temp",
+              window: {
+                range: [-10, 10],
+                unit: "second"
```
lriuui0x0, line 1045: It seems like using time units is a good idea so that we don't duplicate each test case for non time unit windows again. Just note that the time unit is left out of implementation from the first pushdown PR. I should have disabled pushing down those pipelines.

I created a PR to fix this, we should merge that PR before merging these workloads to measure the correct number. https://github.com/10gen/mongo/pull/14804

lriuui0x0 (commented):
The structure looks good. What are the window functions missing from these two files? I suppose those will have dedicated tests, right?

src/workloads/query/SlidingWindows.yml
```diff
@@ -377,6 +377,674 @@ Actors:
...
+      name: DerivativeTimeBased
+      active: [23]
+      operationMetricsName: DerivativeTimeBased
+      nopInPhasesUpTo: *MaxPhases
+      Pipeline:
+          [{$setWindowFields: {
+            partitionBy: "$partitionKey",
+            sortBy: {time: 1},
+            output: {rate: {
+              $derivative: {input: "$temp", unit: "second"},
```
lriuui0x0, line 393: I wonder why some of the units here are seconds and some are hours. We just need to make sure the range is not too small to be not meaningful.

projjal: I think i made a mistake here. I wanted to have around 100-200 documents which comes to around 20 min interval. I mistakenly put 20 sec here.

projjal: Two existing tests had 1 hour range. which amounts to around 500 docs. This should be ok. I will change all to 1 hour range.

projjal:
> What are the window functions missing from these two files? I suppose those will have dedicated tests, right?
 
The ones missing are $linearFill and $locf, for them there is dedicated workloads linearFill.yml and locf.yml.
The some more missing in SlidingWindows.yml don't work with bounded window sizes. Eg, rank, denseRank, documentNumber
For CumulativeWindows.yml I didn't add some like $push, $addToSet for which size will grow unbounded

src/workloads/query/SlidingWindows.yml
lriuui0x0: Sounds good.

lriuui0x0 (approved):
Let's merge this once the pushdown PR is merged in again.

PR #987, opened 2023-08-15
PERF-4546: Add sharded IDHACK to Genny workloads

wbuerger46:
**Whats Changed:**  
Added an IDHack workload on a sharded collection. Since much of the logic is shared with the existing `TenMDocCollection_ObjectId` workload, I factored some of the shared templates into `src/phases/query/IDHack.yml`.

[Evergreen patch](https://spruce.mongodb.com/version/64db88347742ae0c65d8a85e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/phases/query/IDHack.yml
+++ b/src/phases/query/IDHack.yml
@@ -1,0 +1,39 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Defines common configurations for IDHack query workloads.
+
+IDHackObjectIdTemplate:
+  Name: {^Parameter: {Name: "Name", Default: "IDHackObjectId"}}
+  Type: CrudActor
+  Threads: {^Parameter: {Name: "Threads", Default: 1}}
+  Phases:
+    OnlyActiveInPhases:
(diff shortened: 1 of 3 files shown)


src/phases/query/IDHack.yml
```diff
@@ -0,0 +1,39 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Defines common configurations for IDHack query workloads.
```
naama-bareket, line 4: Is the intention that in the future we will add configurations for the other types of Ids here as well? Maybe it would be good to update the comment to specify this is currently only for Object Ids

naama-bareket (approved):
Looks good to me!! The templating made this very clean

src/phases/query/IDHack.yml
wbuerger46: Great question - I left it open-ended since we could certainly refactor more shared configuration. @cswanson310 is that okay with you?

wbuerger46:
evergreen merge

PR #990, opened 2023-08-15
PERF-4561 Add dbCheck canary performance workload

moustafamaher:
**Jira Ticket:** < [PERF-4561](https://jira.mongodb.org/browse/PERF-4561%20Add%20dbCheck%20canary%20performance%20workload) - [SERVER-78927](https://jira.mongodb.org/browse/SERVER-78927)>

**Whats Changed:**  
Add dbCheck canary performance workload.
I had to add a new Actor to run dbcheck, and I needed to make sure dbcheck ends by waiting for the stop health log to be written to each replica set node's local db.

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/64dbe16da4cf47a405f29704/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Related PRs:**   
[Server](https://github.com/10gen/mongo/pull/14916)
[DSI](https://github.com/10gen/dsi/pull/1381)


dalyd:
I'm going to replace myself on the review with Miguel. He's the current person (for these two weeks) to take external reviews.

miguelangelnieto (approved):

src/cast_core/src/DbCheckActor.cpp
```diff
@@ -0,0 +1,90 @@
...
+        : database{db}, 
+          collectionName{phaseContext["Collection"].to<std::string>()},
+          validateMode{phaseContext["ValidateMode"].maybe<std::string>()}  {
+        auto threads = phaseContext.actor()["Threads"].to<int>();
+        if (threads > 1) {
+            std::stringstream ss;
+            ss << "DbCheckActor only allows 1 thread, but found " << threads << " threads.";
+            BOOST_THROW_EXCEPTION(InvalidConfigurationException(ss.str()));
+        }
+    }
```
rtimmons, line 51: Excellent - thanks for adding this.

src/cast_core/src/DbCheckActor.cpp
```diff
@@ -0,0 +1,90 @@
...
+};
+
+void DbCheckActor::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto dbcheckCtx = _dbcheckMetric.start();
+            auto param = config->validateMode? 
+                bsoncxx::builder::stream::document() << "validateMode"
+                << config->validateMode.value() << bsoncxx::builder::stream::finalize
+                : bsoncxx::builder::stream::document() << bsoncxx::builder::stream::finalize;
```
rtimmons, line 65: Consider doing this construction in the PhaseContext constructor since I don't think you need to create a new `param` value each time. You might also wanna do the debug/`to_json` logging at that time as well since it doesn't change over time.

src/gennylib/src/topology.cpp
```diff
@@ -177,7 +177,7 @@ void Topology::computeDataMemberConnectionStrings(DBConnection& connection) {
         bsoncxx::array::view passives_view = passives.get_array();
         for (auto member : passives_view) {
             MongodDescription memberDesc;
-            memberDesc.mongodUri = std::string(member.get_utf8().value);
+            memberDesc.mongodUri = nameToUri(std::string(member.get_utf8().value));
```
rtimmons, line 180: Can you add context about why we need to do this? Is it related to the other parts of this PR?

src/gennylib/src/dbcheck.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <bsoncxx/builder/stream/document.hpp>
+#include <bsoncxx/json.hpp>
+#include <gennylib/dbcheck.hpp>
+#include <gennylib/v1/Topology.hpp>
+
+// Logic in this file is based on the following two js implementations:
+// https://github.com/10gen/workloads/blob/aeaf42b86bb8f1af9bc6ac90198ac0b4ff32bd14/utils/mongoshell.js#L481
+// https://github.com/mongodb/mongo-perf/blob/bd8901a2e76d2fb13d2a6a313f7a9e1bf6be9c04/util/utils.js#L384-L387
+
+namespace genny {
```
rtimmons, line 38: I don't think you need a separate header for this, the Actor is intended to be the interface for running the dbcheck logic. If we need to pull it out for other Actors to use, we can do so later when we have the two use-cases in mind (e.g. perhaps more precise coordination between `Orchestrator`/Phase lifecycle and the (nicely implemented) `WaitForDbCheckVisitor`.

src/gennylib/include/gennylib/dbcheck.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+namespace genny {
+
+/**
+ * Helper function to run dbcheck on a replica set and waiting for it to finish.
+ */
+bool dbcheck(mongocxx::pool::entry& client,
+             mongocxx::database& db,
+             std::string& collName,
+             bsoncxx::document::value& dbCheckParam);
```
rtimmons, line 28: nit: I think this can be const-ref

src/gennylib/src/dbcheck.cpp
```diff
@@ -0,0 +1,150 @@
...
+}  // anonymous namespace
+
+bool dbcheck(mongocxx::pool::entry& client,
+             mongocxx::database& db,
+             std::string& collName,
+             bsoncxx::document::value& dbCheckParam) {
+    /* Only one thread needs to actually do the dbcheck. If
+     * another thread enters the function while a dbcheck
+     * is happening, it waits until the dbcheck finishes
+     * and exits.
```
rtimmons, line 112: We only support one Thread running this, so I think this can go away and we can avoid some complexity.

rtimmons (changes requested):
I'm not super familiar with the actual health-check logic, but I didn't see anything obviously incorrect or confusing. Nice use of the topology visitor.

My biggest suggestion is to move the health-check logic into the Actor. There's also some simplification possibility around locking. Please push back wherever you disagree.

IMHO we should provide more guidance on when/where to create separate headers and "library" functionality versus putting functionality solely into Actors.

moustafamaher:
> I'm not super familiar with the actual health-check logic, but I didn't see anything obviously incorrect or confusing. Nice use of the topology visitor.
> 
> My biggest suggestion is to move the health-check logic into the Actor. There's also some simplification possibility around locking. Please push back wherever you disagree.
> 
> IMHO we should provide more guidance on when/where to create separate headers and "library" functionality versus putting functionality solely into Actors.

I was trying to follow the same design of the [QuiesceActor](https://github.com/mongodb/genny/blob/869fe58310e95d39b6e1d1a63b15f0723f8a2d0d/src/gennylib/src/quiesce.cpp) as we share the same design.

src/gennylib/src/dbcheck.cpp
moustafamaher: yes, but it can be called from different actor in the same phase, right ?

src/gennylib/src/topology.cpp
moustafamaher: I think it was just a bug in the code, as mongodUri should append required username and password and convert the host name to a host URI to be used for the client exactly as the [previous loop](https://github.com/mongodb/genny/blob/869fe58310e95d39b6e1d1a63b15f0723f8a2d0d/src/gennylib/src/topology.cpp#L162)

src/cast_core/src/DbCheckActor.cpp
moustafamaher: Good point.

src/cast_core/src/DbCheckActor.cpp
moustafamaher: Thanks, but for honesty I was following [QuiesceActor](https://github.com/mongodb/genny/blob/869fe58310e95d39b6e1d1a63b15f0723f8a2d0d/src/cast_core/src/QuiesceActor.cpp#L39-L43)

src/gennylib/src/dbcheck.cpp
moustafamaher: I was trying to follow the same design of the [QuiesceActor](https://github.com/mongodb/genny/blob/869fe58310e95d39b6e1d1a63b15f0723f8a2d0d/src/gennylib/src/quiesce.cpp) as we share the same design.

src/gennylib/src/dbcheck.cpp
rtimmons: See my other comment, but I think using this in multiple Actors (i.e. using it in any Actor outside of your DbCheckActor) would be a sign that the Actor(s) involved don't have the right level of granularity.

Actors being single-threaded is how we get away with not having to worry about concurrency in a lot of Actor and Actor-support code. If we have to worry about synchronization across Actors in a way that isn't "Orchestrator-aware" it can lead to very hard to reason-about races or timing bugs since this will compete with the Orchestrator's job of ensuring the cross-Actor state is sane.

If you decide to keep this as a separate header/lib--and if you decide you wanna keep this--then please add commentary that this should never happen because only one DbCheckActor may be instantiated, and this function may only be called from DbCheckActor.

src/gennylib/src/dbcheck.cpp
rtimmons: Ya, that's a fair response. We indeed already have precedent for this in `quiesce`. I'm not sure why this pattern didn't set off a klaxon in my head when we added that.

I'm still rather partial to _not_ having library functions like this. It encourages very fine-grained levels of abstraction that quickly make Genny into a shadow API of the C++ driver.

There may be some notion of logic-re-use across Actors in the future, but I don't think we've found the right solution (or particular need) quite yet. Efforts to slim down / split / extend `CrudActor` might be a proving ground for that more generally.

Please consider moving this into the Actor's cpp. But I wouldn't blame you for wanting to move on and pushing back.

My recommendation--if you don't end up moving this to the Actor--would be to add a comment to both `quiesce` and `dbcheck` indicating that they are intended only for use by their respective Actors.

```c++
// Note: This library function is only intended to be used
//       by the DbCheck Actor.
//
//       If you need to run dbcheck in your workload, you can
//       include the DbCheckActor in your workload config.
//
//       If you believe you need to use this anywhere other
//       than DbCheckActor.{cpp,hpp} please work with the
//       Genny maintainers to discuss how to evolve this API.   
```

And similarly, please add at least high-level docs about what dbcheck (the function) does or ensure it directly references existing docs in the Actor part of the code.

rtimmons: I'll also ping @PyGeek03 here. Ultimately he and the rest of his team are responsible for the future of Genny, so it's totally up to them how to handle this kind of scenario. I'm happy to give input / past experience, but there may be a good reason to encourage this kind of factorization in the future.

moustafamaher: I have already moved it to the Actor.

moustafamaher:
New [evergreen patch](https://spruce.mongodb.com/version/64dee1b4d6d80a83989fd9bb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

rtimmons (approved):
Thanks for iterating!

I think we should "promote" `Topology` out of the `genny::v1` namespace and into just `genny` since `v1` is intended to be Genny-internal and not used directly by Actors, but that's definitely not a concern for this PR.

LGTM.

moustafamaher:
> Thanks for iterating!
> 
> I think we should "promote" `Topology` out of the `genny::v1` namespace and into just `genny` since `v1` is intended to be Genny-internal and not used directly by Actors, but that's definitely not a concern for this PR.
> 
> LGTM.

Thank you very much for your review! 
I think it would be better to create a ticket for that clean up! should I create one ?

moustafamaher:
evergreen merge

moustafamaher:
evergreen merge

PR #992, opened 2023-08-16
PERF-4551 Improve reliability of query stats genny workloads

wqian94:
**Jira Ticket:** PERF-4551

**Whats Changed:**  
Refactor query stats workload to clear query stats between phases.

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/64dd05c63627e053ceea24df)


cswanson310:
So... why are we doing this? I understand it improves reliability, but does it mean that we re no longer testing the worst case scenario for $queryStats? Does this imply that it's worse impact/performance when the cache is full? Should we be ensuring each phase starts with a _full_ query stats store rather than an empty one?

PR #996, opened 2023-08-22
PERF-4583 - Enable setting child-operation properties for BulkWrites

SimonEismann:
The bulk write operation allows setting some properties, such as WriteConcern at the top level instead of for each child-operation within the bulk write:
http://mongocxx.org/api/current/classmongocxx_1_1options_1_1bulk__write.html

Some properties can only be set on the child level, such as upsert:
http://mongocxx.org/api/current/classmongocxx_1_1model_1_1update__one.html 

However, Genny currently ignores the properties of child operations when constructing the model, which makes it impossible to, e.g., set upsert for bulk operations.

This PR uses all OperationOptions that are set for a child-operation when creating the operation model, which is used to construct the bulk operation. 

Aside from making it possible to set these values for bulk operations in the first place, this PR also makes Genny more consistent since an operation will now behave the same within a bulk operation as a standalone operation (previously, e.g., copying an operation definition that includes a hint and using it within a bulk operation meant that the hint would not get used).

@rtimmons Feel free to reassign this PR to someone else in case you are not the right person to review this.

### DIFF
--- a/src/cast_core/src/CrudActor.cpp
+++ b/src/cast_core/src/CrudActor.cpp
@@ -207,6 +207,7 @@ struct InsertOneOperation : public WriteOperation {
 
     mongocxx::model::write getModel() override {
         auto document = _document();
+        // Available options: http://mongocxx.org/api/current/classmongocxx_1_1model_1_1insert__one.html
         return mongocxx::model::insert_one{std::move(document)};
     }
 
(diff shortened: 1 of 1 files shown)


src/cast_core/src/CrudActor.cpp
```diff
@@ -297,7 +312,21 @@ struct UpdateManyOperation : public WriteOperation {
...
+        }
+        if (_options.collation()) {
+            op.collation(_options.collation().value());
+        }
+        if (_options.hint()) {
+            op.hint(_options.hint().value());
+        }
+        if (_options.upsert()) {
+            op.upsert(_options.upsert().value());
+        }
```
rtimmons, line 328: Consider refactoring since these `if` blocks appear copy/paste.

rtimmons (approved):

SimonEismann:
@rtimmons I've refactored the code, could you take a quick look at it? This is my first time dabbling in C++ code and I wanna make sure I'm not committing any atrocities :) 

[This patch](https://spruce.mongodb.com/version/64e72dd50ae606f51b83ecc3) shows that genny is successfully setting the upsert option as we see disk writes when performing upserts against an empty db:
image

PR #997, opened 2023-08-22
PERF-4368 Create a 'bulk-loading' workload

SimonEismann:
This workload tests the performance of the bulkWrite API and compares it to normal writes.
It is modeled as a collection of sensor data. It would be common for this use case to do batching using, e.g., AWS SQS.

It tests the following scenarios:
- BulkWrite API, ~30GB data in the db, single connection, no index
- BulkWrite API, ~30GB data in the db, single connection, with index
- BulkWrite API, ~30GB data in the db, many connections, no index
- BulkWrite API, ~30GB data in the db, many connections, with index
- normal writes, ~30GB data in the db, many connections, no index
- normal writes, ~30GB data in the db, many connections, with index
- BulkWrite Upserts, ~30GB data in the db, many connections, with index

Note: This workload requires the fix from PERF-4583, but wanted to get some eyes on this already.

### DIFF
--- a/src/workloads/scale/BulkLoading.yml
+++ b/src/workloads/scale/BulkLoading.yml
@@ -14,6 +14,8 @@ Description: |
   - normal writes, ~30GB data in the db, many connections, no index
   - normal writes, ~30GB data in the db, many connections, with index
 
+  - BulkWrite Upserts, ~30GB data in the db, many connections, with index
+
 # ---Start Phase Description---
 # 0 Load
 # 1 Measure - BulkWriteSingleConnectionNoIndex
@@ -175,8 +177,7 @@ Actors:
       TemplateParameters:
         Name: Warmup
         Threads: 20
-        #OnlyActiveInPhase: [0, 4, 9, 13, 18, 22, 27]
-        OnlyActiveInPhase: [0]
+        OnlyActiveInPhase: [0, 4, 9, 13, 18, 22, 27]
         Duration: 2 minutes
         Operations: *WriteOps100x
 
@@ -185,8 +186,7 @@ Actors:
     Threads: 1
     Phases:
       OnlyActiveInPhases:
-        #Active: [5, 14, 23, 28]
-        Active: [1]
+        Active: [5, 14, 23, 28]
         NopInPhasesUpTo: *NumPhases
         PhaseConfig:
           Repeat: 1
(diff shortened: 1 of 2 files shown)


SimonEismann:
image
Annotated FTDC data from a local run, full FTDC data attached.
[diagnostic.data.zip](https://github.com/mongodb/genny/files/12409483/diagnostic.data.zip)

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+    Threads: 1
+    Database: *DbName
+    Phases:
+      OnlyActiveInPhases:
+        Active: [3, 8, 12, 17, 21, 26, 31]
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Repeat: 1
+
+  - Name: DeleteCollectionActor
```
jimoleary, line 204: Can you swap the order of the delete and quiesce actors to reflect the orders in which they run ?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+            Duration: { ^Parameter: { Name: "Duration", Default: "unset" } }
+            Database: *DbName
+            Collection: *CollectionName
+            Operations:
+              - OperationName: insertOne
+                OperationCommand:
+                  Document: *Document
+
+Actors:
+  # This loads about 35 GB of data to warm up the DB
```
jimoleary, line 162: nit: Maybe '_Each invocation of this actor bulk loads about 35 GB of data to populate the DB_'?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+  - ActorFromTemplate:
+      TemplateName: BulkWriterTemplate
+      TemplateParameters:
+        Name: Warmup
```
jimoleary, line 166: This isn't really warmup, it's Loading based on your phases descriptions earlier.

Question: Should we have a warmup phase?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+              OperationCommand:
+                drop: *CollectionName
+
+  - ActorFromTemplate:
+      TemplateName: BulkWriterTemplate
+      TemplateParameters:
+        Name: BulkWriteSingleConnectionNoIndex
+        Threads: 1
+        OnlyActiveInPhase: [1]
+        Duration: 1 minute
```
jimoleary, line 225: suggestion: make this duration a constant / yml reference and reuse it.

Is 1 minute correct or should we have a longer duration?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+        Threads: 20
+        OnlyActiveInPhase: [15]
+        Duration: 1 minute
+        Operations: *WriteOps100x
+
+  - ActorFromTemplate:
+      TemplateName: NormalWriterTemplate
+      TemplateParameters:
+        Name: NormalWriteManyConnectionsNoIndex
+        Threads: 30
```
jimoleary, line 259: What's the rationale for different thread levels?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+  - ActorFromTemplate:
+      TemplateName: NormalWriterTemplate
+      TemplateParameters:
+        Name: NormalWriteManyConnectionsWithIndex
+        Threads: 30
+        OnlyActiveInPhase: [24]
+        Duration: 1 minute
+
+  # We only test upserts with an index, as they would require COLSCANs without an index.
+  # 120s * 900,000docs/sec = 108,000,000 docs after warmup (slight overestimation since ids could overlapp).
```
jimoleary, line 272: nit: not warmup ... load.

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+        [
+          *UpsertOps10x, *UpsertOps10x, *UpsertOps10x, *UpsertOps10x, *UpsertOps10x,
+          *UpsertOps10x, *UpsertOps10x, *UpsertOps10x, *UpsertOps10x, *UpsertOps10x,
+        ],
+    }
+
+ActorTemplates:
+  - TemplateName: BulkWriterTemplate
+    Config:
+      Name: { ^Parameter: { Name: "Name", Default: "unset" } }
```
jimoleary, line 125: hmm should the 'unset' values just have a reasonable default instead?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+Actors:
+  # This loads about 35 GB of data to warm up the DB
+  - ActorFromTemplate:
+      TemplateName: BulkWriterTemplate
+      TemplateParameters:
+        Name: Warmup
+        Threads: 20
+        OnlyActiveInPhase: [0, 4, 9, 13, 18, 22, 27]
+        Duration: 2 minutes
+        Operations: *WriteOps100x
```
jimoleary, line 170: Maybe WriteOps100x should actually be the default.

jimoleary (commented):
some questions / suggestions.

I don't think you should use 'unset' as a default given that genny doesn't really error check any of this. 

Besides if you picked some reasonable defaults then there will be less yml.

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+# 28 Index
+# 29 Measure - BulkUpsertsManyConnectionsWithIndex
+# 30 Delete
+# 31 Quiese
+# ---End Phase Description---
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
```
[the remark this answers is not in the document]
SimonEismann: Added comments explaining the values

src/workloads/scale/BulkLoading.yml
SimonEismann: The intention here was to include the term Warmup, to make sure this metric gets ignored by the build barons. I've renamed the parameter to WarmupLoadData to better describe what the actor is used for.

src/workloads/scale/BulkLoading.yml
SimonEismann: Based on your question, I've looked at the variability of the results [here](https://github.com/10gen/product-perf-experimentations/blob/master/users/simon.eismann/PERF-4368/analysis.ipynb). Most measurements are sufficiently stable within 1 minute, but I've made the test longer for the non-bulkload writes as they show a higher variation.

src/workloads/scale/BulkLoading.yml
SimonEismann: Added a comment describing this.

src/workloads/scale/BulkLoading.yml
SimonEismann: This did make the file a bit shorter :)

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,287 @@
...
+
+GlobalDefaults:
+  - &NumPhases 31
+  - &DbName db
+  - &CollectionName coll
+
+  # This document layout leads to ~650% CPU utilization on the genny client,
+  # so this is close to the maximum document complexity we can use here.
+  # Using ^RandomDate for the timestamp is too compute intensive.
+  - &Document {
```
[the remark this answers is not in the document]
jimoleary: Pretty much every other workload  file uses the other syntax (and you use it yourself here for actor etc). 

Although that being said, I think I prefer this syntax also.

src/workloads/scale/BulkLoading.yml
jimoleary: What happens to the remaining 'unsets'. I'd almost prefer a 'Mandatory: true' / 'Required: true' option where genny throws an exception if there is no value.

Maybe this is a genny feature request?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,288 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/product-perf"
```
jimoleary, line 2: Should we be the owners?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,288 @@
...
+          PhaseConfig:
+            Duration: { ^Parameter: { Name: "Duration", Default: *MeasurementDuration } }
+            Database: *DbName
+            Collection: *CollectionName
+            Operations:
+              - OperationName: bulkWrite
+                OperationCommand:
+                  WriteOperations:
+                    { ^Parameter: { Name: "Operations", Default: *WriteOps100x } }
+  - TemplateName: NormalWriterTemplate
```
jimoleary, line 153: nit: maybe add a blank line for consistency?

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,288 @@
...
+              - OperationName: insertOne
+                OperationCommand:
+                  Document: *Document
+
+Actors:
+  # Each invocation of this actor bulk loads about 35 GB of data to populate the DB
+  - ActorFromTemplate:
+      TemplateName: BulkWriterTemplate
+      TemplateParameters:
+        Name: WarmupLoadData
```
jimoleary, line 177: hmm is it warmup or load? I would argue it is load (and this is what you have listed at the top of the file).

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,288 @@
...
+        OnlyActiveInPhase: [0, 4, 9, 13, 18, 22, 27]
+        Duration: 2 minutes
```
jimoleary, line 179: nit:

```
        Duration: &LoaderDuration 2 minutes
```

Or add it to `GlobalDefaults:` , it should also be used in the follow on  ActorFromTemplate

src/workloads/scale/BulkLoading.yml
```diff
@@ -0,0 +1,288 @@
...
+      socketTimeoutMS: -1
+      # The default poolsize of 100 is not enough
+      maxPoolSize: 400
+
+GlobalDefaults:
+  - &NumPhases 31
+  - &DbName db
+  - &CollectionName coll
+  - &MeasurementDuration 1 minute
+
```
jimoleary, line 81: Is it worth adding the default thread levels here (20 and 30 and maybe 1)?

jimoleary (commented):
I'm pretty close to happy on this. Some of my comments are pretty nit-picky !

src/workloads/scale/BulkLoading.yml
SimonEismann: Good question. I think we should be the owners since we implemented it. Therefore, we have the best knowledge of the rational behind the test.

src/workloads/scale/BulkLoading.yml
SimonEismann: I'll open a Genny ticket for this, I've seen a lot of workloads coming up with custom indicators that a parameter does not have a valid default. My dream behavior here would be that the default parameter is optional and that genny errors out if the parameter is not provided and there is no default.

SimonEismann: Opened up [EVG-20735](https://jira.mongodb.org/browse/EVG-20735)

jimoleary: :+1:

src/workloads/scale/BulkLoading.yml
jimoleary: oh github hid this response as the block was outdated .... but I think we should fix the way data is ignored rather than using a work around like this.

jimoleary (approved):
lgtm (I will let you decide if you think we need a new patch)

PR #998, opened 2023-08-23
EVG-20132: `OnSession` is `true` by default for `withTransaction`, `startTransaction` & `commitTransaction`, and `OnSession: false` is an error.

PyGeek03:
**Jira Ticket:** EVG-20132

**Whats Changed:**
Set `OnSession` to `true` by default for `withTransaction`, `startTransaction` & `commitTransaction`, because transaction depends on client session as documented in Drivers API doc [here](https://www.mongodb.com/docs/manual/core/transactions-in-applications/#example%23:~:text=When%20using%20the%20drivers%2C%20each%20operation%20in%20the%20transaction%20must%20be%20associated%20with%20the%20session%20(i.e.%20pass%20in%20the%20session%20to%20each%20operation).). 

Raise `InvalidConfigurationException` if `OnSession: false` is set under `withTransaction`, `startTransaction` or `commitTransaction`. This exception would be caught by Genny's CI test suite at dry-run of all workloads step. Users could also check this with `./run-genny workload --dry-run` after editing their workload.

**Patch testing results:**  
- [sys-perf](https://spruce.mongodb.com/version/64e6ae37306615cee1db91cd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC): Sanity check of affected workloads (`ReplaceMillionDocsInSeparateTxns` and `UpdateMillionDocsInTxn` are part of [`refine_shard_key_transaction_stress`](https://github.com/10gen/dsi/blob/87eaf3633f2577e1c612cbc391dcb78a948fbf02/configurations/test_control/test_control.refine_shard_key_transaction_stress.yml))

`CrudActorYamlTests` is not a workload but a unit test for CrudActor.


src/cast_core/src/CrudActor.cpp
```diff
@@ -1117,19 +1117,24 @@ struct CommitTransactionOperation : public BaseOperation {
...
                              PhaseContext& context,
                              ActorId id)
         : BaseOperation(context, opNode),
-          _onSession{onSession},
           _collectionHandle{std::move(collectionHandle)},
           _operation{operation} {
+        auto& onSession = opNode["OnSession"];
+        if (!onSession || !onSession.to<bool>()){
+            BOOST_THROW_EXCEPTION(InvalidConfigurationException(
+                "'withTransaction' requires explicit 'OnSession: true'."));
```
thessem, line 1131: What is the reasoning behind this? If someone has explicitly set `OnSession=false`, then we should respect their choice and carry out the behavior exactly as requested. Why was this stricter route chosen instead of just setting it to true by default?

Does the YAML parsing system have a way to detect the difference between not being present and being false?

src/cast_core/src/CrudActor.cpp
```diff
@@ -1117,19 +1117,24 @@ struct CommitTransactionOperation : public BaseOperation {
...
+        };
+        _onSession = onSession.to<bool>();
         auto& opsInTxn = opNode["OperationsInTransaction"];
         if (!opsInTxn.isSequence()) {
             BOOST_THROW_EXCEPTION(InvalidConfigurationException(
-                "'WithTransaction' requires an 'OperationsInTransaction' node of sequence type."));
+                "'withTransaction' requires an 'OperationsInTransaction' node of sequence type."));
```
thessem, line 1137: This error message is difficult to understand!

src/cast_core/src/CrudActor.cpp
thessem: (I'm guessing `!onSession` is undefined, the cast to bool is defined but false)

src/cast_core/src/CrudActor.cpp
```diff
@@ -1117,19 +1117,24 @@ struct CommitTransactionOperation : public BaseOperation {
 
 struct WithTransactionOperation : public BaseOperation {
     WithTransactionOperation(const Node& opNode,
-                             bool onSession,
+                             bool,
```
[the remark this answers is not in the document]
thessem: Is it possible to take in this parameter as a pointer to detect the difference between unset and false? I'm assuming that's why you retrieved it from the opNode below.

src/cast_core/test/CrudActorYamlTests.yml
```diff
@@ -651,13 +654,15 @@ Tests:
...
-  - Description: WithTransaction, BulkWrite insert, delete, then re-insert without OnSession on bulkWrite
+  - Description: WithTransaction, BulkWrite insert, delete, then re-insert without session on bulkWrite
     Operations:
       - OperationName: withTransaction
         OperationCommand:
+          OnSession: true
           OperationsInTransaction:
           - OperationName: bulkWrite
             OperationCommand:
+              OnSession: false
```
thessem, line 665: What happened in the old code if `OnSession` alternated between true and false at this level?

thessem (commented):
I have some questions on this but I think we're working in the right area!

src/cast_core/src/CrudActor.cpp
rtimmons: > Does the YAML parsing system have a way to detect the difference between not being present and being false?

Yes. The `.maybe<T>()` gives you a `std::optional<T>` that can tell you if `{true, false, unspecified}`.

PyGeek03: - Indeed `!onSession` is the undefined check.
- [Transaction depends on session](https://www.mongodb.com/docs/manual/core/transactions-in-applications/#:~:text=When%20using%20the%20drivers%2C%20each%20operation%20in%20the%20transaction%20must%20be%20associated%20with%20the%20session%20(i.e.%20pass%20in%20the%20session%20to%20each%20operation).), so setting `OnSession=false` under `withTransaction` should be regarded as invalid configuration. @jimoleary suggested this and I agree with him.

src/cast_core/src/CrudActor.cpp
PyGeek03: Yeah I agree, but I don't know how to fix the overall message. I was just fixing the typo here.

src/cast_core/test/CrudActorYamlTests.yml
PyGeek03: The inner `OnSession` (under `bulkWrite`) would override the outer one (under `withTransaction`). In the old test YAML code, `OnSession` under `withTransaction` wasn't set so `addOperation()` would assume that it's false, and `bulkWrite` would inherit this false value for `OnSession` from `withTransaction`.

src/cast_core/src/CrudActor.cpp
thessem: Okay, if it is going to generate a more difficult to diagnose error, then I agree, failing early here when we have context is the right choice. Happy to consider that resolved.

Where does this function get `onSession` from? Can that just become a `*bool` instead of going into manual mode here? I am uncomfortable with a function taking a parameter that it ignores and gets from somewhere else. This function needs to take all its YAML sourced values as args, or it needs to fetch them manually. I prefer the first if it can be done like that.

src/cast_core/src/CrudActor.cpp
thessem: "OperationsInTransaction should be a list of operations" or "withTransaction expected a OperationsInTransaction with a list of operations" if "withTransaction" or a LOC isn't otherwise in the error message. Do either of those match with your understanding of the code here?

"OperationsInTransaction" probably is always a list in all contexts it's in (guessing it's just this one), and when you're reading the YAML it makes more sense as describing it as containing a list rather than it being a node of sequence type (which makes more sense if you're assembling the AST)

src/cast_core/test/CrudActorYamlTests.yml
thessem: Yuck!

src/cast_core/src/CrudActor.cpp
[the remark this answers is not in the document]
PyGeek03: @rtimmons I've just pushed [some commits](https://github.com/mongodb/genny/pull/998/files/553b05f80aa3bad3a6faa16ee2fca4062d70067b..643f008fcf06a9e7380a55a994ea0f2fd8c9cdd0) to remove all unused parameter from transaction-related operations. However, this requires quite a lot of plumbing changes. 

@thessem We sure can take in this parameter as a pointer instead of a boolean, but 1) it would require a lot of unrelated changes because of generic programming using templates (specifically, all `bool _onSession` in this file must be changed to `const Node& _onSession` including the templates, and their usages must be changed accordingly) and 2) I'm not sure if there's much difference between having the constructor taking a pointer to the node as a parameter, and the constructor getting the node by itself.

I did try to remove the usage of templates in this file, but it doesn't seem to be possible because C++ doesn't allow having typenames (and hence constructors) as values in an unordered_map (unlike Python for example). So the alternative would be massive if-else blocks everywhere the actual constructors of operations are invoked.

src/cast_core/src/CrudActor.cpp
PyGeek03: `onSession` is passed in via the `OpCallback opCreator` variable [here](https://github.com/mongodb/genny/blob/3ca962077cb08b0f42b5e8eeba13616c5b66666a/src/cast_core/src/CrudActor.cpp#L1598-L1613). `opCreator` is the value in [this `opConstructors` unordered_map](https://github.com/mongodb/genny/blob/3ca962077cb08b0f42b5e8eeba13616c5b66666a/src/cast_core/src/CrudActor.cpp#L1221-L1245) that corresponds to the operation's name. `OpCallback` is declared [here](https://github.com/mongodb/genny/blob/3ca962077cb08b0f42b5e8eeba13616c5b66666a/src/cast_core/src/CrudActor.cpp#L139-L140).

src/cast_core/src/CrudActor.cpp
PyGeek03: Yup I agree with you and have updated the error message. Thanks!

src/cast_core/test/CrudActorYamlTests.yml
PyGeek03: Yeah all this implicit false and implicit value inheritance from outer scope is a bit weird 😂

thessem (changes requested):
Minh and I spoke about this PR in person. I have asked him to do the following:

- Modify `addOperation` inside `phaseConfig` to check for `withTransaction` and set the default value for `onSession` to `true` if it’s unspecified, but only for `withTransaction`. Otherwise, it remains `false`.
    - Note that it isn't ideal to do this here, but doing it otherwise would require all operations to become aware of `onSession` at a YAML level (a bigger change we could do later) OR give `withTransaction` a special function signature making it need special case handling.
    - Also note that `startTransaction` and `commitTransaction` need to be caught in this if statement for historical reasons. Fully removing these is out of scope for this PR.
- Ensure that `WithTransactionOperation` retains exactly the same call signature as the other operations, including a `bool onSession`.
- `WithTransactionOperation` is responsible for emitting an error if `onSession` is false because operations should get to decide their validation.
- Explicitly decide to make `withTransaction` have a different default value for `onSession` (i.e., true) vs other operations. This is because requiring an explicit `onSession: true` for `withTransaction` seems annoyingly verbose, difficult to discover for our users (learning about this via CI error is not great), and operations getting to decide this makes more sense.

(Mostly posting this explicitly for @rtimmons's sake, let me know if you disagree with the above here, certainly willing to discuss any of the above, I just don't want Minh to get stuck in the middle of a double sided code-review!)

src/cast_core/src/CrudActor.cpp
```diff
@@ -1273,8 +1286,12 @@ struct DropOperation : public BaseOperation {
 std::unordered_map<std::string, OpCallback&> getOpConstructors() {
     // Maps the yaml 'OperationName' string to the appropriate constructor of 'BaseOperation' type.
     std::unordered_map<std::string, OpCallback&> opConstructors = {
-        {"aggregate", baseCallback<BaseOperation, OpCallback, AggregateOperation>},
         {"bulkWrite", collectionHandleCallback<BaseOperation, OpCallback, BulkWriteOperation>},
+        {"withTransaction",
```
lgfang, line 1290: It appears that you didn't actual change these lines but just moved them around. May I know the reason of adjusting the order of this list?

lgfang (approved):

src/cast_core/src/CrudActor.cpp
PyGeek03: I wanted to group the op constructors that use `collectionHandleCallback` ("bulkWrite" and "withTransaction") closer together, while also keeping the transaction-related op constructors close together.

PyGeek03: I wanted to group the op constructors that use `collectionHandleCallback` ("bulkWrite" and "withTransaction") closer together, while also keeping the transaction-related op constructors close together.

thessem (approved):
This turned out very neatly. Thanks for your work here!

thessem (approved):
This turned out very neatly. Thanks for your work here!

PyGeek03:
evergreen merge

PR #1003, opened 2023-08-29
PERF-4603 Rewrite the QueryStats workloads to track various configurations' performance.

wqian94:
**Jira Ticket:** PERF-4603

**Whats Changed:**  
The `QueryStats.yml` workload is now refactored as a composition of smaller workloads, with proper query stats cache reset and prepopulation.

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/651327f61e2d17297de520b3)


src/workloads/query/query_stats/QueryStats.tmpl
```diff
@@ -0,0 +1,276 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is a template of predefined workload groups for QueryStats workloads.
+
+# Workload "constants."
+ConstantDatabase: &Database {^Parameter: {Name: Database, Default: test}}
```
wbuerger46, line 7: Why describe this as a parameter, if you're never actually using it as a set-able parameter (unless I'm missing where you're setting it)?
I'm not convinced it's worth putting these "constants" in the template file. I understand the intent of trying to factor out as many commonalities as possible, but I think each one makes it slightly more confusing to a reader trying to understand a workload file (by adding one more level of indirection), and each of those small bits is adding up to an experience where it's taking me a while to make sense of all of these workloads since the bits are so de-centralized. I wonder if there's a balance to strike with a bit less abstraction.

wbuerger46 (changes requested):
My comment re the "constants" applies more broadly to all the templating. I think there might be a balance to strike where we can (1) keep leveraging templating for nice abstractions that make it easier to modify these in the future but also (2) make it easy for a reader to comprehend one workload without a lot of digging.

wbuerger46:
It appears one side effect of moving these workloads into `query_stats` folder is that in Evergreen they're showing up with the task name, for example, `query_stats_query_stats`. Not a huge deal, but I don't think that's the most desirable. Just wanted to bring that up - I don't have a strong opinion on fixing or not since I think the folder structure is nice :/

cswanson310 (commented):
Uhh there's a ton going on here... Could we maybe meet up to talk about it? 

Generally, you have to keep in mind that all this templating is very hard for people to read and grok. There is no "jump to definition" within the YAML, and people will be reading these things a lot. 

If I got here after seeing there was a problem in "FindCommand-Cycle25-Random-HMAC-CacheSize300MB-Prepopulate300MB"'s "OperationThroughput", how would I know where to start looking? I cannot actually search for that term, or really any sub-piece of it. "Cycle25" is not meaningful to search for, so I'd have to search for "Cycle" and hope I find the right one? Basically I have to decode your whole template parameter system to debug my issue. That's ... not great.

I often give feedback that for correctness tests it is better to repeat some code and unroll some loops to aid in debugability. For example, in my humble opinion:
```js
for (let testCase of [{x: 1}, {x: 2}]) {
    testFn(testCase);
}
```

Is worse as a test than
```js
testFn({x: 1});
testFn({x: 2});
```

even if there are like 1000 test cases. Even if you could programmatically generate those test cases.  In the first example the stack trace will always point to that line in the for loop. In the second example, a stack trace will tell you which was the failing test case. That's why we have some helpers like this one: [`export function withEachMergeMode(callback)`](https://github.com/mongodb/mongo/blob/6f7a051255ca8b594da4e229a91f70df916a556c/jstests/aggregation/extras/merge_helpers.js#L14)

This patch feels like a similar case where we want to sort of go against our engineering instincts to factor everything out and abstract it away. We don't actually want to reduce the total amount of code. Abstracting away is making it harder for the future readers to un-abstract. So some copy/pasting is worth it IMO.

Can we discuss some ideas for doing this? I don't want to throw away all the templating necessarily, since many of our existing test cases go a little too far in the copy/paste direction. They are copying/pasting information that is not actually very relevant to understanding what's going on and which just bloat the screen. There's gotta be a better balance.

wqian94:
I don't really have any opinions on whether to templatize more or less. It was just easier to develop the tests this way, but this can be easily changed by simply evaluating the workload and dumping that into a file.

I'm more interested in getting this landed because we shouldn't be spending weeks trying to decide what the fixed test suite should look like.

cswanson310:
You can "evaluate the workload and dump it to a file"? What does that look like for one of these?

wqian94:
Well, it'd be a very flat file, for sure. It's around 450 ish lines if the Nops are removed (which they would be, in favor of some syntactical sugar).

```
# This file was generated by running the Genny preprocessor on the workload Main.yml
Clients:
  Default:
    QueryOptions:
      maxPoolSize: 4000
      connectTimeoutMS: 3600000
      socketTimeoutMS: 3600000
    URI: mongodb://localhost:27017
SchemaVersion: '2018-07-01'
Owner: '@mongodb/query'
Description: 'This is the main workload for testing $queryStats performance regressions.

  '
KeyWords:
- queryStats
- redaction
GlobalDefaults:
  ConstantNumGroups: 2
  ConstantStagesPerGroup: 2
  ConstantFixedEntries: 175
  ConstantRandomEntries:
    ^RandomInt:
      min: 1
      max: 350
Actors:
- Name: ClearCollection
  Type: CrudActor
  Database: test
  Threads: 1
  Phases:
  - Repeat: 1
    Threads: 1
    Collection: Collection0
    Operations:
    - OperationName: drop
- Name: InsertData
  Type: Loader
  Threads: 4
  Phases:
  - Repeat: 1
    Database: test
    MultipleThreadsPerCollection: true
    CollectionCount: 1
    DocumentCount: '1e6'
    BatchSize: 1000
    Document:
      a1:
        ^RandomInt:
          min: -100
          max: 100
      a2:
        ^RandomInt:
          min: -100
          max: 100
      string:
        ^RandomString:
          length: 5
    Indexes:
    - keys:
        a1: 1
        a2: 1
- Name: ClearQueryStats-Unique-Fixed-Base-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 0MB
- Name: SetQueryStats-Unique-Fixed-Base-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 300MB
- Name: Quiesce-Unique-Fixed-Base-CacheSize300MB-Prepopulate300MB
  Type: QuiesceActor
  Threads: 1
  Database: test
  Phases:
  - Repeat: 1
  - Repeat: 1
- Name: Prepopulate-Unique-Fixed-Base-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 301 seconds
    Repeat: 30001
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1:
              ^RandomDouble:
                min: 1000000
                max: 10000000000
          - ^Object:
              withNEntries: 10
              havingKeys:
                ^FormatString:
                  format: '%32d'
                  withArgs:
                  - ^Inc:
                      start: 0
                      multiplier: 1
                      step: 32
              andValues:
                $exists: true
              duplicatedKeys: skip
- Name: FindCommand-Unique-Fixed-Base-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 5 minutes
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1: 1
          - a2: 1
          - ^Object:
              withNEntries: 175
              havingKeys:
                ^RandomString:
                  length: 10
              andValues:
                $exists: false
              duplicatedKeys: skip
- Name: ClearQueryStats-Unique-Random-Base-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 0MB
- Name: SetQueryStats-Unique-Random-Base-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 300MB
- Name: Quiesce-Unique-Random-Base-CacheSize300MB-Prepopulate300MB
  Type: QuiesceActor
  Threads: 1
  Database: test
  Phases:
  - Repeat: 1
  - Repeat: 1
- Name: Prepopulate-Unique-Random-Base-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 301 seconds
    Repeat: 30001
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1:
              ^RandomDouble:
                min: 1000000
                max: 10000000000
          - ^Object:
              withNEntries: 10
              havingKeys:
                ^FormatString:
                  format: '%32d'
                  withArgs:
                  - ^Inc:
                      start: 0
                      multiplier: 1
                      step: 32
              andValues:
                $exists: true
              duplicatedKeys: skip
- Name: FindCommand-Unique-Random-Base-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 5 minutes
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1: 1
          - a2: 1
          - ^Object:
              withNEntries:
                ^RandomInt:
                  min: 1
                  max: 350
              havingKeys:
                ^RandomString:
                  length: 10
              andValues:
                $exists: false
              duplicatedKeys: skip
- Name: ClearQueryStats-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 0MB
- Name: SetQueryStats-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 300MB
- Name: Quiesce-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: QuiesceActor
  Threads: 1
  Database: test
  Phases:
  - Repeat: 1
  - Repeat: 1
- Name: Prepopulate-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 301 seconds
    Repeat: 30001
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1:
              ^RandomDouble:
                min: 1000000
                max: 10000000000
          - ^Object:
              withNEntries: 10
              havingKeys:
                ^FormatString:
                  format: '%32d'
                  withArgs:
                  - ^Inc:
                      start: 0
                      multiplier: 1
                      step: 32
              andValues:
                $exists: true
              duplicatedKeys: skip
- Name: FindCommand-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 5 minutes
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1: 1
          - a2: 1
          - ^Object:
              withNEntries: 175
              havingKeys:
                ^RandomString:
                  length: 10
              andValues:
                $exists: false
              duplicatedKeys: skip
- Name: QueryStatsAggregate-Unique-Fixed-HMAC-CacheSize300MB-Prepopulate300MB
  Type: AdminCommand
  Threads: 1
  Phases:
  - Duration: 5 minutes
    SleepBefore: 1 second
    Database: admin
    Operations:
    - OperationMetricsName: QueryStatsMetrics
      OperationName: AdminCommand
      OperationCommand:
        aggregate: 1
        pipeline:
        - $queryStats:
            transformIdentifiers:
              algorithm: hmac-sha-256
              hmacKey:
                ^BinDataSensitive:
                  numBytes: 32
        - $count: numDocuments
        cursor: {}
- Name: ClearQueryStats-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 0MB
- Name: SetQueryStats-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: RunCommand
  Threads: 1
  Phases:
  - Repeat: 1
    Database: admin
    Operations:
    - OperationName: AdminCommand
      OperationCommand:
        setParameter: 1
        internalQueryStatsCacheSize: 300MB
- Name: Quiesce-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: QuiesceActor
  Threads: 1
  Database: test
  Phases:
  - Repeat: 1
  - Repeat: 1
- Name: Prepopulate-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 301 seconds
    Repeat: 30001
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1:
              ^RandomDouble:
                min: 1000000
                max: 10000000000
          - ^Object:
              withNEntries: 10
              havingKeys:
                ^FormatString:
                  format: '%32d'
                  withArgs:
                  - ^Inc:
                      start: 0
                      multiplier: 1
                      step: 32
              andValues:
                $exists: true
              duplicatedKeys: skip
- Name: FindCommand-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: CrudActor
  Threads: 32
  Database: test
  Phases:
  - Duration: 5 minutes
    Collection: Collection0
    Operations:
    - OperationName: find
      OperationCommand:
        Filter:
          $and:
          - a1: 1
          - a2: 1
          - ^Object:
              withNEntries:
                ^RandomInt:
                  min: 1
                  max: 350
              havingKeys:
                ^RandomString:
                  length: 10
              andValues:
                $exists: false
              duplicatedKeys: skip
- Name: QueryStatsAggregate-Unique-Random-HMAC-CacheSize300MB-Prepopulate300MB
  Type: AdminCommand
  Threads: 1
  Phases:
  - Duration: 5 minutes
    SleepBefore: 1 second
    Database: admin
    Operations:
    - OperationMetricsName: QueryStatsMetrics
      OperationName: AdminCommand
      OperationCommand:
        aggregate: 1
        pipeline:
        - $queryStats:
            transformIdentifiers:
              algorithm: hmac-sha-256
              hmacKey:
                ^BinDataSensitive:
                  numBytes: 32
        - $count: numDocuments
        cursor: {}
- Name: PhaseTimingRecorder
  Type: PhaseTimingRecorder
  Threads: 1
AutoRun:
- When:
    mongodb_setup:
      $eq:
      - standalone-query-stats
      - standalone-query-stats-small-rate-limit
      - single-replica-query-stats
      - replica-query-stats
      - atlas-like-replica-query-stats.2022-10
```

src/phases/query/QueryStatsVariants.tmpl
```diff
@@ -0,0 +1,157 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is a template of predefined variants for QueryStats workloads.
```
cswanson310, line 4: What exactly is a "variant" - can you elaborate here? It seems like this is a new layer of abstraction, as opposed to the other files here which are common to genny things. So... needs very clear documentation - it also shares a name with the 'evergreen variant' terminology - is there perhaps a better name? Did you consider others? It seems maybe like a "MacroPhase"? or something?

src/phases/query/QueryStatsVariants.tmpl
```diff
@@ -0,0 +1,157 @@
...
+
+GlobalDefaults:
+  # Local constants.
+  PhasesPerStage: &PhasesPerStage 6
```
cswanson310, line 8: "stage" is different than "variant" - is this intentional? What is the difference if so? Could you add comments?

src/phases/query/QueryStatsVariants.tmpl
```diff
@@ -0,0 +1,157 @@
...
+  EntriesPerObject: &EntriesPerObject {^Parameter: {Name: EntriesPerObject, Default: {^RandomInt: {min: 1, max: 350}}}}
+  Group: &Group {^Parameter: {Name: Group, Default: 1}}
+  StageInGroup: &StageInGroup {^Parameter: {Name: StageInGroup, Default: 1}}
+  # For if there needs to be additional initial phases.
+  InitialPhase: &InitialPhase {^Parameter: {Name: InitialPhase, Default: *SetupPhases}}
+  NumGroups: &NumGroups {^Parameter: {Name: NumGroups, Default: {unused: "Need to specify the number of groups being run."}}}
+  StagesPerGroup: &StagesPerGroup {^Parameter: {Name: StagesPerGroup, Default: {unused: "Need to specify the number of stages per group."}}}
+
+  # Syntactic sugar.
+  CycleLength: &CycleLength {^Parameter: {Name: CycleLength, Default: {unused: "Need to provide the cycle length."}}}
```
cswanson310, line 22: I think I understand what the cycle length is based on our slack history, but for future readers, what is a "cycle"?

src/phases/query/QueryStatsActors.tmpl
```diff
@@ -0,0 +1,117 @@
...
+        Database: admin
+        Operations:
+        - OperationName: AdminCommand
+          OperationCommand:
+            # Configuring the queryStats store size to an arbitrary value which seems reasonable.
+            # This ensures that if run on different sized machines, the test will remain the same.
+            setParameter: 1
+            internalQueryStatsCacheSize: {^Parameter: {Name: Size, Default: {unused: "Need to specify the query stats cache size to change to."}}}
+
+# Syntactic sugar to reduce the size of the workload file.
```
cswanson310, line 37:
```suggestion
# A pretty normal "Quiesce" actor - with syntactic sugar to reduce the size of the workload file.
```

src/phases/query/QueryStatsActors.tmpl
```diff
@@ -0,0 +1,117 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is a template library of actors for QueryStats workloads.
```
cswanson310, line 4: Can you comment each one please?

src/phases/query/QueryStatsStages.tmpl
```diff
@@ -0,0 +1,161 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is a template of helper stages for QueryStats workloads. Each stage represents a single
```
cswanson310, line 4: This sentence is not really making sense... I think it will make sense once I read further, but ideally it makes sense _before_ I read further.

Questions I have when reading this:
* When you say "each stage" - there are 2.5 here in this file? The SetupStage and the QueryStatsStageTemplate. What is this mysetrious "StageTemplate"? Do we need it to be a separate thing?
* Related to my other question but what is "a single workload variant"? A "such as" might help here.
* Stages can be grouped together ... why and how?

src/phases/query/QueryStatsStages.tmpl
```diff
@@ -0,0 +1,161 @@
...
+  workload variant, comprising multiple phases. Multiple stages can be grouped together if they have
+  commonalities in their variations.
+
+GlobalDefaults:
+  # Workload "constants."
```
cswanson310, line 9: Some of these seem to be repeated here and in the actors? DB and Collection. If this is required then it's OK, but a comment about how they need to be in sync between the two places is probably worthwhile.

src/phases/query/QueryStatsVariants.tmpl
```diff
@@ -0,0 +1,157 @@
...
+            ofLength: *CycleLength
+            fromGenerator:
+              ^Object:
+                withNEntries: *EntriesPerObject
+                havingKeys: {^RandomString: {length: 10}}
+                andValues: {$exists: false}
+                duplicatedKeys: skip
+
+CycleHmacStage:
+  ^FlattenOnce:
```
cswanson310, line 72: what's with the flattenings? Comment maybe? Are they necessary?

src/phases/query/QueryStatsVariants.tmpl
```diff
@@ -0,0 +1,157 @@
...
+        MaxPhases: *MaxPhases
+
+        # Template variables.
+        StartingPhase: *GroupAndStageToPhase
+        Suffix: {^PreprocessorFormatString: {format: "Cycle%d-%s-Text", withArgs: [*CycleLength, *Name]}}
+
+        # Payload
+        FindCommandPayload: *CyclePayload
+
+UniqueBaseStage:
```
cswanson310, line 104: what does 'unique' mean in this context?

src/phases/query/QueryStatsWorkloads.tmpl
```diff
@@ -0,0 +1,234 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is a template of predefined workload groups for QueryStats workloads.
```
cswanson310, line 4: what's a "workload group"?

src/phases/query/QueryStatsWorkloads.tmpl
```diff
@@ -0,0 +1,234 @@
...
+
+# Template parameters.
+ConstantGroup: &Group {^Parameter: {Name: Group, Default: 1}}
```
cswanson310, line 7: what is 'constant'?

src/phases/query/QueryStatsWorkloads.tmpl
```diff
@@ -0,0 +1,234 @@
...
+ConstantNumGroups: &NumGroups {^Parameter: {Name: NumGroups, Default: {unused: "Need to specify the total number of groups."}}}
+# Default should be the max of all groups' stage counts.
+ConstantStagesPerGroup: &StagesPerGroup {^Parameter: {Name: StagesPerGroup, Default: 3}}
+
+# Entry count generators.
```
cswanson310, line 12: what is a 'count generator'?

src/workloads/query/query_stats/BaseFixed.yml
```diff
@@ -0,0 +1,41 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This workload tests for performance regressions without concurrent $queryStats aggregates,
+  using fixed-size payloads in the find queries.
```
cswanson310, line 5: fixed size of variable shape or all of the same shape? Or cycling? Or all of the above?

src/workloads/query/query_stats/BaseFixed.yml
```diff
@@ -0,0 +1,41 @@
...
+      Key: AllBaseFixedGroup
+      Parameters:
+        NumGroups: *NumGroups
+        Group: 1
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-query-stats
```
cswanson310, line 40: Can we run this on replica sets as well? And maybe also sharded clusters? I think this is a good template list: https://github.com/mongodb/genny/blob/788eaf26e3b8b08d76c71d54fb0013befee9b032/src/workloads/query/TenMDocCollection_IntId.yml#L188 - would that work? Any concerns?

src/workloads/query/query_stats/Main.yml
```diff
@@ -0,0 +1,84 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: >
+  This is the main workload for testing $queryStats performance regressions.
```
cswanson310, line 4: what makes it "the main one"?

cswanson310 (changes requested):
Ok I left a lot of questions :) I apologize in advance for some dumb-sounding questions which I could definitely deduce the answer to by looking harder and longer. I am hoping to avoid this step for future newcomers.

I think the change to include phase number is good. I also am glad you were able to dump the output - although it is sad it has so many noops :(

Could we include a README.md in this new directory about how to generate that post-preprocessor output and maybe a general overall summary of the tests here?

Finally, this seems to be generating a lot of data... and each phase is running for 5 minutes... Based on a conversation that @wbuerger46 and I had - 5 minutes does helpfully stabilize things, but it seems like things get stable after about 20 seconds in that one example: https://github.com/mongodb/genny/blob/788eaf26e3b8b08d76c71d54fb0013befee9b032/src/workloads/query/TenMDocCollection_IntId.yml#L33-L34

So - maybe we can do 100 seconds here? 

I'm having trouble understanding the dimensionality here also - how many experiments are we running? How many unique (task, test) names are there? I am trying to evaluate whether we want to keep all of these, or maybe restrict down the "AutoRun" on some of these to limit the cost/noise but keep them functional.

src/workloads/query/QueryStats.yml
```diff
@@ -10,233 +10,164 @@ KeyWords:
...
-
-# QueryStats without HMAC.
-- ActorFromTemplate:
-    TemplateName: QueryStatsAggregation
-    TemplateParameters:
-      Name: "QueryStatsAggregationWithoutHMAC"
-      queryStatsParameters: {}
-      OnlyActiveInPhase: 8
+- LoadConfig:
+    Path: ../../phases/query/QueryStats.tmpl
```
wbuerger46, line 38: Most people will come to this point in the workload file to decipher the flow of the workload phases, and even though you've written them out in a useful order, it doesn't make it clear in what phases `ClearCollection`, `InsertData`, or some of the actors run. My suggestion would be to make "Phases" a parameter just like it is for the other actors. Even though the parameter would be technically unnecessary it will make it easier to follow. Alternatively, you could add a comment here to clarify.

wbuerger46 (changes requested):
A couple nits about wording.

My only overall feedback is still some confusion about what's going on in each phase of the workload. It looks to me like phases 3-9 are no-ops? Why are we ignoring those phases? Do we really need 47 phases? That seems like an awful lot of test output to sort through. And most of all, it's taking me a lot of time to piece together which actors pair up with which other actors in which phases. Could you add a (probably verbose) comment at the top to outline the purpose of each phase?

wqian94:
> A couple nits about wording.
> 
> My only overall feedback is still some confusion about what's going on in each phase of the workload. It looks to me like phases 3-9 are no-ops? Why are we ignoring those phases? Do we really need 47 phases? That seems like an awful lot of test output to sort through. And most of all, it's taking me a lot of time to piece together which actors pair up with which other actors in which phases. Could you add a (probably verbose) comment at the top to outline the purpose of each phase?

Empty phases don't generate any substantial amount of output. Given Charlie's earlier point about wanting to distinguish which phase's output corresponds to what kind of data, I figured it'd be easier in this case to consistently put the FindCommand and QueryStats actors in a specific phase. It also makes the math a lot easier. Lemme see if I can figure out how to document the phases more clearly... but that's kind of why I wanted to hide these setup phases from whoever is writing the workload in the first place.

src/phases/query/QueryStats.tmpl
```diff
@@ -0,0 +1,230 @@
...
+    InsertData: (requires MaxPhase, Phase) populates the working collection, should only need to be run once.
+    CacheSizeActor: (requires MaxPhase, Phases, CacheSizeMb) changes the QueryStats cache size.
+    RateLimitActor: (requires MaxPhase, Phases, RateLimitHz) changes the QueryStates rate limit. Should be set to -1 before each QueryStats cache prepopulation to speed things up.
+    QuiesceActor: (requires MaxPhase, Phases) ensures that actors quiesce before continuing.
+    PrepopulateActor: (requires MaxPhase, Phases, CacheSizeMb) prepopulates the QueryStats cache to ensure that it is filled with similarish shapes before every workload execution.
+    FindCommandActor: (requires MaxPhase, Phase, CacheSizeMb, RateLimitHz, Payload; optional Suffix) runs find commands to generate query stats.
+    AggregateActor: (requires MaxPhase, Phase, CacheSizeMb, RateLimitHz; optional Suffix, QueryStatsParameters) runs $queryStats.
+
+  Parameters:
+    # Required parameters.
```
cswanson310, line 17: I'm a little confused by some of this setup reading through for the first time - above you say that certain actors have different required parameters - here it appears that all of these parameters are required and that we'll ben instantiating the template exactly once?

I think maybe just delete this comment? Or if you have a better understanding of it you could summarize, do that :D

src/workloads/query/QueryStats.yml
```diff
@@ -10,233 +10,179 @@ KeyWords:
...
+    Key: ClearCollection  # Drops the collection, only needs to happen once.
+    Parameters:
+      MaxPhase: *MaxPhase
+      Phase: 0
+- LoadConfig:
+    Path: ../../phases/query/QueryStats.tmpl
+    Key: InsertData  # Repopulates the collection, only needs to happen once.
```
cswanson310, line 45: Can we make the data format schema live here where we have the query format schema also? Taking the documents as a template argument to the insertion actor?

Or did you already try this and it didn't work for some reason?

cswanson310: Actually - looking closer - more of the queries are in the template file, so it's fine that the data document is next to those queries - do the "fixed" and "random" payloads need to live in this file? Curious to hear your thought process. I am just trying to cut down on the need to cross-reference the two and have them both open.

cswanson310 (commented):
A couple scattered comments but otherwise this looks much better! I am curious to see how it looks in evergreen also.

src/workloads/query/QueryStats.yml
```diff
@@ -10,233 +10,176 @@ KeyWords:
...
+      QueryStatsParameters: *HmacParameters
+      RateLimitHz: *RateLimitHz
+      Suffix: "Hmac-Random"
 
 AutoRun:
 - When:
     mongodb_setup:
       $eq:
       - standalone-all-feature-flags
       - standalone-query-stats
```
cswanson310, line 174: I think that the first 3 in this list will all provide the same data, yeah? I think we can get away with just running one of these 3, and doing the same for the other platforms. I think maybe just do the non-query-stats ones? That seems like the end state we want, so we won't have to change it up later.

src/workloads/query/QueryStatsLimited.yml
```diff
@@ -0,0 +1,175 @@
...
+      Phase: 47
+      QueryStatsParameters: *HmacParameters
+      RateLimitHz: *RateLimitHz
+      Suffix: "Hmac-Random-RateLimited"
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-query-stats-small-rate-limit
```
cswanson310, line 173: I think given the name of the workload and the name of the actors, should we move this to match the other file? Just always run on the "normal" variants like "standalone" and "replica" and "shard"?

cswanson310 (changes requested):
Can you provide a patch build link again when you get a chance? Curious to do an initial glance at the numbers and some comparisons to see that everything is behaving as we expect.

src/phases/query/QueryStats.tmpl
wqian94: Oh -- some of the required parameters are simply not used in some actors. Genny's idea of "required" is simply that the default value is something that is unlikely to be accepted, so that the runtime will error unless that value is overridden. I can explain that further in the template, sure.

src/workloads/query/QueryStats.yml
wqian94: Ideally, they'd provide the same data, but this isn't a given. The rate-limited query stats variants run with a different server parameter. I'm of the opinion that server parameter differences should still be tested to make sure that the performance doesn't change. This adds an extra layer of protection and information in case something _does_ create a performance difference due to the rate limit parameter.

src/workloads/query/QueryStatsLimited.yml
wqian94: That could also work. :shrug: This isn't specifically tied to any particular variant in practice, since we're manually resetting the rate limit at every turn. We could even put it into the same file and just give the rate-limited actors new names.

cswanson310 (approved):
Looking good! Thanks for all the adjustments! This is some solid coverage and much more comprehensible now, thanks!

wqian94:
Manually merged via evergreen.

PR #1006, opened 2023-08-31
PERF-4637 Create a workload that can be affected by the optimization to expression search for parameter reuse

henrinikku:

**Jira Ticket:** PERF-4637

**Whats Changed:**  
A new workload was added to test the performance of expression search for parameter re-use during parameterization.

**Patch testing results:**  
https://spruce.mongodb.com/version/64f07cbb1e2d170ed4b21ff3/tasks

**Workload Submission form:**  
Submitted

**Related PRs:**   
[SERVER-79092](https://github.com/10gen/mongo/pull/15219)

### DIFF
--- a/src/workloads/query/FilterWithComplexLogicalExpression.yml
+++ b/src/workloads/query/FilterWithComplexLogicalExpression.yml
@@ -25,6 +25,8 @@ Keywords:
 - QuiesceActor
 - insert
 - find
+- matcher
+- expressions
 
 GlobalDefaults:
   Database: &Database test
--- a/src/workloads/query/FilterWithComplexLogicalExpression.yml
+++ b/src/workloads/query/FilterWithComplexLogicalExpression.yml
@@ -25,8 +25,6 @@ Keywords:
 - QuiesceActor
 - insert
 - find
-- matcher
-- expressions
 
 GlobalDefaults:
   Database: &Database test
--- a/src/workloads/query/MatchWithLargeExpression.yml
+++ b/src/workloads/query/MatchWithLargeExpression.yml
@@ -1,0 +1,150 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of expression search for parameter re-use during parameterization. 
+  Before September 2023, parameter re-use had O(n^2) complexity due to using a vector for looking up equivalent expressions. 
+  SERVER-79092 fixed this issue by switching over to a map once the amount of expressions reaches a threshold (currently 50).
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  MaxPhases: &MaxPhases 7
+  DocumentCount: &DocumentCount 10
+
+Keywords:
+- Loader
+- CrudActor
+- QuiesceActor
+- insert
+- Aggregation
+
+ActorTemplates:
+- TemplateName: MatchExpressionWithOrClausesTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "unused"}}
+    Type: CrudActor
+    Database: *Database
+    Threads: 1
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "ActivePhase", Default: -1}}]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: {^Parameter: {Name: "Repeat", Default: 1}}
+          Collection: *Collection
+          Operations:
+          - OperationName: aggregate
+            OperationCommand:
+              Pipeline: [
+                {$match: {$or: {^Array: {of: 
+                  {$and: [
+                    {a: {$gt: {^Inc: {start: 10}}}},
+                    {b: {$gt: {^Inc: {start: 10}}}}
+                  ]},
+                  number: {^Parameter: {Name: "NumberOfClauses", Default: -1}}
+                }}}}
+              ]
+
+Actors:
+# Clear any pre-existing collection state.
+- Name: ClearCollection
+  Type: CrudActor
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: 1
+        Collection: *Collection
+        Operations:
+        - OperationName: drop
+
+- Name: InsertData
+  Type: Loader
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        Collection: *Collection
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: *DocumentCount
+        Document:
+          a: [0, 1]
+          b: [2, 3]
+          c: 4
+          d: 5
+          object: {}
+
+- Name: Quiesce
+  Type: QuiesceActor
+  Threads: 1
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith50Clauses
+      Repeat: 1000
+      ActivePhase: 3
+      NumberOfClauses: 50
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith100Clauses
+      Repeat: 1000
+      ActivePhase: 4
+      NumberOfClauses: 100
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith1000Clauses
+      Repeat: 100
+      ActivePhase: 5
+      NumberOfClauses: 1000
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith10000Clauses
+      Repeat: 10
+      ActivePhase: 6
+      NumberOfClauses: 10000
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith50000Clauses
+      Repeat: 5
+      ActivePhase: 7
+      NumberOfClauses: 50000
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+      - standalone-query-stats
+    branch_name:
+      $gte:
+        v7.0
--- a/src/workloads/query/MatchWithLargeExpression.yml
+++ b/src/workloads/query/MatchWithLargeExpression.yml
@@ -146,7 +146,7 @@ Actors:
     TemplateName: MatchExpressionWithOrClausesTemplate
     TemplateParameters:
       Name: RunMatchExpressionOrWith50000Clauses
-      Repeat: 5
+      Repeat: 10
       ActivePhase: 9
       NumberOfClauses: 50000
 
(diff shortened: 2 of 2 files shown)


src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of expression search for parameter re-use during parameterization. 
+  Before September 2023, parameter re-use had O(n^2) complexity due to using a vector for looking up equivalent expressions. 
+  SERVER-79092 fixed this issue by switching over to a map once the amount of expressions reaches a threshold (currently 50).
```
jimoleary, line 6: Are we planning to commit this before SERVER-79092 so that we can see the improvement? Although it looks like you already have this in the PR for SERVER-79092.

jimoleary: What metrics are important for this workload, throughput, latency, .... etc?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+  Collection: &Collection Collection0
+  MaxPhases: &MaxPhases 7
+  DocumentCount: &DocumentCount 10
+
+Keywords:
+- Loader
+- CrudActor
+- QuiesceActor
+- insert
+- Aggregation
```
jimoleary, line 19: plus matcher and expressions maybe?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+                    {a: {$gt: {^Inc: {start: 10}}}},
+                    {b: {$gt: {^Inc: {start: 10}}}}
+                  ]},
+                  number: {^Parameter: {Name: "NumberOfClauses", Default: -1}}
+                }}}}
+              ]
+
+Actors:
+# Clear any pre-existing collection state.
+- Name: ClearCollection
```
jimoleary, line 50: You shouldn't really need this actor (although I can see how it makes life easier when running the workload locally).

Maybe a default false ```dropFirst``` attribute on the loader actor would be a reasonable feature request to replace this functionality?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        Collection: *Collection
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: *DocumentCount
+        Document:
+          a: [0, 1]
```
jimoleary, line 80: It looks like this data combined with the query means that we never match anything, is that an issue?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+Description: |
+  This workload tests the performance of expression search for parameter re-use during parameterization. 
+  Before September 2023, parameter re-use had O(n^2) complexity due to using a vector for looking up equivalent expressions. 
+  SERVER-79092 fixed this issue by switching over to a map once the amount of expressions reaches a threshold (currently 50).
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  MaxPhases: &MaxPhases 7
+  DocumentCount: &DocumentCount 10
```
jimoleary, line 12: This is a very small number of document? Should we be testing a more realistic volume of data?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+      Name: RunMatchExpressionOrWith50000Clauses
+      Repeat: 5
+      ActivePhase: 7
+      NumberOfClauses: 50000
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
```
jimoleary, line 141: given the size and scale of the workload, this could also be a microbenchmarks test.

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: Would an index make a difference ? (although this is not likely with the number of documents).

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+      - standalone-query-stats
+    branch_name:
+      $gte:
+        v7.0
```
alyacb, line 150: nit: newline

alyacb, line 150: You definitely want to set the $gte to the targeted version of your ticket- which will probably go in for 7.2, since we branched 7.1 already (unless this change will be backported?)

src/workloads/query/MatchWithLargeExpression.yml
alyacb: I thought the usual recommendation was to add all new workloads to genny- has that changed (just for my own knowledge)?

src/workloads/query/MatchWithLargeExpression.yml
alyacb: Probably throughput is more important, but this change will lower the latency per-query from my understanding- why not both?

src/workloads/query/MatchWithLargeExpression.yml
alyacb: This is specifically geared towards testing a case where the slowness is caused by the size of the query rather than the size of the collection I think- that being said, I'd be in favor of doing something like having two runs: one with a small collection size, and one with a larger collection.

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
```
alyacb, line 98: Can we have 1-2 cases with the a clause size below the threshold?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+      Name: RunMatchExpressionOrWith10000Clauses
+      Repeat: 10
+      ActivePhase: 6
+      NumberOfClauses: 10000
+
+- ActorFromTemplate:
+    TemplateName: MatchExpressionWithOrClausesTemplate
+    TemplateParameters:
+      Name: RunMatchExpressionOrWith50000Clauses
+      Repeat: 5
```
alyacb, line 133: This is interesting- are the later queries (5000 clauses) that much slower than the earlier ones that we can only run them 5 times, even if we have the optimization? How long does each one take to run?

src/workloads/query/MatchWithLargeExpression.yml
```diff
@@ -0,0 +1,150 @@
...
+      NumberOfClauses: 50000
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
```
alyacb, line 144: Can you make sure to run this workload with all variants before merging? Its possible (IMO likely) the Bonsai variants won't finish executing this in time, in which case we shouldn't enable this workload for them by default- we can file follow-up tickets to deal with this.

alyacb (commented):
Thanks Henri! Left a few comments

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: We would probably still recommend a larger variant by default.

alyacb: Sorry, bit confused- do you mean a larger evergreen variant here? Not sure how that relates to the question of adding a microbenchmark vs. adding a genny workload. Or did I misunderstand your original comment?

src/workloads/query/MatchWithLargeExpression.yml
henrinikku: I don't think it will be backported, set it to 7.2

henrinikku (commented):
Thanks for the reviews! I think I've now addressed all comments.

src/workloads/query/MatchWithLargeExpression.yml
alyacb: More like I would have leaned towards running them all the same number of times... but then its good to ahve as many datapoints as we can (for stability) in the faster cases. I'd say either 5 or 10 is fine, you can just keep it as is.

It seems like we have a non-linear performance degradation for larger clause sizes: is this correct?

henrinikku: > It seems like we have a non-linear performance degradation for larger clause sizes: is this correct?

I'd say it's somewhat linear with the optimization, and way worse without:

Median time per op (in ms):
```
With optimization
100 - 5.173
1k  - 58.11
10k - 919.9
50k - 13500

Without optimization
100 - 5.809
1k  - 128.4
10k - 8122
50k - 232900
```

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: that should probably be 'We would probably still recommend *the* larger variant by default.' So standalone is fine.

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: > I was thinking of committing this first but can do it either way. Let me know if you have a preference.

Committing first is great! If you take this approach then you should mention on the perf-build-baroning channel that an improvement is expected when the SERVER ticket is committed.
  
> Both operation throughput and latency should be and are improved greatly by SERVER-79092. If one of them isn't, something might be wrong. So I'd say they are both important.

Cool, can you mention in the description that these are the important metrics for this workload?

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: Sounds good! Is it worth mentioning that in the description for this workload?

src/workloads/query/FilterWithComplexLogicalExpression.yml
```diff
@@ -25,6 +25,8 @@ Keywords:
 - QuiesceActor
 - insert
 - find
+- matcher
```
alyacb, line 28: I think these were added to the wrong file?

alyacb (approved):
LGTM on my side, mod open comments/ making sure the Bonsai variants don't crash.

alyacb (approved):
LGTM on my side, mod open comments/ making sure the Bonsai variants don't crash.

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: :+1:

src/workloads/query/MatchWithLargeExpression.yml
alyacb: Got it, thank you for clarifying!

src/workloads/query/FilterWithComplexLogicalExpression.yml
jimoleary: yeah wrong file unless you meant to add to both.

src/workloads/query/MatchWithLargeExpression.yml
jimoleary: created https://jira.mongodb.org/browse/EVG-20779

jimoleary (commented):
There are only a few minor documentation nits left at this stage and the ClearCollection is still in the file right now.

src/workloads/query/FilterWithComplexLogicalExpression.yml
henrinikku: Oops, fixed now.

src/workloads/query/MatchWithLargeExpression.yml
henrinikku: Thank you!

henrinikku (commented):
> There are only a few minor documentation nits left at this stage and the ClearCollection is still in the file right now.

Sorry forgot to remove it. Fixed now.

jimoleary (approved):
LGTM, thanks!

henrinikku:
> LGTM on my side, mod open comments/ making sure the Bonsai variants don't crash.

The [patch](https://spruce.mongodb.com/version/64f59bbb562343cfc9781d0a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) with bonsai variants seems to have succeeded, so I think it's fine to include them as well.

henrinikku:
evergreen merge

henrinikku:
evergreen merge

PR #1010, opened 2023-09-06
SERVER-80783 [streams] introduce new streaming workloads

Aadeshp:

**Jira Ticket:** SERVER-80783

**Whats Changed:**  
Streaming is a new service built on top of mongod, so we'd like to add some new genny workloads to cover this new feature.

**Patch testing results:**  
[Results](https://spruce.mongodb.com/version/64ffd7d461837d528be3dde4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Workload Submission form:**  
Submitted!

**Related PRs:**   
[DSI](https://github.com/10gen/dsi/pull/1402)
[etc/system_perf.yml](https://github.com/10gen/mongo/pull/15502)


src/workloads/streams/CurrencyConversion.yml
```diff
@@ -0,0 +1,73 @@
...
+Owner: "@10gen/altas-streams"
+Description: |
+  Streaming pipeline that applies currency conversion from USD to EU.
+
+Keywords:
+- streams
+
+GlobalDefaults:
+  DatabaseName: &DatabaseName "test"
+  StreamProcessorName: &StreamProcessorName "sp"
```
SimonEismann, line 11: Nit: Most workloads do not use quotation marks for strings in yaml.

src/workloads/streams/CurrencyConversion.yml
```diff
@@ -0,0 +1,73 @@
...
+
+Actors:
+- Name: Setup
+  Type: RunCommand
+  Threads: 1
+  Phases:
+  - Phase: 0
```
SimonEismann, line 18: Nit: Common practice is not to provide the phase numbers and instead infer them from the order of the list.

src/workloads/streams/CurrencyConversion.yml
```diff
@@ -0,0 +1,73 @@
...
+        streams_startStreamProcessor: ""
+        name: *StreamProcessorName
+        pipeline: [
+          { $source: { connectionName: "__testMemory" } },
+          { $project: {
+              auction: 1,
+              bidder: 1,
+              dateTime: 1,
+              price: { $multiply: ["$price", 0.908] }
+          }},
```
SimonEismann, line 34: Nit: This should be spread across two lines and indented correspondingly.

src/workloads/streams/CurrencyConversion.yml
```diff
@@ -0,0 +1,73 @@
...
+  - Phase: 1
+    Nop: true
+  - Phase: 2
+    Repeat: 1
+    Database: *DatabaseName
+    Operations:
+    - OperationMetricsName: Stop
+      OperationName: RunCommand
+      OperationCommand:
+        streams_stopStreamProcessor:
```
SimonEismann, line 47: Nit: on line 25 you add a "" here, let's keep things consistent.

src/workloads/streams/CurrencyConversion.yml
Aadeshp: ahh yeah i did this on purpose just so its much easier to parse exactly what each actor is doing in each phase, b/c i misaligned the nops earlier, so having the phase number explicitly there was easier for me to parse this at first glance personally

SimonEismann:
Thanks for doing this analysis! The general rule of thumb is that everything with a cov of <10% is acceptable, so most tests are very stable (cov of <2%). 

However, some tests are too noisy to be used with the changepoint detection, for example, yahoo_benchmark_Insert_Batch1000x.Insert_Latency50thPercentile has a cov of 0.525, and the test results range from 2,509,007 to 8,391,514. We should take a more in-depth look at this via, e.g., t2 to figure out what the differences between a "good" and a "bad" run are and if there is anything we can do to make these tests more stable.

Aadeshp:
@SimonEismann any chance we could focus on just throughput numbers for now in terms of catching regressions? Latency isn't super important for us right now, just because latency is a bit trickier and complicated to measure with streaming workloads

e.g. latency probably varies a lot here because we're asynchronously inserting documents into the streaming pipeline which shares a lock with the processing thread. So if the inserter is able to insert many batches before the processing thread acquires the lock, the next time the processing thread acquires the lock the subsequent insert will just be a lot slower b/c the processing thread will hold the lock for a longer period proportional to how many batches its processing in a single run. So that probably explains the range of p50 and p95 latencies we're seeing

SimonEismann:
I think that makes sense. We currently don't have a good way to ignore specific metrics, this is planned in [EVG-18404](https://jira.mongodb.org/browse/EVG-18404). You could create a follow-up ticket to ignore the latency metrics that is blocked on [EVG-18404](https://jira.mongodb.org/browse/EVG-18404) and merge this ticket as is.

SimonEismann (approved):

Aadeshp:
evergreen merge

PR #1011, opened 2023-09-07
EVG-17808 add {^UUID} generator

jimoleary:
**Jira Ticket:** EVG-17808

**Whats Changed:**  
Added {^UUID} generator support.

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -22,6 +22,7 @@
 #include <iostream>
 #include <limits>
 #include <map>
+#include <regex>
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
@@ -262,7 +263,6 @@ const static boost::posix_time::ptime max_date{boost::gregorian::date(2150, 1, 1
 // Documentation is at the implementations-site.
 int64_t parseStringToMillis(const std::string& datetime);
 UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs);
-UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs, int64_t d);
 UniqueGenerator<int64_t> int64GeneratorBasedOnDistribution(const Node& node,
                                                            GeneratorArgs generatorArgs);
 UniqueGenerator<double> doubleGenerator(const Node& node, GeneratorArgs generatorArgs);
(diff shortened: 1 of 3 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,53 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+            BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg.str()));
+        }
+        return static_cast<bsoncxx::binary_sub_type>(sub_type);
+    }
+
+    DefaultRandom& _rng;
+    const Node& _node;
+    const UniqueGenerator<std::string> _hexGen;
+    const UniqueGenerator<int64_t> _subTypeGen;
+    std::string hex;
```
jimoleary, line 985: This bit took me the longest to figure out, without it the hex.data array in the returned ```bsoncxx::types::b_binary``` goes out of scope before the value is copied into the generated doc.

I'm happy to take a better approach here if there is one.

src/value_generators/test/DocumentGeneratorTestCases.yml
```diff
@@ -1439,3 +1439,108 @@ Tests:
...
           - "1.0"
+
+  # https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
+  # Check uuid with the following python:
+  #   import base64
+  #   b = b'00000000-0000-0000-0000-000000000000'
+  #   s = base64.b64encode(b).decode()
+  #   assert base64.b64decode(s) == b
+  #
+  - Name: Invalid subType
```
jimoleary, line 1450: not sure if this is valid or should I raise an exception?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,53 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+
+    bsoncxx::types::b_binary evaluate() override {
+        hex = _hexGen->evaluate();
+        return bsoncxx::types::b_binary{generateSubType(),
+                                        static_cast<uint32_t>(hex.length()),
+                                        reinterpret_cast<const uint8_t*>(hex.data())};
+    }
+
+private:
+    bsoncxx::binary_sub_type generateSubType() {
```
jimoleary, line 969: maybe future proofing, maybe overkill

kashkambath (commented):
Thanks so much! LGTM, I wasn't able to approve

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -2117,6 +2169,28 @@ UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorA
...
+ *   a top-level document value i.e. either a scalar or a `^RandomInt` value
+ * @param generatorArgs
+ *   the generator arguments
+ * @param d
+ *   the default value to generate if node is null
+ * @return
+ *   either a `^RantomInt` generator (etc--see `intParsers`)
+ *   or a constant generator if given a constant/scalar.
+ */
+UniqueGenerator<int64_t> intGenerator(const Node& node, GeneratorArgs generatorArgs, int64_t d) {
```
dalyd, line 2183: Do we actually need this thing?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,53 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+ *
+ *  see https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
+ */
+class UuidGenerator : public Generator<bsoncxx::types::b_binary> {
+public:
+    UuidGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _rng{generatorArgs.rng},
+          _node{node},
+          _hexGen{stringGenerator(node["hex"], generatorArgs)},
+          _subTypeGen{intGenerator(node["subType"],
```
dalyd, line 957: Could you move the initialization of _subTypeGen into the constructor body? Then you could test for existence of `node["subType"]` and pass in the default value if the node doesn't exist. Then you don't need to define the new version of intGenerator.  It feels like a lot of machinery to add to allow a default value, when only this generator is using a default value.

dalyd: Presumably you could stay with the inline definition by using a ternary expression here, but those always feel like magic to me.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,53 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+                                   generatorArgs,
+                                   static_cast<int64_t>(bsoncxx::binary_sub_type::k_uuid))} {}
+
+    bsoncxx::types::b_binary evaluate() override {
+        hex = _hexGen->evaluate();
+        return bsoncxx::types::b_binary{generateSubType(),
+                                        static_cast<uint32_t>(hex.length()),
```
dalyd, line 964: @markbenvenuto  is concerned about the static cast if evaluate is returning a string rather than raw bytes. You may need a call to hex2bytes (which may or may not exist). The std::string can be cast to binary, but may not be the bits you want. 

We looked at your examples and put this together: https://gchq.github.io/CyberChef/#recipe=From_Base64('A-Za-z0-9%2B/%3D',true,false)&input=TTJJeU5ERXhNREV0WlRKaVlpMDBNalUxTFRoallXWXROREV6Tm1NMU5qWmhPVFl5

The output should not be readable with dashes after the conversion.

dalyd (changes requested):
I pulled Mark B into this as he was nearby while I was reading it. I've tagged him on a comment. He thinks there's a correctness issue.

src/value_generators/src/DocumentGenerator.cpp
dalyd: I think we are looking at the wrong line, but it's still wrong. It's the reinterpret_cast on the following line, but I think the issue is the same. Additionally, you have not removed the dashes.

thessem (approved):
It looks like my concerns about UUID spec correctness and API composability have been implemented. Thanks for that. I'll leave the correctness and style review to @dalyd who has bitten in already.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,139 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+ *   the array to store the uuid.
+ * @return the output array
+ */
+uint8_t* hex2BinUuid(std::string hex, uint8_t* uuid) {
+    hex.erase(std::remove_if(hex.begin(), hex.end(),
+                             [&](char c) { return c == '-'; } ),
+              hex.end());
+
+    for(std::string::iterator it = hex.begin(); it != hex.end(); ++it) {
+        int pos = std::distance(hex.begin(), it) / 2;
```
dalyd, line 1027: Why not just add a counter to the for loop for the index in addition ot the iterator. Then you can increment it once per for loop iteration. It seems like that has to be better than calling std::distances on the string each time.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,139 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+ *  see https://en.wikipedia.org/wiki/Universally_unique_identifier
+ */
+class UuidGenerator : public Generator<bsoncxx::types::b_binary> {
+public:
+    UuidGenerator(const Node& node, GeneratorArgs generatorArgs)
+        : _rng{generatorArgs.rng},
+          _node{node},
+          _hexGen{stringGenerator(node["hex"], generatorArgs)},
+          // If there is no subType then use bsoncxx::binary_sub_type::k_uuid
+          _subTypeGen{node["subType"]
```
dalyd, line 1050: Nice. Thank you

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -937,6 +938,139 @@ class ObjectIdGenerator : public Generator<bsoncxx::types::b_oid> {
...
+                          ? intGenerator(node["subType"], generatorArgs)
+                          : std::make_unique<ConstantAppender<int64_t>>(
+                                static_cast<int64_t>(bsoncxx::binary_sub_type::k_uuid))} {}
+
+    bsoncxx::types::b_binary evaluate() override {
+        auto hex = _hexGen->evaluate();
```
dalyd, line 1056: I'm not sure we want to validate each and every time. I see why you want to do so. What is the overhead from the call?

src/workloads/docs/Generators.yml
```diff
@@ -332,6 +335,28 @@ Actors:
             # Maps are identified by their "id" are shared.
             freqmap_singleton: { ^TakeRandomStringFromFrequencyMapSingleton: { id: "some_string", from: { "string3":100, "string4":200 } } }
 
+            # UUID
+            uuids: [
+              # Short uuid
+              {^UUID: {hex: {^RandomString: {length: 8, alphabet: "0123456789ABCDEF"}}}},
```
dalyd, line 341: All of your tests above are for full length. I was confused when I got here that this was even possible. Could you add a short uuid passing test a well if this is a thing? 

I went to the wikipedia page and got even more confused, so I'm going to rely on your answer on whether short UUIDs are a thing.

dalyd (approved):
LGTM

jimoleary:
Note: I ran valgrind on the code and there are some 'still reachable' warnings from within external libraries but there are no leaks (to double check I added an explicit 'new' to the hex2Bin and that highlighted as a real leak)

jimoleary:
evergreen merge

PR #1014, opened 2023-09-14
PERF-4698: Set batchsize to collection size in locf.yml

projjal:
Currently limit is set to 100 to prevent cursor leak but this doesn't give a good account of performance


lriuui0x0 (approved):

lriuui0x0:
Could we check whether $linearFill test also has the same problem?

projjal:
> Could we check whether $linearFill test also has the same problem?

I checked earlier that $linearFill uses `cursor: {batchSize: *batchSize}` but now i noticed that the `batchSize` is actually set to 100 while document count is 600.

PR #1016, opened 2023-09-18
PERF-4725 Pause running time-series arbitrary updates on non-all-feat…

YuhongZhang98:
…ure-flag variants

patch:
https://spruce.mongodb.com/version/6508b3182a60edf99822acd5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/TimeSeriesArbitraryUpdate.yml
+++ b/src/workloads/execution/TimeSeriesArbitraryUpdate.yml
@@ -55,9 +55,10 @@ AutoRun:
 - When:
     mongodb_setup:
       $eq:
-      - shard-lite
-      - replica
+      # TODO (PERF-4726): Enable on these variants.
+      # - shard-lite
+      # - replica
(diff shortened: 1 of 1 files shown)


src/workloads/execution/TimeSeriesArbitraryUpdate.yml
```diff
@@ -55,9 +55,10 @@ AutoRun:
 - When:
     mongodb_setup:
       $eq:
-      - shard-lite
-      - replica
+      # TODO (PERF-4726): Enable on these variants.
+      # - shard-lite
+      # - replica
```
gregorynoma, line 60: Can we leave these uncommented since we bumped the version to 7.2?

YuhongZhang98: Discussed offline and will turn off the feature flag on 7.2 as well. So this will stay as is.

PR #1018, opened 2023-09-20
EVG-20698 QuiesceActor reports error if srv mongodb uri is in use.

lgfang:
**Jira Ticket:** EVG-20698

**Whats Changed:**  

QuiesceActor uses Topology, which needs to access cluster members directly. Currently, the URIs for the members are constructed using "protocol://member_fqdn:port", where protocol is either "mongodb" or "mongodb+srv". When "mongodb+srv" is used (for Atlas related tests), the resulted URIs are invalid because mongodb+srv shall not include port numbers.

This change is to report the error during QuiesceAcor initialization (in the constructor) so that:
- The users understand what the issue is.
- Tests with wrong mongo URIs (i.e. the srv ones)  fail at the beginning *not* when QuiesceActor actually start to work.

**Patch testing results:**  
None

### DIFF
--- a/src/cast_core/src/QuiesceActor.cpp
+++ b/src/cast_core/src/QuiesceActor.cpp
@@ -65,7 +65,16 @@ QuiesceActor::QuiesceActor(genny::ActorContext& context)
     : Actor{context},
       _totalQuiesces{context.operation("Quiesce", QuiesceActor::id())},
       _client{context.client()},
-      _loop{context} {}
+      _loop{context} {
+
+    std::string srv = "mongodb+srv://";
+    auto uri_prefix = _client->uri().to_string().substr(0, srv.length());
+
+    if (uri_prefix == srv) {
+        throw InvalidConfigurationException(
+            "QuiesceActor does not support 'mongodb+srv://' connection URI.");
+    }
+}
 
 namespace {
 auto registerQuiesceActor = Cast::registerDefault<QuiesceActor>();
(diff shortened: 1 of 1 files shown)


src/cast_core/src/QuiesceActor.cpp
```diff
@@ -65,7 +65,16 @@ QuiesceActor::QuiesceActor(genny::ActorContext& context)
     : Actor{context},
       _totalQuiesces{context.operation("Quiesce", QuiesceActor::id())},
       _client{context.client()},
-      _loop{context} {}
+      _loop{context} {
+
+    std::string srv = "mongodb+srv://";
+    auto uri_prefix = _client->uri().to_string().substr(0, srv.length());
+
+    if (uri_prefix == srv) {
```
rtimmons, line 73: (Nit: `urlPrefix` since genny uses camelCase)
Also nit: I think you can use `str::find` to look for the srv thing. Might be cleaner.

```
if (_client->uri().to_string().find("...srv") == 1) // found at beginning
```

(At least I think that's right.)

Just a nit. Otherwise looks fine.

rtimmons (approved):
Idea for improvement. Otherwise only blocker is `url_prefix -> urlPrefix`.

src/cast_core/src/QuiesceActor.cpp
lgfang: Will change to camelCase.

I chose (`sub_str` + `==`) instead of `find` because I had a theory that the later might be faster :) . But here the performance difference between these two approach are negligible as the strings are short and is not a concern at all because this is done during initialisation. 

Therefore, after a second thought, I agree with your that using `find` to make it more intuitive is better.

rtimmons (approved):

lgfang:
evergreen merge

PR #1020, opened 2023-09-20
SERVER-81260: custom throughput reporter for streams

Aadeshp:

**Jira Ticket:** [SERVER-81260](https://jira.mongodb.org/browse/SERVER-81260)

**Whats Changed:**  
Introducing a new actor that periodically polls metrics from a mongo streams instance and records the throughput metrics returned in the genny operation metrics. The actor doesn't exit until the conditional expected document count is met. We need this because stream processing is asynchronous, so the best way to measure throughput here is pulling output throughput metrics from the stream process itself.

**Patch testing results:**  

**Related PRs:**   

### DIFF
--- a/src/cast_core/include/cast_core/actors/StreamStatsReporter.hpp
+++ b/src/cast_core/include/cast_core/actors/StreamStatsReporter.hpp
@@ -1,0 +1,82 @@
+// Copyright 2023-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef HEADER_D5EB1FAB_E817_43DE_A126_FCC6995B4C3C_INCLUDED
+#define HEADER_D5EB1FAB_E817_43DE_A126_FCC6995B4C3C_INCLUDED
+
+#include <string_view>
+
+#include <mongocxx/pool.hpp>
+
+#include <gennylib/Actor.hpp>
+#include <gennylib/Orchestrator.hpp>
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+
+#include <metrics/metrics.hpp>
+
+namespace genny::actor {
+
+/**
+ * Periodically polls (every 250ms) the stream processor's stats from the
+ * mongostream instance (`streams_getStats`) until the expected document count
+ * is met. This is needed because stream processing is fully async so we need to
+ * rely on metrics emitted from the stream processor itself in order to get proper
+ * numbers on throughput.
+ *
+ * Throughput for a stream processor is measured based on the input rate of
+ * the stream processor, so in our case it'll be based on the `inputMessageCount`
+ * and `inputMessageBytes` stats in the `streams_getStats` response.
+ *
+ * ```yaml
+ * SchemaVersion: 2017-07-01
+ * Actors:
+ * - Name: StreamStatsReporter
+ *   Type: StreamStatsReporter
+ *   Database: test
+ *   Phases:
+ *   - Repeat: 1
+ *     StreamProcessorName: sp
+ *     ExpectedDocumentCount: 1000000
+ * ```
+ *
+ * Owner: @atlas-streams
+ */
+class StreamStatsReporter : public Actor {
+public:
+    explicit StreamStatsReporter(ActorContext& context);
+    ~StreamStatsReporter() = default;
+
+    void run() override;
+
+    static std::string_view defaultName() {
+        return "StreamStatsReporter";
+    }
+
+private:
+    mongocxx::pool::entry _client;
+
+    // Recorded based on the response of `streams_getStats` from the mongostream
+    // instance.
+    genny::metrics::Operation _throughput;
+
+    /** @private */
+    struct PhaseConfig;
+    PhaseLoop<PhaseConfig> _loop;
+    genny::Orchestrator& _orchestrator;
+};
+
+}  // namespace genny::actor
+
+#endif  // HEADER_D5EB1FAB_E817_43DE_A126_FCC6995B4C3C_INCLUDED
--- a/src/cast_core/src/StreamStatsReporter.cpp
+++ b/src/cast_core/src/StreamStatsReporter.cpp
@@ -1,0 +1,116 @@
+// Copyright 2023-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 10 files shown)


src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    PhaseConfig(PhaseContext& phaseContext, mongocxx::database&& db, ActorId id)
+        : database{db},
+          pollInterval{phaseContext["PollInterval"].to<TimeSpec>()} {
+        int threads = phaseContext.actor()["Threads"].to<int>();
+        assert(threads == 1);
+
+        std::string streamProcessorName = phaseContext["StreamProcessorName"].to<std::string>();
+        bsoncxx::builder::stream::document builder;
+        metricsCommand = builder << "streams_getMetrics" << "" << bsoncxx::builder::stream::finalize;
```
Aadeshp, line 54: so using `getMetrics` rather than `getStats` only because `getMetrics` is more "real time" and `getStats` is only updated periodically (which seems to be very infrequent since the in-memory operator processes everything thats queued up in one run iteration, which could be many many thousand documents)

src/cast_core/include/cast_core/actors/StreamStatsReporter.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * the stream processor.
+ *
+ * ```yaml
+ * SchemaVersion: 2017-07-01
+ * Actors:
+ * - Name: StreamStatsReporter
+ *   Type: StreamStatsReporter
+ *   Database: test
+ *   Phases:
+ *   - PollInterval: 10 seconds
```
Aadeshp, line 47: ik `SleepBefore` and `SleepAfter` exists, so could use that instead and just enforce that its set within the stats reporter phase config

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+};
+
+void StreamStatsReporter::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto throughput = _throughput.start();
+            std::this_thread::sleep_for(genny::Duration(config->pollInterval));
```
rtimmons, line 62: I don't have time for a deep review until tomorrow/Monday, but I think we need to kill this. The Orchestrator class is the keeper of timing and sleeping, and using its mechanisms lets the Actor play nicely with Phase control. I think `SleepBefore` or `SleepAfter` can get what we need, and the constructor can simply assert that a reasonable `Sleep*` value is configured.

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+            bool inputDocuments{false};
+            bool inputBytes{false};
+
+            for (auto c : counters) {
+                auto name = c["name"].get_string().value;
+                int64_t value = c["value"].get_int64();
+
+                if (name == "num_input_documents") {
+                    assert(!inputDocuments);
```
rtimmons, line 76: Instead of `assert` there's `BOOST_THROW_EXCEPTION` that will give us nice stackframes.

src/cast_core/test/StreamStatsReporter_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <gennylib/context.hpp>
+
+namespace genny {
+namespace {
+using namespace genny::testing;
+namespace bson_stream = bsoncxx::builder::stream;
+
+//
+// ⚠️ There is a "known" failure that you should find and fix as a bit of
```
rtimmons, line 34: Plz kill / update

src/cast_core/src/StreamStatsReporter.cpp
rtimmons: You can also just use the rate limiter and have it say `1 per 5 seconds`

Aadeshp: oh sweet!

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <gennylib/context.hpp>
+
+#include <value_generators/DocumentGenerator.hpp>
+
+
+namespace genny::actor {
+
+struct StreamStatsReporter::PhaseConfig {
+    mongocxx::database database;
+    TimeSpec pollInterval;
```
rtimmons, line 43: I think `pollInterval` and related things in the docs/comments can be killed now.

src/cast_core/include/cast_core/actors/StreamStatsReporter.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <gennylib/Actor.hpp>
+#include <gennylib/PhaseLoop.hpp>
+#include <gennylib/context.hpp>
+
+#include <metrics/metrics.hpp>
+
+namespace genny::actor {
+
+/**
+ * Periodically polls (every 250ms) the stream processor's stats from the
```
sandeepgd, line 31: Do this every 1s or 2s instead? Stats polling is not super cheap for stream processor and adding another 2s should not affect the throughput as our throughput test should anyways be running for minutes.

sandeepgd (approved):

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,115 @@
...
+                auto reply = config->database.run_command(config->statsCommand.view());
+                BOOST_LOG_TRIVIAL(debug) << "stats reply: " << bsoncxx::to_json(reply);
+
+                inputMessageCount = reply.view()["inputMessageCount"].get_int64();
+                inputMessageBytes = reply.view()["inputMessageSize"].get_double();
+                if (inputMessageCount >= config->expectedDocumentCount) {
+                    break;
+                }
+
+                std::this_thread::sleep_for(kPollInterval);
```
rtimmons, line 90: If we use the built-in SleepBefore/SleepAfter/GlobalRateLimit then we can get this for free and continue to participate in interrupt handling and the like.

rtimmons (changes requested):

src/cast_core/src/StreamStatsReporter.cpp
Aadeshp: ahh but those built-ins currently require either `Repeat` or `Duration` to be defined alongside right? Is it possible to use one of those out of the box and not make the actor exit until a condition is met (expected document count) rather than after the repeat or duration?

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -0,0 +1,116 @@
...
+};
+
+void StreamStatsReporter::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto throughput = _throughput.start();
+
+            int64_t inputMessageCount{0};
+            double inputMessageBytes{0};
+            while (true) {
```
rtimmons, line 80: I think you want this to be `while(orchestrator->continueRunning())`. This will let your loop end early if there's exceptions reported by other Actors.

rtimmons (changes requested):
Thanks for iterating a bit. Only change is the `while` predicate then LGTM.

src/cast_core/src/StreamStatsReporter.cpp
Aadeshp: ahh good call

rtimmons (approved):

Aadeshp:
evergreen merge

PR #1023, opened 2023-09-25
EVG-20902 Add explicit casting generators to Genny

marksg07:
**Jira Ticket:** [EVG-20902](https://jira.mongodb.org/browse/EVG-20902)

**Whats Changed:**  
Adds ^ConvertToDecimal, ^ConvertToDouble, ^ConvertToInt, ^ConvertToInt32, value generators which convert their string, int, long, or double input into the given type.

**Patch testing results:**  
N/A

**Workload Submission form:**  
N/A

**Related PRs:**   
N/A


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1901,6 +2011,10 @@ const auto [allParsers, arrayParsers, dateParsers, doubleParsers, intParsers, st
          [](const Node& node, GeneratorArgs generatorArgs) {
              return std::make_unique<IncGenerator>(node, generatorArgs);
          }},
+        {"^ConvertToInt",
```
dalyd, line 2014: Shoudl this be `ConvertToInt64t`?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1797,7 +1902,8 @@ Out valueGenerator(const Node& node,
  * If adding a new parsers map, remember to insert its entries into `allParsers` at the end of this
  * lambda function.
  */
-const auto [allParsers, arrayParsers, dateParsers, doubleParsers, intParsers, stringParsers] = [] () {
+const auto [allParsers, arrayParsers, dateParsers, doubleParsers, intParsers,
```
dalyd, line 1905: Should we explicitly rename intParsers to int64Parsers now to go with the int32Parsers? Or at least add a comment explaining?

dalyd (approved):
I don't think I'm competent to review all the C++ in here, but the added generators, test cases, and examples all look good to me. 

@rtimmons @jimoleary would either of you like to take a look. 

And thanks for putting up this PR! We love to see users adding things like this.

jimoleary (commented):
Rather than indvidual casting generators, would it make more sense to support a generic {^Bson:} generator and implement some / all of the [extended json conversion](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table)?


The syntax would be more familiar and we could add support for more types without the need for more casts or generator types as we need them.


```
class BsonNumberLongGenerator : public Generator<bsoncxx::types::bson_value::value> {
public:
    BsonNumberLongGenerator(const Node& node, GeneratorArgs generatorArgs)
        : _numberLongGen{intGenerator(node["$numberLong"], generatorArgs)} {}

    bsoncxx::types::bson_value::value evaluate() override {
        return bsoncxx::types::bson_value::value{bsoncxx::types::b_int64{_numberLongGen->evaluate()}};
    }

private:
    const UniqueGenerator<int64_t> _numberLongGen;
};

class BsonNumberDoubleGenerator : public Generator<bsoncxx::types::bson_value::value> {
public:
    BsonNumberDoubleGenerator(const Node& node, GeneratorArgs generatorArgs)
        : _numberDoubleGen{doubleGenerator(node["$numberDouble"], generatorArgs)} {}

    bsoncxx::types::bson_value::value evaluate() override {
        return bsoncxx::types::bson_value::value{bsoncxx::types::b_double{_numberDoubleGen->evaluate()}};
    }

private:
    const UniqueGenerator<double> _numberDoubleGen;
};

class BsonNumberDecimalGenerator : public Generator<bsoncxx::types::bson_value::value> {
public:
    BsonNumberDecimalGenerator(const Node& node, GeneratorArgs generatorArgs)
        : _numberDecimalGen{stringGenerator(node["$numberDecimal"], generatorArgs)} {}

    bsoncxx::types::bson_value::value evaluate() override {
        return bsoncxx::types::bson_value::value{bsoncxx::types::b_decimal128{_numberDecimalGen->evaluate()}};
    }

private:
    const UniqueGenerator<std::string> _numberDecimalGen;
};

/**
 * @param node
 *   Any node type that evaluates to a bson value
 * @return
 *   The proper generator based on the parsed input.
 */
UniqueGenerator<bsoncxx::types::bson_value::value> bsonGenerator(const Node& node, GeneratorArgs generatorArgs) {
    if (node || node.isMap()) {
        if (node["$numberInt"]) {
            return std::make_unique<BsonNumberIntGenerator>(node, generatorArgs);
        } else if (node["$numberLong"]) {
            return std::make_unique<BsonNumberLongGenerator>(node, generatorArgs);
        } else if (node["$numberDouble"]) {
            return std::make_unique<BsonNumberDoubleGenerator>(node, generatorArgs);
        } else if (node["$numberDecimal"]) {
            return std::make_unique<BsonNumberDecimalGenerator>(node, generatorArgs);
        }
    }
    std::stringstream msg;
    msg << "Node not currently supported " << node;
    BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(msg.str()));
}

/**
 * `{^Bson: ... }`
 *  Generator for bson values, see the conversion-table for syntax.
 *
 *  see https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table
 */
class BsonGenerator : public Generator<bsoncxx::types::bson_value::value> {
public:
    BsonGenerator(const Node& node, GeneratorArgs generatorArgs)
        : _bsonGen{bsonGenerator(node, generatorArgs)} {}

    bsoncxx::types::bson_value::value evaluate() override {
        return _bsonGen->evaluate();
    }

private:
    const UniqueGenerator<bsoncxx::types::bson_value::value> _bsonGen;
};
```


Example:
```
  - Name: numberDecinal
    GivenTemplate:
      a: {^Bson: {"$numberDecimal": 42}}
    ThenReturns:
      - {a: { "$numberDecimal" : "42" }}

```

PyGeek03:
Would this PR disable implicit conversion, for example [EVG-17777](https://jira.mongodb.org/browse/EVG-17777)? Would be interesting if we can hit two birds with one stone.

marksg07:
@jimoleary 

> Rather than indvidual casting generators, would it make more sense to support a generic {^Bson:} generator and implement some / all of the [extended json conversion](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table)?
> 
> The syntax would be more familiar and we could add support for more types without the need for more casts or generator types as we need them.

The only issue I see with this solution as it stands is that it would only handle conversions from one specific type to number. This makes some sense if we make that specific string for all the converters, since the extended BSON spec expects the value to be a string representation for all of the `{"$numberXXX": <value>}` formats. If we did this, we would need to convert numeric types to string before passing them into these converters. If we wanted this solution to handle more conversions from number to number, we would still need to have something like the try-catch generator conversion logic that I have in this PR, since we'd need to handle various types of generators. 

I do think that it makes sense to do these conversions into BSON values rather than different YAML types, as these conversion generators will almost always be the last step of value generation (i.e. we probably aren't going to be doing more transformations on the converted values). I can take a crack at this, or you can if you think it makes more sense? Thank you for your feedback.

src/value_generators/src/DocumentGenerator.cpp
marksg07: I think if I did this, it would be correct to also rename `^RandomInt` to `^RandomInt64`, but I'm reluctant to do that as it would be a breaking change. We could add an alias `^RandomInt64` and deprecate `^RandomInt` (throwing some type of warning at load time), but that feels hefty. Changing all the things named `intXXX` internally to `int64XXX` (i.e. `intParsers` to `int64Parsers` as you mentioned below) sounds good to me though.

marksg07:
@PyGeek03 Currently it doesn't. I could try disabling this implicit conversion and running a patch to make sure we don't break anything though. 

Also, this seems like it would be cleanly solved by @jimoleary 's solution -- If we added a `BsonStringGenerator`, we could do
```
...
$gte: {^Bson: "000000"}
...
```
in the failing example given in that ticket, which would give us a `bson_value` holding the correct string which would never get converted.

dalyd (approved):
Still LGTM with caveat that someone with stronger c++ than I should also look. 

At this point I'd argue for the shortest path to merging. I've found myself wanting to point people to this stuff in the past few weeks :) We can always add more.

thessem:
Is @johndaniels (on leave this week) or @jimoleary comfortable saying they're confident reviewing the C++ aspects of this? If not, I can have a go since I'm fairly familiar with what the code is doing, just not doing it in C++ :)

(cc @dalyd)

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1113,6 +1116,108 @@ class FastRandomStringGenerator : public StringGenerator {
...
 
+// The following is a replacement for partial template specialization for functions, which is not
+// currently supported in C++.
+template <typename F, typename T>
+struct convert{};
+
+// Conversion used between int, long, double
+template <typename F, typename T>
+T convertToNumeric(convert<F, T>, const F& from)
+{
```
rtimmons, line 1127: Nit: formatting - please run `clang-format` (the versions aren't standard so if you get a bunch of diffs, just do your best to match the existing style; the brace on the new line threw this off for me.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1113,6 +1116,108 @@ class FastRandomStringGenerator : public StringGenerator {
...
+    return stol(s);
+}
+int32_t convertToNumeric(convert<std::string, int32_t>, const std::string& s) {
+    return stoi(s);
+}
+
+// This is the only overload we run directly, and it will route to the correct template.
+template <typename F, typename T>
+T convertToNumeric(const F& from) {
+    return convertToNumeric(convert<F, T>{}, from);
```
rtimmons, line 1157: This is amazing and terrible.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1113,6 +1116,108 @@ class FastRandomStringGenerator : public StringGenerator {
...
+}
+
+// overloaded struct for use with std::variants.
+template<class... Ts>
+struct overloaded : Ts... { using Ts::operator()...; };
+
+template<class... Ts>
+overloaded(Ts...) -> overloaded<Ts...>;
```
rtimmons, line 1165: I think it's worth pointing to the cppreference page that details the mechanics of this: https://en.cppreference.com/w/cpp/utility/variant/visit

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1113,6 +1116,108 @@ class FastRandomStringGenerator : public StringGenerator {
...
+            return;
+        } catch (const InvalidConversionException& e) {}
+          catch (const UnknownParserException& e) {}
+        try {
+            _generator = doubleGenerator(node["from"], generatorArgs);
+            return;
+        } catch (const InvalidConversionException& e) {}
+          catch (const UnknownParserException& e) {}
+        try {
+            _generator = int32Generator(node["from"], generatorArgs);
```
rtimmons, line 1189: I don't think this catches ambiguous values that could parse as multiple options. Do we want to try them all and then barf if more than one works?

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1113,6 +1116,108 @@ class FastRandomStringGenerator : public StringGenerator {
...
+            return;
+        } catch (const InvalidConversionException& e) {}
+          catch (const UnknownParserException& e) {}
+        BOOST_THROW_EXCEPTION(InvalidValueGeneratorSyntax(
+            "Numeric conversion generator only supports conversions from the string, double, int, "
+            "and int32 types. The type of the \"from\" field did not match any of those types."));
+    }
+    T evaluate() override {
+        return std::visit(overloaded{
+            [&](UniqueGenerator<std::string>& gen) {
```
rtimmons, line 1199: I'm confused why you need the reference capture in the lambdas. Does it need `[&]` for something or would plain 'ol `[]` work?

Not sure if it matters that much tbh, but fewer places for values to potentially escape the stack seems better.

rtimmons (approved):
Code looks good! Nice use of `std::visit`. 

FWIW the original value-generators leaned heavily into `std::visit`. We removed it because it became difficult to use with `^Verbatim` and with recursive parsers, but in the "lower-level" (primitives where we can't nest parsers) case this makes a lot of sense.

None of my comments are worth holding up merging, but please consider them.

src/value_generators/src/DocumentGenerator.cpp
marksg07: Indeed. I added a brief explanation in the comment, let me know if you think that's sufficient.

src/value_generators/src/DocumentGenerator.cpp
marksg07: Done.

src/value_generators/src/DocumentGenerator.cpp
marksg07: Done; however I added a special case for constant values, as those can parse successfully as multiple types (for example, `{^ConvertToInt: {from: "1"}}` can parse the "from" node successfully with any generator, because of how `Node::to<T>` operates), and can always be parsed as strings.

src/value_generators/src/DocumentGenerator.cpp
marksg07: You're right, changed to `[]`

rtimmons (approved):
Nicely done!

thessem (approved):
Happy with the reviews this has already seen, I'm getting out of the way of merging now

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1293,6 +1297,154 @@ class FastRandomStringGenerator : public StringGenerator {
...
+        } catch (const InvalidConversionException& e) {
+        } catch (const UnknownParserException& e) {
+        }
+
+        // Otherwise, "from" should be a generator -- try all of the possible generators for from,
+        // and ensure that exactly one works.
+        bool succeeded = false;
+        try {
+            _generator = stringGenerator(node["from"], generatorArgs);
+            succeeded = true;
```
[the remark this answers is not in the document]
marksg07: We are currently failing if multiple generators succeed -- Do you mean that we should keep track of exactly which succeed and throw a more detailed error with that info?

rtimmons (approved):
Knowing which conversion(s) worked in the case of multiple seems nice-to-have but also something that could be added later without much headache. LGTM.

marksg07:
evergreen merge

PR #1025, opened 2023-09-26
SERVER-81260: support ^ClientURI in preprocessing

Aadeshp:

**Jira Ticket:** [SERVER-81260](https://jira.mongodb.org/browse/SERVER-81260?filter=-1)

**Whats Changed:**  
Support `^ClientURI` replacement which will replace any instances of `^ClientURI` with the corresponding registered client's URI. The plan is to use this for streams in DSI and have a streams instance create a source operator that consumes a change stream from a mongo VM within the same DSI VPC. But in order for that to work, I need the client URI to be exposed.

Unfortunately the `URI` for each client will be populated during preprocess time in genny `preprocess.py` so the `URI` is not known pre-compile time so we can't anchor/reference the `URI`.

e.g.
```
Clients:
  Default:
    QueryOptions:
      maxPoolSize: 10
  Stream:
    QueryOptions:
      maxPoolSize: 10
   
Actors:
- Name: Setup
  Type: RunCommand
  ClientName: Stream
  Threads: 1
  Phases:
  - Phase: 0
     ...
    Operations:
    - OperationMetricsName: CreateStreamProcessor
      OperationName: RunCommand
      OperationCommand:
        streams_startStreamProcessor: ""
        name: *StreamProcessorName
        pipeline: [
          { $source: { connectionName: "db", db: "test", coll: "input" } },
          ...
        ]
        connections: [{ name: "db", type: "atlas", options: { uri: {^ClientURI: { Name: "Default" } } } }]


- Name: Loader
  Type: Loader
  ClientName: Default
  Threads: 1
  Phases:
  - Phase: 0
     ...
```

**Patch testing results:**  

**Related PRs:**   

### DIFF
--- a/src/lamplib/src/genny/tasks/preprocess.py
+++ b/src/lamplib/src/genny/tasks/preprocess.py
@@ -11,7 +11,7 @@ from omegaconf import OmegaConf
 import yaml
 import structlog
 import numexpr
-from typing import Union
+from typing import Any, Dict, Optional, Union
 
 SLOG = structlog.get_logger(__name__)
 # Cannot be in the default config because yaml merges overwrite lists instead of appending.
@@ -98,6 +98,7 @@ class _ContextType(Enum):
 
     Parameter = (1,)
     ActorTemplate = (2,)
+    Client = (3,)
 
 
 class _ParseMode(Enum):
@@ -192,7 +193,7 @@ class _WorkloadParser(object):
         default_uri: str,
         source: YamlSource = YamlSource.File,
         path: Union[Path, str] = "",
-        parse_mode: _ParseMode =_ParseMode.Normal,
+        parse_mode: _ParseMode = _ParseMode.Normal,
     ):
         """Parse the yaml input, assumed to be a file by default."""
 
@@ -243,6 +244,8 @@ class _WorkloadParser(object):
             out = self._replace_flattenonce(value)
         elif key == "^PreprocessorFormatString":
             out = self._replace_formatstr(value)
+        elif key == "^ClientURI":
+            out = self._replace_clienturi(value)
         elif key == "ActorTemplates":
             self._parse_templates(value)
         elif key == "ActorFromTemplate":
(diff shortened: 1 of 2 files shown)


johndaniels:
Wanted to check in on approach here. Is there going to be an actor doing the streams work using the client URI within Genny? If so, then the actor should be able to make a client using context.client directly, which may be easier. I can point to examples.

Aadeshp:
> Wanted to check in on approach here. Is there going to be an actor doing the streams work using the client URI within Genny? If so, then the actor should be able to make a client using context.client directly, which may be easier. I can point to examples.

@johndaniels there'll be an actor that talks to the `Stream` client and another actor that talks to the `Default` client, but the former actor talking to the `Stream` client is just telling a mongostream instance to asynchronously start a streaming pipeline and to consume data from a MongoDB instance located in another VM, which would be `Default` client URI. So the first actor (client `Stream`) is just creating a config and passing it to mongostream via `RunCommand`, and part of that config i need the other client (`Default`)'s URI. Is that what `context.client` might be able to help with?

johndaniels:
So we talked in Slack and it sounds like the advantage of doing it this way is that it is available to be referenced elsewhere in the YAML by actor configuration, as opposed to needing a new C++ actor to do so.

johndaniels (changes requested):

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -381,6 +388,29 @@ def _replace_formatstr(self, input):
             )
             raise ParseException(msg)
 
+    def _replace_clienturi(self, input, out):
```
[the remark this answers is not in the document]
Aadeshp: 🤦🏽 yeah hah, woops

src/lamplib/src/genny/tasks/preprocess.py
```diff
@@ -381,6 +388,29 @@ def _replace_formatstr(self, input):
             )
             raise ParseException(msg)
 
+    def _replace_clienturi(self, input):
+        OP_KEY = "^ClientURI"
+
+        name: Optional[str] = input.get("Name", None)
```
thessem, line 394: `Optional` does not appear to have been imported.

thessem (changes requested):
Very minor nit with the type annotations, you fixing that will clear this review, but I would've given a tick otherwise.

Thanks for writing the tests and giving a usage example in the PR.

src/lamplib/src/genny/tasks/preprocess.py
Aadeshp: ahh rip haha, good catch, bad merge conflict 😭

thessem (approved):

johndaniels (approved):

Aadeshp:
evergreen merge

PR #1028, opened 2023-09-27
EVG-17757: Add function to autodetect distro.

johndaniels:
If a distro is not specified, we should be able to autodetect it given the finite number of actual configurations we support.

**Jira Ticket:** EVG-17757

**Whats Changed:**  
Autodetects the value to use for distro if it is not specified

**Patch testing results:**  
[This patch](https://spruce.mongodb.com/version/65147562d1fe074855e94159/tasks) shows all the compile tasks running successfully without the distro set

### DIFF
--- a/src/lamplib/src/genny/tasks/compile.py
+++ b/src/lamplib/src/genny/tasks/compile.py
@@ -2,6 +2,7 @@ from typing import List
 
 import structlog
 import os
+import platform
 import subprocess
 
 from genny.cmd_runner import run_command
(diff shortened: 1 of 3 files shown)


thessem:
What's the reasoning for not going directly to some Python library like `platform.system_alias`? Is getting the info via Genny for single source-of-truth reasons?

thessem:
Does `platform.system_alias` do anything useful for you in this context?

johndaniels:
I don't think system_alias does much. All of the unaliased names seem to make sense and system_alias mostly deals with setting the system name to "Solaris" or "Windows" depending on the underlying system name.

src/lamplib/src/genny/tasks/compile.py
```diff
@@ -32,6 +33,85 @@ def _sanitizer_flags(sanitizer: str, genny_repo_root: str):
...
+            raise DistroDetectionError(f"Invalid machine type for Amazon 2: {machine}")
+    else:
+        raise DistroDetectionError(f"Invalid version for Amazon Linux: {freedesktop_version}")
+
+
+def detect_distro():
+    SLOG.info(f"Distro not specified. Detecting distro.")
+
+    system = platform.system()
+    machine = platform.machine()
```
PyGeek03, line 94: Nit: Maybe can move line 94 to between line 100 & 101, so that the `machine` variable is closer to where it's actually used?

src/lamplib/src/genny/tasks/compile.py
```diff
@@ -32,6 +33,85 @@ def _sanitizer_flags(sanitizer: str, genny_repo_root: str):
...
+    if system == "Darwin":
+        distro = "not-linux"
+    elif system == "Linux":
+        freedesktop_release = platform.freedesktop_os_release()
```
PyGeek03, line 98: A risk involved with using `platform.freedesktop_os_release()` is that [it's only available from Python 3.10](https://docs.python.org/3/library/platform.html#:~:text=New%20in%20version%203.10.), and we haven't decided what's the lowest version of Python that we want Genny to support. Evergreen uses Python 3.9 IIRC, but if the `-d` flag is set for `./run-genny install` then it would be the same as before, so should be okay for automation on Evergreen. However, for manual use on MacOS, Genny users might be used to not setting this flag, so if they're using a Python version before 3.10 then this would be a UX change for them. 

Maybe we can add a Python version check using [`platform.python_version_tuple()`](https://docs.python.org/3/library/platform.html#platform.python_version_tuple), and error out to tell the user to set the `-d` flag if they are using a Python version before 3.10?

johndaniels: This is a good point. It might be a little more complicated to get the version without that, but hopefully not that much more complicated?

johndaniels: I fixed this by adding our own `_freedesktop_os_release` function. Since it's relatively easy to test if all our distros are working, this seems not too crazy bad. I also tried to make clear in the comment the _why_ of it, so hopefully future code readers know they can get rid of it.

src/lamplib/src/genny/tasks/compile.py
johndaniels: Fixed

dalyd:
John, thanks for picking this one up and doing. I don't think I'm needed on this PR though. I think you have the right people on it and it will be good. I'm going to try to pull myself off. We'll see if I can succeed.

PyGeek03 (approved):
LGTM, thank you John!

johndaniels:
evergreen merge

PR #1029, opened 2023-09-28
SERVER-81719: add new kafka loader actor

Aadeshp:

**Jira Ticket:** [SERVER-81719](https://jira.mongodb.org/browse/SERVER-81719?filter=-1)

**Whats Changed:**  
Introducing a new actor that inserts data into a specified kafka cluster. This won't be used in DSI automatically right now since DSI doesn't support kafka at the moment, but this is moreso for local testing.

**Related PRs:**   
vcpkg PR to install librdkafka for the genny toolchain (https://github.com/10gen/vcpkg/pull/29)

### DIFF
--- a/src/cast_core/src/KafkaLoader.cpp
+++ b/src/cast_core/src/KafkaLoader.cpp
@@ -1,0 +1,113 @@
+// Copyright 2023-present MongoDB Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cast_core/actors/KafkaLoader.hpp>
+
+#include <memory>
+
+#include <yaml-cpp/yaml.h>
+
+#include <bsoncxx/json.hpp>
+
+#include <mongocxx/client.hpp>
+#include <mongocxx/collection.hpp>
+#include <mongocxx/database.hpp>
+
+#include <boost/log/trivial.hpp>
+#include <boost/throw_exception.hpp>
+
+#include <gennylib/Cast.hpp>
+#include <gennylib/MongoException.hpp>
+#include <gennylib/context.hpp>
+
+#include <value_generators/DocumentGenerator.hpp>
+
+namespace genny::actor {
+
+namespace {
+
+// Kafka producer flush timeout. Flush is called after each phase.
+static constexpr int kKafkaFlushTimeoutMs = 10'000;
+
+};
+
+struct KafkaLoader::PhaseConfig {
+    DocumentGenerator documentExpr;
+
+    PhaseConfig(PhaseContext& phaseContext, ActorId id) :
+        documentExpr{phaseContext["Document"].to<DocumentGenerator>(phaseContext, id)} {}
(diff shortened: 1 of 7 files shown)


thessem:
I have added myself and @PyGeek03 as reviewers on this. I have added @johndaniels and @lgfang to CC them in on this review, but they're optional (unless they are requesting changes)

PyGeek03:
Hi @Aadeshp, can you update the [TOOLCHAIN_BUILD_ID](https://github.com/mongodb/genny/blob/02a4d42b04a041d415c7184825815b9297ee1efe/src/lamplib/src/genny/toolchain.py#L202) to be `patch_c0f42baafd4845b1f77e0cc19875eba3660d64ba_651c32eca4cf47f8f46c7bd8_23_10_03_15_27_42`, so that it matches the [patch for your 10gen/vcpkg PR](https://spruce.mongodb.com/task/genny_toolchain_macos_1100_arm64_t_compile_patch_c0f42baafd4845b1f77e0cc19875eba3660d64ba_651c32eca4cf47f8f46c7bd8_23_10_03_15_27_42/logs?execution=0)? This would enable the Genny CI tests to run.

src/cast_core/src/KafkaLoader.cpp
```diff
@@ -0,0 +1,113 @@
...
+};
+
+void KafkaLoader::run() {
+    for (auto&& config : _loop) {
+        for (const auto&& _ : config) {
+            auto document = config->documentExpr();
+            std::string json = bsoncxx::to_json(document.view());
+
+            auto inserts = _totalInserts.start();
+            BOOST_LOG_TRIVIAL(debug) << " KafkaLoader Inserting " << json;
```
PyGeek03, line 59: Logging might introduce unpredictable performance variance, so I think we should do any logging outside of the code path being measured as much as possible:

```suggestion
            BOOST_LOG_TRIVIAL(debug) << " KafkaLoader Inserting " << json;
            auto inserts = _totalInserts.start();
```

src/cast_core/test/KafkaLoader_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <testlib/helpers.hpp>
+
+#include <gennylib/context.hpp>
+
+namespace genny {
+namespace {
+using namespace genny::testing;
+namespace bson_stream = bsoncxx::builder::stream;
+
+TEST_CASE_METHOD(KafkaTestFixture, "KafkaLoader successfully connects to a Kafka broker.", "[streams][KafkaLoader]") {
```
PyGeek03, line 33: An Actor test case would only be run if one of the square-bracketed tags match one of the tags defined in a YAML file in [this directory](https://github.com/mongodb/genny/tree/master/src/resmokeconfig). The existing defined tags are `[sharded]`, `[single_node_replset]`, and `[three_node_replset]` (`[SelfTestActor]` is to test Genny code, not Actor code, so it should not be used). The square-bracketed tags are not arbitrary. So we have 2 options here:

1. Add one of the three existing defined tags above.
2. Add a new YAML file under the above directory to define either `[streams]` or `[KafkaLoader]`.

On a related note, I found that [this actor test case](https://github.com/mongodb/genny/blob/77f0c993ec2a0d389da9f140133cf28dcef90a42/src/cast_core/test/StreamStatsReporter_test.cpp#L34) is also not being run because neither `[streams]` nor `[StreamStatsReporter]` are defined tags. But I think this should be fixed in another PR instead of in this PR.

PyGeek03 (changes requested):

thessem:
Hey @Aadeshp, this PR has been sitting in draft for a while, what are the plans for this one?

thessem (commented):
I am significantly more onboarded with Genny than I was 7 months ago, let me know if this is something you want to push forward with by requesting a review. Otherwise I'd appreciate it if this could be closed.

PR #1030, opened 2023-09-29
[WIP] Merge CreateIndex.yml and CreateIndexSharded.yml

alicedoherty:


src/workloads/execution/CreateIndexSharded.yml
```diff
@@ -1,3 +1,4 @@
+# TODO Delete this file
```
SimonEismann, line 1: Let's delete this

src/lamplib/src/genny/tasks/dry_run.py
```diff
@@ -24,6 +24,7 @@ def dry_run_workload(
         "MixedWorkloadsGennyStress.yml",
         "ClusteredCollection.yml",
         "ClusteredCollectionLargeRecordIds.yml",
+        "CreateIndex.yml",
```
SimonEismann, line 27: Let's not do this and see if anything breaks as a way to validate that we actually still need to do this

PR #1031, opened 2023-10-02
PERF-4783 Add a quiesce phase after each $queryStats phase.

wqian94:
**Jira Ticket:** PERF-4783

**Whats Changed:**
Add a quiesce phase after each $queryStats phase.

**Patch testing results:**  
[here](https://spruce.mongodb.com/version/651b102b3066151ab2c219a8)

### DIFF
--- a/src/workloads/query/QueryStats.yml
+++ b/src/workloads/query/QueryStats.yml
@@ -83,11 +83,13 @@ Actors:
 - LoadConfig:
     Path: ../../phases/query/QueryStats.tmpl
     Key: QuiesceActor  # Wait for all threads to quiesce before running prepop or query stats.
+                       # Also quiesce at the end of each set.
     Parameters:
       MaxPhase: *MaxPhase
       Phases: [
         13, 23, 33, 43, 53, 63,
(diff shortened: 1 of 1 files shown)


cswanson310:
Nice! Can you just run a patch build to test this out? Rather than testing it live and maybe needing another ticket?

cswanson310:
Hmm no dice. My next theory is that there are two/multiple mongos processes involved, maybe we aren't hitting each of them. Probably we are not... Is there any precedent for configuring a parameter in a sharded environment?

wqian94:
I reran with a separate quiesce actor on the `admin` database. Let's see if this works before moving on.

wqian94:
Ok, time to investigate your hypothesis about configuring parameters in a sharded environment.

wqian94:
Looks like your hunch has a lot of merit to it. Looking at the logs again, it seems only one of the shards actually processes the admin command, and when this shard changes, the test crashes.

cswanson310:
Silenced this in a separate PR, so I'll close this PR. We have a ticket open to look into the idea mentioned in the comments.

PR #1032, opened 2023-10-03
PERF-4782 Add workloads to stress bucket unpacking and investigate differences in tsbs 

galon1:
**Jira Ticket:** [PERF-4782](https://jira.mongodb.org/browse/PERF-4782)

**Whats Changed:**  
Adding workloads to stress bucket unpacking. Still need to decide if we are going to commit the tsbs similar query workload, so I'm putting this PR as a draft for now. 

**Patch testing results:** 
[Patch link](https://spruce.mongodb.com/version/6527fd0232f417bda9ede681/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) (on standalone)

Running multiple runs right now on repl sets variant (links to multiple runs):
[Patch 1](https://spruce.mongodb.com/version/652d5ad9d6d80aa3a77ef7bd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[Patch 2](https://spruce.mongodb.com/version/652d5ba41e2d17ff724698c2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Workload Submission form:**  
Form submitted!

** Related PRs:**
[DSI PR](https://github.com/10gen/dsi/pull/1422)
[Mongo PR](https://github.com/10gen/mongo/pull/16184)

**Results from Local Runs:**
All results are average latency:
10 buckets: 31.58 seconds
100 buckets: 249.2 milliseconds 
1000 buckets: 1395 milliseconds
There was not a big difference between conjunction and simple predicates.


src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,228 @@
...
+              OperationName: aggregate
+              OperationCommand:
+                # With 600 documents per bucket, 1000 buckets will have 'measurement0' <= 600000 (600 * 1000).
+                Pipeline: [{ $match: { "measurement0": { $gte: 600100 } } }]
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - standalone-all-feature-flags
```
galon1, line 221: This variant is picked so we target the current SBE project. I think when the project is done we should update this to include replica set. Does that sound reasonable? The same applies for the variant in the other workload.

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,435 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
+  This workload tests running identical queries as tsbs_query in a Genny workload. During
```
IrinaYatsenko, line 4: I'd change the description to say:
"This workload runs queries from the tsbs_query benchmark in Genny. It allows us to cross-check the results of the TSBS benchmarks, collect additional measurements, profile individual queries, and extend some of the queries as needed.
The data is preloaded by the dsi configuration. See 'configurations/test_control/test_control.tsbs_query.yml' for all the details. There are 20736000 documents in the collection."

src/phases/query/GetHostNameArray.yml
```diff
@@ -0,0 +1,24 @@
+SchemaVersion: 2018-07-01
```
IrinaYatsenko, line 1: Why in a separate file? It's of no use outside of tsbs_query workloads and depends heavily on the schema of the collection used there (e.g. having exactly 100 hosts)

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,435 @@
...
+#  The min and max time of the pre-loaded tsbs dataset.
+MinDate:
+  &MinDate {
+    ^RandomDate: { min: "2016-01-01T00:00:00Z", max: "2016-01-24T23:59:50Z" },
+  }
+
+MaxDate:
+  &MaxDate { ^RandomDate: { min: *MinDate, max: "2016-01-24T23:59:50Z" } }
+
+HostName1: &HostName1
```
IrinaYatsenko, line 33: Please add comments along the lines of:
```
# Generates a single string like "host_42", with random integer in range [0,99].
```
Same for HostName8 (think about the future readers of this file who have _no prior_ knowledge of the TSBS benchmarks).

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,435 @@
...
+          Collection: *collection
+          Operations:
+            - OperationName: aggregate
+              OperationCommand:
+                Pipeline:
+                  [
+                    {
+                      $match:
+                        {
+                          "tags.hostname": { $in: *HostName1 },
```
IrinaYatsenko, line 79: I see you've dropped the useless `"measurement":"cpu"` filter... While I totally agree it's not adding value to the query, it does mean that these queries are diverging from the ones in tsbs_query. We should at least call it out in the Description section.

src/workloads/query/TimeseriesTSBSQuery.yml
IrinaYatsenko: Also, think about whether there are any queries we'd like to run _in addition_ to the ones from tsbs_query against this dataset. E.g. non-trivial match on a measurement?

IrinaYatsenko (changes requested):
Feedback on the TSBS (haven't looked at the unpacking yet)

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,1188 @@
...
+                          usage_iowait: { $first: "$usage_iowait" },
+                          usage_irq: { $first: "$usage_irq" },
+                          usage_nice: { $first: "$usage_nice" },
+                          usage_idle: { $first: "$usage_idle" },
+                          usage_guest_nice: { $first: "$usage_guest_nice" },
+                        },
+                    },
+                  ]
+
+  # The queries below are not from tsbs_query, but are typical user queries.
```
IrinaYatsenko, line 1031: I'm not sure about the "typical" claim. I'd replace it with a description of what kind of optimizations/behaviours we are trying to validate with these pipelines. For example, your non-selective query essentially stresses unpacking, given you are adding a separate workload just for that, is it worth adding it here? 

For the other two, I presume the goal is to test bucket-targeting, right? What are our expectations for the distributions of the measurements across the buckets? In the tsbs dataset I'm seeing ~161K total buckets and ~31K buckets that match a filter like `{"control.min.usage_nice": {$lte: 80}, "control.max.usage_nice": {$gte: 80}}` (which gets inserted for `{$match: {usage_nice: 80}}`). That is, almost 20% of the buckets, and that's for a point-query. Extend the query to a range and you might end up with having to open half of the buckets if not more, rendering the test to be very slow and not really giving you good signal on bucket targeting.

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,1188 @@
...
+                          min__usage_steal: { $min: "$usage_steal" },
+                          min_usage_guest: { $min: "$usage_guest" },
+                          min_usage_user: { $min: "$usage_user" },
+                          min_usage_system: { $min: "$usage_system" },
+                        },
+                    },
+                    { $sort: { _id: 1 } },
+                  ]
+
+  - Name: match_on_measurement_meta_time
```
IrinaYatsenko, line 1114: What optimization does this pipeline target? Would it use index on the metaField or a collscan?

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,1188 @@
...
+                          _id: { $dateTrunc: { date: "$time", unit: "hour" } },
+                          min_usage_softirq: { $min: "$usage_softirq" },
+                          min__usage_steal: { $min: "$usage_steal" },
+                          min_usage_guest: { $min: "$usage_guest" },
+                          min_usage_user: { $min: "$usage_user" },
+                          min_usage_system: { $min: "$usage_system" },
+                        },
+                    },
+                    { $sort: { _id: 1 } },
+                  ]
```
IrinaYatsenko, line 1188: Other optimizations that we should cover (not necessarily on this dataset, but if we can do it here, so much the better):
1. whole bucket filter
2. empty include list on the unpack stage
3. count/min/max rewrites (unless there are already workloads for these)

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,1188 @@
...
+HostName8:
+  &HostName8 { ^Array: { of: *GenHostNames, number: 8, distinct: true } }
+HostNameAll:
+  &HostNameAll { ^Array: { of: *GenHostNames, number: 100, distinct: true } }
+
+Actors:
+  # The data is already loaded into the collection from the dsi configuration.
+  - Name: Quiesce
+    Type: QuiesceActor
+    Threads: 1
```
IrinaYatsenko, line 58: TSBS does either 4 or 8 threads. I don't think we need to run both configs, but let's do one of them.

IrinaYatsenko (changes requested):
Feedback for TSBS workload. Still haven't looked at the stress unpacking.

src/workloads/query/TimeseriesTSBSQuery.yml
IrinaYatsenko: Right now I'm more worried about having workloads for all the optimizations we do on time-series. Benchmarking typical user queries would be nice but not as pressing, imho.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+          Repeat: 1
+          Database: *database
+          CollectionCount: 1
+          Threads: 1
+          DocumentCount: 1000000
+          BatchSize: 1000
+          Document:
+            # Since step = 900 ms and documentCount = 1,000,000, the first and last document
+            # would be 900,000,000 ms apart (roughly 10 days).
+            time: { ^IncDate: { start: "2022-01-01", step: 900 } }
```
IrinaYatsenko, line 69: This means there will be no "out-of-order" events, making for very neat buckets. Maybe add a comment why this is OK/desireable for this workload?

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+            temp: { ^RandomDouble: { min: -30, max: 120 } }
+            # This is the value we are going to query against. We will have ~1600 buckets in this
+            # collection and roughly 600 documents per bucket. By making this an increasing value, we can write
+            # predicates to target certain number of buckets.
+            measurement0: { ^Inc: { start: 1 } }
+            # The fields below create larger documents to stress unpacking.
+            measurement1: { ^RandomDouble: { min: -100.0, max: 900.00 } }
```
IrinaYatsenko, line 76: Given the different nature of the fields, I'd call them something like `measurement` (the one we'll filter on) and `payloadN` (the ones we don't filter on). The issue with the randomized ones is that they are not changing continuously, making the bucket controls for them mostly useless.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+        Active: [3]
+        NopInPhasesUpTo: *maxPhases
+        PhaseConfig:
+          Repeat: 1
+
+  # Run queries that target a different number of buckets.
+  - Name: SimpleMatchMeasurement10Buckets
+    Type: CrudActor
+    Database: *database
+    Threads: 10
```
IrinaYatsenko, line 99: Why this number of threads?

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+        PhaseConfig:
+          Repeat: *repeat
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                # With 600 documents per bucket, 10 buckets will have 'measurement0' <= 6000 (600 * 10).
+                # We'll add 100 just to be safe.
```
IrinaYatsenko, line 113: It's not clear what we are trying to be safe from. If you want to _guarantee_ exactly 10 buckets being selected, let's test it does happen for whatever number is chosen in the filter.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+                Pipeline: [{ $match: { "measurement0": { $lte: 6100 } } }]
```
IrinaYatsenko, line 114: This pipeline will produce a kExclude unpacking stage and won't be eligible for lowering to SBE. To lower to SBE we have to either project or group.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+
+  - Name: ConjunctionMatchMeasurement10Buckets
```
IrinaYatsenko, line 116: What signal would this test provide in addition to the test above?

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+        PhaseConfig:
+          Repeat: *repeat
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                # With 600 documents per bucket, 100 buckets will have 'measurement0' <= 60000 (600 * 100).
+                Pipeline: [{ $match: { "measurement0": { $gt: 60100 } } }]
```
IrinaYatsenko, line 164: $lte?

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,227 @@
...
+        PhaseConfig:
+          Repeat: *repeat
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                # With 600 documents per bucket, 1000 buckets will have 'measurement0' <= 600000 (600 * 1000).
+                Pipeline: [{ $match: { "measurement0": { $gte: 600100 } } }]
```
IrinaYatsenko, line 214: $lte?

IrinaYatsenko (changes requested):
Feedback for stress unpacking

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,179 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
+  This test stress tests bucket unpacking. Each document is large with many top-level fields.
```
IrinaYatsenko, line 4: Thinking about the SBE's specifics of unpacking, I think we need to cover one more parameter here, namely, how many fields need to be unpacked from each bucket. Something like: 1, 10, 20. It might also be of interest to compare queries that project _all_ fields vs kExclude that project none but have to unpack everything anyway (these queries will run in classic).

galon1: Done!

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,749 @@
...
+  Default:
+    QueryOptions:
+      maxPoolSize: 4000 # With 8 threads we need a larger pool.
+
+GlobalDefaults:
+  Database: &database cpu
+  Collection: &collection point_data
+  maxPhases: &maxPhases 20
+  Repeat: &repeat 500
+  # TSBS constructs a time windows of various sizes (1 hour, 8 hours, 12 hours) from a uniformly random start-time within
```
galon1, line 32: The tasks were timing out on evergreen and taking so long to run (over 2 hours), so I'm testing reducing the number of times we repeat each query.

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,749 @@
...
+          Operations:
+            - OperationName: aggregate
+              OperationCommand:
+                Pipeline:
+                  # All usage_* fields are random integers between 0-100.
+                  [
+                    {
+                      $match:
+                        {
+                          $or:
```
IrinaYatsenko, line 693: In my local experiment with the tsbs dataset, this query generates the following filter on the bucket collection:
```
"$or" : [
	{"$and" : [
		{"control.max.usage_nice" : {"$_internalExprGt" : 0}},
		{"control.min.usage_nice" : {"$_internalExprLt" : 5}}
	]},
	{"$and" : [
		{"control.max.usage_guest" : {"$_internalExprGte" : 80}},
		{"control.min.usage_guest" : {"$_internalExprLt" : 85}}
	]}
]
```
And it selects 56,288 buckets out of 161,448, which is **~35%** and not ~1% as you claim in the comment.

I haven't tried all of the measurements but it seems that a single measurement value hits ~20% of the buckets. Assuming independent distributions, I'd expect that $and of three $eq on different measurements would yield ~0.8% and for two measurements would get ~4%. Either is OK for this test (but I'd check against the real data before commiting to a choice of filter).

src/workloads/query/TimeseriesTSBSQuery.yml
IrinaYatsenko: let's move these into the "optimizations" workload as well (after all, bucket-targeting is a form of optimization :))
btw, what's the workload for `$count`?

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,166 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
+  This workload runs different time-series optimizations with the tsbs dataset. During development, we targeted optimizations
+  that are most relevant to pushdown time-series queries to SBE. Further optimizations should be added to this workload.
+  The data is preloaded by the dsi configuration. See 'configurations/test_control/test_control.tsbs_query.yml' for all 
+  the details. There are 20736000 documents in the collection.
```
IrinaYatsenko, line 7: I'd collect and add here a few more statistics about the dataset: 
- the number of buckets, 
- the p50 and p80 of the number of documents per bucket 
- how many buckets are v1 vs v2, 
- what % of buckets a time window of 1 hour is expected to hit, 
- what % of buckets a single value of a measurement is expected to hit.

I'd also mention which field is meta and which field is time, the bounds on the time, and which indexes exist. Basically, do not expect the future maintainers to already know the particulars of the dataset (or know where to find this information). Collecting and outlining statistics like this also helps you, the author of the workload, to validate your assumptions about the perf characteristics of the queries.

IrinaYatsenko (changes requested):
review in progress

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,385 @@
...
+          Operations:
+          - OperationName: aggregate
+            OperationCommand:
+              Pipeline: [
+                    {
+                      $match:
+                        {
+                          "measurement0":
+                            {
+                              $lte:
```
IrinaYatsenko, line 48: nit: even though it's adding a little more duplication to the actors, I'd use as a parameter the whole predicate, that is `Predicate: {$lte: 6000}`. I think it will help with readability of the individual actors (also, technically speaking, a "predicate" is a boolean operator and not a number).

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,166 @@
...
+        NopInPhasesUpTo: *maxPhases
+        PhaseConfig:
+          Repeat: *repeat
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                Pipeline:  [ { $match: { "time": { $lt: *dateForMatchQuery } } }]
```
IrinaYatsenko, line 166: To make it eligible for SBE lowering, it must project.

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,166 @@
...
+  - group
+  - sort
+
+GlobalDefaults:
+  Database: &database cpu
+  Collection: &collection point_data
+  maxPhases: &maxPhases 20
+  repeat: &repeat 100
+
+DateForMatchQuery: &dateForMatchQuery
```
IrinaYatsenko, line 21: Why not use `{^Date: "2016-01-15T01:49:23"}` at the call site?

IrinaYatsenko (changes requested):
I'm super curious to see the Evergreen results of these workloads for all-feature-flags vs classic!

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,385 @@
...
+        Active: [3]
+        NopInPhasesUpTo: *maxPhases
+        PhaseConfig:
+          Repeat: 1
+
+  # Run queries that target a different number of buckets.
+  - ActorFromTemplate:
+      TemplateName: RunMatchProjectPipeline
+      TemplateParameters:
+        Name: "match_10_buckets_include_1_field"
```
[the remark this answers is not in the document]
galon1: Based on local testing the average latency in milliseconds:
match 10 buckets:
2 fields: 798.4
11 fields: 3476
15 fields: 4226
exclude: 2200

There is a significant difference between 2 and 11 fields, but not so much for 11 and 15 fields. I think we can just have 2 and 11 fields then. I made those changes to the workload.

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,286 @@
...
+  that are most relevant to pushdown time-series queries to SBE. Further optimizations should be added to this workload.
+  The data is preloaded by the dsi configuration. See 'configurations/test_control/test_control.tsbs_query.yml' for all 
+  the details. Below are helpful statistics about the collection:
+  1. There are 20736000 documents and 161448 buckets in the collection. 
+  2. "tags" is the metaField and "time" is the timeField.
+  2. The 50th percentile of the number of documents per bucket is 93.
+  4. The 80th percentile of the number of document per bucket is 274.96.
+  3. 40435 buckets are version 1 and 121013 buckets are version 2.
+  4. Predicates of bucket at a time window of 1 hour is expected to hit 0.2% of buckets.
+  5. Predicates on a single measurement value, such as {usage_nice: 5} will hit around 18% of buckets.
```
IrinaYatsenko, line 14: I'd extend this point to say that all measurements vary in [X, Y] range and have similar distributions. I'd also swap points 5 and 6 so that time related points are next to each other.

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,286 @@
...
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                Pipeline: [
+                  { 
+                    $group: 
+                    { 
+                      _id: "$tags.hostname",
```
IrinaYatsenko, line 92: Is it intentional that `GroupRewriteMax` groups by `$tags` and `GropuRewriteMin` groups by `$tags.hostname`?

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,286 @@
...
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                Pipeline: [
+                  { 
+                    $group: 
+                    { 
+                      _id: 5,
```
IrinaYatsenko, line 123: nit: it might be more idiomatic to use `_id: null`

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,286 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
+  This workload runs different time-series optimizations with the tsbs dataset. During development, we targeted optimizations
+  that are most relevant to pushdown time-series queries to SBE. Further optimizations should be added to this workload.
+  The data is preloaded by the dsi configuration. See 'configurations/test_control/test_control.tsbs_query.yml' for all 
+  the details. Below are helpful statistics about the collection:
+  1. There are 20736000 documents and 161448 buckets in the collection. 
+  2. "tags" is the metaField and "time" is the timeField.
```
IrinaYatsenko, line 9: Please add that `tags.hostname` has uniformly distributed 100 values.

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,286 @@
...
+          Repeat: *repeat
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName:
+              OperationName: aggregate
+              OperationCommand:
+                Pipeline:  [ 
+                  { 
+                    $match: { "time": { $lt: {^Date: "2016-01-15T01:49:23"} } }
```
IrinaYatsenko, line 195: Could you please add a comment re how many buckets this filter hits? (for the optimization to becomes noticeable, the number of hit buckets should be relatively high)

IrinaYatsenko (changes requested):
Almost there. Need changes in `TimeseriesStressUnpacking`. 

When you have the workloads finalized, please run them in Evergreen a few times (5?) to assess how noisy they are and whether the numbers make sense. We also should compare the results to the current tsbs_query benchmarks and make sure they are consistent.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,296 @@
...
+          Repeat: 1
+          Database: *database
+          CollectionCount: 1
+          Threads: 1
+          DocumentCount: 1000000
+          BatchSize: 1000
+          Document:
+            # Since step = 900 ms and documentCount = 1,000,000, the first and last document
+            # would be 900,000,000 ms apart (roughly 10 days).
+            # Add events in this way means we have no out-of-order events and neat buckets. However, this is acceptable for this workload,
```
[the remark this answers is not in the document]
galon1: Genny uses mongodb server style guidelines, but has a very limited linter. I ran the linter and the formatter suggested by the perf team (esbenp.prettier-vscode), but then that gave me the aggressive spacing, so I went in manually and fixed the aggressive spacing. But to match the style of the rest of the file, and fit the line width I could only remove some but not all of the aggressive spacing.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,296 @@
...
+            # because we are testing the speed of bucket unpacking. Inserting documents in order allows us to deterministically control how
+            # many documents are in each bucket, the number of overall buckets, and how many buckets each predicate will unpack.
+            time: { ^IncDate: { start: "2022-01-01", step: 900 } }
+            # This is the value we are going to query against. We will have ~1600 buckets in this
+            # collection and roughly 600 documents per bucket. By making this an increasing value, we can write
+            # predicates to target certain number of buckets.
+            measurement: { ^Inc: { start: 1 } }
+            # The fields below create larger documents to stress unpacking.
+            payload1: { ^RandomDouble: { min: -100.0, max: 900.00 } }
+            payload2: { ^RandomString: { length: 10 } }
```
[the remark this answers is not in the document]
galon1: I originally wanted to make the documents larger, so I wanted to add strings to make unpacking heavier, but I changed all of the fields to be numeric. This does mean that more documents will fit in each bucket. With all of the values being numeric we see less of an effect but still significant effect of unpacking more fields, which makes sense. Now locally I'm seeing:
100 buckets 1 field: 195.3 milliseconds avg latency
100 buckets 10 fields: 569.3 milliseconds avg latency
100 buckets exclude: 209.5 milliseconds avg latency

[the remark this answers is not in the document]
IrinaYatsenko: The exclude runs in classic and has an optimization to unpack directly to bsonObject which might explain why it's faster than include with the same set of fields. What is the latency of "100 buckets 10 fields" in classic?

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,749 @@
...
+
+GlobalDefaults:
+  Database: &database cpu
+  Collection: &collection point_data
+  maxPhases: &maxPhases 20
+  Repeat: &repeat 1000 # Matches the number of queries run in 'configurations/test_control/test_control.tsbs_query.yml'.
+  # TSBS constructs a time windows of various sizes (1 hour, 8 hours, 12 hours) from a uniformly random start-time within
+  # this time-period. Ideally, we would do this as well, but currently genny doesn't support storing randomized date values as
+  # variables, so we will fix a starting value and increment it each time the query runs.
+  StartDate: &StartDate "2016-01-04T21:49:27Z"
```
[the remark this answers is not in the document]
IrinaYatsenko: For what it's worth, the last incarnation of these anchors still uses inconsistent casing :)

src/workloads/query/TimeseriesTSBSQuery.yml
IrinaYatsenko: Intresting, so the values for the different "usage_*" metrics aren't uniformly distributed. Well, then yes, we have to double-check the actual selectivities of all queries we use here and be careful when/if changing them. Thank you for running the experiments!

src/workloads/query/TimeseriesTSBSOptimizations.yml
IrinaYatsenko: In a different comment thread Gil posted results that show that the distributions aren't actually similar...

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,282 @@
...
+  See 'configurations/test_control/test_control.tsbs_query_genny.yml' for all  the details.
+  
+  Below are helpful statistics about the collection:
+  1. There are 20736000 documents and 161448 buckets in the collection.
+  2. "tags" is the metaField and "time" is the timeField.
+  3. "tags.hostname" has 100 values that are uniformly distributed.
+  4. The 50th percentile of the number of documents per bucket is 93.
+  5. The 80th percentile of the number of document per bucket is 274.96.
+  6. 40435 buckets are version 1 and 121013 buckets are version 2.
+  7. Predicates on a single measurement value, such as {usage_nice: 5} will hit around 18% of buckets.
```
IrinaYatsenko, line 16: Remove this point (because it seems that the distributions of values for different "usage_*" metrics aren't the same and some of them aren't uniform)?

IrinaYatsenko (approved):
Hurray! I love having these workloads in Genny. Thank you so much for building them!

src/workloads/query/TimeseriesStressUnpacking.yml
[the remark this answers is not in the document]
galon1: I need to run an updated patch on evergreen but locally here is what I got from different runs. All of the times are average latency in milliseconds.

Results with the feature flag on: 
100 buckets include 1: 301.9
100 buckets include 10: 972.8
100 buckets exclude: 232.5

Results from classic:
100 buckets include 1: 252.6
100 buckets include 10: 570.2
100 buckets exclude: 262.9

So it seems like classic is faster and that difference becomes more apparent the more fields we need to unpack. I'll confirm this with an evergreen run however.

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,276 @@
...
+        Active: [12]
+        # With 988 documents per bucket, 1000 buckets will have 'measurement' <= 988000 (988 * 1000).
+        Predicate: { $lte: 988000 }
+        # Without a $project stage, $_internalUnpackBucket will have an "exclude" parameter.
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - standalone-all-feature-flags
```
banarun, line 269: I think we should use the same variants as all the other Time-series workloads. The `replica` and `replica-all-feature-flags`. Makes it easier to do comparisons.

https://github.com/mongodb/genny/blob/2f11236c169a8a5abf2b6b38202340057276b665/src/workloads/query/TimeseriesCount.yml#L124-L127

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,276 @@
...
+          - standalone
+      branch_name:
+        $neq:
+          - v4.0
+          - v4.2
+          - v4.4
+          - v5.0
```
banarun, line 276: Can we also disallow `v6.0`? I think keeping `v7.0` makes sense so that we can do master vs 7.0 comparison, but I don't see much benefit in running these workloads on 6.0.

banarun (commented):
I just have a couple of comments about where we should run these workloads.

src/workloads/query/TimeseriesStressUnpacking.yml
IrinaYatsenko: there was quite a bit of work for time-series in 7.0, no? given the scarcity of coverage why no run these workloads in 6.0 to see the cross-version trend?

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,283 @@
...
+                      $project:
+                        {
+                          _id: 1,
+                          tags: 1,
+                          usage_softirq: 1,
+                          usage_steal: 1,
+                          usage_guest: 1,
+                        },
+                    },
+                  ]
```
jimoleary, line 283: Don't these need a branch_name?

src/workloads/query/TimeseriesTSBSQuery.yml
```diff
@@ -0,0 +1,950 @@
...
+
+Actors:
+  # The data is already loaded into the collection from the dsi configuration.
+  - Name: Quiesce
+    Type: QuiesceActor
+    Threads: 1
+    Database: *database
+    Phases:
+      OnlyActiveInPhases:
+        Active: [0]
```
jimoleary, line 68: should this be executed between phases?

src/workloads/query/TimeseriesStressUnpacking.yml
```diff
@@ -0,0 +1,277 @@
...
+            # The fields below create larger documents to stress unpacking.
+            payload1: { ^RandomDouble: { min: -10, max: 100 } }
+            payload2: { ^RandomDouble: { min: -10, max: 100 } }
+            payload3: { ^RandomDouble: { min: -10, max: 100 } }
+            payload4: { ^RandomInt: { min: -10, max: 100 } }
+            payload5: { ^RandomInt: { min: -10, max: 100 } }
+            payload6: { ^RandomInt: { min: -10, max: 100 } }
+            payload7: { ^RandomInt: { min: -10, max: 100 } }
+
+  - Name: Quiesce
```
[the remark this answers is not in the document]
galon1: added!

src/workloads/query/TimeseriesTSBSOptimizations.yml
```diff
@@ -0,0 +1,283 @@
...
+
+Actors:
+  # The data is already loaded into the collection from the dsi configuration.
+  - Name: Quiesce
+    Type: QuiesceActor
+    Threads: 1
+    Database: *database
+    Phases:
+      OnlyActiveInPhases:
+        Active: [0]
```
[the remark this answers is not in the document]
galon1: added!

src/workloads/query/TimeseriesTSBSQuery.yml
galon1: added!

src/workloads/query/TimeseriesStressUnpacking.yml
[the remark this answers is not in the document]
IrinaYatsenko: I thought "Quiesce" forces writes to be flushed to disk and such so that there is no background activity to interfere with the tests. These tests are all read-only, is "Quiesce" still useful?

src/workloads/query/TimeseriesTSBSOptimizations.yml
IrinaYatsenko: It looks like the "AutoRun" workloads get named after their yml file by transforming Capitalization to underscore. That is,  "FooBar.yml" would map to "foo_bar" task. To make the analysis of the perf results easier, maybe, let's name this file "TimeseriesTsbsOptimizations.yml" and name the corresponding task in DSI "timeseries_tsbs_optimizations"?

jimoleary (approved):
lgtm mod changing the file names to be consistent (plus maybe another patch to be sure it's working if you change the names ... have a look at [skipping compilation on sys perf projects](https://docs.devprod.prod.corp.mongodb.com/performance/patch_testing/#skipping-compilation-on-sys-perf-projects) if you want to get an hour back).

In terms of commit order it's probably:

1. genny
2. dsi
3. mongodb

jimoleary:
I also created EVG-21091 for the OperationMetricsName null weirdness but it's pretty minor.

src/workloads/query/TimeseriesTSBSOptimizations.yml
IrinaYatsenko: BTW, connecting the workload definitions to the results in Evergreen might be an argument to create separate DSI tasks for each workload (using Genny's AutoRun naming scheme). Otherwise, one has to search in DSI repo for the mentioning of this *.yml file to see task name which then can be searched in Evergreen... (I do hope the proposal about tagging workloads with "Context" would help with stuff like this).

galon1: done! Confirmed with Jim offline and split the configurations between TsbsOptimizations workload and TsbsQuery workload. Note that the names of the dsi configurations are slightly different than the genny workloads to match the dsi configuration name style, which is different than the genny workload name style. 

TsbsOptimizations.yml dsi configuration is tsbs_query_optimizations
TsbsQuery.yml dsi configuration is tsbs_query_genny

galon1:
evergreen merge

PR #1033, opened 2023-10-03
SERVER-79756 Create workload targetted at plans which require a shard filter

bshteinfeld:

**Jira Ticket:** SERVER-79756

**Whats Changed:**  
Added new workload to target plans with a shard filter stage.

**Patch testing results:**  
🌲 [patch](https://spruce.mongodb.com/version/651c6fe32a60eda358ef736e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Workload Submission form:**  
Done

**Related PRs:**   
n/a

### DIFF
--- a/src/workloads/query/ShardFilter.yml
+++ b/src/workloads/query/ShardFilter.yml
@@ -9,7 +9,7 @@ GlobalDefaults:
   Collection: &Collection Collection0
   Namespace: &Namespace test.Collection0
   MaxPhases: &MaxPhases 8
-  DocumentCount: &DocumentCount 1000
+  DocumentCount: &DocumentCount 1e6
 
 Actors:
 - Name: CreateShardedCollection
@@ -99,3 +99,44 @@ Actors:
         - OperationName: find
           OperationCommand:
             Filter: {sKey: {$in: [123, 5000, 6789]}}
+
+- Name: PointQueryByNonShardKey
+  Type: CrudActor
+  Database: *Database
+  Threads: 16
+  Phases:
+    OnlyActiveInPhases:
+      Active: [5]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: 5000}
+
+- Name: AllDocumentsQueryByNonShardKey
+  Type: CrudActor
+  Database: *Database
+  Threads: 16
+  Phases:
+    OnlyActiveInPhases:
+      Active: [6]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: {$gt: -1}}
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - shard
+      - shard-heuristic-bonsai
+    branch_name:
+      $gte: v7.2
--- a/src/workloads/query/ShardFilter.yml
+++ b/src/workloads/query/ShardFilter.yml
@@ -1,0 +1,101 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of queries which need to perform shard filtering.
+Keywords:
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+  MaxPhases: &MaxPhases 8
+  DocumentCount: &DocumentCount 1000
+
+Actors:
+- Name: CreateShardedCollection
+  Type: AdminCommand
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: admin
+        Operations:
+        - OperationMetricsName: EnableSharding
+          OperationName: AdminCommand
+          OperationCommand:
+            enableSharding: *Database
+        - OperationMetricsName: ShardCollection
+          OperationName: AdminCommand
+          OperationCommand:
+            shardCollection: *Namespace
+            key: {sKey: 1}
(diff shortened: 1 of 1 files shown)


src/workloads/query/ShardFilter.yml
```diff
@@ -0,0 +1,101 @@
...
+    OnlyActiveInPhases:
+      Active: [4]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {sKey: {$in: [123, 5000, 6789]}}
```
alyacb, line 101: nit: newline?

alyacb, line 101: Shouldn't we have an AutoRun section here? This should include the bonsai 3-shard variant, as well as the baseline 3-shard variant if we want it to run in waterfall with Bonsai + master so we can compare the two in the dashboard. ([example](https://github.com/mongodb/genny/blob/9af1ccd6887c5320cd08801b6bc02a1b9a13af83/src/workloads/sharding/MultiShardTransactionsWithManyNamespaces.yml#L243-L248))

That's probably why it didn't run in the patch you linked.

src/workloads/query/ShardFilter.yml
```diff
@@ -0,0 +1,101 @@
...
+Description: |
+  This workload tests the performance of queries which need to perform shard filtering.
+Keywords:
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  Namespace: &Namespace test.Collection0
+  MaxPhases: &MaxPhases 8
+  DocumentCount: &DocumentCount 1000
```
alyacb, line 12: Why do we want such a small collection?

src/workloads/query/ShardFilter.yml
```diff
@@ -0,0 +1,101 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: {$gt: 5000}}
+
+- Name: QueryByShardKeyExpression
```
alyacb, line 87: I think Xiaochen brought up at some point that we don't have many concurrent mixed-query workloads- might be worth adding one phase where we run some/all of these queries together in parallel. On the other hand, could be quite noisy.

src/workloads/query/ShardFilter.yml
```diff
@@ -0,0 +1,101 @@
...
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {sKey: {$gt: 5000}}
```
alyacb, line 69: Are we making an intentional choice here about the cardinality of this query (50% of the collection)? If so, we should document it in a comment.

src/workloads/query/ShardFilter.yml
```diff
@@ -0,0 +1,101 @@
...
+    OnlyActiveInPhases:
+      Active: [3]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {a: {$gt: 5000}}
```
alyacb, line 85: Do we want to test point queries/different cardinalities in this workload as well? Having a range of query-types may help us to find any problems that apply to shard-filtering specifically, since we'll be collecting data about more than just shard-filtering for each query run.

alyacb (changes requested):
Thanks Ben! I have a few comments/questions, but mainly I think we need an AutoRun section.

bshteinfeld (commented):
Thanks for the review!

src/workloads/query/ShardFilter.yml
alyacb: Mostly I just want to see multiple cardinalities of queries being tested, unless we're trying to force a particular scenario to occur. I think the extra tests you added should cover that.

Dumb question- do we still shard-filter if we match no documents? If so, maybe a no-match case would be helpful.

src/workloads/query/ShardFilter.yml
alyacb: LGTM! Thanks!

src/workloads/query/ShardFilter.yml
alyacb: Perfect, thanks! Small request to run the 3-shard version of your workload in your patch as well, as a sanity check.

src/workloads/query/ShardFilter.yml
alyacb: It matters because of what we're measuring. Because we don't have optimizer metrics yet, we only get stats about client reported metrics (overall query latency and throughput basically). This means that depending on what size of collection you pick, different results will be visible in the dashboard (i.e. its possible that either a slow execution time or a slow planning time will be masked by something else).

Small collection = planning time will probably dominate based on the other workloads I've looked at for Bonsai, while a larger collection = execution time will probably dominate. If you want to focus on one or the other, that can inform your choice of cardinality- for instance, we recently added a small-collection workload with large queries designed to stress the planner. If you want both data points, I would recommend having a "small" and "large" version of each query.

From a shipping M2 perspective, I'd say both is desirable.

alyacb (approved):
Thanks Ben! I still have a couple of questions/small requests, but otherwise LGTM!

src/workloads/query/ShardFilter.yml
bshteinfeld: https://spruce.mongodb.com/version/651dbed03627e0894809bf3b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

src/workloads/query/ShardFilter.yml
allenwux: Let's keep this one more targeted. We can use other queries (maybe the ones we got from help tickets) for that purpose.

allenwux (approved):

PR #1042, opened 2023-10-12
SERVER-81326 Production-ize memory heavy workload

lia394126:
**Jira Ticket:** SERVER-81326

**Whats Changed:**  
Added a workload that tests performance of the server under heavy memory pressure.

**Patch testing results:**  
https://spruce.mongodb.com/version/652544039ccd4e8ae4c52a74/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
https://spruce.mongodb.com/version/65283c2b3e8e86e8b9ba6f31/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

**Workload Submission form:**  
TODO

**Related PRs:**   
N/A

### DIFF
--- a/src/workloads/scale/MemoryStress.yml
+++ b/src/workloads/scale/MemoryStress.yml
@@ -1,0 +1,90 @@
+SchemaVersion: 2018-07-01
+Owner: Service Architecture
+Description: |
+  Used to test performance of the server under heavy memory pressure.
+
+GlobalDefaults:
+  dbname: &DBName memorystress
+  MaxPhases: &MaxPhases 20
+
+  LoadThreads: &LoadThreads 128
+  LoadBatchSize: &LoadBatchSize 500
+
+  NumberOfDocuments: &NumDocs 20096
+  # NumberOfDocumentsPerWorker = NumberOfDocuments / LoadThreads.
+  NumberOfDocumentsPerWorker: &DocsPerThread 157
+  DocSize: &DocSize 520000
+  Document: &Doc
+    # Document number/ID.
+    a: {^Inc: {start: -157, step: 1, multiplier: *DocsPerThread}}
+    # Random number from [0, NumberOfDocuments].
+    b: {^RandomInt: {min: 0, max: *NumDocs}}
+    # Fill 520 KB.
+    c: {^RandomString: {length: *DocSize}}
+
+  # Number of documents to sort by each thread. Tuned so that *SortStep * *DocSize < 100 MB,
+  # which is the maximum memory a sort query is allowed to use before spilling to disk.
+  SortStep: &SortStep 180
+  SortBatchSize: &SortBatchSize 180
+  SortThreads: &SortThreads 250
+  SortRepeat: &SortRepeat 5
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
+      maxPoolSize: 500
+    
+Actors:
+- Name: LoadDocuments
+  Type: Loader
+  Threads: *LoadThreads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        CollectionCount: 1
+        Database: *DBName
+        Repeat: 1
+        Document: *Doc
+        MultipleThreadsPerCollection: true
+        DocumentCount: *NumDocs
+        BatchSize: *LoadBatchSize
+
+- Name: SortMany
+  Type: RunCommand
+  Threads: *SortThreads
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *SortRepeat
+        Database: *DBName
+        Operations:
+        - OperationMetricsName: SortMany
+          OperationName: RunCommand
+          OperationCommand:
+            # Loader default collection name.
+            aggregate: Collection0
+            # Sort *SortStep number of documents starting from a randomly chosen 'a'.
+            pipeline:
+              [{$match:
+                {$expr:
+                  {$let:
+                    {vars:
+                      {start: {^RandomInt: {min: 0, max: *NumDocs}}},
+                    in: {$and: [$gte: ["$a", "$$start"], $lt: ["$a", {$add: ["$$start", *SortStep]}]]}
+                    }
+                  }
+                }
+              },
+              {$sort: {b: 1}}]
+            cursor: {batchSize: *SortBatchSize}
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - shard-lite


src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
+SchemaVersion: 2018-07-01
+Owner: Service Architecture
+Description: |
+  Used to test performance of the server under heavy memory pressure.
```
lia394126, line 4: As an effect of the purpose described here, the memory heavy operations should cause the hosts associated with the latest mongod binaries to crash. This is because we currently do not have anything to prevent memory overload.

The ideal purpose for putting this crash inducing workload into master is so that other server engineers can find this as an Evergreen task and run to test their implementations against. But because this would cause red/purple-ness on sys-perf with the current mongod binaries, I want to leave this whole workload commented out. Future engineers who want to use this workload would simply uncomment and run `schedule_patch_auto_tasks`.

In the future, I expect this to be uncommented as mongod gets better at handling memory pressure (and therefore would no longer cause failures).

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
+SchemaVersion: 2018-07-01
```
lia394126, line 1: Apparently compile needs SchemaVersion

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
...
+                      {start: {^RandomInt: {min: 0, max: *NumDocs}}},
+                    in: {$and: [$gte: ["$a", "$$start"], $lt: ["$a", {$add: ["$$start", *SortStep]}]]}
+                    }
+                  }
+                }
+              },
+              {$sort: {b: 1}}]
+            cursor: {batchSize: *SortBatchSize}
+
+# AutoRun:
```
lia394126, line 86: As an effect of the purpose described in `Description` in the header, the memory heavy operations should cause the hosts associated with the latest mongod binaries to crash. This is because we currently do not have anything to prevent memory overload.

### Why commented out

The ideal purpose for putting this crash inducing workload into master is so that other server engineers can find this as an Evergreen task and run to test their implementations against. But because this would cause red/purple-ness on sys-perf with the current mongod binaries, I want to leave this workload commented out. Future engineers who want to use this workload would simply uncomment and run `schedule_patch_auto_tasks`.

In the future, I expect this to be uncommented as mongod gets better at handling memory pressure (and therefore would no longer experience these failures). Let me know what you think about this approach of commenting the workload out.

### Alternatives

Other options:
(1) Use task annotations to mark the workload as non problematic.
(2) Leave as YML file linked to the ticket, and engineers wishing to run the workload would put the file into `src/workloads/scale/` and run `schedule_patch_auto_tasks`, no merging to master required.
(3) Create a new temporary (or permanent) project with its own configurations (e.g. copy something like `system_perf.yml` and configure post run actions, run as `evergreen patch -p <new_project_name>`). This may be a bit more heavy handed but it would not cause any noise on `sys-perf` master.

Please do let me know if there are better ways of doing any of this.

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
...
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
+      maxPoolSize: 500
+    
+Actors:
+- Name: LoadDocuments
+  Type: Loader
+  Threads: *LoadThreads
```
lia394126, line 41: Currently, this workload just mainly tests read only operations (agg sort), which is consistent in causing OOM. But this can be extended in the future to test other memory intensive operations, such as insert (just increasing the amount of loader threads), index creation, updates, etc.

jimoleary (commented):
This workload is cool!

But I had an issue with it until I got to your end comment about not being able to auto-schedule it as it will always be purple which makes triage difficult.

I think this shows that we are missing genny functionality, essentially the genny equivalent of :

```
with pytest.raises(ValueError, match='must be 0 or None'):
    raise ValueError('value must be 0 or None')
```

Maybe we should raise a ticket to implement such a feature.

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
...
+  LoadThreads: &LoadThreads 128
+  LoadBatchSize: &LoadBatchSize 500
+
+  NumberOfDocuments: &NumDocs 20096
+  # NumberOfDocumentsPerWorker = NumberOfDocuments / LoadThreads.
+  NumberOfDocumentsPerWorker: &DocsPerThread 157
+  DocSize: &DocSize 520000
+  Document: &Doc
+    # Document number/ID.
+    a: {^Inc: {start: -157, step: 1, multiplier: *DocsPerThread}}
```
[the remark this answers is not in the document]
lia394126: This is an unfortunate result of the equation for `^Inc = start + ActorId * multiplier + i * step`. I think ActorId starts at 1,  therefore in order for the true start of `^Inc` to be 0, it would mean `0 = start + 1*157 + 0*1` -> `start=-157`. I tried to use the preprocessor to do something like `start:-*DocsPerThread`, but that doesn't seem to pass `run-genny evaluate`.

I can put a comment here describing the above, or I could start the [sort at line 77](https://github.com/mongodb/genny/blob/e301a291f620b10d547e23a050f9c874597d021d/src/workloads/scale/MemoryStress.yml#L77) at `*DocSize` and end at `*DocSize + *NumDocs`. lmk what you think

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
...
+    # Random number from [0, NumberOfDocuments].
+    b: {^RandomInt: {min: 0, max: *NumDocs}}
+    # Fill 520 KB.
+    c: {^RandomString: {length: *DocSize}}
```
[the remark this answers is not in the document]
lia394126: Done!

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,90 @@
+SchemaVersion: 2018-07-01
+Owner: Service Architecture
+Description: |
+  Used to test performance of the server under heavy memory pressure.
+
```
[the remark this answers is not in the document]
lia394126: done

src/workloads/scale/MemoryStress.yml
lia394126: Will look into this, thanks!

src/workloads/scale/MemoryStress.yml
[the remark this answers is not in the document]
lia394126: done! left out language related to type of exception, because the actual exception genny encounters is "generic server error" due to "Could not find host matching read preference" (which is a result of mongods crashing from OOM). Added a sentence about how expectation of failure is temporary. LMK if language should be changed in any way

src/workloads/scale/MemoryStress.yml
jimoleary: I've created EVG-21102 to look into exception handling ... but I suspect it is a relatively large feature so I wouldn't depend or wait on it.

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,103 @@
...
+  SortRepeat: &SortRepeat 5
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
+      maxPoolSize: 500
+    
+Actors:
+- Name: LoadDocuments
```
jimoleary, line 52: Note: I've seen the following [setup phase](https://github.com/mongodb/genny/blob/ba047273ee7b8d5b693ca70dab51d97e4b948bc2/src/workloads/query/GroupSpillToDisk.yml#L23-L37)  added to some workloads. It can make local testing simpler.

jimoleary (commented):
A few suggestions and I created EVG-21102 to investigate exception handling.

src/workloads/scale/MemoryStress.yml
lia394126: > From [here](https://github.com/evergreen-ci/evergreen/blob/ff46048d63feb595caedb68fd4d9fc85ab73a96b/model/project_parser.go#L435) and [here](https://github.com/evergreen-ci/commit-queue-sandbox/blob/b92cc103abfbef75b1529fde02952b0b48570e99/evergreen.yml#L42C9-L42C24) I think we can add a task to a variant so that it can only be executed manually:

Added a `memory_overload` task (naming must be different to avoid conflict with schedule_global_auto_tasks AKA generate.tasks, which will create MemoryStress.yml -> memory_stress), PR: https://github.com/10gen/mongo/pull/16354

Note: I believe the conventional way to select workloads to run is to use the [AutoRun feature](https://jira.mongodb.org/browse/TIG-2041) with the intention of making workloads easy to run without changing or adding a bunch of project configurations. Perhaps we could keep this task temporarily while it doubles as a performance and correctness test? Then as we add features and testing to prevent OOM, we can remove this as a special task (instead just running with the `schedule_patch_auto_task` like the other workloads)?

src/workloads/scale/MemoryStress.yml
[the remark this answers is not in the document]
lia394126: Done!

src/workloads/scale/MemoryStress.yml
lia394126: Done!

src/workloads/scale/MemoryStress.yml
jimoleary: > Note: I believe the conventional way to select workloads to run is to use the [AutoRun feature](https://jira.mongodb.org/browse/TIG-2041) with the intention of making workloads easy to run without changing or adding a bunch of project configurations. Perhaps we could keep this task temporarily while it doubles as a performance and correctness test? Then as we add features and testing to prevent OOM, we can remove this as a special task (instead just running with the `schedule_patch_auto_task` like the other workloads)?

Generally it is preferable to use an AutoRun ... but it is perfectly valid to create an explicit task and run it with more control (to control how and when it runs).

So your current approach is perfectly valid.

I had assumed you might want to set `activate: false` on the sys perf task ... but you seem to be suggesting that you want to actively develop and run it. **Is that correct?**

If this is the case then ... you should let the buildbarons know (on the \#perf-build-baroning slack channel) that  this test is expected to fail with a system failure. It looks like there can be 2 different system failure types so we should let them know about that also.

It would also make sense to see if we can get EVG-21102 started as:

1.  this would remove the system exception
2. should allow us to add the exception count (or whatever metrics is created) to signal processing long term.

jimoleary (commented):
Looking good, but we would need to inform the build-barons before pushing it.

src/workloads/scale/MemoryStress.yml
jimoleary: also I only just now noticed your question on evergreen-users about the name. Yes it is the way that the genny file name gets translated into a name that then clashes with the test name. I generally try to use '-' rather than underscores in system_perf.yml.

So:
```
  - name: memory-stress
    commands:
      - func: f_run_dsi_workload
        vars:
          test_control: auto_genny_workload
          auto_workload_path: ./src/genny/dist/etc/genny/workloads/scale/MemoryStress.yml
```

That is probably preferrable because otherwise it can be confusing when you are looking for the file and or test.

jasonjhchan: I think this is intended only for local development, and shouldn't be pushed into the mainline waterfall (since we don't want to purposely push tests that will show up as failures, even if it's meant to happen).

lia394126: So with the project configuration here: https://github.com/10gen/mongo/pull/16354, any developer can run this as a checkable box task. The cron will also not periodically run this task because of `activate: false`, and therefore there shouldn't really be any alarm. I will also check in with #perf-build-baron to notify that this task is expected to fail in case someone accidentally runs this task. IMO, this doesn't really cause any noise and it just makes it easier for workload management people to test out any solutions/POCs. Is this an acceptable solution?

Otherwise we could just do the commenting out AutoRun, which IMO would be a bit more confusing for developers who don't use genny. lmk your thoughts

jasonjhchan: I think we probably would rather avoid having the box checkable by anyone if it's going to fail just to avoid any confusion. I'd rather the workflow be that for the small amount of people who know that they want this particular workload (and so know what the intended task status is) need to do the manual step of checking out genny and uncommenting some lines in code.

I want to avoid the situation where people may decide to blindly run some set of workloads and then spend time digging into why a task is failing.

src/workloads/scale/MemoryStress.yml
```diff
@@ -0,0 +1,103 @@
+SchemaVersion: 2018-07-01
+Owner: Service Architecture
+Description: |
+  Used to test performance of the server under heavy memory pressure. This workload is
+  expected to fail due to host(s) being unreachable as a result of mongod(s) running out
+  of memory. When the server becomes more suited to handle memory pressure, the
```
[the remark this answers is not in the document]
lia394126: done!

src/workloads/scale/MemoryStress.yml
lia394126: Okay that makes sense. I will add a comment in the JIRA ticket after this is merged to basically to this: uncomment the lines, run evg patch sys-perf, then check `schedule_patch_auto_task` on the shard-lite variant.

jimoleary: what ever works for you guys.

jimoleary (approved):
LGTM

src/workloads/scale/ReadMemoryStressUntilFailure.yml
```diff
@@ -0,0 +1,131 @@
...
+- out of memory
+
+GlobalDefaults:
+  dbname: &DBName memorystress
+  MaxPhases: &MaxPhases 20
+
+  LoadThreads: &LoadThreads 128
+  LoadBatchSize: &LoadBatchSize 500
+
+  NumberOfDocuments: &NumDocs 20096
```
jasonjhchan, line 33: Is this number significant in any way other than being some "large" number? If not, what's the criteria that we want to document for what this number should be?

src/workloads/scale/ReadMemoryStressUntilFailure.yml
```diff
@@ -0,0 +1,131 @@
...
+  # which is the maximum memory a sort query is allowed to use before spilling to disk.
+  SortStep: &SortStep 180
+  SortBatchSize: &SortBatchSize 180
+  SortThreads: &SortThreads 250
+  SortRepeat: &SortRepeat 5
+
+Clients:
+  Default:
+    QueryOptions:
+      socketTimeoutMS: -1
```
jasonjhchan, line 59: what does a socketTimeoutMs of -1 mean?

src/workloads/scale/ReadMemoryStressUntilFailure.yml
```diff
@@ -0,0 +1,131 @@
...
+      PhaseConfig:
+        CollectionCount: 1
+        Database: *DBName
+        Repeat: 1
+        Document: *Doc
+        MultipleThreadsPerCollection: true
+        DocumentCount: *NumDocs
+        BatchSize: *LoadBatchSize
+
+# Spawn many threads to sort 180 documents to test server's capacity to handle memory pressure.
```
jasonjhchan, line 95:
```suggestion
# Spawn many threads to sort enough documents to test server's capacity to handle memory pressure.
```
Since 180 is a constant defined above that is hardcoded and may change down the road. The Actor should only care about running a sort that is meant to generate large memory-consuming queries.

src/workloads/scale/ReadMemoryStressUntilFailure.yml
```diff
@@ -0,0 +1,131 @@
...
+                      {start: {^RandomInt: {min: 0, max: *NumDocs}}},
+                    in: {$and: [$gte: ["$a", "$$start"], $lt: ["$a", {$add: ["$$start", *SortStep]}]]}
+                    }
+                  }
+                }
+              },
+              {$sort: {b: 1}}]
+            cursor: {batchSize: *SortBatchSize}
+
+# AutoRun:
```
jasonjhchan, line 127: Maybe add a comment for why this is commented out?

jasonjhchan (changes requested):
last few comments, thanks!

src/workloads/scale/ReadMemoryStressUntilFailure.yml
lia394126: `NumberOfDocuments` is just a multiple of `NumberOfDocumentsPerWorker`. This number is large enough that `NumberOfDocumentsPerWorker` and `DocumentSize` contribute to an almost 100MB document. These numbers are tunable, but if one number is tuned, then the others need to be as well in order to match the constraints (of 100MB).

So it is just a large number, and it can be changed. `DocumentSize` and `NumberOfDocumentsPerWorker` are both arbitrary but reasonable numbers.

src/workloads/scale/ReadMemoryStressUntilFailure.yml
lia394126: done

src/workloads/scale/ReadMemoryStressUntilFailure.yml
lia394126: I assumed many workloads use -1 as max timeout but I can't find proof of that this is max timeout, will look for it

jasonjhchan (approved):

lia394126:
evergreen merge

PR #1047, opened 2023-10-23
SERVER-82359: Add genny workloads to test IDHack with identity view and via aggregation

wbuerger46:
Adds 3 IDHack-related workloads:

1. IDHack via find over an identity view (a no-op view)
2. IDHack via basic agg ({$match: {_id: x}})
3. IDHack via agg over an identity view

[Evergreen patch](https://spruce.mongodb.com/version/653672dad1fe0769389d15d7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


wbuerger46:
[Sys-perf patch with more runs through mongos](https://spruce.mongodb.com/version/65392258c9ec4447b40c53d5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

src/phases/query/IDHack.yml
```diff
@@ -22,6 +23,47 @@ IDHackObjectIdTemplate:
...
+IDHackIntIdAggTemplate:
+  Name: {^Parameter: {Name: "Name", Default: "IDHackIntIdAgg"}}
+  Type: CrudActor
+  Threads: {^Parameter: {Name: "Threads", Default: 1}}
+  Phases:
+    OnlyActiveInPhases:
+      Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: {unused: "please specify in which phases this actor should be active."}}}]
+      NopInPhasesUpTo: {^Parameter: {Name: "NopInPhasesUpTo", Default: 99}}
+      PhaseConfig:
+          # Our testing suggested the first ~20 seconds of this actor shows a lot of noise before
```
cswanson310, line 56: This is reading as slightly dishonest :D Do you want to double check? or just remove this comment or reword it to say "matching the query timeline" or something?

src/phases/query/IDHack.yml
```diff
@@ -39,3 +81,50 @@ LoadObjectIdTemplate:
...
+          _id: {^Inc: {start: -2000000, multiplier: 1000000}}
+          # [10M, 20M)
+          a: {^Inc: {start: -1000000, multiplier: 1000000}}
+          # [20M, 30M)
+          b: {^Inc: {start: 0, multiplier: 1000000}}
+        Indexes:
+          - keys: {a: 1}
+          - keys: {b: 1}
+
+# Creates an identity view (view with an empty pipeline that acts as a no-op) over the given collection.
```
cswanson310, line 116: If we want this in the future, feels like it deserves it's own little template file. I could see it coming up again and it feels odd to load it via something called "IDHack.yml"

cswanson310 (approved):
LGTM! As mentioned, let's make sure this works on sharded variants first. Then I'm curious to see/double check that the numbers are as we expect - find should be faster than agg should be faster than identity view - right?

src/phases/query/IDHack.yml
wbuerger46: Ah good point. Updated to say `100 seconds to match the find query timeline.`

src/phases/query/IDHack.yml
wbuerger46: Moved to `Views.yml` -- @cswanson310 does that sound good?

wbuerger46:
evergreen merge

wbuerger46:
evergreen merge

PR #1049, opened 2023-10-24
SERVER-82400: rename streams/Lookup.yml to not collide with query/Lookup.yml

Aadeshp:
This seems to cause issues in genny workload auto run, which uses the yaml file as the unique identifier for that workload across all directories, so we can't actually have two workloads with the same file name.


sandeepgd (approved):

Aadeshp:
evergreen merge

Aadeshp:
evergreen merge

PR #1052, opened 2023-10-30
SERVER-82359: Enable aggregation IDHack workloads on sharded environments

wbuerger46:
We originally omitted sharded variants from the aggregation IDHack auto-runs since the reported statistics were suspiciously low. After an investigation, we found the root cause to be in mongos and filed SERVER-82566. Now we'll enable those workloads since the issue has been shown to not be with the workload itself.


cswanson310 (approved):
You ran a patch and these worked, correct? It's just that the numbers were suspicious?

wbuerger46:
evergreen merge

PR #1054, opened 2023-11-01
DEVPROD-2171: Tests AutoRun logic in Genny commit queue

PyGeek03:
**Jira Ticket:** [DEVPROD-2171](https://jira.mongodb.org/browse/DEVPROD-2171)

**Whats Changed:**  
Tests AutoRun logic in all workload YAML files in Genny commit queue, by adding a dry-run mode for `./run-genny auto-task`.  The way this dry-run mode works is quite similar to [auto-task's existing unit tests](https://github.com/mongodb/genny/blob/e412bf1375a405e09a8a60ea830252ee6ae1770c/src/lamplib/src/tests/test_auto_tasks.py#L467): by mocking the `expansions.yml` file. 

Also improves the error message for the case in BF-28651.

**Patch testing results:**  
- [Verifies that this change would have caught BF-28651 pre-commit](https://spruce.mongodb.com/version/6542dcc69ccd4eeaefec6f17/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -488,31 +499,40 @@ class ConfigWriter:
...
         """
         :param config: The configuration to write
         :param output_file: What file to write to.
         """
-        
         success = False
         raised = None
         try:
             out_text = config.to_json()
+            if dry_run:
```
johndaniels, line 513: Honestly I'd consider pulling this out into the main method, since the entire purpose of this function is to write the config, and having a parameter to not do that seems confusing and involves more threading of parameters into functions.

src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -169,10 +155,18 @@ class Workload:
...
+        workspace_root: str,
+        file_path: str,
+        is_modified: bool,
+        reader: YamlReader,
+        dry_run: bool,
+    ):
         self.workspace_root = workspace_root
         self.file_path = file_path
         self.is_modified = is_modified
+        self.dry_run = dry_run
```
johndaniels, line 169: Is this used anywhere?

johndaniels (changes requested):
It seems like dry_run only needs to be considered in the main method, and avoid adding complexity to the interior methods. Let me know if you think that's not right.

src/lamplib/src/genny/tasks/auto_tasks.py
PyGeek03: I initially used it inside `Workload.variant_tasks()`, but then realized that mocking `expansions` is a better way to achieve the same thing (exercising the variant comparison logic). I've removed it from this class.

src/lamplib/src/genny/tasks/auto_tasks.py
PyGeek03: I agree. I've removed dry_run from `write_config()`.

johndaniels (approved):

PyGeek03:
evergreen merge

PyGeek03:
evergreen merge

PR #1059, opened 2023-11-07
SERVER-77343 Add Boolean simplification benchmarks

aligusnet:

**Jira Ticket:** SERVER-77343

**Whats Changed:**  
Add benchmarks that measure performance of boolean expressions which can be simplified by the Boolean Simplifier. 

**Patch testing results:**  [link](https://spruce.mongodb.com/version/6555d34861837d4298a2e825/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/phases/query/BooleanSimplifier.yml
+++ b/src/phases/query/BooleanSimplifier.yml
@@ -1,0 +1,239 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-optimization"
+
+Description: |
+  This workload measures performance of boolean expressions which can be simplified by
+  the Boolean Simplifier. It is designed to track effectiveness of the simplifier.
+
+RandomSeed: 172547222
+
+GlobalDefaults:
+  Database: &Database {^Parameter: {Name: Database, Default: test}}
+  Collection: &Coll Collection0
+  DocumentCount: &DocumentCount {^Parameter: {Name: DocumentCount, Default: 1e5}}
+  Threads: &Threads 1
+  MaxPhases: &MaxPhases 13
+  Duration: &Duration 2 seconds
+  QueryStringIn: &QueryStringIn [ "HrHCEtBAGnv",
+                                  "pCqgFttJBou",
+                                  "tjOnAuHAOpi",
+                                  "PmqoJGpDFjB",
+                                  "jEkEGEBnICu",
+                                  "kjPIJuEMGjm",
+                                  "nBKiBqKPCJN",
+                                  "BFkCIrIECBt",
+                                  "glhsCMqJPEs",
+                                  "lEMoKmmCiKv"
+                                ]
+  LargeFilterTerms: &LargeFilterTerms 100
+
+ActorTemplates:
+- TemplateName: BenchmarkTemplate
+  Config:
+    Name: {^Parameter: {Name: "Name", Default: "unused"}}
+    Type: CrudActor
+    Database: *Database
+    Threads: 1
+    Phases:
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "ActivePhase", Default: -1}}]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Duration: *Duration
+          Collection: *Coll
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {unused: "please specify a valid filter"}}}
+
+Actors:
+- Name: ClearCollection
+  Type: CrudActor
+  Database: *Database
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [0]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: 1
+        Collection: *Coll
+        Operations:
+        - OperationName: drop
+
+- Name: InsertData
+  Type: Loader
+  Threads: 4
+  Phases:
+    OnlyActiveInPhases:
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        MultipleThreadsPerCollection: true
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document:
+          a1: {^RandomInt: {min: 0, max: 100}}
+          b1: {^Inc: {start: 11, step: 2}}
+          a2: {^RandomInt: {min: 0, max: 1000}}
+          b2: 3
+          c2: {^RandomInt: {min: 0, max: 5}}
+          d2: {^RandomInt: {min: 0, max: 5}}
+          e2: 4
+          a3: {^RandomInt: {min: 0, max: 100}}
+          b3: {^RandomInt: {min: 0, max: 100}}
+          c3: {^Array: {of: {^FastRandomString: {length: 11}}, number: 10}}
+        Indexes:
+        - keys: {b1: 1}
+        - keys: {c2: 1, d2: 1}
+
+# Ensure all data is synced to disk.
+- Name: Quiesce
+  Type: QuiesceActor
+  Threads: 1
+  Database: *Database
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Threads: 1
+
+# The simplifier optimizes out an $or's branch to single predicate {b1: 2} and makes an Index Scan plan possible.
+- ActorFromTemplate:
+    TemplateName: BenchmarkTemplate
+    TemplateParameters:
+      Name: SimplifiedOutOrBranch.BM
+      ActivePhase: 3
+      Filter: {$or: [{$and: [{a1: 1}, {a1: {$ne: 1}}]}, {b1: 2}]}
+
+# The simplifier factorizes out common terms and makes a plan with one fetch possible instead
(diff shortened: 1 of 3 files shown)


src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,208 @@
...
+                                    {c3: {$nin: *QueryStringIn}},
+                                ]
+                            }
+                        ]
+                    },
+                ]
+            }
+
+
+- Name: TrivialOr.BM
```
alyacb, line 154: I don't think we need the .BM suffix- IMO, genny metrics names are quite verbose enough

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,208 @@
...
+    OnlyActiveInPhases:
+      Active: [3]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: *Duration
+        Collection: Collection0
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter:  {$or: [{$and: [{a1: 1}, {a1: {$ne: 1}}]}, {b1: 2}]}
```
alyacb, line 102: Might be helpful to include the "simplified" query here for reference (so, just mention its {b1: 2}, in case its not obvious to someone). Could do the same for other queries.

aligusnet: done

aligusnet (commented):
Thank you for your review, Alya. Please take a look at the updated workflow.

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,128 @@
...
+      $neq:
+      - M30-repl
+    branch_name:
+      $neq:
+      - v4.0
+      - v4.2
+      - v4.4
+      - v5.0
+      - v7.0
+      - v7.1
```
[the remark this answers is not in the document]
aligusnet: done

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,222 @@
...
+      PhaseConfig:
+        Duration: *Duration
+        Collection: *Coll
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter:  {a2: 1, $or: [{c2: 1, d2: 2, b2: 3}, {c2: 1, d2: 2, e2: 4}]}
+
+# The simplifier simplifes the filter expression to {a3: 0, b3: 0} which makes evaluating the Filter
+# stage of Collection Scan plan mach cheaper.
```
alyacb, line 124: nit: much- here & elsewhere

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,222 @@
...
+                                    {c3: {$in: *QueryStringIn}},
+                                    {c3: {$nin: *QueryStringIn}},
+                                ]
+                            }
+                        ]
+                    },
+                ]
+            }
+
+# The simplifier reduces the filter expression to {c2: 1, d2: 1} which makes passible a covered
```
alyacb, line 190: nit: possible

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,222 @@
...
+      Active: [7]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: *Duration
+        Collection: *Coll
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter:  {$or: [{c3: "HrHCEtBAGnv", c2: 1, d2: 1}, {c3: {$ne: "HrHCEtBAGnv"}, c2: 1, d2: 1}]}
+            Projection: {_id: 0, c2: 1, d2: 1}
```
alyacb, line 207: Could we add a test which uses a larger but also simplifiable predicate (I'm thinking in the range of 100+ clauses- here's an example of a [template](https://github.com/mongodb/genny/blob/28d4e7b8e7ca2171d260b17e92b45d78542b2f8d/src/workloads/query/FilterWithComplexLogicalExpression.yml#L58-L72 which generates an arbitrary number of clauses that would be helpful here)?

If we could add this test before index creation, we could also add the bonsai variants to AutoRun/ and could tag the data generated by that workload for the dashboard. This would allow us to see how Bonsai differs against master after your change/ if we need to address any difference here sooner rather than later.

I'd also say that for that particular workload, we could have a small collection & a large collection for the same workload- so, first we would create the small collection, say 100 or 500 docs, run this new workload, and then big collection with 100k docs, run the same workload again, then create indexes, and then run the remaining workloads on the big collection.

CC @timourk since this is relevant to the discussion yesterday around having two sizes of collection in workloads.

alyacb (changes requested):
Thanks Alex! This is almost LGTM, mod one request for a Bonsai-friendly workload (see my comment for details).

src/workloads/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,222 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-optimization"
+
+Description: |
+  This workload measures performance of boolean expressions which can be simplified by
+  the Boolean Simplifier. It is designed to track effetiveness of the simplifier.
```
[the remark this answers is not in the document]
aligusnet: done

src/workloads/query/BooleanSimplifier.yml
aligusnet: done

src/workloads/query/BooleanSimplifier.yml
aligusnet: done

src/workloads/query/BooleanSimplifier.yml
aligusnet: I am afraid I cannot add these tests right now, because the simplifier is enabled only for some restricted cases:

- because of performance we wouldn't even attempt to simplify a complex expression at the moment, this can be changed in the future
- it is disabled for Bonsai. Because, Bonsai tests are flaky in general. There is a server ticket to enable the simplifier for Bonsai https://jira.mongodb.org/browse/SERVER-76509

aligusnet: Actually, there is one collection scan test already added, so once Bonsai enabled the simplification we will see the results and let me try to add a big query

aligusnet: Added 2 benchmarks as you suggested 

LargeFilterCollectionScan and LargeFilterIndexScan. The simplifier actually managed to simplify them and shows around 2 times better throughput comparing to non-simplified version.

alyacb: Thank you for adding the large filter tests!

Can you also add a smaller collection size version of the LargeFilterCollectionScan workload? (see my first comment for an explanation)

> Actually, there is one collection scan test already added, so once Bonsai enabled the simplification we will see the results and let me try to add a big query

Bonsai (M2) won't run if there are any indexes other than _id (regardless of whether or not the query could use them), which is why I wanted the queries to run before creating indexes (running them after deleting the existing indexes is fine too). Otherwise, there's no point in running this on a Bonsai variant at all.

Could you also add the Bonsai variants to AutoRun after making this change?
```yml
      - standalone-heuristic-bonsai
      - standalone-sampling-bonsai
```

>  The simplifier actually managed to simplify them and shows around 2 times better throughput comparing to non-simplified version.

Very nice! 🎉

alyacb (changes requested):

antkorsh:
I have removed myself from the reviewers list as I discussed my concerns with Alex in person and they were addressed fully or partially.

src/workloads/query/BooleanSimplifier.yml
aligusnet: @alyacb I don't see a way how to run the benchmarks twice on collection of different sizes without copy-pasting everything in some way or another.

aligusnet: Actually, I believe that you mean to run only collection scans on smaller workload, which is like 2-3 actors only.

src/phases/query/BooleanSimplifier.yml
```diff
@@ -0,0 +1,231 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-optimization"
+
+Description: |
```
timourk, line 4: Could you make it clear in the comment the distinction between measuring the effectiveness of the simplifier (that is, how much it improved the final plan, so that the plan got faster) vs the efficiency of the simplifier (that is, how fast does the simplifier does its job). The latter would be tested via large predicates, and very small or empty datasets, while the former would be tested via various input predicates over reasonably large dataset.

aligusnet: These benchmarks are not designed to test efficiency of the simplifier, only effectiveness, which is already stated in the comment. For example, these benchmarks test the simplifier's efficiency: https://github.com/10gen/mongo/blob/dd975024e8e11bf250c9a75310020bb33a689e5d/src/mongo/db/query/boolean_simplification/match_expression_bm.cpp#L175-L313

src/workloads/query/BooleanSimplifier.yml
aligusnet: Added this second workload [BooleanSimplifierSmallDataset.yml](https://github.com/mongodb/genny/pull/1059#diff-b38d492b33cb9807ed8501874b543718bdd61b8d68a64dadfeb9f1c192e877ea) to test on smaller collections. Also the indexes are created after running collection scans now.

alyacb (approved):
LGTM

PR #1064, opened 2023-11-15
DEVPROD-2654: Don't generate tasks if no AutoRun block.

johndaniels:
We shouldn't generate a global task for any workload that doesn't have an AutoRun block. These tasks aren't usable and pollute the namespace.

**Jira Ticket:** DEVPROD-2654

**Whats Changed:**  
Don't generate a global task definition for Genny Workloads that don't have an AutoRun block. There is no reason to do this and it pollutes the global task namespace.

**Patch testing results:**  

https://spruce.mongodb.com/version/655432d83066159f7eed78fa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

The above patch is the fix for a naming conflict that was happening as part of PM-3362, since the "search" task name that was generated by Genny conflicted with an existing task.

**Workload Submission form:**  

**Related PRs:**   

### DIFF
--- a/src/lamplib/src/genny/tasks/auto_tasks.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks.py
@@ -246,7 +246,7 @@ class Workload:
         :return: all possible tasks irrespective of the current build-variant etc.
         """
         if not self.auto_run_info:
-            return [GeneratedTask(self.snake_case_base_name, None, None, self)]
+            return []
 
         tasks = []
         for block in self.auto_run_info:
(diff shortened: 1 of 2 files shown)


src/lamplib/src/genny/tasks/auto_tasks.py
```diff
@@ -246,7 +246,7 @@ def all_tasks(self) -> List[GeneratedTask]:
         :return: all possible tasks irrespective of the current build-variant etc.
```
PyGeek03, line 246: Maybe we can add some documentation here like "except if the task has no AutoRun block"?

dalyd (approved):
Change and intent LGTM unless I'm missing something. 
Have we patched tested it to make sure nothing surprising happens?

dalyd:
Oh, you are going to need to update the python tests that check the autotask functionality.

johndaniels:
evergreen merge

PR #1069, opened 2023-11-24
PERF-4889 Remove the 'numInitialChunks' parameter used with 'ShardCollection' command

robsndr:
**Jira Ticket:** [PERF-4889](https://jira.mongodb.org/browse/PERF-4889)

**Whats Changed:**  
The numInitialChunks parameter was deprecated in [SERVER-74747](https://jira.mongodb.org/browse/SERVER-74747), and now 1 chunk per shard is created by default. 
Adjust workloads accordingly. 

**Patch testing results:**  
Example patch run: https://spruce.mongodb.com/version/655f8149e3c3312d2f4f6616/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/query/LookupColocatedData.yml
+++ b/src/workloads/query/LookupColocatedData.yml
@@ -138,7 +138,7 @@ Actors:
           OperationName: AdminCommand
           OperationCommand:
             enableSharding: *Database
-        # This collection will have data distributed across 6 chunks.
+        # This collection will have data distributed evenly across the shards.
         - OperationMetricsName: ShardLocalCollection
           OperationName: AdminCommand
           OperationCommand:
--- a/src/workloads/query/LookupColocatedData.yml
+++ b/src/workloads/query/LookupColocatedData.yml
@@ -144,14 +144,12 @@ Actors:
           OperationCommand:
             shardCollection: test.ShardedLocal
             key: {*LocalShardKey: hashed}
-            numInitialChunks: 6
         # Here we use 1 chunk. This ensures that all the data for this coll is on one shard.
         - OperationMetricsName: ShardLocalCollectionAllOnOneShard
           OperationName: AdminCommand
           OperationCommand:
             shardCollection: test.ShardedLocalAllOnOneShard
-            key: {*LocalShardKey: hashed}
-            numInitialChunks: 1
+            key: {*LocalShardKey: 1}
         # Disable the balancer so that it can't skew results while the $lookups are running.
         - OperationMetricsName: DisableBalancer
           OperationName: AdminCommand
--- a/src/workloads/query/TenMDocCollection_ObjectId_Sharded.yml
+++ b/src/workloads/query/TenMDocCollection_ObjectId_Sharded.yml
@@ -38,7 +38,6 @@ Actors:
           OperationCommand:
             shardCollection: *Namespace
             key: {_id: "hashed"}
-            numInitialChunks: 12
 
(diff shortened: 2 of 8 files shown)


src/workloads/query/LookupColocatedData.yml
```diff
@@ -144,14 +144,12 @@ Actors:
...
             key: {*LocalShardKey: hashed}
-            numInitialChunks: 6
         # Here we use 1 chunk. This ensures that all the data for this coll is on one shard.
         - OperationMetricsName: ShardLocalCollectionAllOnOneShard
           OperationName: AdminCommand
           OperationCommand:
             shardCollection: test.ShardedLocalAllOnOneShard
-            key: {*LocalShardKey: hashed}
-            numInitialChunks: 1
+            key: {*LocalShardKey: 1}
```
pierlauro, line 152: Pointing out for other reviewers the chose around this shard key change: since `numInitialChunks` was being ignored, the chunks were actually created on all shards so the distributed lookup was faster (because it was not happening on just one node) and performances improved.
Changing the shard key to non-hashed actually forces the collection to be created on one node and brings performances back to "normal"

pierlauro (approved):
LGTM

src/workloads/query/LookupColocatedData.yml
jimoleary: Thanks! I was wondering why the numInitialChunks params were being deleted.

src/workloads/query/TenMDocCollection_ObjectId_Sharded.yml
```diff
@@ -38,7 +38,6 @@ Actors:
           OperationCommand:
             shardCollection: *Namespace
             key: {_id: "hashed"}
-            numInitialChunks: 12
```
jimoleary, line 41: Will older mongo versions still have a default of 2 chunks per shard and will that be significant?

jimoleary (commented):
The changes look good to me (I'll defer to your superior knowledge) but I have a couple of questions:

1. These tests run on older versions (most 6 and up, some 5 and up) , will these changes work as expected on them?
2. Is this change likely to cause a change in performance? If yes then you should let the perf buildbarons know before this PR is merged.

src/workloads/query/TenMDocCollection_ObjectId_Sharded.yml
robsndr: If the updated workload is ran against an older version, the default of 2 chunks per shard will created. It seems that using the default value should not affect the performance in this case.

jimoleary: Can you confirm that with a quick patch?

robsndr: The performance of those workloads shall be the same, because data will still be evenly distributed across all shards, regardless of the number of chunks used.  

A patch using 6.0 is running [here](https://spruce.mongodb.com/version/6560a5632a60ed45b4d9ec1f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

jimoleary (approved):
lgtm

There are some perf differences in the sharding commands that can be ignored, maybe some of the other changes will create change points / BFs but we can probably deal with them afterwards if this change needs to be merged anyway.

Maybe let the buildbarons know.

robsndr:
evergreen merge

PR #1079, opened 2023-12-07
SERVER-83897 Add versions of FilterWithComplexLogicalExpression and MatchFilters that run on collections of small sizes

henrinikku:
Thanks for submitting a PR to the Genny repo. Please include the following fields (if relevant) prior to submitting your PR.

**Jira Ticket:** SERVER-83897

**Whats Changed:**  
Add versions of FilterWithComplexLogicalExpression and MatchFilters that run on collections of small sizes (100 and 10K documents).

**Patch testing results:**  
Running [here](https://spruce.mongodb.com/version/6576f3cd2fbabe12e36fbaa5/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

**Workload Submission form:**  
Done.

**Related PRs:**   
- https://github.com/mongodb/genny/pull/1081
- https://github.com/mongodb/genny/pull/1080

PR #1081, opened 2023-12-08
SERVER-83895 Add benchmarks with more complex CNF, DNF and mixed predicates with multiple levels of nestedness

henrinikku:

**Jira Ticket:** SERVER-83895

**Whats Changed:**  
Added benchmarks that execute queries with complex CNF, DNF and mixed predicates with multiple levels of nestedness on collections of different sizes (100, 10K and 1M documents).

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/65772e26d1fe077537b625da/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Workload Submission form:**  
Done.

**Related PRs:**   
- https://github.com/mongodb/genny/pull/1080
- https://github.com/mongodb/genny/pull/1079


miguelangelnieto (approved):

src/phases/query/CollScanComplexPredicate.yml
```diff
@@ -0,0 +1,521 @@
...
+    "e.h", 
+    "e.i",
+    "nonexistent", 
+    "nullVal"
+  ]
+  QueryOps: &QueryOps ["$eq", "$lt", "$lte", "$gt", "$gte"]
+  
+  # Constructs an $and query with N $or clauses, where each $or contains 1-10 randomly generated
+  # comparisons.
+  CNFQuery: &CNFQuery
```
alyacb, line 39: This could prove problematic for anyone investigating a regression in this workload, since it would be very difficult to derive the query being run here directly by looking at it (and consequently, very difficult to repro locally independent of genny). I see two options here.
1) obtain the query being generated once locally, and hard-code it directly in this file instead.
2) log the explain for this query on the command-line. there should be an actor that lets you do this (LoggingActor or similar). the downside is we will be collecting metrics for something we don't care about.

My vote would be for 1).

src/phases/query/CollScanComplexPredicate.yml
```diff
@@ -0,0 +1,521 @@
...
+                                {$and: [{f: {$gt: 0}}, {f: {$lt: 100}}, {p: 99}]},
+                                {$or: [{f: {$gt: 90}}, {f: {$lt: 200}}, {q: 99}]},
+                                {$and: [{f: {$gt: 190}}, {f: {$lt: 300}}, {r: 99}]}
+                              ]}
+                            ]}
+                          ]}
+                        ]},
+                        {$and: [{h: {$gt: 90}}, {h: {$lt: 200}}, {b: 99}]},
+                        {$or: [
+                          {$and: [{a: {$lt: 11}}, {c: {$gt: 12}}, {f: {$in: ["1", "2", 1, 2]}}, {d: null}]},
```
alyacb, line 205: I think Genny might still have a bug where the literal "1" is turned into the number 1 directly. I'd suggest using a different string. Speaking of strings- can we make one of the fields be a string field? Could have one "shortString" field and one "longString" field, since the collscan benchmarks doc mentioned that we don't have enough coverage on the string type.

src/phases/query/CollScanComplexPredicate.yml
```diff
@@ -0,0 +1,521 @@
...
+            b: {^Inc: {start: 11, step: 2}}
+            c: {^RandomInt: {min: 0, max: 1000}}
+            d: {^RandomInt: {min: 0, max: 100}}
+            e: {
+              a: {^RandomInt: {min: 0, max: 10}},
+              b: {^Inc: {start: 11, step: 2}},
+              c: {^RandomInt: {min: 0, max: 1000}},
+              d: {^RandomInt: {min: 0, max: 100}},
+              e: {a: {^RandomInt: {min: 0, max: 1000}}, b: {^Inc: {start: 11, step: 2}}},
+              f: [{^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}],
```
alyacb, line 367: Any chance we could make one of these arrays have variable length?

src/phases/query/CollScanComplexPredicate.yml
```diff
@@ -0,0 +1,521 @@
...
+              g: {^RandomInt: {min: 0, max: 10}},
+              h: {^RandomInt: {min: 0, max: 1000}},
+              i: {^RandomInt: {min: 0, max: 10000}},
+            }
+            f: [{^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}]
+            g: {^RandomInt: {min: 0, max: 10}}
+            h: {^RandomInt: {min: 0, max: 1000}}
+            i: {^RandomInt: {min: 0, max: 10000}}
+          f: [{^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}]
```
alyacb, line 376: Could also have one of these fields include an array of nested docs with fields that would be matched on by the query above, just to exercise the speed of path traversals, e.g. [{a: 1, ...}, [[a: 2]] ...]

src/phases/query/CollScanComplexPredicate.yml
```diff
@@ -0,0 +1,521 @@
...
+              c: {^RandomInt: {min: 0, max: 1000}},
+              d: {^RandomInt: {min: 0, max: 100}},
+              e: {a: {^RandomInt: {min: 0, max: 1000}}, b: {^Inc: {start: 11, step: 2}}},
+              f: [{^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}],
+              g: {^RandomInt: {min: 0, max: 10}},
+              h: {^RandomInt: {min: 0, max: 1000}},
+              i: {^RandomInt: {min: 0, max: 10000}},
+            }
+            f: [{^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}, {^RandomInt: {min: 0, max: 10}}]
+            g: {^RandomInt: {min: 0, max: 10}}
```
alyacb, line 373: Request for a at least one of these fields to be made into a double.

alyacb (changes requested):
Thanks Henri! Left some initial comments- note that when I suggest adding a different type of field, it would be good to check that we have predicates that include that type as well.

I would also note that it looks like some of the queries here do match on a few documents (see the DocumentsTotal metric in TrendCharts). This is fine, but I wanted to make sure that it is intentional.

src/phases/query/CollScanComplexPredicate.yml
henrinikku: I implemented 1) and hardcoded the queries in a separate file `CollScanComplexPredicateQueries.yml`.

henrinikku (commented):
Thanks for the reviews! I think I addressed all of Alya's comments. @miguelangelnieto does the way large queries are hard coded and loaded seem ok to you?

alyacb (approved):
LGTM! Looks like what we discussed offline.

miguelangelnieto (approved):

henrinikku:
evergreen merge

PR #1082, opened 2023-12-11
PERF-4969 Add TS task with compound $and and $or

albymassari:
Apart for adding a new test where we have an explicit $and of $or using different comparisons (because the other $or inside an $and in the test MatchOnMultipleNestedFields is being converted into a $in given that is an equality of two constants), I also moved the active phases to avoid running two tests in the same phase


albymassari:
evergreen merge

PR #1083, opened 2023-12-12
Devprod 1626

lgfang:
**Jira Ticket:** DEVPROD-1626

**Whats Changed:**

-   Download and the genny toolchain which merged the upstream tag '2023.11.20'.

-   Modify cmake files to reduce noises

-   Remove 'factory.makePool()' from TEST\_CASE("PoolFactory behavior") for CXX-2705.
    
    We were making a pool using a bogus URI, which is expected to encounter an SRV lookup error. Previously it was not an issue because such errors were ignored on pool construction (but would result in a crash when using the pool). In MongoCxx 3.9.0, the test fails because an exception is thrown on this error due to CXX-2705.

**Patch testing results:**

TODO

**Related PRs:**

-   [Corresponding genny toolchain PR](https://github.com/10gen/vcpkg/pull/32)


thessem (changes requested):
This is going to potentially be a change with some side-effects in Genny, so therefore I want to make sure this is a welll formed and a well-described commit. I would like to see the following in this PR before moving it forwards.
1. A descriptive title with the big-ticket upgrades mentioned (mongoc, mongocpp etc)
2. A descriptive body with all the deps being changed due to the vcpkg update
3. A brief analysis of potential perf effects this may cause.

thessem (dismissed):
I am a bit worried about the 30000% perf change on some of the perf results here. If this is a result of the C++ driver updates making it do something clever, we need to determine whether we've invalidated the test. Alternatively, we need to validate that we're still actually testing something and not NOPing. We should ask Product Perf about this.

lgfang:
> I am a bit worried about the 30000% perf change on some of the perf results here. If this is a result of the C++ driver updates making it do something clever, we need to determine whether we've invalidated the test. Alternatively, we need to validate that we're still actually testing something and not NOPing. We should ask Product Perf about this.

As per Simon, this actually confirms that a known issue is fixed by updating the toolchain. (see
https://mongodb.slack.com/archives/C0606DZASJV/p1702614668402969?thread_ts=1702610409.667489&cid=C0606DZASJV ).

PyGeek03 (approved):
LGTM!

lgfang:
evergreen merge

PR #1101, opened 2024-01-08
SERVER-84400 Add large projection workloads

RuoxinXu:

**Jira Ticket:** SERVER-84400

**Whats Changed:**  
Add a workload runs queries with a large projection.

**Patch testing results:**  
https://evergreen.mongodb.com/task/sys_perf_linux_intel_standalone_heuristic_bonsai_large_projections_patch_868736311bb45191097442b4fd545fa0e0cb9909_659bd9cd850e61bedb674d51_24_01_08_11_22_27


src/workloads/query/LargeProjections.yml
```diff
@@ -0,0 +1,204 @@
...
+    mongodb_setup:
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+    branch_name:
+      $neq:
```
alyacb, line 199: I thought there was a $gte operator? Could do $gte: 7.0 if that's the case

src/workloads/query/LargeProjections.yml
```diff
@@ -0,0 +1,204 @@
...
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document: {
```
alyacb, line 54: It might be more valuable from a testing perspective if some of these fields had different types/were randomized. There are some examples in other workloads, e.g. [here](https://github.com/mongodb/genny/blob/12953ac9a76cea4de4becdddbcc8d8da4ea887cb/src/phases/query/CollScanComplexPredicate.yml#L279).

src/workloads/query/LargeProjections.yml
```diff
@@ -0,0 +1,204 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *Repeat
+        Collection: *Collection
+        Operations:
+        - OperationName: find
+          OperationCommand:
+            Filter: {}
+            Options:
+              Projection: {c: 1, d: 1, e: 1, f: 1, g: 1, h: 1, i: 1, j: 1, k: 1, l: 1, m: 1, n: 1, o: 1, p: 1, q: 1, r: 1, s: 1, t: 1, u: 1, v: 1}
```
alyacb, line 96: Can we have some large exclusion queries as well? Also for the nested case.

src/workloads/query/LargeProjections.yml
```diff
@@ -0,0 +1,204 @@
...
+- Loader
+- CrudActor
+- QuiesceActor
+- insert
+- find
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  DocumentCount: &DocumentCount 1e6
```
alyacb, line 16: I think we should add a small/medium version of this workload as well. Seeing when a regression only affects a larger workload vs when it only affects a smaller workload (or how much it affects each collection size) can give you a hint about whether its execution or optimization time that is problematic.

alyacb (changes requested):
Thanks Roy! I have a few requests.

Do all of these queries run with tryBonsai? (We should keep them even if they don't, but I wanted to check since we should only tag the ones that definitely do as optimizer_m2 for the dashboard).

src/phases/query/CollScanLargeProjection.yml
```diff
@@ -0,0 +1,243 @@
...
+        MultipleThreadsPerCollection: true
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document: {
+          a: {
+            b: {
+              c: {
+                d: {
+                  e: [{a: {^RandomInt: {min: 0, max: 1000}}, b: {^Inc: {start: 11, step: 2}}}],
```
RuoxinXu, line 106: This part of leaf fields is a big dup code. Tried to dedup it with `GlobalDefults` (Please check the 3rd commit) but I failed to run the workload. Not sure if there's a better approach to dedup the code.

RuoxinXu:
Thanks Alya for the comments!  I pushed a few commits that run the workload in different data set size. These queries don't always run with 'tryBonsai'. They also run with `mongod_setup` like `standalone-sbe` that doesn't specify the knob.

src/phases/query/CollScanLargeProjection.yml
alyacb, line 1: nit: can we just name this CollScanProjection.yml?

src/phases/query/CollScanLargeProjection.yml
alyacb: I'm ok with this being duplicated for now, but if you wanted to factor it out I think you could create yet another phases yml and import it here, similar to how this [query](https://github.com/mongodb/genny/commit/091adc58b38bbe099bf1697076491e55685cccab#diff-48c7368ab151bb9b136bbac855dcca7e723fc2d481343154e9db3f33c4a96309R75) is defined in one phase yml and imported in [another](https://github.com/mongodb/genny/commit/091adc58b38bbe099bf1697076491e55685cccab#diff-90994296f65c6485553a8e1f170e1fa3330b7035a66f84c1413e280e3dd09afcR23-R26) then referenced [here](https://github.com/mongodb/genny/commit/091adc58b38bbe099bf1697076491e55685cccab#diff-90994296f65c6485553a8e1f170e1fa3330b7035a66f84c1413e280e3dd09afcR351). Might facilitate writing new queries.

alyacb (approved):
LGTM! Thanks Roy! 🎉

RuoxinXu:
evergreen merge

PR #1104, opened 2024-01-08
SERVER-83901 Add conjunctive queries having predicates with different selectivities

henrinikku:

**Jira Ticket:** SERVER-83901

**Whats Changed:**  
Add conjunctive queries, where predicate order matters a lot due to different selectivity of the predicates. This will test the ability of the optimizer to order the predicates according to their selectivity.

**Patch testing results:**  
https://spruce.mongodb.com/version/659c3d9a32f4170682c7baf4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

**Workload Submission form:**  
Submitted

### DIFF
--- a/src/phases/query/CollScanPredicateSelectivity.yml
+++ b/src/phases/query/CollScanPredicateSelectivity.yml
@@ -7,13 +7,19 @@ Description: |
   Similarly, an 'indistinguishable' case means that heuristic CE defines the same selectivity for all predicates
   in the query. For this workload, average latency is the most important metric to look at.
 
+Keywords:
+- Loader
+- CrudActor
+- QuiesceActor
+- insert
+- find
 
 GlobalDefaults:
   Database: &Database {^Parameter: {Name: Database, Default: unused}}
   DocumentCount: &DocumentCount {^Parameter: {Name: DocumentCount, Default: -1}}
   Repeat: &Repeat {^Parameter: {Name: Repeat, Default: -1}}
   Collection: &Collection Collection0
-  MaxPhases: &MaxPhases 32
+  MaxPhases: &MaxPhases 52
   Billion: &Billion 1_000_000_000
   NegativeBillion: &NegativeBillion -1_000_000_000
   Document: &Document {
--- a/src/phases/query/CollScanPredicateSelectivity.yml
+++ b/src/phases/query/CollScanPredicateSelectivity.yml
@@ -188,7 +188,7 @@ Actors:
       Filter:
         &HighSelectivityGoodCaseFourClauses {
           onePercentConstantOne: 1, # Selectivity = .01
-          quarterPercentPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .25
+          quarterPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .25
           quarterNegativeInt: {$gt: 0}, # Selectivity = .75
           tenPercentConstantOne: {$ne: 1}, # Selectivity = .9
         }
--- a/src/phases/query/CollScanPredicateSelectivity.yml
+++ b/src/phases/query/CollScanPredicateSelectivity.yml
@@ -64,9 +64,9 @@ Actors:
         BatchSize: 1000
         Document:
           {
-            onePercentConstantOne: {^Choose: {from: ["not int", 1], weights: [99, 1]}},
-            tenPercentConstantOne: {^Choose: {from: ["not int", 1], weights: [90, 10]}},
-            quarterConstantOne: {^Choose: {from: ["not int", 1], weights: [75, 25]}},
+            onePercentNull: {^Choose: {from: ["non-null", null], weights: [99, 1]}},
+            tenPercentNull: {^Choose: {from: ["non-null", null], weights: [90, 10]}},
+            quarterNull: {^Choose: {from: ["non-null", null], weights: [75, 25]}},
             onePercentPositiveInt: {^Choose: {
               from: &PositiveAndNegativeIntegers [{^Inc: {start: 1, step: 1}}, {^Inc: {start: -1, step: -1}}], 
               weights: [1, 99]
@@ -79,9 +79,9 @@ Actors:
             onePercentNegativeInt: {^Choose: {from: *PositiveAndNegativeIntegers, weights: [99, 1]}},
             nested: [
               {
-                onePercentConstantOne: {^Choose: {from: ["not int", 1], weights: [99, 1]}},
-                tenPercentConstantOne: {^Choose: {from: ["not int", 1], weights: [90, 10]}},
-                quarterConstantOne: {^Choose: {from: ["not int", 1], weights: [75, 25]}},
+                onePercentNull: {^Choose: {from: ["non-null", null], weights: [99, 1]}},
+                tenPercentNull: {^Choose: {from: ["non-null", null], weights: [90, 10]}},
+                quarterNull: {^Choose: {from: ["non-null", null], weights: [75, 25]}},
                 onePercentPositiveInt: {^Choose: {from: *PositiveAndNegativeIntegers, weights: [1, 99]}},
                 tenPercentPositiveInt: {^Choose: {from: *PositiveAndNegativeIntegers, weights: [10, 90]}},
                 quarterPositiveInt: {^Choose: {from: *PositiveAndNegativeIntegers, weights: [25, 75]}},
@@ -113,12 +113,12 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: HighSelectivityGoodCaseTwoClauses
+      Name: HighSelHeuristicGoodTwoClauses
       ActivePhase: 3
       Filter:
         # Here (0, 1e9) is the most selective predicate, both in reality and according to heuristic CE.
         # Note that without predicate reordering, predicates are ordered by first comparing field names.
-        &HighSelectivityGoodCaseTwoClauses {
+        &HighSelHeuristicGoodTwoClauses {
           onePercentPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .01
           onePercentNegativeInt: {$gt: 0}, # Selectivity = .99
         }
@@ -126,10 +126,10 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: MediumSelectivityGoodCaseTwoClauses
+      Name: MediumSelHeuristicGoodTwoClauses
       ActivePhase: 4
       Filter:
-        &MediumSelectivityGoodCaseTwoClauses {
+        &MediumSelHeuristicGoodTwoClauses {
           tenPercentPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .1
           tenPercentNegativeInt: {$gt: 0}, # Selectivity = .9
         }
@@ -137,10 +137,10 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: LowSelectivityGoodCaseTwoClauses
+      Name: LowSelHeuristicGoodTwoClauses
       ActivePhase: 5
       Filter:
-        &LowSelectivityGoodCaseTwoClauses {
+        &LowSelHeuristicGoodTwoClauses {
           quarterPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .25
           quarterNegativeInt: {$gt: 0}, # Selectivity = .75
         }
@@ -148,12 +148,12 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: HighSelectivityBadCaseTwoClauses
+      Name: HighSelHeuristicBadTwoClauses
       ActivePhase: 6
       Filter:
-        # Here (-1e9, 0) is the most selective predicate according heuristic CE. However, (-inf, 0)
+        # Here (-1e9, 0) is the most selective predicate according to heuristic CE. However, (-inf, 0)
         # is more selective in reality.
-        &HighSelectivityBadCaseTwoClauses {
+        &HighSelHeuristicBadTwoClauses {
           onePercentPositiveInt: {$gt: *NegativeBillion, $lt: 0}, # Selectivity = .99
           onePercentNegativeInt: {$lt: 0}, # Selectivity = .01
         }
@@ -161,10 +161,10 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: MediumSelectivityBadCaseTwoClauses
+      Name: MediumSelHeuristicBadTwoClauses
       ActivePhase: 7
       Filter:
-        &MediumSelectivityBadCaseTwoClauses {
+        &MediumSelHeuristicBadTwoClauses {
           tenPercentPositiveInt: {$gt: *NegativeBillion, $lt: 0}, # Selectivity = .9
           tenPercentNegativeInt: {$lt: 0}, # Selectivity = .1
         }
@@ -172,10 +172,10 @@ Actors:
 - ActorFromTemplate:
     TemplateName: FindQueryTemplate
     TemplateParameters:
-      Name: LowSelectivityBadCaseTwoClauses
+      Name: LowSelHeuristicBadTwoClauses
       ActivePhase: 8
       Filter:
-        &LowSelectivityBadCaseTwoClauses {
+        &LowSelHeuristicBadTwoClauses {
           quarterPositiveInt: {$gt: *NegativeBillion, $lt: 0}, # Selectivity = .75
           quarterNegativeInt: {$lt: 0}, # Selectivity = .25
         }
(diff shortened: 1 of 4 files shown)


src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+          tenPercentNegativeInt: {$gt: 0}, # Selectivity = .9
+        }
+
+- ActorFromTemplate:
+    TemplateName: FindQueryTemplate
+    TemplateParameters:
+      Name: LowSelectivityGoodCaseTwoClauses
+      ActivePhase: 5
+      Filter:
+        &LowSelectivityGoodCaseTwoClauses {
```
alyacb, line 143: I hate making the names even longer, but... could we do something like:
- GoodCase -> HeuristicGood
- Selectivity -> Sel

I think this would help when reading dashboard metrics

alyacb (approved):
LGTM mod comments/requests for mixed conjunctive cases (we don't have to be exhaustive, but lets have a few)

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+      Active: [1]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: 1
+        Database: *Database
+        MultipleThreadsPerCollection: true
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document:
```
jimoleary, line 65: Would it make sense to pull the document definition into the GlobalDefaults?

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+          {
+            onePercentNull: {^Choose: {from: ["non-null", null], weights: [99, 1]}},
```
jimoleary, line 67: question: I *think* the yml spec should result in this being the value null rather than the string null. Can you confirm this is the case (assuming it is significant)?

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "ActivePhase", Default: -1}}]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: *Repeat
+          Collection: *Collection
+          Operations:
+          - OperationName: find
+            OperationCommand:
+              Filter: {^Parameter: {Name: "Filter", Default: {invalid: "invalid"}}}
```
jimoleary, line 33: What happens if the default filter is ever used? I looks like it will cause an unindexed collection scan ... the results may not be different enough from a low selectivity scan to be caught.

I'm not sure how to deal with this ... it may be a genny feature request (e.g. make a parameter mandatory and fail  if there isn't one) or double checking if this is the current behaviour. I'm pretty sure its not.

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of conjunctive collection scan queries where the order of
```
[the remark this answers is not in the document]
jimoleary: What are the important metrics? Can you add this information to the description?

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+# $eq < closed range (e.g. {$gt: 10, $lte: 20}) < open range (e.g. {$gt: 10}) < $ne
+
+- ActorFromTemplate:
+    TemplateName: FindQueryTemplate
+    TemplateParameters:
+      Name: HighSelHeuristicGoodTwoClauses
+      ActivePhase: 3
+      Filter:
+        # Here (0, 1e9) is the most selective predicate, both in reality and according to heuristic CE.
+        # Note that without predicate reordering, predicates are ordered by first comparing field names.
```
jimoleary, line 120: for my own benefit does 'predicates are ordered by first comparing field names' mean that that onePercentNegativeInt would be applied first?

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
...
+          onePercentNull: "non-null", # Selectivity = .99
+          quarterPositiveInt: {$gt: *NegativeBillion, $lt: 0}, # Selectivity = .75
+          halfNegativeInt: {$lt: 0}, # Selectivity = .5
+          quarterNull: {$ne: "non-null"}, # Selectivity = .25
+        }
+
+- ActorFromTemplate:
+    TemplateName: FindQueryTemplate
+    TemplateParameters:
+      Name: HighSelIndistinguishableByHeuristics
```
jimoleary, line 264: can you explain / introduce what you mean by Indistinguishable somewhere?

src/workloads/query/CollScanPredicateSelectivityLarge.yml
```diff
@@ -0,0 +1,26 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of conjunctive collection scan queries where the order of
+  predicates matters due to selectivity of the predicates.
+
+LoadConfig:
+  Path: ../../phases/query/CollScanPredicateSelectivity.yml
+  Parameters:
+    Database: CollScanPredicateSelectivityLarge
```
jimoleary, line 10: Do you really care about the database names in the different workload versions? (Or is this only to speed up you own local testing)

src/workloads/query/CollScanPredicateSelectivityMedium.yml
```diff
@@ -0,0 +1,26 @@
...
+Description: |
+  This workload tests the performance of conjunctive collection scan queries where the order of
+  predicates matters due to selectivity of the predicates.
+
+LoadConfig:
+  Path: ../../phases/query/CollScanPredicateSelectivity.yml
+  Parameters:
+    Database: CollScanPredicateSelectivityMedium
+    DocumentCount: 1e4
+    Repeat: 100
```
jimoleary, line 12: Is the repeat significant in terms of a constant _DocumenCount_ * _Repeat_ between the workloads? And if so then should it be 1_000 (workload time permitting).

jimoleary (commented):
The workloads themselves look good to me, most of my comments relate to futher explanation / missing sections. 

How noisy are the tests?

src/phases/query/CollScanPredicateSelectivity.yml
alyacb: Could just set the filter to the value null by default, which would crash the query?

jimoleary: We could use null assuming that it gives a reasonable error in the task. 

Note: I'm not proposing holding up the review.... If it's a feature request we can just add a TODO to come back to.

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,427 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  This workload tests the performance of conjunctive collection scan queries where the order of
+  predicates matters due to selectivity of the predicates.
+
```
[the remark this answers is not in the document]
henrinikku: Done!

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,688 @@
...
+- ActorFromTemplate:
+    TemplateName: FindQueryTemplate
+    TemplateParameters:
+      Name: HighSelHeuristicGoodTwoEqClauses
+      ActivePhase: 4
+      Filter:
+        &HighSelHeuristicGoodTwoEqClauses {
+          onePercentNull: null, # Selectivity = .01
+          onePercentStr: {$ne: "non-null"}, # Selectivity = .99
+        }
```
henrinikku, line 148: @alyacb I added another version of each two-predicate query. The reordering PR will probably reorder the $eq/$ne queries, but as we discussed briefly offline, it might not reorder e.g. `HighSelHeuristicGoodTwoRangeClauses` because `{$gt: 0, $lt: *Billion}` is seen as a conjunction of open-range intervals instead of one closed-range interval.

henrinikku (commented):
Thanks for the comments, I think I've addressed everything now.

src/phases/query/CollScanPredicateSelectivity.yml
jimoleary: thanks.

src/workloads/query/CollScanPredicateSelectivityLarge.yml
jimoleary: Thats seems reasonable.

src/phases/query/CollScanPredicateSelectivity.yml
jimoleary: thanks for checking!

src/phases/query/CollScanPredicateSelectivity.yml
jimoleary: That works for me, I've also raised DEVPROD-3649 to request mandatory parameter support.

jimoleary: That works for me, I've also raised DEVPROD-3649 to request mandatory parameter support.

src/workloads/query/CollScanPredicateSelectivitySmall.yml
```diff
@@ -0,0 +1,26 @@
...
+Description: |
+  This workload tests the performance of conjunctive collection scan queries where the order of
+  predicates matters due to selectivity of the predicates.
+
+LoadConfig:
+  Path: ../../phases/query/CollScanPredicateSelectivity.yml
+  Parameters:
+    Database: CollScanPredicateSelectivitySmall
+    DocumentCount: 100
+    Repeat: 100
```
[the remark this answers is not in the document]
jimoleary: yeah 100K repeats does seem a lot!

However, that being said, the current [small versions phases](https://evergreen.mongodb.com/task_log_raw/sys_perf_linux_standalone.2022_11_coll_scan_predicate_selectivity_small_patch_1c2ce4380e3e05ddc1e843db811ef95f1cd6aa8d_659c3d9a32f4170682c7baf4_24_01_08_18_29_45/0?type=T#L1445) take less than half a second each so a larger number would be better.

jimoleary (commented):
Thanks for the updates, the workload files look good to me.

But .. the large version of this workload takes about 7 minutes (for phase 3 - 31) while the medium version takes 50 seconds and the small version takes 8 seconds (for phase 3 - 31).

I'd be concerned that the shorter versions might be noisy.

Could you adjust the repeats to bring these versions up to at least a couple of minutes overall runtime and / or use the multipatch tool to figure out how noisy the rests are?

You can find the multipatch tool [here](https://github.com/10gen/product-perf-experimentations/tree/master/applications/multipatch) and the noise analysis notebook [here](https://github.com/10gen/product-perf-experimentations/blob/master/applications/multipatch/NoiseAnalysis.ipynb).

henrinikku:
I now see I forgot to address this question yesterday:
> How noisy are the tests?

I don't know yet. I got pretty stable results from the 100 docs version locally. I started three patches with the latest changes, and I'll compare the results once they finish:
- [patch 1](https://spruce.mongodb.com/version/659d8751c9ec446ed3708711/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [patch 2](https://spruce.mongodb.com/version/659d875b61837d71df9b56d7/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [patch 3](https://spruce.mongodb.com/version/659d8769850e61719765dfec/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

jimoleary:
> > How noisy are the tests?
> 
> I don't know yet. I got pretty stable results from the 100 docs version locally. I started three patches with the latest changes, and I'll compare the results once they finish:
> 
> * [patch 1](https://spruce.mongodb.com/version/659d8751c9ec446ed3708711/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
> * [patch 2](https://spruce.mongodb.com/version/659d875b61837d71df9b56d7/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
> * [patch 3](https://spruce.mongodb.com/version/659d8769850e61719765dfec/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

Thanks, have a look at the multipatch next time, it's just a fairly simple wrapper around the evergreen cli and it helps automate the checks.

jimoleary:
Sorry, fat fingered the wrong button!

src/phases/query/CollScanPredicateSelectivity.yml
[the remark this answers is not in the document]
alyacb: OperationThroughput would probably be the most important, but we look at all the latency measurements as well. They will become increasingly important once plan-cache is enabled, since maxlatency/latency99thpercentile will probably be the only place where optimization time (without hitting the plan cache) is tracked until the optimizer metrics project (and relevant perf-infra work) completes.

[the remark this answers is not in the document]
henrinikku: Good points, I didn't take plan cache into account. I updated the description.

src/phases/query/CollScanPredicateSelectivity.yml
alyacb: Good point, but I have a follow-up question- do we know if SBE/classic use the original order in any of these cases?

src/phases/query/CollScanPredicateSelectivity.yml
```diff
@@ -0,0 +1,688 @@
...
+      Filter:
+        &MediumSelHeuristicGoodTwoRangeClauses {
+          tenPercentPositiveInt: {$gt: 0, $lt: *Billion}, # Selectivity = .1
+          tenPercentNegativeInt: {$gt: 0}, # Selectivity = .9
+        }
+
+- ActorFromTemplate:
+    TemplateName: FindQueryTemplate
+    TemplateParameters:
+      Name: MediumSelHeuristicGoodTwoEqClauses
```
alyacb, line 164: I think some of these names might be misleading, since $ne for e.g. is not an equality predicate. Could say EqNe instead or similar.

alyacb (approved):
LGTM, mod one overarching concern, which is type-bracketing. An open-range predicate is never really translated to a single open-range (e.g. I think {$gt: 0} actually because something like {$gt: 0, $lt: <whatever-the-max-int-type-is>).

I think this is ok because we have no way outside of manually constructing an ABT to force this sort of case, and $eq/closed range intervals are sufficiently differently estimated by heuristics that these test cases should remain useful, but it may be worth an additional note.

henrinikku (commented):
Thanks again for the comments! I think I responded to everything.

@jimoleary I ran the noise analysis notebook for the patches ([link](https://github.com/henrinikku/product-perf-experimentations/blob/10e78dd8e2450527c1653e36587919ef09cb6cfa/applications/multipatch/NoiseAnalysis.ipynb)). The `medium` and `large` workloads seem to be the noisiest ones. I haven't looked at variability numbers of tests before so I can't say how much is really a lot, but relative range of ~0.5 for some queries in the `large` workload doesn't seem great. Do you think it would help to repeat the larger workloads even more times? Both medium and large took around 20min to complete with the latest changes.


> But .. the large version of this workload takes about 7 minutes (for phase 3 - 31) while the medium version takes 50 seconds and the small version takes 8 seconds (for phase 3 - 31).
> I'd be concerned that the shorter versions might be noisy.

I increased repeat on the small workload to 20K. It should take around 12min now, which is the same order of magnitude as the other sizes.

> mod one overarching concern, which is type-bracketing. An open-range predicate is never really translated to a single open-range (e.g. I think {$gt: 0} actually because something like {$gt: 0, $lt: ).

Good point! I recall we used some some heuristic like comparing types in the heuristic CE to detect type bracketed comparisons. The min/max bounds for numeric types are NaN (double) and "" (string), which are both different from `NumberInt32`/`NumberInt64`, so I think the heuristic works correctly for integer comparisons.

jimoleary:
> Thanks again for the comments! I think I responded to everything.
> 
> @jimoleary I ran the noise analysis notebook for the patches ([link](https://github.com/henrinikku/product-perf-experimentations/blob/10e78dd8e2450527c1653e36587919ef09cb6cfa/applications/multipatch/NoiseAnalysis.ipynb)). The `medium` and `large` workloads seem to be the noisiest ones. I haven't looked at variability numbers of tests before so I can't say how much is really a lot, but relative range of ~0.5 for some queries in the `large` workload doesn't seem great. Do you think it would help to repeat the larger workloads even more times? Both medium and large took around 20min to complete with the latest changes.
> 
> > But .. the large version of this workload takes about 7 minutes (for phase 3 - 31) while the medium version takes 50 seconds and the small version takes 8 seconds (for phase 3 - 31).
> > I'd be concerned that the shorter versions might be noisy.
> 
> I increased repeat on the small workload to 20K. It should take around 12min now, which is the same order of magnitude as the other sizes.
> 

The noise level looks quite bad for these tests, when the noise level is above 10% it's difficult to triage any changes points.  More test runs might be useful but I'd be surprised if there was any change in the noise levels.

Do you have any idea where the noise maybe coming from or if it is likely that the noise levels could be lowered in the futue?

It's still possible to deploy the test as is .... but we would need to tell the buildbarons to ignore any change points until the noise issue is resolved.

henrinikku:
> Do you have any idea where the noise maybe coming from or if it is likely that the noise levels could be lowered in the futue?

Not really, the only explanation that comes to mind right now is the execution order of predicates being non-deterministic, which would be really odd/surprising.

I started some patches against the latest master just to double check the results. If the results seem equally noisy, I'll try to investigate it further.
- [patch #1](https://spruce.mongodb.com/version/65a029480305b943d5a044bd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [patch #2](https://spruce.mongodb.com/version/65a0295d3e8e864661e7d321/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [patch #3](https://spruce.mongodb.com/version/65a029673066159a235d8026/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- [patch #4](https://spruce.mongodb.com/version/65a029739ccd4e6cef018aa0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

henrinikku:
@jimoleary The new patches I started seem equally noisy ([analysis notebook](https://github.com/henrinikku/product-perf-experimentations/blob/915d2c5ef579630d446b9bd3c46173554766e56d/applications/multipatch/NoiseAnalysis.ipynb)). I also did similar measurements locally and got similarly noisy results. We talked about this in a team standup and decided that we don't want to merge these until the noisiness is resolved, so I'll keep on trying to find some explanation for the noise.

What seems odd to me is that while there is always a handful of really noisy queries (30-39% variance), those queries can be completely different ones if anything changes in the setup (e.g. build flag, checkout to a different recent commit etc). However if nothing changes, those queries seem consistently noisy even when run in isolation. Have you previously seen similar issues in any other benchmarks?

jimoleary:
> @jimoleary The new patches I started seem equally noisy ([analysis notebook](https://github.com/henrinikku/product-perf-experimentations/blob/915d2c5ef579630d446b9bd3c46173554766e56d/applications/multipatch/NoiseAnalysis.ipynb)). I also did similar measurements locally and got similarly noisy results. We talked about this in a team standup and decided that we don't want to merge these until the noisiness is resolved, so I'll keep on trying to find some explanation for the noise.
> 
> What seems odd to me is that while there is always a handful of really noisy queries (30-39% variance), those queries can be completely different ones if anything changes in the setup (e.g. build flag, checkout to a different recent commit etc). However if nothing changes, those queries seem consistently noisy even when run in isolation. Have you previously seen similar issues in any other benchmarks?

@henrinikku An alternative to holding off on merging the workload is to create a server ticket to deal with the noise later, inform the buildbarons to ignore these test results for now and merge the workload in the meantime.

Tests can be noisey for a number of reasons (especially if there are no backing indexes). However, your workload is running for long enought that I would have expected much of the randomness to have averaged out.

Have you looked at the patch data to see if the noise is coming from the server or the client?

henrinikku:
I did some more experimenting, and at least when running on my virtual machine, the less selective queries seem to be by far the noisiest ones. I got the below numbers locally after [changing](https://github.com/mongodb/genny/pull/1104/commits/d97fb40ddc9443991dc2eb108f5353b76d23d74e) the low / medium selectivity queries to be much more selective (from 25% / 10% to 10% / 5% respectively). Two queries had around 10% noise and others had way less, which seems like a nice improvement to me (many queries used to have >30%, even when running locally). The downside of this is of course that improvements achieved by predicate reordering may not be as noticeable.

<details closed>
<summary>Relative range (n=4)</summary>

```
ElemMatchLowSelHeuristicBadEqNe.find 0.1023
ElemMatchLowSelHeuristicGoodEqNe.find 0.1006
ElemMatchLowSelHeuristicGoodFourClauses.find 0.08220
LowSelHeuristicBadEqNe.find 0.07976
LowSelHeuristicGoodEqNe.find 0.07879
MediumSelHeuristicBadEqNe.find 0.07742
ElemMatchMediumSelHeuristicGoodEqNe.find 0.07002
LowSelHeuristicGoodFourClauses.find 0.0654
ElemMatchLowSelHeuristicBadTwoRangeClauses.find 0.06449
ElemMatchMediumSelHeuristicBadEqNe.find 0.06417
ElemMatchLowSelHeuristicGoodTwoRangeClauses.find 0.05706
LowSelHeuristicGoodTwoRangeClauses.find 0.05596
MediumSelHeuristicGoodEqNe.find 0.05137
LowSelHeuristicBadTwoRangeClauses.find 0.05137
ElemMatchMediumSelHeuristicGoodTwoRangeClauses.find 0.04681
ElemMatchMediumSelHeuristicBadTwoRangeClauses.find 0.04643
MediumSelHeuristicGoodTwoRangeClauses.find 0.03767
ElemMatchLowSelIndistinguishableByHeuristics.find 0.03726
MediumSelHeuristicBadTwoRangeClauses.find 0.03657
HighSelHeuristicGoodEqNe.find 0.03543
LowSelIndistinguishableByHeuristics.find 0.03294
HighSelHeuristicBadEqNe.find 0.02944
HighSelHeuristicBadTwoRangeClauses.find 0.02889
ElemMatchHighSelHeuristicBadEqNe.find 0.02843
ElemMatchHighSelHeuristicBadFourClauses.find 0.02751
ElemMatchDNFHeuristicBad.find 0.02327
ElemMatchHighSelHeuristicGoodEqNe.find 0.02287
DNFHeuristicBad.find 0.02263
DNFHeuristicGood.find 0.02197
ElemMatchDNFHeuristicGood.find 0.02118
MediumSelHeuristicGoodFourClauses.find 0.01945
ElemMatchMediumSelHeuristicGoodFourClauses.find 0.01940
ElemMatchMediumSelHeuristicBadFourClauses.find 0.01902
ElemMatchHighSelHeuristicBadTwoRangeClauses.find 0.01685
ElemMatchHighSelHeuristicGoodFourClauses.find 0.0166
MediumSelHeuristicBadFourClauses.find 0.01598
HighSelHeuristicGoodTwoRangeClauses.find 0.01593
CNFHeuristicGood.find 0.01352
MediumSelIndistinguishableByHeuristics.find 0.01344
ElemMatchHighSelHeuristicGoodTwoRangeClauses.find 0.01342
CNFHeuristicBad.find 0.01230
ElemMatchLowSelHeuristicBadFourClauses.find 0.01157
LowSelHeuristicBadFourClauses.find 0.01003
HighSelHeuristicBadFourClauses.find 0.009979
HighSelHeuristicGoodFourClauses.find 0.009255
ElemMatchMediumSelIndistinguishableByHeuristics.find 0.008589
HighSelIndistinguishableByHeuristics.find 0.006156
ElemMatchCNFHeuristicGood.find 0.005425
ElemMatchCNFHeuristicBad.find 0.004779
ElemMatchHighSelIndistinguishableByHeuristics.find 0.004579
```
</details>

alicedoherty:
@henrinikku Thanks for looking into the noise - Jim is off this week so I'll review on behalf of him. Like you said I think this workload is good to merge, just let the BBs know. If you're struggling to get value from the workloads due to noise then I'd follow up with a server ticket to further look into this. As a side note, if you haven't already, it could be good to run a couple of patches on the new PERF variants as we've seen a good few workloads becoming less noisy on them. (Some of the mappings from old to new variants aren't super clear so feel free to reach out if you're not sure).

src/workloads/query/CollScanPredicateSelectivityLarge.yml
```diff
@@ -0,0 +1,26 @@
...
+      $eq:
+      - standalone
+      - standalone-all-feature-flags
+      - standalone-classic-query-engine
+      - standalone-heuristic-bonsai
+      - standalone-sampling-bonsai
+      - standalone-sbe
+    branch_name:
+      $gte:
+        v7.3
```
alicedoherty, line 26: Does this workload need to be backported to v7.3 branch? That being said I don't think we run any of the variants listed above on our release branches.

henrinikku: No need to backport this. I'll adjust the predicate to make it more clear that this should only run on master.

alicedoherty (approved):
LGTM - maybe just keep an eye on the waterfall noise and open a follow on ticket if that becomes an issue

henrinikku:
evergreen merge

PR #1106, opened 2024-01-10
DEVPROD-3617: Allow users to specify a task to run on task generation.

johndaniels:
This adds an activate_generated_tasks parameter, which allows for specifying variant/task pairs as part of the version parameters. auto_task_all will generate all of those variant/task pairs with activate:true s that they are automatically run.

**Jira Ticket:** DEVPROD-3617

**Whats Changed:**  
This adds an activate_generated_tasks parameter, which allows
for specifying variant/task pairs as part of the version
parameters. auto_task_all will generate all of those variant/task
pairs with activate:true so that they are automatically run.

**Patch testing results:**  

https://spruce.mongodb.com/version/659dcd8fa4cf477e674acaa1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

array_traversal and large_scale_long_lived in the patch above were activated automatically by this code.

**Workload Submission form:**  

### DIFF
--- a/src/lamplib/src/genny/tasks/auto_tasks_all.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks_all.py
@@ -11,13 +11,17 @@ from genny.tasks.auto_tasks import (
     Repo,
     ConfigWriter,
     YamlReader,
+    VariantTask,
 )
 
 SLOG = structlog.get_logger(__name__)
 
-def get_all_builds(global_expansions: dict[str, str], project_file_path: str) -> List[CurrentBuildInfo]:
+
+def get_all_builds(
+    global_expansions: dict[str, str], project_file_path: str
+) -> List[CurrentBuildInfo]:
     """
-    Attempts to compute the expansions 
+    Attempts to compute the expansions
     """
     with open(project_file_path) as project_file:
         project = yaml.safe_load(project_file)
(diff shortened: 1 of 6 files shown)


src/lamplib/src/genny/tasks/auto_tasks_all.py
```diff
@@ -32,36 +36,52 @@ def get_all_builds(global_expansions: dict[str, str], project_file_path: str) ->
...
-        SLOG.error(f"Evergreen expansions file {os.path.join(workspace_root, 'expansions.yml')} does not exist. Ensure this file exists and that it is in the correct location.")
+        SLOG.error(
+            f"Evergreen expansions file {os.path.join(workspace_root, 'expansions.yml')} does not exist. Ensure this file exists and that it is in the correct location."
+        )
         sys.exit(1)
     execution = int(global_expansions["execution"])
+    activate_tasks = set(
+        VariantTask(*item.strip().split(":"))
+        for item in global_expansions.get("activate_generated_tasks", "").split(",")
+    )
```
SimonEismann, line 77: While correct, this is a bit hard to parse. My personal preference would be to split this into 2-3 lines so it reads can be read top down, while right now it needs to be read "bottom-up". Feel free to ignore this.

src/lamplib/src/genny/tasks/auto_tasks_all.py
```diff
@@ -11,13 +11,17 @@
...
 SLOG = structlog.get_logger(__name__)
 
-def get_all_builds(global_expansions: dict[str, str], project_file_path: str) -> List[CurrentBuildInfo]:
+
+def get_all_builds(
+    global_expansions: dict[str, str], project_file_path: str
+) -> List[CurrentBuildInfo]:
     """
-    Attempts to compute the expansions 
+    Attempts to compute the expansions
```
SimonEismann, line 24: It seems your formatter disagrees with some of the existing formatting, which results in formatting-only changes like this that make the PR harder to parse.

My preference here would be that we either:
a) apply no formatter at all
b) set up an agreed-upon formatter as part of the CI tests (not as part of this ticket)

SimonEismann (approved):
Looks good, let's make sure to document this aggressively and announce it in #performance-tooling-users. Could this be used in some of our tools to schedule tasks (e.g., when kicking off profiling builds for generated tasks)? If so, let's open corresponding tickets.

src/lamplib/src/genny/tasks/auto_tasks_all.py
johndaniels: I attempted to what was set up as part of the repo, and explicitly ran `run-genny lint-python --fix`, which _should_ match what we do in CI. I'm a bit mystified by the fact that it fixed anything, but I figured I'd add it.

johndaniels: That is, we're supposed to be doing 2 above but it seems like something is wonky.

src/lamplib/src/genny/tasks/auto_tasks_all.py
johndaniels: Agreed. Will clean up.

johndaniels:
evergreen merge

johndaniels:
evergreen merge

PR #1115, opened 2024-01-18
SERVER-84676 Add queries over collections with a large number of fields

henrinikku:

**Jira Ticket:** SERVER-84676

**Whats Changed:**  
Add queries over collections with a large number of fields

**Patch testing results:**  
[Patch](https://spruce.mongodb.com/version/65a97a9b9ccd4e79469f8a98/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/phases/query/CollScanLargeNumberOfFields.yml
+++ b/src/phases/query/CollScanLargeNumberOfFields.yml
@@ -16,7 +16,7 @@ GlobalDefaults:
   DocumentCount: &DocumentCount {^Parameter: {Name: DocumentCount, Default: -1}}
   Repeat: &Repeat {^Parameter: {Name: Repeat, Default: -1}}
   Collection: &Collection Collection0
-  MaxPhases: &MaxPhases 18
+  MaxPhases: &MaxPhases 26
 
   # Generates a random string of 5-50 characters.
   RandomShortString: &RandomShortString {^FastRandomString: {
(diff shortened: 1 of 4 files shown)


src/phases/query/CollScanLargeNumberOfFields.yml
```diff
@@ -0,0 +1,359 @@
...
+  ]
+  FieldValues: &FieldValues [
+    &RandomBool {^Choose: {from: [true, false, null]}},
+    &RandomInt {^RandomInt: {min: 0, max: 1_000_000_000}},
+    &RandomDouble {^RandomDouble: {min: 0, max: 1_000_000_000}},
+    &RandomString {^Choose: {from: [*RandomShortString, *RandomLongString, null]}},
+    &RandomObjectId {^ObjectId: {^RandomString: {length: 24, alphabet: "0123456789ABCDEF"}}},
+    *RandomUUID,
+    &RandomDate {^RandomDate: {min: "1980-01-01", max: "2025-01-01"}},
+    &RandomObject {^Object: {
```
alyacb, line 58: Does this mean we will only have documents with one level of nesting?

alyacb (approved):
LGTM, though there is some question in my mind about:
- projections? Would be good to add some tests for really wide projections here actually, but it may overlap with the Projection workload Roy added
- Is it worth adding some smaller queries? Might help reveal inefficiencies where we scan more of each doc than needed

src/phases/query/CollScanLargeNumberOfFields.yml
henrinikku: Yep, all RandomObjects only have one level of nesting. Overall, we have paths with up to 4 levels (e.g. `arr.a.object1.randomfield`).

henrinikku (commented):
Thanks for the comments!
> projections? Would be good to add some tests for really wide projections here actually, but it may overlap with the Projection workload Roy added

I added inclusion and exclusion projections with different sizes (99, 50, 10). There may be some overlap, but not much imo. Roy's workload has projections that are more complex (e.g. more nested, mixes of renaming/inclusion projections).

> Is it worth adding some smaller queries? Might help reveal inefficiencies where we scan more of each doc than needed

I added queries with just 1 predicate. Also, the projection queries have an empty filter.

henrinikku:
evergreen merge

PR #1118, opened 2024-01-24
SERVER-85660: fix streams passthrough change stream source workload

Aadeshp:
**Jira Ticket:** SERVER-85660

**Whats Changed:**  
The change stream source should only be consumed after all documents have been written and durably persisted in the mongo collection, so the stream processor should be started after all the documents have been added to the source collection. In order to this we need to use `startAtOperationTime` in the $source.config to tell the stream processor to consume the documents in the change stream source from the beginning. It doesn't seem like there is a great way to generate a BSON timestamp object that mongo IDL likes through genny YAML or through the existing generators, so added `{^NowTimestamp: {}}` generator for BSON timestamps, similar to BSON date generators. (NOTE: None of the BSON date generators work here because the mongo IDL explicitly wants a BSON timestamp object for `startAtOperationTime`).

https://spruce.mongodb.com/version/65b1402f5623438e33194272/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -1480,6 +1480,17 @@ private:
     std::string _actorId;
 };
 
+/** `{^NowTimestamp: {}}` */
+class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp> {
+public:
+    NowTimestampGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_timestamp evaluate() override {
+        auto now = std::chrono::system_clock::now();
+        uint32_t nowSinceEpoch = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch()).count();
+        return bsoncxx::types::b_timestamp{.increment = 1, .timestamp = nowSinceEpoch};
+    }
+};
+
 /** `{^Now: {}}` */
 class NowGenerator : public Generator<bsoncxx::types::b_date> {
 public:
@@ -2142,6 +2153,7 @@ Out valueGenerator(const Node& node,
  */
 const auto [allParsers,
             arrayParsers,
+            timestampParsers,
             dateParsers,
             doubleParsers,
             intParsers,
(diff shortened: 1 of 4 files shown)


src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1480,6 +1480,17 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
 };
 
+/** `{^NowTimestamp: {}}` */
+class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp> {
+public:
+    NowTimestampGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_timestamp evaluate() override {
+        auto now = std::chrono::system_clock::now();
+        uint32_t nowSinceEpoch = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch()).count();
+        return bsoncxx::types::b_timestamp{.increment = 1, .timestamp = nowSinceEpoch};
```
thessem, line 1490: I don't fully understand why we have `.increment = 1` here? The sensible default seems to be `0` to me, and if this needs to be configurable it should be an argument to the generator. This might be a `mongod` thing I'm not understanding.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1480,6 +1480,17 @@ class ActorIdStringGenerator : public Generator<std::string> {
     std::string _actorId;
 };
 
+/** `{^NowTimestamp: {}}` */
+class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp> {
```
thessem, line 1484: This could be more generally useful if it could accept a `bsoncxx::types::b_date` from `DateGenerator` (for example) and only default to `now()` if it is not passed in. This would make it properly testable too rather than just checking it doesn't throw

thessem (changes requested):
Thanks for requesting the review. A few comments on the generator which I think make it more immediately useful and testable, if I've accidentally asked for something difficult here let me know.

mongodb-matthew-normyle:
re-opened here:
https://github.com/mongodb/genny/pull/1148#event-12108928905

src/value_generators/src/DocumentGenerator.cpp
harendra-c: Could you please clarify? 
Is you suggestion to use 
`class NowTimestampGenerator : public Generator<bsoncxx::types::b_date>`
instead of 
`class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp>` ??

PR #1120, opened 2024-01-30
PERF-5078 Improve Genny LookupUnwind.yml and add LookupOnly.yml

kevin-cherkauer:
Improve Genny workload LookupUnwind.yml to:
1. Use 20,000 docs in the local collection instead of just one
2. Use join keys in 0-9 instead of always 1

Add a sister Genny workload LookupOnly.yml that does the same tests as LookupUnwind.yml except without the $unwind stage of the pipeline to make the performance difference with and without $unwind visible. ($lookup-$unwind, aka $LU, is a macro stage that does a $lookup followed immediately by an $unwind that unwinds the lookup's "as" result array, but implemented so as never to materialize that array at all.)

Currently both of these benchmarks use LookupStrategy::kHashJoin for all cases.
 
Change 1 is needed so the cost of building the hash table has a positive return on investment. (I also opened SERVER-85964 in QO to improve the join strategy selection so as not to pick HashJoin when the local collection's cardinality is too small.)

Change 2 is needed so the queries in the new LookupOnly.yml do not fail from exceeding the 16 MB BSON object size limit when constructing the "as" output array.

Both changes make the benchmark better reflect typical real-world situations, whereas the original was testing more of an edge case.

Local runs of these two benchmarks show $LU is about 6x as fast as plain $lookup in both Classic and SBE execution engines, implying that most of the time of a $lookup is spent aggregating the foreign matches for a given local doc into a single array output.

### DIFF
--- a/src/phases/query/LookupCommands.yml
+++ b/src/phases/query/LookupCommands.yml
@@ -149,7 +149,21 @@ NLJLookupCmdTemplate:
   allowDiskUse: false
   cursor: {batchSize: 2000}
 
-# The command template for lookup/unwind.
+# The command template for workloads/query/LookupOnly.yml.
+LookupOnlyCmdTemplate: &LookupOnlyCmdTemplate
+  aggregate: {^Parameter: {Name: "l", Default: "local"}}
+  pipeline:
+    [
+      {$lookup: {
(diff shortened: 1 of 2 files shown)


aelhelw (approved):
LGTM

kevin-cherkauer:
evergreen merge

PR #1123, opened 2024-02-05
DEVPROD-4562 Update timeout_secs

SimonEismann:
Auto-generated tasks are currently generated with a 7200 (2h) idle timeout, which can not be overridden by users.

Since this timeout [duplicates DSI's no_output_ms ](https://github.com/10gen/dsi/blob/62d5572cebf887e260fe25231e4decc95db9d9f5/configurations/test_control/defaults.original.yml#L8), my suggestion would be to make timeout_secs equal to exec_timeout_secs, which defacto disables the evergreen idle timeout for these tasks. This way, we only have the DSI idle timeout that can be more easily configured by users.

### DIFF
--- a/src/lamplib/src/genny/tasks/auto_tasks.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks.py
@@ -578,7 +578,7 @@ class ConfigWriter:
                 [
                     CommandDefinition()
                     .command("timeout.update")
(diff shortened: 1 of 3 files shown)


johndaniels (approved):

SimonEismann:
evergreen merge

PR #1128, opened 2024-02-16
SERVER-84079 Introduce query settings Genny performance tests

denis631:
**Jira Ticket:** SERVER-84079

**Whats Changed:**  
Introduce a findOne by index queries with and without query settings to measure the performance

### DIFF
--- a/src/workloads/query/QuerySettings.yml
+++ b/src/workloads/query/QuerySettings.yml
@@ -1,7 +1,7 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  Measure the performance for find command with-/out query settings as well as checking both hit and miss cases.
+  Measure the performance for "find" command with or w/o query settings as well as checking both query settings hit and miss cases.
 
 KeyWords:
 - querySettings
--- a/src/workloads/query/QuerySettings.yml
+++ b/src/workloads/query/QuerySettings.yml
@@ -15,10 +15,10 @@ Clients:
 GlobalDefaults:
   Database: &Database test
   Collection: &Collection querySettingsColl
-  DocumentCount: &DocumentCount 100000
+  DocumentCount: &DocumentCount 1
   MaxPhases: &MaxPhases 12
-  HitFilter: &HitFilter {a1: {^RandomInt: {min: 1, max: 350}}}
-  MissFilter: &MissFilter {a2: {^RandomInt: {min: 1, max: 350}}}
+  HitFilter: &HitFilter {a1: 1}
+  MissFilter: &MissFilter {a2: 1}
   Threads: &Threads 64
 
 Actors:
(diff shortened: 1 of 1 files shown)


denis631:
[🌲 sys-perf with genny build (query settings variants only)](https://spruce.mongodb.com/version/65cf7f5e306615a113cccde6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

src/workloads/query/QuerySettings.yml
```diff
@@ -0,0 +1,164 @@
...
+    OnlyActiveInPhases:
+      Active: [3]
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Duration: 1 minute
+        Collection: *Collection
+        Operations:
+        - OperationName: findOne
+          OperationCommand:
+            Filter: *HitFilter
```
mmalinauskas, line 86: Even if the pseudo-random number generator seed is fixed for Genny test run, it is not clear if the filter matches a document. That likely can become very important if we want to compare results between no query settings, matching and non-matching query settings due to low resource consumption of the "find" command. I find this randomness and uncertainty unnecessary in this already messy world :).
I would suggest explicitly testing a scenario in which at least one document matches.

src/workloads/query/QuerySettings.yml
```diff
@@ -0,0 +1,164 @@
...
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 400
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection querySettingsColl
+  DocumentCount: &DocumentCount 100000
```
mmalinauskas, line 18: Is there a value of having so many documents in the test collection? Insert of 1000s of docs takes time.

src/workloads/query/QuerySettings.yml
```diff
@@ -0,0 +1,164 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  Measure the performance for find command with-/out query settings as well as checking both hit and miss cases.
```
mmalinauskas, line 4: "find" command

mmalinauskas, line 4: "both hit and miss cases" -> both query settings hit and miss cases"

mmalinauskas, line 4: with-/out -> "with or w/o"

mmalinauskas (commented):
Looks good @denis631!

src/workloads/query/QuerySettings.yml
denis631: Would you propose inserting 1000 documents instead?

denis631 (commented):
Thank you for your review, Mindaugas!

src/workloads/query/QuerySettings.yml
mmalinauskas: That's correct.

src/workloads/query/QuerySettings.yml
mmalinauskas: The absolute minimum, whatever it is.

denis631:
[🌲 genny run](https://spruce.mongodb.com/version/65d8c5463066157819014382/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

PR #1130, opened 2024-02-21
PERF-5125 Add perf workload for limit and skip

Adityav369:
**Jira Ticket:** PERF-5125

**Whats Changed:**  
Added performance workload for limit and skip with different constants.

**Patch testing results:**  

Latest (2/27, commit 6): https://spruce.mongodb.com/version/65de299a0305b90dd89576ba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

(2/26, commit 5): https://spruce.mongodb.com/version/65dcf4800ae6063ec621f36c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

https://spruce.mongodb.com/version/65d76cccc9ec44d86c88b3f2/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

https://spruce.mongodb.com/version/65d66ae7a4cf47f3c3d27148/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,134 @@
...
+      pipeline: [{$match: {$and: [{hello: "hellohellohellohellobye"}, {zero: 0}]}}, {$skip: *limitSkipNum}, {$project: {_id: 0}}]
+
+- ActorFromTemplate:
+    TemplateName: RunAggCommand
+    TemplateParameters:
+      name: "LimitSkipVaryConstant"
+      active: [5]
+      pipeline: [{$match: {$and: [{hello: "hellohellohellohellobye"}, {zero: 0}]}}, {$skip: *limitSkipNum}, {$limit: *limitSkipNum} , {$project: {_id: 0}}]
+
+AutoRun:
```
Adityav369, line 120: Do we want to run this on any other build variants or versions?

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,134 @@
...
+        Repeat: 1
+        Threads: 1
+
+# Phase 3-N: Run pipelines with limit, skip, and limit + skip.
+- ActorFromTemplate:
+    TemplateName: RunAggCommand
+    TemplateParameters:
+      name: "LimitVaryConstant"
+      active: [3]
+      pipeline: [{$match: {$and: [{hello: "hellohellohellohellobye"}, {zero: 0}]}}, {$limit: *limitSkipNum}, {$project: {_id: 0}}]
```
Adityav369, line 104: Would we want an even simpler pipeline to stress limit and skip? This is a stripped down customer inspired query from a help ticket.

Adityav369 (commented):
@Fefer-Ivan I left a few questions about the design of this workload.

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,134 @@
...
+
+- ActorFromTemplate:
+    TemplateName: RunAggCommand
+    TemplateParameters:
+      name: "LimitSkipVaryConstant"
+      active: [5]
+      pipeline: [{$match: {$and: [{hello: "hellohellohellohellobye"}, {zero: 0}]}}, {$skip: *limitSkipNum}, {$limit: *limitSkipNum} , {$project: {_id: 0}}]
+
+AutoRun:
+- When:
```
Fefer-Ivan, line 121: Let's also add a pipeline with sort, as it should absorb limit and also should be parameterized.

src/workloads/query/LimitSkip.yml
Fefer-Ivan: No, I think that's fine

src/workloads/query/LimitSkip.yml
Fefer-Ivan: No, the pipeline looks fine.

Fefer-Ivan (changes requested):
Good work. Thanks, Adi. 
I have requested one more pipeline to be added, but that's about it

src/workloads/query/LimitSkip.yml
Adityav369: Done.

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,142 @@
...
+- classic
+- sbe
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  DocumentCount: &DocumentCount 1e6
+  Threads: &Threads 1
+  MaxPhases: &MaxPhases 6
+  LimitSkipNumber: &limitSkipNum {^Inc: {start: 1, step: 100}}
```
Fefer-Ivan, line 21: Let's make it random and not incremental

Fefer-Ivan, line 21: Let's make it random and not incremental

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,142 @@
...
+    Phases:
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "active", Default: [0]}}
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: 1000
+          Database: *Database
+          Collection: *Collection
+          Operations:
+          - OperationMetricsName: sbe
```
Fefer-Ivan, line 40: Do we really need it? I think we can just remove this parameter and have auto-generated metric names from actor name.

Fefer-Ivan (changes requested):
I am pretty sure this is the last (or second to last) round of this review.
Just left a couple of small comments.

src/workloads/query/LimitSkip.yml
Adityav369: I thought of making it random initially, then realized, we may get inconsistent results across different runs. With ^Inc and a set number of "Repeat," we can be sure that this workload runs deterministically. Do you think this would matter? If not, then going down the random route is better.

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,142 @@
...
+        Repeat: 1
+        Database: *Database
+        MultipleThreadsPerCollection: true
+        CollectionCount: 1
+        DocumentCount: *DocumentCount
+        BatchSize: 1000
+        Document:
+          zero: 0
+          hello: "hellohellohellohellobye"
+          randomInt: {^Choose: {from: [1, 2, 3, 4, 5], deterministic: true}}
```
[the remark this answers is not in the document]
Adityav369: I did this so `$sort` has a deterministic amount of work to do. With deterministic = true, we can guarantee the spread of the data.

src/workloads/query/LimitSkip.yml
Adityav369: Done. I added `sort -> limit -> skip` and `sort -> skip -> limit`.

src/workloads/query/LimitSkip.yml
```diff
@@ -17,30 +17,30 @@ GlobalDefaults:
...
-        Active: {^Parameter: {Name: "active", Default: [0]}}
+        Active: {^Parameter: {Name: "active", Default: {unused: "Invalid phase number."}}}
         NopInPhasesUpTo: *MaxPhases
         PhaseConfig:
           Repeat: 1000
           Database: *Database
           Collection: *Collection
           Operations:
-          - OperationMetricsName: sbe
+          - OperationMetricsName: *Name
```
Fefer-Ivan, line 40: This will happen automatically, so you don't need to specify this
https://github.com/mongodb/genny/blob/dcab85f33da584a2a2c7e2334c2499eb584627d5/src/cast_core/src/CrudActor.cpp#L1637

src/workloads/query/LimitSkip.yml
Fefer-Ivan: Okay, that makes sense. Thanks.

However, how would that work for cases where we have both $limit and $skip - it would mean that we will always have skip == limit and the query will return no results ever.

Maybe we can separate skip and limit num definitions and start limit at 100 and skip at 0?

Fefer-Ivan (changes requested):

src/workloads/query/LimitSkip.yml
Adityav369: Oh that's nice, done!

src/workloads/query/LimitSkip.yml
Adityav369: Done.

src/workloads/query/LimitSkip.yml
```diff
@@ -0,0 +1,151 @@
...
+- classic
+- sbe
+
+GlobalDefaults:
+  Database: &Database test
+  Collection: &Collection Collection0
+  DocumentCount: &DocumentCount 1e6
+  Threads: &Threads 1
+  MaxPhases: &MaxPhases 8
+  LimitNumber: &limitNum {^Inc: {start: 100, step: 100}}
```
Fefer-Ivan, line 21: Given that we do Repeat: 1000, the amount of documents, returned by the query will change from 100 too 100 000 during run time.

It feels a little bit too much variety. Maybe leave the default step of 1?

Fefer-Ivan (approved):
Approved with only one small change required: adjust step constant.

src/workloads/query/LimitSkip.yml
Adityav369: Done.

Adityav369:
evergreen merge

PR #1090, opened 2023-12-19
PERF-4992 Clean up references to queryStats configurations

shaargtz:
**Jira Ticket:** [PERF-4992
](https://jira.mongodb.org/browse/PERF-4992)
**Whats Changed:**  
As the queryStats project is wrapping up, the specific variants used for its testing have been removed, and so its corresponding configuration files from DSI are being removed as well. Because of this, references to said configurations are being removed from Genny workloads too. No testing is being done as only autorun calls are being removed.

**Related PRs:**   
https://github.com/10gen/dsi/pull/1499
https://github.com/10gen/mongo/pull/17634

PR #1131, opened 2024-02-22
PERF-5152 Update query-stats setups in AutoRun

shaargtz:
This patch is adding the mongodb_setups for the query stats variants to the relevant workloads' AutoRuns. This is a sort of revert of #1090. This is a follow up to https://github.com/10gen/dsi/pull/1545 and https://github.com/10gen/dsi/pull/1545.

sys-perf patch pending

### DIFF
--- a/src/workloads/query/CollScanSimplifiablePredicateSmall.yml
+++ b/src/workloads/query/CollScanSimplifiablePredicateSmall.yml
@@ -19,6 +19,7 @@ AutoRun:
       - standalone-all-feature-flags
       - standalone-classic-query-engine
       - standalone-heuristic-bonsai
+      - standalone-query-stats
       - standalone-sampling-bonsai
       - standalone-sbe
     branch_name:
--- a/src/workloads/docs/RunCommand-Simple.yml
+++ b/src/workloads/docs/RunCommand-Simple.yml
@@ -28,4 +28,5 @@ AutoRun:
       - standalone
       - standalone-classic-query-engine
       - standalone-dsi-integration-test
+      - standalone-query-stats
       - standalone-sbe
--- a/src/workloads/execution/ClusteredCollection.yml
+++ b/src/workloads/execution/ClusteredCollection.yml
@@ -21,11 +21,15 @@ AutoRun:
       $eq:
       - standalone
       - standalone-all-feature-flags
+      - standalone-query-stats
       - replica
       - replica-all-feature-flags
       - replica-query-stats-rate-limit
       - atlas-like-replica.2022-10
+      - atlas-like-replica.2023-09
+      - atlas-like-replica-query-stats.2023-09
       - shard
       - shard-lite-all-feature-flags
+      - shard-query-stats
     branch_name:
       $gte:  v5.3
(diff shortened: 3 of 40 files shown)


src/workloads/query/CollScanSimplifiablePredicateSmall.yml
```diff
@@ -19,6 +19,7 @@ AutoRun:
       - standalone-all-feature-flags
       - standalone-classic-query-engine
       - standalone-heuristic-bonsai
+      - standalone-query-stats
```
awagenmaker, line 22: Looks like this workload's auto-run is configured to only run on v7.3 and above. Was it added in v7.3? We should either remove our config from this list or change the branch name, if it's possible to do that. Looks like this applies to all the other `CollScanSimplifiablePredicate*` workloads as well.

awagenmaker (commented):
Overall looks good, thanks @shaargtz ! I left one question about the branch name filter on some of the workloads, and also noticed that we didn't touch the following workloads from the last project's cleanup: 

`MatchFiltersMedium.yml` 
`MatchWithLargeExpression.yml`
`LargeIndexedIns.yml`
`MajorityReads10KThreads.yml`
`MixedWorkloadsGenny.yml`
`MixedWorkloadsGennyRateLimited.yml`

Were these intentionally left out? Just want to confirm that we aren't going to be missing anything.

shaargtz:
> Overall looks good, thanks @shaargtz ! I left one question about the branch name filter on some of the workloads, and also noticed that we didn't touch the following workloads from the last project's cleanup:
> 
> `MatchFiltersMedium.yml` `MatchWithLargeExpression.yml` `LargeIndexedIns.yml` `MajorityReads10KThreads.yml` `MixedWorkloadsGenny.yml` `MixedWorkloadsGennyRateLimited.yml`
> 
> Were these intentionally left out? Just want to confirm that we aren't going to be missing anything.

Oh I forgot to add `UpdateWith.yml` to the document so I miscounted, I missed `MatchFiltersMedium.yml`. `MatchWithLargeExpression.yml` is marked for 7.2 onwards because of a performance improvement, and in a slack thread @cswanson310 said it was fine to skip it. `MajorityReads10KThreads.yml` and `MixedWorkloadsGenny.yml` are already clickable tasks in the variants so I skipped the AutoRun to avoid double running it, but if we'd rather have it in AutoRun I can go ahead and change that. I also missed `MixedWorkloadsGennyRateLimited.yml` as it was at the bottom of the list. I'll go ahead and update as necessary.

src/workloads/query/CollScanSimplifiablePredicateSmall.yml
shaargtz: I didn't notice that. The collscan benchmarks appear to be added in 7.3, I'll ask the author for the proper course of action.

shaargtz: @henrinikku said it's ok to mark it for v7.0, done

awagenmaker (approved):
Nice work, LGTM!

shaargtz:
evergreen merge

PR #1134, opened 2024-02-27
PERF-5111 Multiplanner simple case

dhsegel:


src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,400 @@
...
+    OnlyActiveInPhases:
+      Active: [3, 5, 7]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        # TODO more repeats
+        Repeat: 2
+        Database: *db
+        Collection: *coll
+        Operations:
+        - OperationMetricsName: RunQueries
```
dpercy, line 239: We should probably split this actor into three: one for each phase, so that when we use a different number of indexes the metrics go to a different OperationMetricsName.

On this patch it looks like they're being aggregated when I filter by RunQueries, AverageLatency there's only on row: https://spruce.mongodb.com/version/65de69655623437ddd50da76/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC%3BNAME%3AASC

The numbers I got so far are:
| build variant | latency (not sure what unit) |
|--------|--------|
| standalone | 21.9 million |
| standalone all feature flags | 22.5 million |
| standalone classic | 22.1 million | 

Doesn't it seem weird that these numbers would be so close?  The SBE multiplanner should be worse, because we have 1 good plan and 62 bad plans.  I wonder if it's getting lucky, and enumerating the good plan first.  If it limits each trial execution by the best upper bound so far, then seeing the best plan first lets it avoid wasting time on the bad plans.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,419 @@
...
+        - OperationName: RunCommand
+          OperationCommand:
+            setParameter: Collection0
+            # Disable the plan cache so we multiplan on every request.
+            internalQueryDisablePlanCache: true
+            # Disable index intersection so we can more easily control the number of competing plans.
+            # We should get one plan per index.
+            internalQueryPlannerEnableIndexIntersection: false
+
+- Name: "CreateDataset"
```
Adityav369, line 69: Let's move this part to `src/phases/..`. This way, for future tickets like PERF-5112, PERF-5113 we can import the InsertActor and just make the different indexes. 

You could make the indexes field a `^Parameter` and the whole actor can be moved. 
See `src/phases/query/AggregateExpressions.yml` for example.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,419 @@
...
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+
+- Name: Queries
+  Type: CrudActor
+  Threads: 1
+  Phases:
```
Adityav369, line 230: A cleaner way would be to use OnlyActiveInPhases like the other phases and then redefine the Actor for phase 5 and 7 with different names like `SimpleMultiplannerWithXIndexes`. This way you can also define a find query template and parameterize the `Active` and `Name`.

Adityav369 (changes requested):
Thanks Daniel, I added a couple comments. Lmk if you have questions.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,419 @@
...
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+        Database: *db
+        Operations:
+        - OperationName: RunCommand
+          OperationCommand:
+            drop: Collection0
+
+- Name: SetQueryKnobs
```
dpercy, line 47: I think this actor will be identical in the other scenarios, so it can be pulled out into a separate file.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,419 @@
...
+      - OperationMetricsName: RunQueriesB
+        OperationName: find
+        OperationCommand:
+          Filter: *filter
+    - *Nop
+    - Repeat: 2
+      Database: *db
+      Collection: *coll
+      Operations:
+      - OperationMetricsName: RunQueriesC
```
dpercy, line 321: It would be helpful to make these names more meaningful, like "query with 63 indexes".

Even though you still need to understand the workload to understand the output, having clearer labels would reduce amount of flipping back and forth between the output and workload definition.

src/workloads/query/multiplanner/Simple.yml
dpercy: Ah, this must be because all three are using the classic multiplanner:
* standalone, the default configuration = classic+classic
* standalone all feature flags = classic+SBE, now that our project's feature flag is included here
* standalone classic = classic+classic

So I'll add this test to another build variant, for standalone-sbe.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,400 @@
...
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {x13: 1}}}}
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {x14: 1}}}}
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {x15: 1}}}}
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {x16: 1}}}}
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone  # At time of writing this will be SBE.
```
Fefer-Ivan, line 416: This is not SBE
This is SBE with no other feature flags
```
standalone-sbe
```

src/workloads/query/multiplanner/Simple.yml
Fefer-Ivan: You also don't need to specify OperationMetricsName. It is an optional field, so if it is not something meaningful, let's just delete it and it will default to OperationName

src/workloads/query/multiplanner/Simple.yml
dhsegel: Removing altogether per Ivan's comment below. Will keep clear labels in the `Name` field.

src/phases/query/Multiplanner.yml
```diff
@@ -0,0 +1,194 @@
...
+Description: |
+  This file defines templates to use in multiplanner performance tests.
+
+QueryKnobTemplate:
+  Name: SetQueryKnobs
+  Type: RunCommand
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: {^Parameter: {Name: "active", Default: [0]}}
```
Adityav369, line 12: For `default`, we could use incorrect values like a string. This way, we can ensure we do not miss a parameter when using this template. Something like https://github.com/mongodb/genny/blob/master/src/workloads/query/LimitSkip.yml#L34.
This comment applies to other places Parameter is used.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,277 @@
...
+      nopInPhasesUpTo: *maxPhase
+      collection: *coll
+
+- Name: CreateDataset
+  LoadConfig:
+    Path: "../../../phases/query/Multiplanner.yml"
+    Key: InsertTemplate
+    Parameters:
+      database: *db
+      threadsA: 1
```
Adityav369, line 64: As per another comment in template file, no need for `threadsA` and `threadsB`.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,277 @@
...
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-sbe
+      - standalone-all-feature-flags  # At time of writing this will enable PM-3591.
+      - standalone-classic-query-engine
+    branch_name:
+      $gte: v7.3
+
```
Adityav369, line 277: [nit] format

Adityav369 (changes requested):
Thanks Daniel, just added some comments about style. In terms of the logic and actual workload, this looks about right!

src/workloads/query/multiplanner/Simple.yml
dhsegel: What is the problem here? Should `v7.3` be on its own line?

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -0,0 +1,277 @@
...
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [4]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+        Database: *db
+        Operations:
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {x17: 1}}}}
```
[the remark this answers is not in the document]
dhsegel: I'm not sure this is possible. I don't believe I can use `^FormatString` in conjunction with `^Inc` to succinctly name all the indexes. Is this similar to what you had in mind?
`- {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {{^FormatString: {"{x%d_1", "withArgs:": {^Inc: {start: 17}}}}}}}}`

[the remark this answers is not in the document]
dpercy: For example, imagine trying to factor it out like this:

```
place_for_macro_definitions:
  my_macro:
    OperationName: RunCommand
    OperationCommand:
      collMod: *coll
      index: 
        hidden: true
        keyPattern: {^Parameter: {Name: "keyPattern", Default: "invalid"}}
...
Operations:
- {invoke_a_macro: 'my_macro', args: {keyPattern: {x1: 1}}}
- {invoke_a_macro: 'my_macro', args: {keyPattern: {x2: 1}}}
...
- {invoke_a_macro: 'my_macro', args: {keyPattern: {x63: 1}}}
```

There would still be repetition, which we need to verify by eye, but it also adds an indirection that can make it less clear what an individual command looks like.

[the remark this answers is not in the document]
dhsegel: That makes sense to me, but is that preferred stylistically? It would be more lines of code and less readable than just repeating the same 64 lines where it is very clear that the same operation is being done to each index.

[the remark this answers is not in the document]
dpercy: Right, to be clear I was arguing that it wouldn't be an improvement overall to factor it out that way.

src/workloads/query/multiplanner/Simple.yml
Adityav369: I think there is an extra space on L27. There is a red circle which indicates that a blank line is not left at the end of file (which is standard practice for git). If you remove this space, it should go away.

src/workloads/query/multiplanner/Simple.yml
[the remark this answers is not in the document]
Adityav369: That's an interesting approach, and I agree with @dpercy that it is better to be explicit. I was thinking of something similar to [this](https://github.com/mongodb/genny/pull/1134#discussion_r1513153261), but it is not important. Feel free to ignore this suggestion, better to be explicit.

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -155,6 +153,74 @@ Actors:
       database: *db
       collection: *coll
       selectivity: *selectivity
```
Adityav369, line 155: This is no longer relevant, right? Since you have extracted the query from the template and are supplying the selectivity value directly? I also think that the predicate on x1 will default to 1 in this case. We can now directly use selectivity.

Adityav369 (approved):
Thanks Daniel! LGTM from my side, but I left one last comment. This looks great, will make the subsequent tickets much easier to implement. 

I have no further comments as of now and will defer to the assignee for the final LGTM.

src/workloads/query/multiplanner/Simple.yml
dhsegel: Thanks for catching that!

dpercy (approved):

src/workloads/query/multiplanner/Simple.yml
Adityav369: @dhsegel pls fix this before you're ready to merge.

dhsegel: Done. I had done this to `Simple.yml` like you saw but not to `Multiplanner`.

Fefer-Ivan (approved):

dhsegel:
evergreen merge

PR #1136, opened 2024-02-28
Remove AutoRun from unnecessary variants in QueryStats.yml

shaargtz:
The only analysis we'll do for the 7.0 QueryStats backport that involves this workload is in th M60-like variant, so running it elsewhere is not needed.


awagenmaker (approved):
LGTM

shaargtz:
evergreen merge

PR #1137, opened 2024-02-28
Devprod-2165 redirect and review Poplar outputs

lgfang:
**Jira Ticket:** DEVPROD-2165

**Whats Changed:**

Redirect the output of Poplar into a log file. When the process finishes, review the log file to:

-   Print any "fatal error" to stderr
-   If the error OOM (or alike), print a message to suggest that the user uses a larger workload instance.

Previously the outputs of Poplar are log at the INFO level. Normally this is not an issue because Poplar prints almost nothing. But, when Poplar hits the OOM issue, it may print a huge stack trace, making it hard to spot the issue. See [this log](https://parsley.mongodb.com/evergreen/sys_perf_perf_1_node_replSet_longRunning.arm.aws.2023_11_llt_mixed_patch_798243406d1f483e46669d7d20f29f4d8f50ba01_65ddb0c20ae6063ec624c105_24_02_27_10_09_37/0/task?bookmarks=0,11604) for an example.

**Patch testing results:**

See the log for [llt\_mixed](https://spruce.mongodb.com/task/sys_perf_perf_1_node_replSet_longRunning.arm.aws.2023_11_llt_mixed_patch_798243406d1f483e46669d7d20f29f4d8f50ba01_65ddb0c20ae6063ec624c105_24_02_27_10_09_37/logs?execution=5)

**Related PRs:**

None

### DIFF
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -190,10 +190,10 @@ def poplar_grpc(cleanup_metrics: bool, workspace_root: str, genny_repo_root: str
 
 
 def _report_poplar_error(log: TextIO):
-    """Review the log file for fatal errors and print them to stderr.
+    """
+    Review the log file for fatal errors and print them to stderr.
 
     We don't want to print the entire log, just highlight the fatal error(s).
-
     """
 
     log.seek(0)
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -1,3 +1,4 @@
+import sys
 import errno
 import socket
 import os
@@ -6,7 +7,7 @@ import subprocess
 import datetime
 import time
 from uuid import uuid4
-from typing import Optional
+from typing import Optional, TextIO
 
 import structlog
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


src/lamplib/src/genny/curator.py
```diff
@@ -145,45 +146,82 @@ def poplar_grpc(cleanup_metrics: bool, workspace_root: str, genny_repo_root: str
...
     finally:
         os.chdir(prior_cwd)
 
 
+def _report_poplar_error(log: TextIO):
+    """Review the log file for fatal errors and print them to stderr.
+
+    We don't want to print the entire log, just highlight the fatal error(s).
+
+    """
```
thessem, line 197:
```suggestion
    """
    Review the log file for fatal errors and print them to stderr.

    We don't want to print the entire log, just highlight the fatal error(s).
    """
```

src/lamplib/src/genny/curator.py
```diff
@@ -145,45 +146,82 @@ def poplar_grpc(cleanup_metrics: bool, workspace_root: str, genny_repo_root: str
...
+
+    is_oom = False
+
+    for line in log:
+        line = line.strip().lower()
+        if "fatal error" in line:
+            # Error messages should go to sys.stderr. Especially, from the
+            # perspective of DSI, all Genny SLOG log entries are just the stdout of
+            # the child process regardless the log level.
+            print(f"Poplar fatal error: {line}", file=sys.stderr)
```
thessem, line 209: As this is written this seems like it's going to print something out twice, just-in-case? This seems sub-optimal. If SLOG is redirected to a log file, it seems to me that the correct thing to do here is to print this error into the error log?

Additionally, can't we create a bound copy of SLOG that will print out on `stderr`? I'd rather not use `print` logging if we can avoid, I don't want two styles of logging flying around.

Is this change needed for this PR to function?

src/lamplib/src/genny/curator.py
```diff
@@ -145,45 +146,82 @@ def poplar_grpc(cleanup_metrics: bool, workspace_root: str, genny_repo_root: str
...
-                        # We expect to get connection refused if poplar is not up yet.
-                        # If we get something else, bail
-                        raise socket_error
-                finally:
-                    sock.close()
-            if not connected:
-                raise OSError(f"Poplar not listening on port 2288")
-            yield poplar
-        finally:
+        with open("./build/WorkloadOutput/poplar_grpc.log", "w+") as log_file_obj:
```
thessem, line 149: Where did this filename come from?

thessem (changes requested):

src/lamplib/src/genny/curator.py
lgfang: I chose this file name because the enclosing function is named `poplar_grpc`

thessem: Sorry, I should've been more specific, is `./build/WorkloadOutput` a magic basedir name that makes this meaningful? I ask because I think evergreen very recently added "pick up logs from a directory" and I was wondering whether this was to enable that, or whether this is a required directory etc.

src/lamplib/src/genny/curator.py
lgfang: The rationale behind printing error messages to stderr is to enable DSI to classify such messages as errors ( in https://github.com/10gen/dsi/pull/1575).

While is It is feasible to eliminate redundancy and maintain a consistent logging style, I chose not to for two reasons: a) I wanted to keep the changes to a minimum. b) in the future, if we decide to redirect logs to log files and reserve stderr/stdout solely for user interaction (maybe via [DEVPROD-4672](https://jira.mongodb.org/browse/DEVPROD-4672)),   this will no longer pose a problem.

Having said that, I acknowledge the validity of your request and hence have made the necessary updates to the code in the latest commit. [Here](https://parsley.mongodb.com/evergreen/sys_perf_perf_1_node_replSet_longRunning.arm.aws.2023_11_llt_mixed_patch_798243406d1f483e46669d7d20f29f4d8f50ba01_65dfd77c0305b9f4950e16b8_24_02_29_01_05_23/3/task?bookmarks=0,1697&selectedLineRange=L1475-L1518&shareLine=1475) is an example log when run in DSI.

src/lamplib/src/genny/curator.py
lgfang: My understanding is: Genny writes test results, metrics etc. into this directory; DSI downloads the content of this directory during `make_artifacts`. Since this log file maybe needed for debugging, I put it into this directory.

thessem (approved):
I think we may need to add some strings to check for based on some of the other BFs coming through but I am happpy we have a general method for this now

PR #1140, opened 2024-03-04
PERF-5184 Add workload reproducing an OOM on mongos caused by a cursor storm

robsndr:
Jira Ticket: [PERF-5184](https://jira.mongodb.org/browse/PERF-5184)

Whats Changed:
Add a new workload reproducing an OOM on mongos caused by a cursor storm.

Workload Submission form:
If applicable, only required if is a new workload being added. Form can be found [here](https://docs.google.com/forms/d/1r0kOHvFUa7rJxVmX_wp9prxYU5K155yKyZ1y3d5yhZg/edit)

### DIFF
--- a/src/workloads/scale/CursorStormMongos.yml
+++ b/src/workloads/scale/CursorStormMongos.yml
@@ -5,7 +5,7 @@ Description: |
   many heavy cursors. The workload is expected to fail due to host(s) being unreachable as a
   result of mongos running out of memory.
 
-  To achieve this, many threads are then spawned to run aggregate sort on a number of documents.
+  To achieve this, many threads are spawned to run aggregate sort on a number of documents.
   The document size and the number of documents to sort by each thread are tuned so, that the
   operations do not spill to disk [<100MB], and mongos is able to exhaust the cursors on shards
   when pre-filling its buffers [<16MB per shard]. As a result, memory pressure on the shards
@@ -83,7 +83,7 @@ Actors:
           OperationCommand: {dropDatabase: 1}
 
 # Create collection
-- Name: Create
+- Name: CreateCollection
   Type: RunCommand
   Threads: 1
   Phases:
@@ -99,7 +99,7 @@ Actors:
             # Loader default collection name.
             create: Collection0
 
-# Enable sharding
+# Enable sharding & shard collection
 - Name: ShardCollection
   Type: AdminCommand
   Threads: 1
@@ -167,7 +167,7 @@ Actors:
                   }
                 }
               },
-              # Merge-sorting results increases the number of heavy cursors on mongos
+              # Merge-sorting results imposes additional memory pressure on mongos
               {$sort: {b: 1}}]
             cursor: {batchSize: *SortStep}
 
--- a/src/workloads/scale/CursorStormMongos.yml
+++ b/src/workloads/scale/CursorStormMongos.yml
@@ -5,11 +5,10 @@ Description: |
   many heavy cursors. The workload is expected to fail due to host(s) being unreachable as a
   result of mongos running out of memory.
 
-  To achieve this, many threads are spawned to run aggregate sort on a number of documents.
-  The document size and the number of documents to sort by each thread are tuned so, that the
-  operations do not spill to disk [<100MB], and mongos is able to exhaust the cursors on shards
-  when pre-filling its buffers [<16MB per shard]. As a result, memory pressure on the shards
-  remains low, while it's kept large on the mongos.
+  To achieve this, many threads are spawned to run an unfiltered find on a collection.
+  The number and size of documents in that collection are tuned such, that the mongos is able
+  to exhaust cursors on shards when pre-filling its buffers [<16MB per shard]. As a result, 
+  memory pressure on the shards remains low, while it's kept large on the mongos.
 
 Keywords:
 - scale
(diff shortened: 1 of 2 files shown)


src/workloads/scale/CursorStormMongos.yml
```diff
@@ -0,0 +1,180 @@
...
+      NopInPhasesUpTo: *MaxPhases
+      PhaseConfig:
+        Repeat: *SortRepeat
+        Database: *DBName
+        Operations:
+        - OperationMetricsName: SortMany
+          OperationName: RunCommand
+          OperationCommand:
+            # Loader default collection name.
+            aggregate: Collection0
```
josefahmad, line 158: Can we simply stream results from the shard, e.g. a find resulting in a plain index scan or collection scan. We don't need to tax the shards, all we need is to generate a batch to feed the mongos.

src/workloads/scale/CursorStormMongos.yml
```diff
@@ -0,0 +1,180 @@
...
+                  {$let:
+                    {vars:
+                      {start: {^RandomInt: {min: 0, max: *NumDocs}}},
+                    in: {$and: [$gte: ["$a", "$$start"], $lt: ["$a", {$add: ["$$start", *SortStep]}]]}
+                    }
+                  }
+                }
+              },
+              # Merge-sorting results imposes additional memory pressure on mongos
+              {$sort: {b: 1}}]
```
josefahmad, line 171: Wasn't aware of this. How so?

josefahmad (commented):
Thanks for putting up this PR. I think we can simplify the workload on the shard, in that all I need is shards returning results to the mongos.

src/workloads/scale/CursorStormMongos.yml
robsndr: I've kept the sort based on the observation that without it, the workload takes ~2x longer to reach OOM, with ~2x more cursors being opened on the mongos. Could it be that using the sort-stage affects how data is fetched/buffered on the mongos? 

Shall we simplify the workload regardless?

src/workloads/scale/CursorStormMongos.yml
robsndr: Please see above.

src/workloads/scale/CursorStormMongos.yml
```diff
@@ -0,0 +1,180 @@
...
+
+  # The size of documents to process in a given step should approach 16MB per shard. Therefore, the
+  # number of documents to sort per shard is tuned such that SortStepPerShard * DocSize -> 16 MB.
+  SortStepPerShard: &SortStepPerShard 90
+  # With data distributed evenly across shards, the total size of documents to process in a
+  # given step should be equall to NumOfShards * SortStepPerShard.
+  SortStep: &SortStep
+    ^NumExpr:
+      withExpression: "num_shards * sort_step_per_shard"
+      andValues: {num_shards: *NumShards, sort_step_per_shard: *SortStepPerShard}
```
josefahmad, line 59: IIUC all of this is to set a specific batch size. Why don't we just use the default batch size?

src/workloads/scale/CursorStormMongos.yml
```diff
@@ -0,0 +1,180 @@
...
+Description: |
+  Used to test performance of the router under memory pressure caused by accumulating
+  many heavy cursors. The workload is expected to fail due to host(s) being unreachable as a
+  result of mongos running out of memory.
+
+  To achieve this, many threads are spawned to run aggregate sort on a number of documents.
+  The document size and the number of documents to sort by each thread are tuned so, that the
+  operations do not spill to disk [<100MB], and mongos is able to exhaust the cursors on shards
+  when pre-filling its buffers [<16MB per shard]. As a result, memory pressure on the shards
+  remains low, while it's kept large on the mongos.
```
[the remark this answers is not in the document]
robsndr: Got rid of the sort, as it didn't have its former effect on the workload after removing the aggregation.

src/workloads/scale/CursorStormMongos.yml
robsndr: I've assumed that we might see some effects related to caching if the exact same query is run by many threads, many times. However, this seems to not be the case! Removed some noise from the workload.

> Can we simplify this workload by doing a simple unfiltered find + sort stage on an index and avoiding a blocking sort on the shards?

Updated.

> Out of curiosity, how long does it take for this stage to OOM with and without the sort stage?

It was ~40s vs ~80s. However, now its ~20s, and I could barely see any difference with or without the sort. Therefore, I've just removed the sort altogether.

josefahmad: Perhaps the speed-up is due to https://jira.mongodb.org/browse/DEVPROD-4593? In any case, I'm happy without the sort stage!

josefahmad (approved):
LGTM! Thanks Robert, and sorry for the delay in reviewing

src/workloads/scale/CursorStormMongos.yml
```diff
@@ -0,0 +1,147 @@
+SchemaVersion: 2018-07-01
+Owner: Service Architecture
```
josefahmad, line 2: @robsndr I just come across [this slack thread|https://mongodb.slack.com/archives/C0685CU3EPJ/p1708101747222789] suggesting the owner should be Query.

robsndr: Changed the owning team to Query.

josefahmad (approved):
LGTM!

robsndr:
evergreen merge

PR #1143, opened 2024-03-05
PERF-5112 Multiplanner compound indexes

dhsegel:


src/workloads/query/multiplanner/CompoundIndexes.yml
```diff
@@ -0,0 +1,347 @@
...
+  Threads: 1
+  Phases:
+    OnlyActiveInPhases:
+      Active: [4]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+        Database: *db
+        Operations:
+        - {OperationName: RunCommand, OperationCommand: {collMod: *coll, index: {hidden: true, keyPattern: {tenantId: 1, x17: 1}}}}
```
Fefer-Ivan, line 241: You can reference indexes by name instead of keyPattern in coll mod.

And I think it might be a good idea to have indexes named the same way in all our workloads of this type.

If we do this, we can move this huge actor into phases file and re-use.

src/workloads/query/multiplanner/CompoundIndexes.yml
```diff
@@ -0,0 +1,347 @@
...
+        x56: *distribution,
+        x57: *distribution,
+        x58: *distribution,
+        x59: *distribution,
+        x60: *distribution,
+        x61: *distribution,
+        x62: *distribution,
+        x63: *distribution,
+      }
+      indexes:
```
Fefer-Ivan, line 135: Here you can add 'name' as a parameter after keys:

Fefer-Ivan (changes requested):

src/workloads/query/multiplanner/CompoundIndexes.yml
dhsegel: Done. Moved `Hide48Indexes` and `Hide14Indexes`.

Fefer-Ivan (approved):

dhsegel:
evergreen merge

dhsegel:
evergreen merge

PR #1145, opened 2024-03-07
PERF-5115 Create multi-planner performance workload where we hit 'getTrialPeriodMaxWorks'

dhsegel:
Only change here compared to `Simple.yml` is on L222, adding `{no_such_field: "none"}` to `query`.


Fefer-Ivan (approved):

dhsegel:
evergreen merge

PR #1146, opened 2024-03-08
PERF-5117 Create multi-planner performance workload with blocking sort

dhsegel:


src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -0,0 +1,404 @@
...
+  Threads: 1
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+
+- Name: SimpleMultiplannerWith64Indexes
```
Adityav369, line 271: Change the name of the actors. These are the same as Simple.yml. I suggest something like `MultiplannerWith64IndexesBlockingSort`. Same for other actor names.

src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -0,0 +1,404 @@
...
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone-sbe
+      - standalone-all-feature-flags  # At time of writing this will enable PM-3591.
+      - standalone-classic-query-engine
+    branch_name:
+      $gte: v7.3
```
Adityav369, line 404: [nit] Leave a blank line at end of file.

Adityav369 (changes requested):
Thanks @dhsegel. I added a few comments.

src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -0,0 +1,404 @@
...
+          x56: {$lte: 1},
+          x57: {$lte: 1},
+          x58: {$lte: 1},
+          x59: {$lte: 1},
+          x60: {$lte: 1},
+          x61: {$lte: 1},
+          x62: {$lte: 1},
+          x63: {$lte: 1}
+        },
+        Options: {Sort: '{"no_such_field": {"$numberInt": "1"}}'}
```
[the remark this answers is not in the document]
dhsegel: My inclination is to say yes, since the engine would still attempt to find and sort any instances of this field. @dpercy could you weigh in here? I'm not entirely sure.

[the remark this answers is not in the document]
dpercy: > > Would we get a blocking sort?
>
> My inclination is to say yes, since the engine would still attempt to find and sort any instances of this field.

That's what I was originally thinking.  The only way the sort could be non-blocking is if the server knew up front that the data is already sorted.  It can't tell from the query, and it can't tell from the contents of any index.  So this would only change if the server started maintaining some kind of exact schema information.

But if it's not too much work, it would probably be better to sort by an actual field, to be more realistic.  For example, [some sorting algorithms](https://en.wikipedia.org/wiki/Timsort) detect and exploit monotonically increasing runs, so we might not want to test such a happy path.

src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -0,0 +1,404 @@
...
+    Key: QueryTemplate
+    Parameters:
+      active: [3]
+      nopInPhasesUpTo: *maxPhase
+      repeat: *queryRepeats
+      database: *db
+      collection: *coll
+      query: &query {
+        Filter: {
+          x1: {$lte: 1},
```
Fefer-Ivan, line 283: Why are indexes all the same?

src/workloads/query/multiplanner/BlockingSort.yml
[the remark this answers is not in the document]
dhsegel: Sounds good, I'll go ahead and remove the last index so we can sort on an actual unindexed field.

src/workloads/query/multiplanner/BlockingSort.yml
dhsegel: The goal here is to ensure every plan is a blocking plan by adding a sort on an unindexed field. To accomplish this I've maintained the same structure as the simplest case but taken out `index2` so that we can sort on that field.

src/phases/query/Multiplanner.yml
```diff
@@ -26,15 +26,15 @@ QueryKnobTemplate:
 
 InsertTemplate:
   Name: CreateDataset
-  Type: Loader
-  Threads: &threads {^Parameter: {Name: "threads", Default: {unused: "Invalid thread count."}}}
+  Type: MonotonicLoader
```
Adityav369, line 29: I wonder if this might cause issues with existing tests that define an _id field. Maybe better to do this as a separate ticket for all our workloads?

Adityav369 (approved):
LGTM with one change that might cause issues with other tests. 
One more thing, very optional, maybe we could drop the `Simple` out of the actor names, since it's more than just simple with the sort and everything.

src/phases/query/Multiplanner.yml
dhsegel: Yeah, I had originally done this to maintain consistency with other workloads that I'm changing in a different ticket, but I see your point. I'll revert this here and make the change in PERF-5269.

Fefer-Ivan (approved):

PR #1147, opened 2024-03-13
PERF-5198 Rework AggregationsOutput workload to properly load initial documents

jordist:
- Get rid of `MonotonicSingleLoader` usages. Use `Loader`s instead.
- Based on PM-3229 performance analysis work observations (WRITING-17072), I redefined some of the workloads:
-- Test single thread
-- Test single document size
-- Test different data set sizes (limiting using $limit). This helps show constant overheads vs O(dataSetSize) components.

:evergreen_tree: [EVG](https://spruce.mongodb.com/version/65f075ae0305b939a52d7047/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/query/AggregationsOutput.yml
+++ b/src/workloads/query/AggregationsOutput.yml
@@ -2,31 +2,17 @@ SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
   This test exercises both $out and $merge aggregation stages. It does this based on all combinations of the following:
-    A. The src collection is sharded or unsharded
-    B. The document size is either 1K, 4K, or 16K.
-    C. There are 1 thread, 4 threads, or 16 threads doing the same operation.
-  Each actor does the following operations concurrently with other actors at the same stage:
-    A. $out to a collection unique amongst actors.
-       This measures $out performance.
-    B. $merge to the source collection with an updated field value.
-       This measures $merge contention on the same collection.
-    C. $merge to a different unsharded collection unique for each Actor with an updated field value.
-       This measures $merge contention on a separate collection shared by all threads of an Actor.
+    A. The src collection is sharded or unsharded.
+    B. The output collection being sharded or unsharded (only for $merge, as $out does not support sharded output collections).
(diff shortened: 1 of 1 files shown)


src/workloads/query/AggregationsOutput.yml
```diff
@@ -35,480 +21,184 @@ Clients:
...
-        Database: *Database
-        Collection: *Unsharded16K
-        DocumentCount: *NumDocs
-        BatchSize: 100
-        Document: *16KDoc
+      OnlyActiveInPhases:
+        Active: [{^Parameter: {Name: "ActivePhase", Default: 0}}]
+        NopInPhasesUpTo: *NumPhases
+        PhaseConfig:
+          Repeat: 10
```
jordiolivares, line 44: Are we sure only doing 10 repetitions will be enough? The first one should be quite quick to execute and I would expect a lot of noise for the first metric with only 10 executions. The 10k limit would be fine since there's a lot of docs to process and the noise would be dwarfed by the processing.

jordiolivares: To verify the noisiness I'd look at the output of multiple remote executions running this. If we see that it isn't noisy there I'd be open to mark this as resolved.

Otherwise what you suggest (increasing the number of repetitions) is fine since we have to get the mean value somehow and the only solution is brute-force in this case.

jordist: I've run a patch 4 times and the short operation metrics don't look super noisy when compared to the other ones. I'll increase the number of repetitions to 20 anyway. Does this sound good?

jordist: Forgot to attach the results:
Screenshot 2024-03-18 at 12 21 03

jordiolivares (approved):

src/workloads/query/AggregationsOutput.yml
jordiolivares: SGTM

jordist:
evvergreen merge

jordist:
evergreen merge

PR #1148, opened 2024-03-13
SERVER-85660: fix streams passthrough change stream source workload

harendra-c:
Jira Ticket: SERVER-85660

Whats Changed:
The change stream source should only be consumed after all documents have been written and durably persisted in the mongo collection, so the stream processor should be started after all the documents have been added to the source collection. In order to this we need to use startAtOperationTime in the $source.config to tell the stream processor to consume the documents in the change stream source from the beginning. It doesn't seem like there is a great way to generate a BSON timestamp object that mongo IDL likes through genny YAML or through the existing generators, so added {^NowTimestamp: {}} generator for BSON timestamps, similar to BSON date generators. (NOTE: None of the BSON date generators work here because the mongo IDL explicitly wants a BSON timestamp object for startAtOperationTime).

https://spruce.mongodb.com/version/65eb93e00305b9f009bb7f7e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/value_generators/src/DocumentGenerator.cpp
+++ b/src/value_generators/src/DocumentGenerator.cpp
@@ -1480,6 +1480,20 @@ private:
     std::string _actorId;
 };
 
+/** `{^NowTimestamp: {}}` */
+// For Timesamp, the default start value for .increment is 1 and .timestamp is current time since
+// UNIX epoch in seconds. 
+// see https://www.mongodb.com/docs/mongodb-shell/reference/data-types/#timestamp for more details.
+class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp> {
+public:
+    NowTimestampGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_timestamp evaluate() override {
+        auto now = std::chrono::system_clock::now();
+        uint32_t nowSinceEpoch = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch()).count();
+        return bsoncxx::types::b_timestamp{.increment = 1, .timestamp = nowSinceEpoch};
+    }
+};
+
 /** `{^Now: {}}` */
 class NowGenerator : public Generator<bsoncxx::types::b_date> {
 public:
@@ -2142,6 +2156,7 @@ Out valueGenerator(const Node& node,
  */
 const auto [allParsers,
             arrayParsers,
+            timestampParsers,
             dateParsers,
             doubleParsers,
             intParsers,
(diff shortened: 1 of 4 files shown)


mongodb-matthew-normyle (approved):
It looks good to me, but I don't have experience with the genny code. @thessem [reviewed @Aadeshp's prior change](https://github.com/mongodb/genny/pull/1118) a while back, it would be good to address @thessem's comments (if not already) and let @thessem take another look.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1480,6 +1480,17 @@ class ActorIdStringGenerator : public Generator<std::string> {
     std::string _actorId;
 };
 
+/** `{^NowTimestamp: {}}` */
```
thessem, line 1483: I have a feeling this could be more general. Let me know if I'm wrong with my thinking here, but isn't this closer in spirit to a type conversion of an existing set of generators we have? Aren't we going to need to add `^RandomTimestamp`, `^Timestamp` and `^IncTimestamp`? I think a decent API is something along the lines of `{^ConvertToTimestamp: {from: ^Now: {}}}` to copy the integer cast cases.

jimoleary (commented):
The general implementaiton looks good to me but I do have a concern about the usability of this implementation.

It would be very difficult for an end user to generate x different timestamps predictably. However this could be fixed by allowing the increment to be passed in.

OTOH, this could also be easily fixed later.

src/value_generators/src/DocumentGenerator.cpp
harendra-c: This change is to test the performance of Atlas stream processing. Just to give the background, the `Passthrough_ChangeStreamSource.yml` does processing on the documents that are inserted into a collection. The idea behind adding the `timestamp` is to keep track of the point `(StartTimestamp)` in time when we started the insertion. The stream processing job will start consuming the documents for processing from `(StartTimestamp)`, after all the documents are ingested.

This allows us to track the performance of the stream processing job without the delay caused by inserting documents in the collection.

The `timestamp` doesn't need to be unique.

src/value_generators/src/DocumentGenerator.cpp
```diff
@@ -1480,6 +1480,17 @@ class ActorIdStringGenerator : public Generator<std::string> {
...
+class NowTimestampGenerator : public Generator<bsoncxx::types::b_timestamp> {
+public:
+    NowTimestampGenerator(const Node& node, GeneratorArgs generatorArgs) {}
+    bsoncxx::types::b_timestamp evaluate() override {
+        auto now = std::chrono::system_clock::now();
+        uint32_t nowSinceEpoch = std::chrono::duration_cast<std::chrono::seconds>(now.time_since_epoch()).count();
+        return bsoncxx::types::b_timestamp{.increment = 1, .timestamp = nowSinceEpoch};
```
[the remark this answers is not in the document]
harendra-c: As per my understanding, bsoncxx::types::b_timestamp uses .increment to order (or compare) two timestamps in case the .timestamp is same.
For example,
if we have two documents with timestamp as
Doc 1 - Timestamp({ .timestamp: 1710804294, .increment: 1 }) 
Doc 2 - Timestamp({ .timestamp: 1710804294, .increment: 2 })
we use .increment to order these 2 documents.
For reference, [this](https://www.mongodb.com/docs/mongodb-shell/reference/data-types/#timestamp) is what I looked at.

I tried removing the .increment (as good as setting it to 0), the stream processing job did not pick any document from the collection.

jimoleary (approved):
LGTM with a document suggestion

PR #1154, opened 2024-03-18
PERF-5244 Create multiplanner performance workload with varied selectivity

dhsegel:


src/phases/query/Multiplanner.yml
```diff
@@ -147,3 +147,94 @@ Hide14IndexesTemplate:
         - {OperationName: RunCommand, OperationCommand: {collMod: *coll2, index: {hidden: true, name: index14}}}
         - {OperationName: RunCommand, OperationCommand: {collMod: *coll2, index: {hidden: true, name: index15}}}
         - {OperationName: RunCommand, OperationCommand: {collMod: *coll2, index: {hidden: true, name: index16}}}
+
+fraction: &fraction {^NumExpr: {withExpression: "resultCount / docCount", andValues: {
+                      resultCount: &resultCount {^Parameter: {Name: "resultCount", Default: {unused: "Invalid result count."}}},
+                      docCount: {^Parameter: {Name: "docCount", Default: {unused: "Invalid docCount."}}}}}}
+lowBound: &lowBound {^NumExpr: {withExpression: "random * (1 - fraction)", andValues: {
+                      random: 0.5,
```
dhsegel, line 155: For now, I have `random` set manually to 0.5. This is because `^NumExpr` is preprocessed, making it incompatible with `^RandomDouble` which is a generator. Furthermore, each use of a "variable" `^RandomDouble` is a literal substitution that re-generates a new random double, making it challenging to reference the same random number more than once. 

Can anyone think of a workaround to this problem? For now, the workload is functional but doesn't average out to the expected N.

Adityav369: Off the top of my head: Would `^FixedGeneratedValue` be useful? It calls the generator once and fixes the value it gets.

dhsegel: Unfortunately, this also doesn't work because the preprocessor attempts to parse the `^NumExpr` before the generation, and it errors out upon receiving "{^FixedGeneratedValue: {fromGenerator: {^RandomDouble: {min: 0.0, max: 1.0}}}}" rather than a numerical value.

```
genny.tasks.preprocess.ParseException: Invalid values for 'andValues' in '^NumExpr', only numerical values are allowed.
Node eval: {'random': {'^FixedGeneratedValue': {'fromGenerator': {'^RandomDouble': {'distribution': 'uniform', 'min': 0.0, 'max': 1.0}}}}, 'fraction': 0.0001}
```

dhsegel: Resolved as discussed in standup.

src/workloads/query/multiplanner/VariedSelectivity.yml
```diff
@@ -0,0 +1,368 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run a
```
chiihuang, line 4: Add the description about how it tests with varied selectivity.

chiihuang (changes requested):
Overall looks good. I just have one request to update the workload description.

src/phases/query/Multiplanner.yml
```diff
@@ -147,3 +147,95 @@ Hide14IndexesTemplate:
...
         - {OperationName: RunCommand, OperationCommand: {collMod: *coll2, index: {hidden: true, name: index15}}}
         - {OperationName: RunCommand, OperationCommand: {collMod: *coll2, index: {hidden: true, name: index16}}}
+
+fraction: &frac {^NumExpr: {withExpression: "resultCount / docCount", andValues: {
+                      resultCount: &resultCount {^Parameter: {Name: "resultCount", Default: {unused: "Invalid result count."}}},
+                      docCount: {^Parameter: {Name: "docCount", Default: {unused: "Invalid docCount."}}}}}}
+
+FindWithNExpectedResults:
+  Name: FindWithNExpectedResults
+  Type: RunCommand
```
Fefer-Ivan, line 157: As I mentioned on slack, the fundamental problem with this is that we don't exhaust or close the cursor, returned by the find command here.

This leads to two problems:
1. We don't test actual whole query execution.
2. We leave open cursors that may lead to RAM overflow.

We can solve it by setting batchSize to be bigger than the number of expected results.
However, if result set is bigger than 16 MBs, it will lead to the same problems.

The biggest issue with this is that the problem is "silent" - we won't notice it and just get bad data.

So I suggest that you file a ticket to support let option in CrudActor and do it.
It will be useful addition to genny and you get to write some C++ code instead of genny yaml files!

We can do this workload later after you are done with this.

If we want to start gathering perf data for this,
let's add cursor batchSize option to this query, so we can start getting data while figuring out let in CrudActors.

src/phases/query/Multiplanner.yml
```diff
@@ -147,3 +147,95 @@ Hide14IndexesTemplate:
...
+        Database: {^Parameter: {Name: "database", Default: {unused: "Invalid database name."}}}
+        Operation: 
+          {OperationName: RunCommand, OperationCommand: {
+            find: {^Parameter: {Name: "collection", Default: {unused: "Invalid collection name."}}},
+            let: {
+              lowBound: {$multiply: [{$subtract: [1, *frac]}, {^RandomDouble: {distribution: uniform, min: 0.0, max: 1.0}}]}, 
+              size: *frac
+            },
+            filter: {$expr: {$and: [
+                      # Confirm that lowBound <= x1 < lowBound + size
```
Fefer-Ivan, line 174: Did you check with explain() that optimizer is smart enough to constant fold this and extract correct index bounds?

src/workloads/query/multiplanner/VariedSelectivity.yml
```diff
@@ -0,0 +1,368 @@
...
+    Parameters:
+      active: [8]
+      nopInPhasesUpTo: *maxPhase
+      repeat: *queryRepeats
+      database: *db
+      collection: *coll
+      docCount: *docCount
+      resultCount: 300
+
+- Name: SimpleMultiplannerWith1000ExpectedResults
```
Fefer-Ivan, line 347: [optional] Do we need `Simple` here?

Fefer-Ivan (changes requested):

src/phases/query/Multiplanner.yml
dhsegel: Yep, everything looks good in that regard.

Fefer-Ivan (approved):

dhsegel:
evergreen merge

chiihuang (approved):

dhsegel:
evergreen merge

PR #1153, opened 2024-03-18
PERF-5246 Add multiplanner workload where query predicates match no documents

Adityav369:
**Jira Ticket:** PERF-5246

**Whats Changed:**  
Add multiplanner workload where query predicates match no documents

**Patch testing results:**  
https://spruce.mongodb.com/version/65f876e0aaa0f80007c5f141/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1155, opened 2024-03-18
PERF-5269 Replace Loader with MonotonicLoader in existing multiplanner workloads

dhsegel:


Fefer-Ivan (approved):
Awesome

Adityav369 (approved):
LGTM. 
Sorry for jumping in on the review uninvited but heads up: https://github.com/mongodb/genny/pull/1153 is in the commit queue. Please merge master and add `NoResults.yml`. As for the other PRs, seems like this would be in commit queue before them, so we could merge master on those.

dhsegel:
evergreen merge

PR #1157, opened 2024-03-19
PERF-5245 Create multiplanner performance workload where all plans are bad

dhsegel:


src/workloads/query/multiplanner/NoGoodPlan.yml
```diff
@@ -0,0 +1,403 @@
...
+  Threads: 1
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+
+- Name: SimpleMultiplannerWith64Indexes
```
Fefer-Ivan, line 271: Simple?

src/workloads/query/multiplanner/NoGoodPlan.yml
```diff
@@ -0,0 +1,403 @@
...
+    Key: QueryTemplate
+    Parameters:
+      active: [3]
+      nopInPhasesUpTo: *maxPhase
+      repeat: *queryRepeats
+      database: *db
+      collection: *coll
+      query: &query {
+        Filter: {
+          x1: {$lt: *selectivity},
```
Fefer-Ivan, line 283: So this looks like a good plan?

Fefer-Ivan (changes requested):

src/workloads/query/multiplanner/NoGoodPlan.yml
dhsegel: Removing.

src/workloads/query/multiplanner/NoGoodPlan.yml
dhsegel: What do you mean? Wouldn't marking `index1` as hidden mean that there is no good plan?

Fefer-Ivan: Oh, now I see.
That is too easy to miss, given copy-paste nature of this workloads.
Why do we go this way?
Let's just add new "unindexed" field, for example?

Fefer-Ivan: This will be more readable.

src/workloads/query/multiplanner/NoGoodPlan.yml
```diff
@@ -16,14 +16,14 @@ GlobalDefaults:
   # Collection name used for queries.
   coll: &coll Collection0
 
-  docCount: &docCount 1e5
+  docCount: &docCount 1e2
```
Fefer-Ivan, line 19: Should this be reverted before merge?

Fefer-Ivan (approved):
LGTM with a comment

dhsegel:
evergreen merge

PR #1159, opened 2024-03-19
PERF-5120 Add two multiplanner focused clustered collection workloads

Fefer-Ivan:
**Jira Ticket:** PERF-5120

**Whats Changed:**  
ClusteredCollection workload that has huge _id.
UseClusteredIndex workload that has predicate on _id as selective as predicate on x1, making clustered index tied with x1 index.

**Patch testing results:**  
https://spruce.mongodb.com/version/65fc278f3a0f5b000781f94d/tasks

Note: it depends on the monotonic loader PR #1155

### DIFF
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -275,7 +275,7 @@ Actors:
       PhaseConfig:
         Repeat: 1
 
-- Name: MultiplannerWith64Indexes
+- Name: ClusteredCollectionMultiplannerWith64Indexes
   LoadConfig:
     Path: "../../../phases/query/Multiplanner.yml"
     Key: QueryTemplate
@@ -364,7 +364,7 @@ Actors:
       database: *db
       collection: *coll
 
-- Name: MultiplannerWith16Indexes
+- Name: ClusteredCollectionMultiplannerWith16Indexes
   LoadConfig:
     Path: "../../../phases/query/Multiplanner.yml"
     Key: QueryTemplate
@@ -387,7 +387,7 @@ Actors:
       database: *db
       collection: *coll
 
-- Name: MultiplannerWith2Indexes
+- Name: ClusteredCollectionMultiplannerWith2Indexes
   LoadConfig:
     Path: "../../../phases/query/Multiplanner.yml"
     Key: QueryTemplate
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -1,13 +1,9 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/query"
 Description: |
-  The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run a
-  query that makes all of them eligible, so we get as many competing plans as possible.
-
-  The original goal of this test was to demonstrate weaknesses of the SBE multiplanner when compared to
-  the classic multiplanner. Mainly, the SBE multiplanner can't round-robin between plans, which means it
-  has to run the list of plans sequentially, which means we can't short-circuit when the shortest-running
-  plan finishes.
+ The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run a
+  query that makes all of them eligible, so we get as many competing plans as possible. Here, we do this on a
+  clustered collection that has very large strings as _id.
 
   We expect classic to have better latency and throughput than SBE on this workload,
   and we expect the combination of classic planner + SBE execution (PM-3591) to perform about
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -1,0 +1,410 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run a
+  query that makes all of them eligible, so we get as many competing plans as possible.
+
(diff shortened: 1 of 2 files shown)


src/workloads/query/multiplanner/ClusteredCollection.yml
```diff
@@ -0,0 +1,410 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
```
chiihuang, line 3: Update the test description to reflect how it tests with cluster index.

src/workloads/query/multiplanner/UseClusteredIndex.yml
```diff
@@ -0,0 +1,410 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
```
chiihuang, line 3: ditto

src/workloads/query/multiplanner/UseClusteredIndex.yml
```diff
@@ -0,0 +1,410 @@
...
+    Key: InsertTemplate
+    Parameters:
+      database: *db
+      threads: 1
+      active: [1]
+      nopInPhasesUpTo: *maxPhase
+      repeat: 1
+      collection: *coll
+      docCount: *docCount
+      document: {
```
chiihuang, line 73: [question] Does this also needs `_id: {^FastRandomString: {length: 512}},` for large field ? Or is it intended to make `_id` and `x` share similar size?

src/workloads/query/multiplanner/ClusteredCollection.yml
```diff
@@ -0,0 +1,410 @@
...
+  Threads: 1
+  Database: *db
+  Phases:
+    OnlyActiveInPhases:
+      Active: [2]
+      NopInPhasesUpTo: *maxPhase
+      PhaseConfig:
+        Repeat: 1
+
+- Name: MultiplannerWith64Indexes
```
chiihuang, line 278: Make the actor name unique among tests such as `MultiplannerWith64IndexesClusteredCollection`. Same for the 16 and 2 indexes ones in this workload file.

chiihuang (changes requested):
Overall looks good. Just a few change requests on comments and a question regarding the large field in UseClusteredIndex workload.

src/workloads/query/multiplanner/UseClusteredIndex.yml
Fefer-Ivan: In this workload we want to use clustered index, so incremental _id's, generated by the monotonic loader should be good.

src/workloads/query/multiplanner/ClusteredCollection.yml
Fefer-Ivan: I am not sure if we want to do that.
In performance data the name of the workload would also be present, so in the end we will get something like
UserClusteredIndex.MultiplannerWith64Indexes - which is pretty clear. 
Adding a second mention of clustered index sounds like an overkill

dhsegel: I think either would work, but it should be consistent across all of our workloads. I have been giving each actor a distinct name up to now, but if we maintained consistency, the data might look something like this:
| variant                                              | task                          | test                           |
|------------------------------------------------------|-------------------------------|--------------------------------|
| perf-standalone-all-feature-flags.arm.aws.2023-11    | multiplanner_simple           | MultiplannerWith64Indexes.Crud |
| perf-standalone-classic-query-engine.arm.aws.2023-11 | multiplanner_simple           | MultiplannerWith64Indexes.Crud |
| perf-standalone-all-feature-flags.arm.aws.2023-11    | multiplanner_compound_indexes | MultiplannerWith64Indexes.Crud |
| perf-standalone-classic-query-engine.arm.aws.2023-11 | multiplanner_compound_indexes | MultiplannerWith64Indexes.Crud |

It effectively comes down to a question of whether we want to be referencing both `task` and `test` or whether we want to get all the identification information from `test`. I would see a benefit in maintaining both as we could filter down to workloads' data for `MultiplannerWith64Indexes` if that is something we want to do.

dhsegel (approved):
LGTM, barring description and naming updates.

src/workloads/query/multiplanner/ClusteredCollection.yml
chiihuang: Makes sense. Either way seems good to me. @dpercy in terms of notebook analysis, which way is preferable? 

One benefit of avoiding second mentioning in data analysis is to group the tests across tasks, or having a table like:

| task \ test |  MultiplannerWith2Indexes | MultiplannerWith16Indexes | MultiplannerWith64Indexes |
| -- | -- | -- | -- |
| multiplanner_simple | 50ms | 51ms | 52ms |
| multiplanner_compound_indexes | 150ms | 200ms | 250ms |

Fefer-Ivan: I will follow the current pattern. Let's discuss during stand-up if we should do an update on that

Fefer-Ivan: Done

src/workloads/query/multiplanner/UseClusteredIndex.yml
Fefer-Ivan: Done

src/workloads/query/multiplanner/ClusteredCollection.yml
Fefer-Ivan: Done

chiihuang (approved):

Fefer-Ivan:
evergreen merge

PR #1167, opened 2024-03-21
PERF-5289 Ensure all high-value workloads run on all-feature-flags variants

BlakeIsBlake:
**Jira Ticket:** PERF-5289

**Whats Changed:**  
Self-explanatory

**Patch testing results:**  
None

**Workload Submission form:**  
None

**Related PRs:**   
None

### DIFF
--- a/src/workloads/query/ExpressiveQueries.yml
+++ b/src/workloads/query/ExpressiveQueries.yml
@@ -189,6 +189,7 @@ AutoRun:
       - atlas
       - atlas-like-replica.2022-10
       - replica
+      - replica-all-feature-flags
       - single-replica
       - standalone
       - standalone-classic-query-engine
--- a/src/workloads/query/ArrayTraversal.yml
+++ b/src/workloads/query/ArrayTraversal.yml
@@ -273,7 +273,6 @@ AutoRun:
       - standalone-sampling-bonsai
       - standalone-sbe
       - replica
-      - replica-all-feature-flags
     branch_name:
       $neq:
       - v4.0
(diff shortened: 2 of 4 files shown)


src/workloads/query/ExpressiveQueries.yml
```diff
@@ -189,6 +189,7 @@ AutoRun:
       - atlas
       - atlas-like-replica.2022-10
       - replica
+      - replica-all-feature-flags
```
jimoleary, line 192: Should this also run on standalone-all-feature-flags?

jimoleary (commented):
Apologies for possibly maybe making this much larger but the full list of genny HVW is [here](https://github.com/10gen/performance-tooling-docs/blob/573e9907173efa78e95d859f04bc7ee6ed1f32d9/workloads/high-value-workloads.md?plain=1#L180). 

Should these also run on a '_*-all-feature-flags_' variant?
And what's the rationale between _standalone / standalone-all-feature-flags_ v _replica / replica-all-feature-flags_ ?

BlakeIsBlake:
The tests I included are only high-value on 3-node replica set variants. In fact, I removed ArrayTraversal since in that same high-value workload list I missed that it's only high-value on standalone, which is already included.

BlakeIsBlake:
And I don't think where you linked is the high-value workloads, I think it's an appendix of all of the Genny workloads that Build Barons track!

jimoleary (approved):
LGTM

BlakeIsBlake:
evergreen merge

PR #1168, opened 2024-03-25
PERF-5294 Change VariedSelectivity.yml to use 2 indexes instead of 64

dhsegel:


src/workloads/query/multiplanner/VariedSelectivity.yml
```diff
@@ -1,12 +1,10 @@
...
-  The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run
-  a query that makes all of them eligible, so we get as many competing plans as possible. We run
-  this query 7 times, each one with a different selectivity value that we are comparing against x1,
-  calcuated based on the number of documents we want the query to match. This will help us measure
-  the overhead of throwing out the result set gathered during multi-planning when the result set
-  exceeds 101 documents.
+  The goal of this test is to exercise multiplanning. We run the same query 7 times, each one with a
+  different selectivity value that we are comparing against x1, calcuated based on the number of
+  documents we want the query to match. This will help us measure the overhead of throwing out the
+  result set gathered during multi-planning when the result set exceeds 101 documents.
```
dpercy, line 7:
```suggestion
  result set gathered during multi-planning when the result set exceeds 101 documents.
  
  Unlike many of the other multiplanner/ workloads, we only test with 2 indexes here, because
  2 indexes is a worst case for throwing away results. Having more indexes increases planning time,
  but not query execution time, so having more indexes makes the *relative* cost of throwing away
  results smaller.
```

dpercy (approved):

dhsegel:
evergreen merge

dhsegel:
evergreen merge

PR #1169, opened 2024-03-26
PERF-5302: Properly process port number environment variables captured by Poplar

wqian94:
This fixes a potential bug with the Genny hook via Curator into Poplar. [Poplar](https://github.com/mongodb/curator/blob/a1293258b30e04ea503e6b91b6b1628789e90093/operations/poplar.go#L21C34-L21C49) accepts the `POPLAR_RPC_HOST` and the `POPLAR_RPC_PORT` environment variables and defaults to `localhost` and `2288` if those environment variables are not set.

This change properly handles the `POPLAR_RPC_PORT` environment variable when setting up Genny to connect to Poplar. Prior to this change, Genny would attempt to connect to port `2288`, regardless of where the Poplar service was actually listening.

### DIFF
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -150,6 +150,7 @@ def poplar_grpc(cleanup_metrics: bool, workspace_root: str, genny_repo_root: str
             poplar = subprocess.Popen(
                 args, stdout=log_file_obj, stderr=subprocess.STDOUT, text=True
             )
+            poplar_port = int(os.environ.get("POPLAR_RPC_PORT", 2288))
             if poplar.poll() is not None:
                 raise OSError("Failed to start Poplar.")
             try:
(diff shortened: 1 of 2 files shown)


dalyd:
I'm going to remove myself from this one as I don't understand Poplar well enough to evaluate this change. Maybe @johndaniels or @zamj do. Minh has today and Monday off and should then be back online.

thessem:
I added @PyGeek03 as a reviewer on behalf of @mongodb/devprod-performance-infrastructure

wqian94:
> I'm going to remove myself from this one as I don't understand Poplar well enough to evaluate this change. Maybe @johndaniels or @zamj do. Minh has today and Monday off and should then be back online. 

Ah sorry, just meant to tag you here is all. We had a brief discussion about this a while back and I figured you'd be interested to know that it's possible to make concurrent Genny runs happen.

thessem (approved):
Please see previous comment, leaving this as a tick now

PR #1170, opened 2024-03-26
PERF-5304 Add best-case scenario for SBE multi-planner, compared to Classic

Fefer-Ivan:
**Jira Ticket:** PERF-5304

**Whats Changed:**  
Added QueryWithMultiplanningAndTieBreakingHeuristics that runs a query for which tie breaking heuristics will guess the correct index. Expected result would be that SBE multi planner will perform better for this kind of queries.

**Patch testing results:**  
https://spruce.mongodb.com/version/6602ecc5f7a3490007adbe2a/tasks

### DIFF
--- a/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
+++ b/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
@@ -4,6 +4,13 @@ Description: >
   Create collection with multiple indexes and run queries for which multi planner needs to read a lot
   of data in order to pick the best plan.
 
+  First CrudActor QueryWithMultiplanning will use queries specifically design to trick tie
+  breaking heuristics, so that if the planner doesn't reach enough data, it will pick the wrong
+  index.
+
+  Second CrudActor QueryWithMultiplanningAndTieBreakingHeuristics will use queries for which tie
+  breaking heuristics are guessing the correct index.
+
 Keywords:
   - indexes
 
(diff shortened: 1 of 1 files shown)


src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
```diff
@@ -4,6 +4,13 @@ Description: >
   Create collection with multiple indexes and run queries for which multi planner needs to read a lot
   of data in order to pick the best plan.
 
+  First CrudActor QueryWithMultiplanning will use queries specifically design to trick tie
+  breaking heuristics, so that if the planner doesn't reach enough data, it will pick the wrong
+  index.
+
+  Second CrudActor QueryWithMultiplanningAndTieBreakingHeuristics will use queries for which tie
+  breaking heuristics are guessing the correct index.
+
```
dpercy, line 13:
```suggestion
    In both cases, heuristics lead us to pick the (flag_a, flag_b) index.  In the first case, this is a bad choice:
    the two boolean predicates select 25% of documents while the int_a predicate is much narrower.
    In the second case, we still choose the (flag_a, flag_b) index, but this time it is the correct choice,
    because the int_a predicate matches >50% of documents.
```

dpercy (approved):

Fefer-Ivan:
evergreen merge

src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
dpercy: Paraphrasing what you explained in Slack

Thanks!  I was confused because I thought the new case would be trying to trick the heuristics into picking the better index, (int_a); actually it's about changing the query so that the chosen (flag_a, flag_b) is the right choice.

PR #1171, opened 2024-03-26
DEVPROD-5757 Require owners, descriptions, and keywords in all workload files.

ahartschen:
**Jira Ticket:** DEVPROD-5757

**Whats Changed:**  
We now require that all workload files have these required fields. We made some exceptions for older files that did not have keywords. Any future workloads will require all the fields. I also changed things up to disable the formatting check because this was disabled about 2 years ago and there were ~13k errors that need corrected. I opened DEVPROD-5682 to have this addressed.

**Patch testing results:**  
Running the CLI locally, no output since there are no errors:

```
(genny_venv) ➜  genny git:(DEVPROD-5757) ./run-genny lint-yaml --no-format
(genny_venv) ➜  genny git:(DEVPROD-5757) 
```

**Workload Submission form:**  

**Related PRs:**   
N/A

PLEASE READ AND REMOVE THE SECTION BELOW BEFORE SUBMITTING YOUR PR FOR REVIEW

1.  Once (1) you get approval from the STM or the Product Perf team for your change and (2) all your CI tests pass, you can initiate a merge by entering a comment on your PR with the text

    > evergreen merge

    This kicks your PR into the [evergreen commit queue][ecq]. Evergreen will automatically squash-merge your commit after checking that you have PR approval and that the CI tests all pass.

[ecq]: https://github.com/evergreen-ci/evergreen/wiki/Commit-Queue


thessem (approved):
From a perf infra point of view everything here looks good. I'd still wait for somebody from Product Perf on the test-config implications.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -26,37 +32,47 @@ def main(genny_repo_root: str):
...
-def check_description(yaml_path):
+def check_required_fields(yaml_path):
     workload = _load_yaml(yaml_path)
-    if "Description" not in workload:
-        SLOG.error(f"Genny workload {yaml_path} lacks a Description field.")
-        return False
-    return True
+    no_errors = True
+    for field in REQUIRED_FIELDS:
+        if field not in workload:
```
MAhmadShah, line 70: NIT: Do you think it's also worth checking that the field actually has some text in it? There could be a loop hole here where users can just put the key in with an empty value.

MAhmadShah:
Just one comment other wise LGTM!

MAhmadShah (approved):

docs/using.md
```diff
@@ -450,12 +450,14 @@ Try using `python test_result_summary.py --help` for more options.
 2.  Run the self-tests:
 
     ```bash
-	./run-genny lint-yaml  # Lint all YAML files
+	./run-genny lint-yaml  --no-format # Lint all YAML files
```
jimoleary, line 453: Do we need ```--no-format``` here if the default is false?

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -26,37 +32,47 @@ def main(genny_repo_root: str):
...
-        yamllint_argv=yamllint_argv,
-    )
+        SLOG.info(
+            "Linting workload YAML files with yamllint",
+            count=len(all_yamls),
+            yamllint_argv=yamllint_argv,
+        )
 
-    yamllint.cli.run(yamllint_argv)
+        yamllint.cli.run(yamllint_argv)
```
jimoleary, line 60: How are errors handled here? What do we do if ```yamllint.cli.run(yamllint_argv)``` fails?

jimoleary (changes requested):
The ```no_format``` variable name looks a bit odd, can we change it to ```lint_format```?

This hurts my head less.

I assume you went do this path so as not to override the builtin ```format``` function.

docs/using.md
jimoleary: oh yeah, let's make that the default for now.

src/lamplib/src/genny/cli.py
```diff
@@ -572,11 +572,14 @@ def self_test(ctx: click.Context):
 
 
 @cli.command(name="lint-yaml", help="Run pylint on all workload and phase yamls")
+@click.option(
+    "--no-format", is_flag=True, help="Do not lint the yaml formatting.", default=False, type=bool
+)
 @click.pass_context
-def lint_yaml(ctx: click.Context):
+def lint_yaml(ctx: click.Context, no_format: bool):
```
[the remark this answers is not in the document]
jimoleary: lint_format defaulting to False or no_format to True would be the sensible state .

jimoleary (approved):
LGTM

I think we are probably on the hook for MongosyncScripts.yml, if I find out anything different after this merges ... I'll change it.

PR #1174, opened 2024-03-27
PERF-5293 Add subplanning workload

Fefer-Ivan:
**Jira Ticket:** PERF-5293

**Whats Changed:**  
Add subplanning workload

**Patch testing results:**  
https://spruce.mongodb.com/version/660452f15813710007edf328/tasks

### DIFF
--- a/src/workloads/query/multiplanner/Subplanning.yml
+++ b/src/workloads/query/multiplanner/Subplanning.yml
@@ -4,7 +4,7 @@ Description: |
   The goal of this test is to exercise subplanning. We create as many indexes as possible, and run a
   query that makes all of them eligible, so we get as many competing plans as possible.
 
-  The workload creates an $or query with 8 clauses each containing 8 predicates. Each branch have
+  The workload uses an $or query with 8 clauses each containing 8 predicates. Each branch have
   only one selective predicate.
 
   We expect classic to have better latency and throughput than SBE on this workload,
--- a/src/workloads/query/multiplanner/Subplanning.yml
+++ b/src/workloads/query/multiplanner/Subplanning.yml
@@ -1,0 +1,357 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  The goal of this test is to exercise subplanning. We create as many indexes as possible, and run a
+  query that makes all of them eligible, so we get as many competing plans as possible.
+
+  The original goal of this test was to demonstrate weaknesses of the SBE multiplanner when compared to
+  the classic multiplanner. Mainly, the SBE multiplanner can't round-robin between plans, which means it
+  has to run the list of plans sequentially, which means we can't short-circuit when the shortest-running
+  plan finishes.
+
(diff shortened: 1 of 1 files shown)


src/workloads/query/multiplanner/Subplanning.yml
```diff
@@ -0,0 +1,357 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query"
+Description: |
+  The goal of this test is to exercise subplanning. We create as many indexes as possible, and run a
+  query that makes all of them eligible, so we get as many competing plans as possible.
+
+  The original goal of this test was to demonstrate weaknesses of the SBE multiplanner when compared to
```
dhsegel, line 7: Could you update this with a more specific description for this workload?

src/workloads/query/multiplanner/Subplanning.yml
```diff
@@ -0,0 +1,357 @@
...
+        x56: *distribution,
+        x57: *distribution,
+        x58: *distribution,
+        x59: *distribution,
+        x60: *distribution,
+        x61: *distribution,
+        x62: *distribution,
+        x63: *distribution
+      }
+      indexes:
```
dhsegel, line 133: [optional] If we're not using index hiding, we could shorten this by removing the names of all the indexes. But it might be better to maintain consistency across all our workloads.

src/workloads/query/multiplanner/Subplanning.yml
```diff
@@ -0,0 +1,357 @@
...
+  LoadConfig:
+    Path: "../../../phases/query/Multiplanner.yml"
+    Key: QueryTemplate
+    Parameters:
+      active: [3]
+      nopInPhasesUpTo: *maxPhase
+      repeat: *queryRepeats
+      database: *db
+      collection: *coll
+      query: &query
```
dhsegel, line 282: No need for `&query` as we're only using this once

dhsegel (approved):
LGTM, with comments

src/workloads/query/multiplanner/Subplanning.yml
Fefer-Ivan: I don't think that this is necessary.
The query in the workload is pretty straight-forward and I don't want to have to change description every time we tweak the query.

src/workloads/query/multiplanner/Subplanning.yml
Fefer-Ivan: Consistency

src/workloads/query/multiplanner/Subplanning.yml
Fefer-Ivan: If we were to add index hiding in the future, it would be useful. Let's just leave it here, again, for consisntency

src/workloads/query/multiplanner/Subplanning.yml
Fefer-Ivan: I have added the query description

Fefer-Ivan:
evergreen merge

PR #1176, opened 2024-03-29
SERVER-88741 Create 8.0-to-enable feature flag performance variants

BlakeIsBlake:


src/workloads/execution/ClusteredCollectionLargeRecordIds.yml
```diff
@@ -20,15 +20,18 @@ AutoRun:
...
       - standalone-all-feature-flags
       - standalone-query-stats
       - replica
+      - replica-80-feature-flags
       - replica-all-feature-flags
       - replica-query-stats-rate-limit
       - atlas-like-replica.2022-10
       - atlas-like-replica.2023-09
       - atlas-like-replica-query-stats.2023-09
       - shard
```
erin2722, line 33: Do we want `configurations/mongodb_setup/mongodb_setup.shard-80-feature-flags.yml` wherever there is the `shard` setup present?

erin2722: Asking for the `shard` setup specifically because it does not have an equivalent `shard-all-feature-flags` setup to grep for right now.

erin2722 (approved):
LGTM! Just left one question.

I reviewed by searching the genny workloads directory for `all-feature-flags` and confirming that your changes were present everywhere.

src/workloads/execution/ClusteredCollectionLargeRecordIds.yml
erin2722: Although I am guessing we likely don't want that, but just confirming :)

BlakeIsBlake: I'm not sure yet... I think the answer should be yes but I'm conflicted on whether that should be part of this ticket and if we'd need approval from performance teams to do so. You can say on one hand that since we're adding in everything else _but_ genny workloads this is completeness though.

erin2722: Yeah, lets break it off into another ticket then, or just leave it for the perf team to decide 👍

BlakeIsBlake: I added them. I chose workloads that do shard-lite-feature flags and also workloads that do shard that seem relevant.

BlakeIsBlake:
evergreen merge

src/workloads/execution/ClusteredCollectionLargeRecordIds.yml
erin2722: Ok nice!

PR #1177, opened 2024-03-29
DEVPROD-5910 Add generated documentation for workload files.

ahartschen:
**Jira Ticket:** DEVPROD-5910

**Whats Changed:**  

I added a new command to generate markdown documentation based on the workload yamls. There is an evergreen task that will check in the PR and the merge queue that the documentation has been updated with each commit. I also changed a few of the workload yaml's descriptions to be better formatted in the markdown, this included using `|` instead of `>` when doing multi-line strings because this respects newlines when parsing.

**Patch testing results:**  
[Example patch](https://spruce.mongodb.com/version/6605cfe230fe360007d4dce1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) that failed when the markdown didn't match

**Workload Submission form:**  

**Related PRs:**   

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -32,6 +32,16 @@ Lint Workload and other YAML:
 ./run-genny lint-yaml
 ```
 
+### Workload Documentation Generation
+
+Markdown documentation is generated for the workload YAML files. When a workload is updated run the following command:
+
+```sh
+./run-genny generate-docs
+```
+
+An evergreen task verifies that docs have been updated for each commit. If this task fails you need to run the `generate-docs` command and include the updated documentation in your commit.
+
 ### C++ Unit-Tests
 
 ```sh
--- a/src/lamplib/src/genny/tasks/documentation_generator.py
+++ b/src/lamplib/src/genny/tasks/documentation_generator.py
@@ -1,0 +1,75 @@
+from os import path
+import os
+from re import S
+from typing import NamedTuple
+
+from jinja2 import Environment, PackageLoader, select_autoescape
+import structlog
+import yaml
+
+SLOG = structlog.get_logger(__name__)
+
+
+class Workload(NamedTuple):
+    name: str
+    owner: str
+    description: str
+    keywords: list[str]
+    path: str
+
+
+def main(genny_repo_root: str):
+    workload_dirs = [
+        path.join(genny_repo_root, "src", "workloads"),
+        path.join(genny_repo_root, "src", "phases"),
+    ]
+    SLOG.info("Gathering workloads from files.", workload_dirs=workload_dirs)
+
+    workloads = [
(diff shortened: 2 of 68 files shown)


jimoleary:
I've added brad as well as I am on vacation next week. I'll do an initial read now anyway.

docs/developing.md
```diff
@@ -32,6 +32,16 @@ Lint Workload and other YAML:
 ./run-genny lint-yaml
 ```
 
+### Workload Documentation Generation
+
+Markdown documentation is generated for the workload YAML files. When a workload is updated run the following command:
```
jimoleary, line 37: ~do we automatically regenerate a new docs suite whenever a workload changes?~

Based on ```f_check_generated_docs``` it looks like we don't but the merge fails until you run it.

src/phases/execution/ClusteredCollection.yml
```diff
@@ -1,6 +1,6 @@
 SchemaVersion: 2018-07-01
 Owner: "@mongodb/server-execution"
-Description: >
+Description: |
```
jimoleary, line 3: folded block scalars (>) v literal block scalars (|) are going to get tiresome and ugly really quickly ... do we have any way of enforcing the literal block scalars?

src/lamplib/src/genny/tasks/documentation_generator.py
```diff
@@ -0,0 +1,75 @@
...
+            for filename in filenames:
+                if filename.endswith(".yml"):
+                    yaml_files.append(path.join(dirpath, filename))
+    return yaml_files
+
+
+def get_workload_from_file(yaml_path: str) -> Workload:
+    with open(yaml_path, "r") as f:
+        workload_yaml = yaml.safe_load(f)
+        workload_name = f.name.split("/")[-1].split(".")[0]
```
jimoleary, line 56: Suggestion: [pathlib.Path.stem](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)  would be a better way to get the workload name.

jimoleary (approved):
LGTM with a suggestion.

Thanks for fixing all the descriptions!

docs/developing.md
jimoleary: I prefer this UX to me forgetting and having broken docs! Automated would be better, but there's no way to guarantee that it is run.

src/phases/execution/ClusteredCollection.yml
jimoleary: kk lets leave that for later then.

jimoleary (approved):
still LGTM.

One note on the pathlib .. I think it requires python >= 3.4 but I think we are >= 3.7 anyway.

PR #1180, opened 2024-04-02
SERVER-88846 Fix Streams perf harness system failures due to missing 'options' field

harendra-c:
Jira Ticket: SERVER-88846

Whats Changed:  
The start_StreamProcessor command has an options field which is not an optional field anymore. That is causing the workloads to fail. This change is to fix the broken stream workloads. Adding an empty options field in the phase which starts a stream processor.

https://spruce.mongodb.com/version/660b8c6b1f4649000769fdd7/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


mongodb-matthew-normyle (approved):
lgtm

PR #1182, opened 2024-04-03
DEVPROD-6041 Documentation generation improvements.

ahartschen:

**Jira Ticket:** DEVPROD-6041

**Whats Changed:**  
1. Made separate documentation files for workloads and phases.
2. Added hyperlinks to the respective workload and phase yamls in the docs.

### DIFF
--- a/docs/generated/workloads.md
+++ b/docs/generated/workloads.md
@@ -3915,8 +3915,7 @@ it uses a ^Once generator to generate the arrays once during initialization.
 
 
 
-<<<<<<< HEAD
-## LargeIndexedInsMatchingDocuments
+## [LargeIndexedInsMatchingDocuments](https://www.github.com/mongodb/genny/blob/master/src/workloads/scale/LargeIndexedInsMatchingDocuments.yml)
 ### Owner
 @mongodb/product-perf
 ### Description
@@ -3951,10 +3950,7 @@ The important metrics for this workload are:
 Loader, CrudActor, find, $in, matching documents using $in 
 
 
-## LargeScaleLongLived
-=======
 ## [LargeScaleLongLived](https://www.github.com/mongodb/genny/blob/master/src/workloads/scale/LargeScaleLongLived.yml)
->>>>>>> d9bc52e3 (DEVPROD-6041 Documentation generation improvements.)
 ### Owner
 Storage Engines
 ### Description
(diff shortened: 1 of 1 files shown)


src/lamplib/src/genny/templates/documentation.md.j2
```diff
@@ -1,6 +1,6 @@
-# Workload Documenation
+# Workload Documentation 
 {% for workload in workloads %}
```
dalyd, line 2: Can we add a line break inside this loop, either at the top or bottom? I think 
```
---
```
Would be sufficient.

dalyd (approved):
The change LGTM. I have two requests that can be in this PR or separate. 
1. Can we add a horizontal line between entries? 
2. Can we link to the generated docs from https://github.com/mongodb/genny/blob/3765880c20407c696af4c56568fa932b4940c342/docs/start.md?

src/lamplib/src/genny/templates/documentation.md.j2
```diff
@@ -1,6 +1,6 @@
-# Workload Documenation
+# Workload Documentation
```
SimonEismann, line 1: The added space here makes me think that we aren't linting this. That's probably okay in the commit queue, but running some linter over this locally might be a good idea.

docs/generated/workloads.md
```diff
@@ -1,925 +1,6 @@
-# Workload Documenation
+# Workload Documentation
```
SimonEismann, line 1: We need some introductory text here:
* What workloads does this cover?
* What information is available for each workload?
* Where does this documentation come from?
* What to do if documentation is missing/bad?
* ...

docs/generated/workloads.md
SimonEismann, line 1: This is a bit ambitious, so feel free to ignore.

We could have two sections here:
* Active Workloads (workloads with an autorun section that run on sys-perf)
* Other Workloads (the rest)

My concern is that users might find a workload on this page and then go looking for it in sys-perf and not find it.

SimonEismann (changes requested):

src/lamplib/src/genny/templates/documentation.md.j2
dalyd: Ah, yeah -- the heading including the line does make that confusing.

SimonEismann (approved):

docs/generated/workloads.md
SimonEismann: Done: https://jira.mongodb.org/browse/DEVPROD-6160

PR #1185, opened 2024-04-09
DEVPROD-5910 - Generate documentation from genny workload files

MAhmadShah:
**Jira Ticket:** https://jira.mongodb.org/browse/DEVPROD-5910

### Whats Changed
We reformat some of the Genny docs to exist under more easily accessible headers.

### DIFF
--- a/docs/generated/workloads.md
+++ b/docs/generated/workloads.md
@@ -4256,11 +4256,11 @@ Measures the performance of running dbcheck's modes and its effect on crud opera
 ### Keywords
 dbcheck, collections, indexes, crud
 
-
+## Replication Startup
 ## 1_0_5GB
-### Owner
+#### Owner
 mongodb/server-replication"
-### Description
+#### Description
 Loads the data for the light phase.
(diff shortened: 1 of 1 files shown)


dalyd:
@MAhmadShah I'm confused by this PR. I only see one changed file, and I thought that was supposed to be generated. Shouldn't be also be updating the code that generates it?

thessem (commented):
Agreeing with @dalyd here. Was the generated file inconsistent or something?

MAhmadShah:
@dalyd @thessem sorry for the light description on this one. The generated file didn't have the clearest distinction between some of the TPCH workloads that have the same name (i.e. [Q1 normalized scale 1](https://github.com/mongodb/genny/blob/master/src/workloads/tpch/normalized/1/Q1.yml) vs [Q1 denormalized scale 1](https://github.com/mongodb/genny/blob/master/src/workloads/tpch/denormalized/1/Q1.yml)) so I made changes to organize it better. I agree though this work should've been done via the generated script incase we want to regenerate the doc, I'll close this one and sync with Austin who did that work originally.

PR #1187, opened 2024-04-11
PERF-5241 Update AutoRun in workloads relevant to the QueryStats 6.0 backport

awagenmaker:
### What's Changed
Updating the auto-run branches on tasks that we added query stats auto-run configs for in PERF-5152 to include 6.0. Also, adding a query stats variant to the MetricSecondaryIndexTimeseriesCollection.yml auto-run because we'd like to get data from it on both backport branches. 

SERVER-87943 is adding the query stats variants that these will be run under to v6.0.

### Patch Testing Results
https://spruce.mongodb.com/version/6617fa2e8ce8480007b1695c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/query/QueryStats.yml
+++ b/src/workloads/query/QueryStats.yml
@@ -230,4 +230,4 @@ AutoRun:
       - atlas-like-replica-query-stats.2023-09
     branch_name:
       $gte:
-        v7.0
+        v6.0
--- a/src/workloads/query/QueryStats.yml
+++ b/src/workloads/query/QueryStats.yml
@@ -228,6 +228,8 @@ AutoRun:
     mongodb_setup:
       $eq:
       - atlas-like-replica-query-stats.2023-09
+      # Note: this is just using the configuration used for all of our 3-node repl query stats variants.
+      # The workload will still override the rate limit regardless of what the base configuration is.
       - replica-query-stats-rate-limit
     branch_name:
       $gte:
--- a/src/workloads/query/QueryStats.yml
+++ b/src/workloads/query/QueryStats.yml
@@ -228,6 +228,7 @@ AutoRun:
     mongodb_setup:
       $eq:
       - atlas-like-replica-query-stats.2023-09
+      - replica-query-stats-rate-limit
     branch_name:
       $gte:
         v6.0
--- a/src/workloads/query/CollScanComplexPredicateLarge.yml
+++ b/src/workloads/query/CollScanComplexPredicateLarge.yml
@@ -25,4 +25,4 @@ AutoRun:
       - standalone-sbe
     branch_name:
       $gte:
-        v7.0
+        v6.0
--- a/src/workloads/query/CollScanComplexPredicateMedium.yml
+++ b/src/workloads/query/CollScanComplexPredicateMedium.yml
@@ -25,4 +25,4 @@ AutoRun:
       - standalone-sbe
     branch_name:
       $gte:
-        v7.0
+        v6.0
--- a/src/workloads/query/CollScanComplexPredicateSmall.yml
+++ b/src/workloads/query/CollScanComplexPredicateSmall.yml
@@ -25,4 +25,4 @@ AutoRun:
       - standalone-sbe
     branch_name:
       $gte:
-        v7.0
+        v6.0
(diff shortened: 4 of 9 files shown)


src/workloads/query/QueryStats.yml
```diff
@@ -228,6 +228,7 @@ AutoRun:
     mongodb_setup:
       $eq:
       - atlas-like-replica-query-stats.2023-09
+      - replica-query-stats-rate-limit
```
shaargtz, line 231: @cswanson310 i remember from one of the conversations we had that we determined we were only going to run this on the atlas-like against itself. of course we can't do that right now given the lack of an atlas-like in 6.0, but i don't remember if there was a particular reason why we chose that specific variant.

awagenmaker: Good callout @shaargtz, I forgot to note this here. We can change this to the real Atlas variant if SERVER-89277 proves to be fruitful, but thought it would be better to start getting some data on it in case that ticket doesn't work out or takes a while.

jimoleary (commented):
I am assuming that the [QueryStats 7.0 Backport perf build variants and tasks](https://docs.google.com/document/d/1G0g6WE1TFaMRT-Mmb_o3PFVAjsYUAAC4s03EKu2KQuI/edit#heading=h.eslkeb2n1d6r) doc also covers the variants / tasks required for v6.0. 

The MetricSecondaryIndexTimeseriesCollection change will result in this workload running on the rate limited replset variants on **all matching branch names** (so v7.x, v8.x and master will also change). 

Generally PRs are ordered to commit to the mongo repo last so this workload will begin executing immediately on the non-rate limited replica set variants (PERF 3-Node ReplSet ARM AWS 2023-11 / PERF 3-Node ReplSet Intel AWS 2023-11) on v6.x and additionally on the rate limited replset variants on the other matching branches. 

Is this change deliberate? It looks like this might be fixing an over sight from the original ticket.

jimoleary:
> Would you rather that we make the auto-run here specific to v6.0+v7.0?

No, I think the way you have defined it is simpler / clearer. If it becomes a problem it can be fixed. 

> I'm not intending to merge this until we've merged SERVER-87943 to v6.0 (in review right now). We need data from non-query stats variants to compare to the query stats variants, so this is desirable behavior.

Sounds good thanks.

src/workloads/query/QueryStats.yml
cswanson310: Oh right ok. Umm I think the different rate limited platforms are not valuable here, since this test manually controls the rate limit (see line 29). So... this change will run this test repeatedly in environments which are identical, since all our various different levels of rate limit will all fall under `replica-query-stats-rate-limit`, correct? Does that sound right @shaargtz ?

shaargtz: oh yeah that is absolutely right. i don't know what other variant could be a good placeholder as i'm not sure how important it is that we run the workload on a replica set. maybe we can skip this for now and add it until we figure out the real atlas or the atlas-like?

awagenmaker: Yeah, running more tests than we need is a drawback to doing things this way. replica-query-stats-rate-limit is the dsi configuration that we use for the normal default/100-rate-limited intel and arm replica set variants, so I agree that we may not have a better placeholder

Unfortunately it sounds like the conclusion on the atlas/atlas-like variants is that we can't create one for query stats (see comment on SERVER-89277 - will also discuss more at standup). Our alternative here would be to not auto-run this workload at all and instead run a multi-patch on the real atlas variant at towards the end of the project.

cswanson310: Ok gotcha, so we just want it to run somewhere? I think anything is fine then, but we definitely want to put a comment in here indicating that the rate limit/query stats part of the name is not relevant since the test will override them.

cswanson310: We'll need to match the variant here: https://github.com/10gen/PerformanceReports/blob/a370f666bd144de38190c8278656924046d06582/scripts/perf_reports_query_stats_perf.yml#L294

awagenmaker: Sounds good, added a comment

shaargtz (approved):
LGTM

jimoleary (approved):
LGTM

awagenmaker:
SERVER-87943 has now merged to 6.0, so merging this in as well

PR #1188, opened 2024-04-11
DEVPROD-6453 Fix MultikeyIndexes.yml indexes so that multi-planning is exercised as intended

dstorch:
I noticed that all of the queries were using COLLSCAN plans. I think this helps to explain the current results for this workload presented here: https://github.com/10gen/product-perf-experimentations/blob/master/investigations/PERF-5121-compare-multiplanners/full-results.ipynb. Hopefully the results tell us something more meaningful about the three multi-planning code paths after this fix!

I'm verifying that the workload runs correctly locally, but haven't done any other testing of this change.

I will alert the perf build barons that change points (probably improvements) are expected for this workload as a consequence of this change!

### DIFF
--- a/src/workloads/query/multiplanner/MultikeyIndexes.yml
+++ b/src/workloads/query/multiplanner/MultikeyIndexes.yml
@@ -129,131 +129,131 @@ Actors:
         x63: *distribution
       }
       indexes:
-      - keys: {tenantId: 1, x1: 1}
+      - keys: {x1: 1}
         options: {name: index1}
-      - keys: {tenantId: 1, x2: 1}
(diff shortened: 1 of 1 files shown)


jimoleary (approved):
Oops, this looks like a cut'n'pasta issue from [workloads/query/multiplanner/CompoundIndexes.yml](https://github.com/mongodb/genny/blob/4e19ebf52faa7a51e8f3c99866d9a827f8b94253/src/workloads/query/multiplanner/CompoundIndexes.yml#L65).

Generally I would ask for a patch but I think in this case it should be fine but it must surely generate large change points.

PR #1189, opened 2024-04-12
SERVER-88684: Mixed10KThreads find operations limited to 10

miguelangelnieto:
The read queries match on an increasing set of documents over time because we insert more data that matches the query predicate over time. Read throughput is directly influenced by write throughput. To reduce that influence, the find query will only return 10 documents per query.

### DIFF
--- a/src/workloads/scale/Mixed10KThreads.yml
+++ b/src/workloads/scale/Mixed10KThreads.yml
@@ -23,8 +23,6 @@ Keywords:
 dbname: &dbname mixed_10k
 DocumentCount: &NumDocs 10000
 CollectionCount: &NumColls 1
-Limit: &limit 100
-
 NumPhases: &NumPhases 2
 Document: &doc
   a: {^RandomInt: {min: 0, max: *NumDocs}}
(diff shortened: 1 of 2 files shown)


miguelangelnieto (commented):
Patch test running here: https://spruce.mongodb.com/version/6618f6d5b363c7000724da3e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

thessem (approved):
We're fixing the code review required from us on docs updates, so leaving a tick to get out of the way.

src/workloads/scale/Mixed10KThreads.yml
```diff
@@ -22,6 +23,7 @@ Keywords:
 dbname: &dbname mixed_10k
 DocumentCount: &NumDocs 10000
 CollectionCount: &NumColls 1
+Limit: &limit 100
```
jimoleary, line 26: This should be 10 and used on line 77 or deleted?

jimoleary (commented):
Remove the limit anchor on line 20 ... it's not being used.

src/workloads/scale/Mixed10KThreads.yml
miguelangelnieto: Removed, thanks.

jimoleary (approved):
LGTM 

We need to inform the Build Barons about the related jump in throughput.

PR #1192, opened 2024-04-16
DEVPROD-6525 Rename TPCH workloads to make names unique.

ahartschen:
**Jira Ticket:** DEVPROD-6525

### Whats Changed

For the workload documentation project we need to ensure that all workloads have a unique name to ensure a 1 to 1 map from a workload to a piece of documentation. This PR changes the name of the TCPH workloads to make them all unique.

We have taken the sub-directories that are part of the workload path and appended them to the filename to ensure uniqueness.  I checked with the workload owner, @alyacb, that this naming scheme was okay with her.

For posterity I did this with a couple shell commands:

Change file names:

```
for file in *; do mv "$file" "${file%.*}_$(basename $(pwd)).${file##*.}"; done
```

Update references to phases in workloads:

Normalized
```
find normalized -name '*.yml' -exec sed -i '' 's|\(../../../../phases/tpch/normalized/\)\(.*\)\(.yml\)|\1\2_normalized\3|' {} \;
```

Denormalized
```
find denormalized -name '*.yml' -exec sed -i '' 's|\(../../../../phases/tpch/denormalized/\)\(.*\)\(.yml\)|\1\2_denormalized\3|' {} \;
```

### Follow Up Work

This should not be merged until I update [this test control file](https://github.com/10gen/dsi/blob/be106823fd492aba7091eba46f0c76a72ec1f464/configurations/test_control/test_control.tpch.yml#L13) to include the new file name attributes.

[Here](https://github.com/10gen/dsi/pull/1675) is the PR to update the test control files.


docs/generated/phases.md
MAhmadShah, line 1: Nit: Is it possible to have the headers show up in numerical order instead of alphabetical order?

src/workloads/tpch/normalized/10/Q9_normalized_10.yml
```diff
@@ -21,5 +21,5 @@ Actors:
 #   Database: tpch
 #   Phases:
 #   - LoadConfig:
-#       Path: ../../../../phases/tpch/normalized/Q19.yml
```
MAhmadShah, line 24: Looks like a typo from the original workload, probably should be Q9 instead of Q19.

docs/generated/phases.md
MAhmadShah: Not blocking, could be an annoying one to solve for

thessem (approved):
My positive review here is not sufficient for a merge, I don't think @mongodb/devprod-performance-infrastructure needs to review this so I am getting out of the way. I am fixing the CODEOWNERS here: https://github.com/mongodb/genny/pull/1193

MAhmadShah (approved):

jimoleary:
> For the workload documentation project we need to ensure that all workloads have a unique name to ensure a 1 to 1 map from a workload to a piece of documentation. This PR changes the name of the TCPH workloads to make them all unique.

Quest: Does this apply to **all** files in all directories? If I add a new file in a different directory with the same basename, would we have the same problem. If this is the case then wouldn't it be better to fix the actual problem rather than forcing a rename of existing files.

dalyd (commented):
The change LGTM, but the commit coordination worries me across repos. 

Could you do this in three steps? 
1. PR as it is plus symbolic links from the old file names to the new file names. This will make it safe to merge this ahead of the DSI change. 
2. Update the DSI configuration files
3. Delete the symbolic links

It's an annoying, but it ensures that we don't end up with any commits on the waterfall having manifests that cannot run these tests.

jimoleary:
> Yes, I've made [DEVPROD-6603](https://jira.mongodb.com/browse/DEVPROD-6603) to lint for unique workload names. This needs to go in regardless of that change though.

Why does it have to be unique on basename rather than the full path name?

Sorry, if I am getting a bit too caught up on this but it seems a bit odd to force this restriction.

Will that mean that all basenames have to be unique across all tests? How will this apply across multiple git repos (given that we have genny workloads and PrivateWorloads)?

src/workloads/tpch/denormalized/1/Q10_denormalized_1.yml
```diff
@@ -0,0 +1,50 @@
+SchemaVersion: 2018-07-01
```
alyacb, line 1: It looks like we have multiple identicalworkloads now:

- src/workloads/tpch/denormalized/1/Q10.yml
- src/workloads/tpch/denormalized/1/Q10_denormalized_1.yml

Is this necessary?

alyacb (changes requested):

dalyd:
> > > @dalyd @MAhmadShah @alyacb @jimoleary I've discussed this with @johndaniels and we are in agreement that we should close this PR and not proceed with renaming these files. The general rule of thumb we agreed on is that we don't need to map workloads to documentation for workloads that are missing an `AutoRun` section. In the case of TPCH the actual documentation for the task will live in the [test control](https://github.com/10gen/dsi/blob/master/configurations/test_control/test_control.tpch.yml#L13) file. What are your thoughts on this?
> > 
> > 
> > Will we still have multiple copies of the Q10.yml documentation and others, and the test_control will point to them? I'd hate to disregard the documentation already in the file (that seems surprising), but it looks like we would actually be okay. It looks like the code gives it unique anchors (e.g. [Q10 normalized with scale factor 1](https://github.com/mongodb/genny/blob/master/docs/generated/workloads.md#q10-2) )
> > Assuming a human can still manually link to the right place, this sounds just fine to me. And then we can point out the issue to the workload owners and let them handle it if they want to.
> 
> Yeah, my thought was to manually create a markdown file to put in the test control file and then provide links to the Genny docs there, which as you pointed out we can do due to the unique anchors.
> 
> Overall, that seems cleaner than what I've done in this PR.

SGTM. 
I think we can use the existing metadata field structure in the test_control. Such as [here](https://github.com/10gen/dsi/blob/a328e7ac171baf29c6ebae4ad00433d4d6ec37c6/configurations/test_control/test_control.ycsb.2023-09.yml#L4)  and [here](https://github.com/10gen/dsi/blob/a328e7ac171baf29c6ebae4ad00433d4d6ec37c6/configurations/test_control/test_control.ycsb.2023-09.yml#L17) in YCSB assuming we are or will be parsing those.

jimoleary:
> The reason why we need the basename to be unique is that we treat each workload defined here as a task. Our project requires that we can map a specific task to a specific piece of documentation. You can imagine that we have an API like `GET /task/{task_name]/documentation` - this won't work if we have duplicate task names.

Can the routing glob the full path name rather than the basename? 

so  GET /task/{task_name]/documentation where **task_name** could be unique:

1. tpch/normalized/1/Q1
2. tpch/denormalized/1/Q1

This approach would leave the filenames as they currently are and not impose restrictions on someone needing to know all the other  basenames.

MAhmadShah:
Just catching up here, the approach of a separate markdown page linking to the unique anchors for each TPCH workload sounds good to me, I can link to that page then in the TPCH test control file in this [PR](https://github.com/10gen/dsi/pull/1661). This approach seems to rock the boat the least.

PR #1195, opened 2024-04-19
DEVPROD-5933 Validate team names and slack contact information for workloads.

ahartschen:
**Jira Ticket:** DEVPROD-5933 

### Whats Changed

For the workload ownership project we are going to require that all workloads are owned by a valid team and have valid contact information for support. We are using the Mothra repo as our source of truth for this information. This PR introduces a `MothraService` that uses the GitHub API to gather the yaml files representing teams at Mongo and cross checks the owners and slack channels in each workload against this data.

I created [this spreadsheet](https://docs.google.com/spreadsheets/d/1e02apqAU77fC-cQ-HxvITKpa-6ch1U5ptE3SxdcbSYM/edit#gid=0) to map existing workload owners to a team represented in Mothra and gathered slack channels for support for each team. I updated all the existing workload owners to their Mothra representation and added contact information. The two exceptions to this rule are `@mongodb/query` and `@mongodb/product-query` as these two need updated to one of `Query Optimization`, `Query Integration`, or `Query Execution`. There are 336 such instances that still need updated and I will work with a query team member to fix these. 

### Testing

The `./run-genny lint-yaml` command succeeds locally for me as I have a GH token configured locally. The PR check is currently failing because I don't yet have permissions to add a secret to the Evergreen project and I'm waiting on IT to issue me an API key.

image

[Here](https://spruce.mongodb.com/task/genny_amazon2_t_lint_workloads_standalone_patch_67f55a09328efd294de9fa9fbdf5243bef1f27f4_6622f06acbb2230007d6d406_24_04_19_22_30_03/logs?execution=0) is the new standalone task that I ran. It failed due to the access token not being set yet.

###  Prerequisites

- Before merging we need [this PR](https://github.com/10gen/mothra/pull/1003) in Mothra to be merged as it includes the Slack support channels necessary for our validation. 
- I have an open request with IT to get a Github token created to use for the Genny evergreen project. This secret needs to be configured before merging.

### DIFF
--- a/src/lamplib/src/genny/cli.py
+++ b/src/lamplib/src/genny/cli.py
@@ -582,17 +582,17 @@ def self_test(ctx: click.Context):
 )
 @click.pass_context
 def lint_yaml(ctx: click.Context, lint_format: bool):
-    from genny.tasks import yaml_linter
+    from genny.tasks.yaml_linter import YamlLinter
 
-    yaml_linter.main(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"], lint_format=lint_format)
+    YamlLinter().lint(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"], lint_format=lint_format)
 
 
 @cli.command(name="generate-docs", help="Generate documentation for all workloads.")
 @click.pass_context
 def generate_docs(ctx: click.Context):
-    from genny.tasks import documentation_generator
+    from genny.tasks.documentation_generator import DocumentationGenerator
 
-    documentation_generator.main(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"])
+    DocumentationGenerator().generate(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"])
 
 
 @cli.command(
--- a/src/lamplib/src/genny/tasks/mothra_service.py
+++ b/src/lamplib/src/genny/tasks/mothra_service.py
@@ -18,12 +18,9 @@ class Team:
     support_slack_channel_name: Optional[str] = None
     support_slack_channel_id: Optional[str] = None
 
-    def __init__(self, **kwargs):
-        # Filter out any kwargs that are not in the class
-        filtered_kwargs = {k: v for k, v in kwargs.items() if k in self.__annotations__}
-
-        for k, v in filtered_kwargs.items():
-            setattr(self, k, v)
+    @classmethod
+    def create(cls, **kwargs) -> "Team":
+        return cls(**{k: v for k, v in kwargs.items() if k in cls.__annotations__})
 
 
 class MothraService:
@@ -37,7 +34,7 @@ class MothraService:
     def _load_team_map(self) -> dict[str, Team]:
         SLOG.info("Cloning Mothra repository to get team information.")
         with tempfile.TemporaryDirectory() as temp_dir:
-            subprocess.run(["git", "clone", MOTHRA_REPO, temp_dir])
+            subprocess.run(["git", "clone", "--depth=1", MOTHRA_REPO, temp_dir])
             teams = []
             team_map = {}
             for file in TEAMS_FILES_PATHS:
@@ -45,7 +42,7 @@ class MothraService:
                     teams.extend(yaml.safe_load(f)["teams"])
 
             for team in teams:
-                team_data = Team(**team)
+                team_data = Team.create(**team)
                 team_map[team_data.name] = team_data
 
             return team_map
--- a/src/lamplib/src/genny/tasks/yaml_linter.py
+++ b/src/lamplib/src/genny/tasks/yaml_linter.py
@@ -13,128 +13,128 @@ from genny.tasks.yaml_linter_constants import (
 from genny.tasks.mothra_service import MothraService
 
 SLOG = structlog.get_logger(__name__)
-mothra_service = MothraService()
 
 
-def main(genny_repo_root: str, lint_format: bool):
-    workload_dirs = [
-        path.join(genny_repo_root, "src", "workloads"),
-        path.join(genny_repo_root, "src", "phases"),
-    ]
+class YamlLinter:
 
-    has_errors = False
-    resmoke_dirs = [path.join(genny_repo_root, "src", "resmokeconfig")]
-    evergreen_dirs = [path.join(genny_repo_root, "evergreen.yml")]
+    def __init__(self):
+        self.mothra_service = MothraService()
 
-    workload_yamls, workload_error = _traverse_yamls(workload_dirs)
-    resmoke_yamls, resmoke_error = _traverse_yamls(resmoke_dirs)
-    evergreen_yamls, evergreen_error = _traverse_yamls(evergreen_dirs)
-    all_yamls = workload_yamls + resmoke_yamls + evergreen_yamls
+    def lint(self, genny_repo_root: str, lint_format: bool):
+        workload_dirs = [
+            path.join(genny_repo_root, "src", "workloads"),
+            path.join(genny_repo_root, "src", "phases"),
+        ]
 
-    if workload_error or resmoke_error or evergreen_error:
-        SLOG.error(
-            "Found invalidly-named yaml files. Please correct and rerun ./run-genny lint-yaml."
-        )
-        has_errors = True
+        has_errors = False
+        resmoke_dirs = [path.join(genny_repo_root, "src", "resmokeconfig")]
+        evergreen_dirs = [path.join(genny_repo_root, "evergreen.yml")]
 
-    all_have_descriptions = True
-    for workload_yaml in workload_yamls:
-        if not check_required_fields(workload_yaml):
-            all_have_descriptions = False
-    if not all_have_descriptions:
-        SLOG.error(
-            f"""
-            The above YAML workloads lack one or more of the required {", ".join(REQUIRED_FIELDS)} fields. 
-            These fields should be populated with human-readable values. After doing so, please re-run ./run-genny lint-yaml
-            """
-        )
-        has_errors = True
+        workload_yamls, workload_error = self._traverse_yamls(workload_dirs)
+        resmoke_yamls, resmoke_error = self._traverse_yamls(resmoke_dirs)
+        evergreen_yamls, evergreen_error = self._traverse_yamls(evergreen_dirs)
+        all_yamls = workload_yamls + resmoke_yamls + evergreen_yamls
 
-    if lint_format:
-        config_file_path = path.join(genny_repo_root, ".yamllint")
-        yamllint_argv = ["--strict", "--config-file", config_file_path] + all_yamls
+        if workload_error or resmoke_error or evergreen_error:
+            SLOG.error(
+                "Found invalidly-named yaml files. Please correct and rerun ./run-genny lint-yaml."
+            )
+            has_errors = True
 
-        SLOG.info(
-            "Linting workload YAML files with yamllint",
-            count=len(all_yamls),
-            yamllint_argv=yamllint_argv,
-        )
+        all_have_descriptions = True
+        for workload_yaml in workload_yamls:
+            if not self.check_required_fields(workload_yaml):
+                all_have_descriptions = False
+        if not all_have_descriptions:
+            SLOG.error(
+                f"""
+                The above YAML workloads lack one or more of the required {", ".join(REQUIRED_FIELDS)} fields. 
+                These fields should be populated with human-readable values. After doing so, please re-run ./run-genny lint-yaml
+                """
+            )
+            has_errors = True
 
-        yamllint.cli.run(yamllint_argv)
+        if lint_format:
+            config_file_path = path.join(genny_repo_root, ".yamllint")
+            yamllint_argv = ["--strict", "--config-file", config_file_path] + all_yamls
 
-    if has_errors:
-        sys.exit(1)
-    else:
-        SLOG.info("All YAML files are correctly formatted.")
+            SLOG.info(
+                "Linting workload YAML files with yamllint",
+                count=len(all_yamls),
+                yamllint_argv=yamllint_argv,
+            )
 
+            yamllint.cli.run(yamllint_argv)
 
-def check_required_fields(yaml_path):
-    workload = _load_yaml(yaml_path)
-    no_errors = True
-    for field in REQUIRED_FIELDS:
-        if field not in workload:
-            if field == "Keywords" and yaml_path in GRANDFATHERED_WORKLOADS_WITHOUT_KEYWORDS:
-                continue
-            SLOG.error(f"Genny workload {yaml_path} lacks the {field} field.")
-            no_errors = False
+        if has_errors:
+            sys.exit(1)
+        else:
+            SLOG.info("All YAML files are correctly formatted.")
 
-    # Grandfathered workloads do not need to have the Owner field match a team in Mothra or contain contact information.
-    if workload["Owner"] in GRANDFATHERED_WORKLOADS_OWNERS_NOT_IN_MOTHRA:
+    def check_required_fields(self, yaml_path):
+        workload = self._load_yaml(yaml_path)
+        no_errors = True
+        for field in REQUIRED_FIELDS:
+            if field not in workload:
+                if field == "Keywords" and yaml_path in GRANDFATHERED_WORKLOADS_WITHOUT_KEYWORDS:
+                    continue
+                SLOG.error(f"Genny workload {yaml_path} lacks the {field} field.")
+                no_errors = False
+
+        # Grandfathered workloads do not need to have the Owner field match a team in Mothra or contain contact information.
+        if workload["Owner"] in GRANDFATHERED_WORKLOADS_OWNERS_NOT_IN_MOTHRA:
+            return no_errors
+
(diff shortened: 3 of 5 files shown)


src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,58 @@
...
+            team_data = Team(**filtered_data)
+            team_map[team_data.name] = team_data
+
+        return team_map
+
+    def _get_github_token(self) -> str:
+        load_dotenv()
+        token = os.getenv("GITHUB_ACCESS_TOKEN")
+        if not token:
+            raise ValueError("GITHUB_ACCESS_TOKEN is not set in the environment")
```
thessem, line 56: Please add to this error message the file I could add `GITHUB_ACCESS_TOKEN` to, so I can fix this error without looking up documentation. `GITHUB_ACCESS_TOKEN is not set in the environment, try setting it in .env.local` or similar?

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,58 @@
...
+        for team in teams:
+            # Filter out any keys that are not part of the Team model
+            filtered_data = {key: value for key, value in team.items() if key in Team._fields}
+            team_data = Team(**filtered_data)
+            team_map[team_data.name] = team_data
+
+        return team_map
+
+    def _get_github_token(self) -> str:
+        load_dotenv()
```
thessem, line 53: Is this better done in a more global spot? I see `.env` in the repo and I assume I can set `GRPC_PORT` or whatever (made up example), and this isn't going to work unless I've linted a yaml file, right?

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,58 @@
...
+        return self.team_map.get(team_name, None)
+
+    def _load_team_map(self) -> dict[str, Team]:
+        github = Github(auth=Auth.Token(self._get_github_token()))
+        repo = github.get_repo(MOTHRA_REPO)
+        teams = []
+        team_map = {}
+        for file_path in TEAMS_FILES_PATHS:
+            # TODO: Remove hardcoded ref
+            content = repo.get_contents(file_path, ref="DEVPROD-5933")
```
thessem, line 37: Is it possible we're going to fill up a GitHub rate limit doing this?

docs/using.md
```diff
@@ -450,12 +450,25 @@ Try using `python test_result_summary.py --help` for more options.
...
+    ./run-genny resmoke-test  # Run Actor integration tests - only necessary if adding/editing Actors
     ```
 
-    Note the current [issue](#orgb084b49) running resmoke-test. Also, note that there is no schema-checking of the yaml.
+    Note: the current [issue](#orgb084b49) running resmoke-test. 
+
+    Linting requires special configuration to run locally, see [here](./developing.md#lint-workload-and-other-yaml) for more information before proceeding with this command:
+
+    ```bash
+    ./run-genny lint-yaml  # Lint all YAML files,
```
thessem, line 462: I have a weak preference for this being called `validate_yaml` or `validate-workloads` (or `-` whatever we're doing by default in this repo). It gives us the namespace room to `lint-yaml`, e.g., make sure it's formatted properly, people aren't accidentally specifying octals, we're using consistent indentation etc etc.


It is surprising to me that YAML linting would a GitHub token.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -70,8 +73,56 @@ def check_required_fields(yaml_path):
         if field not in workload:
             if field == "Keywords" and yaml_path in GRANDFATHERED_WORKLOADS_WITHOUT_KEYWORDS:
```
thessem, line 74: It's fine if the answer is no, but is it possible we should've done this via the `SchemaVersion` at the top of each YAML file? We're changing required fields in the workload schema, and if that doesn't require a bump, what's the point of even having that field?

src/workloads/sharding/multi_updates/MultiUpdates.yml
```diff
@@ -1,5 +1,7 @@
 SchemaVersion: 2018-07-01
-Owner: "@mongodb/sharding"
+Owner: Cluster Scalability
+SlackSupportChannelName: "#server-sharding"
+SlackSupportChannelId: C8PK5KZ5H
```
thessem, line 4: What's the reasoning for making this a raw ID? Having this as a URL seems more usable to me (e.g., https://mongodb.enterprise.slack.com/archives/C8PK5KZ5H) since I'm not aware in slack of any way of turning an ID into a URL without manually doing it.

thessem (changes requested):
I am very much looking forward to having this, I am out-of-the-loop on the design of this, and my main concern is the Python code and repository config, so any comments I have on schema are less important.

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self):
+        self.team_map = self._load_team_map()
+
+    def get_team(self, team_name: str) -> Optional[Team]:
+        return self.team_map.get(team_name, None)
+
+    def _load_team_map(self) -> dict[str, Team]:
+        github = Github(auth=Auth.Token(self._get_github_token()))
+        repo = github.get_repo(MOTHRA_REPO)
```
jimoleary, line 32: Question: I'm ignorant in this but isn't this just a fancy git clone ? If so then could we not just clone the repo and avoid the github API?

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,58 @@
...
+
+            # The get_contents method returns a list if the path is a directory, but we know the path is a file and assert here to keep the linter happy.
+            assert isinstance(content, ContentFile)
+
+            teams.extend(yaml.safe_load(content.decoded_content.decode())["teams"])
+
+        for team in teams:
+            # Filter out any keys that are not part of the Team model
+            filtered_data = {key: value for key, value in team.items() if key in Team._fields}
+            team_data = Team(**filtered_data)
```
jimoleary, line 47: suggestion: it might be simpler to let the NamedTuple instance select the fields itself. The code isn't terrifically different (see [here](https://stackoverflow.com/questions/17622419/creating-a-namedtuple-object-using-only-a-subset-of-arguments-passed)) but it looks less distracting where you use it.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -70,8 +73,56 @@ def check_required_fields(yaml_path):
...
                 continue
+            if (
```
jimoleary, line 76: Suggestion: Would it make sense at this stage to create a `Workload` model class and put all this vaidation into a method on the instance? 

I would also suggest that Pydantic might help (although it depends on whether it is complicated enough at this stage).

src/phases/HelloWorld/ExamplePhase2.yml
```diff
@@ -1,5 +1,7 @@
 SchemaVersion: 2018-07-01
-Owner: "@mongodb/product-perf"
+Owner: Product Performance
+SlackSupportChannelName: "#performance"
+SlackSupportChannelId: C0V3KSB52
```
jimoleary, line 4: Question: Could  `SlackSupportChannelId` ,  `SlackSupportChannelName` be stored in the mothra  teams meta data? That way it only needs to be changed / stored in one place.

If it ever changed then updating it just involves chaning the mothra repos teams data rather than 30 to 40 workload files.

jimoleary (commented):
I've mainly left some questions / suggestions.

Of all of them ... moving the SlackSupportChannelId , SlackSupportChannelName into mothra might be of the most benefit.

src/phases/HelloWorld/ExamplePhase2.yml
jimoleary: I would be inclined to do it that way ... I can see how I would know about the performance channel name but it might be more difficult to know where the C0V3KSB52 came from (particularly for a new channel).

src/lamplib/src/genny/tasks/mothra_service.py
jimoleary: Personally I think simpler is better ... but you have documented how to get a token and shown that it's unlikely that we'll ever hit the API limit ... so I'm happy to go with either.

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,51 @@
...
+
+@dataclass
+class Team:
+    name: str
+    support_slack_channel_name: Optional[str] = None
+    support_slack_channel_id: Optional[str] = None
+
+    def __init__(self, **kwargs):
+        # Filter out any kwargs that are not in the class
+        filtered_kwargs = {k: v for k, v in kwargs.items() if k in self.__annotations__}
```
thessem, line 23: [non-blocking nit]

I'm not a python guy so let me know if this is actually good python. It seems like manually setting these (and letting kwargs absorb the rest) is a better option.

This is actually a perferct use-case for Pydantic, given these are being loaded straight from serialisation, but I'd understand not wanting to pick up that dep in this PR

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import dataclass
+import subprocess
+import tempfile
+from typing import Optional
+import structlog
+import yaml
+
+MOTHRA_REPO = "git@github.com:10gen/mothra.git"
```
thessem, line 8: SSH repositories are going to stop working within evergreen soon (DEVPROD-351). We will need to get evergreen to provide us a short-lived SSH token. 

I am okay with this code getting merged in as-is so long as we have a plan for conversion and we carry it out soon (cc @lgfang who is handling this for my team).

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self):
+        self.team_map = self._load_team_map()
+
+    def get_team(self, team_name: str) -> Optional[Team]:
+        return self.team_map.get(team_name, None)
+
+    def _load_team_map(self) -> dict[str, Team]:
+        SLOG.info("Cloning Mothra repository to get team information.")
+        with tempfile.TemporaryDirectory() as temp_dir:
+            subprocess.run(["git", "clone", MOTHRA_REPO, temp_dir])
```
thessem, line 40: I like this approach, we may run into some perf issues and need to make this shallower or something, but this is easy enough to update later when we notice.

src/lamplib/src/genny/tasks/mothra_service.py
```diff
@@ -0,0 +1,51 @@
...
+        for k, v in filtered_kwargs.items():
+            setattr(self, k, v)
+
+
+class MothraService:
+
+    def __init__(self):
+        self.team_map = self._load_team_map()
+
+    def get_team(self, team_name: str) -> Optional[Team]:
```
thessem, line 34: [non-blocking comment] Not 100% sure this function is needed (up to you!), but, I want to thank you for not overloading an operator to do this!

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -6,11 +6,14 @@
...
 import yamllint.cli
 from genny.tasks.yaml_linter_constants import (
+    GRANDFATHERED_WORKLOADS_OWNERS_NOT_IN_MOTHRA,
     REQUIRED_FIELDS,
     GRANDFATHERED_WORKLOADS_WITHOUT_KEYWORDS,
 )
+from genny.tasks.mothra_service import MothraService
 
 SLOG = structlog.get_logger(__name__)
+mothra_service = MothraService()
```
thessem, line 16: Can we do this in main instead? I don't like that importing this module is going to eventually do a `git clone`.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -72,6 +77,35 @@ def check_required_fields(yaml_path):
...
+        SLOG.error(
+            f"""
+            The Owner field, {workload["Owner"]}, in the workload at {yaml_path} does not match any team in the Mothra repository.
+            Please correct the Owner field to match a team in the Mothra repository or add the team to the Mothra repository.
+
+            Mothra Repository: https://github.com/10gen/mothra
+            """
+        )
+        no_errors = False
+        return no_errors
```
thessem, line 96: `return False`? This reads as an accidental return otherwise.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -72,6 +77,35 @@ def check_required_fields(yaml_path):
...
+
+    team = mothra_service.get_team(workload["Owner"])
+    if not team:
+        SLOG.error(
+            f"""
+            The Owner field, {workload["Owner"]}, in the workload at {yaml_path} does not match any team in the Mothra repository.
+            Please correct the Owner field to match a team in the Mothra repository or add the team to the Mothra repository.
+
+            Mothra Repository: https://github.com/10gen/mothra
```
thessem, line 92: This is a great error message, thanks.

src/lamplib/src/genny/tasks/mothra_service.py
thessem, line 1: This is a brand new service being added, is it possible to get some tests? I am concerned that (a) it's tested and (b) it's written in a way that _is_ testable, which having at least some tests would validate.

src/lamplib/src/genny/tasks/mothra_service.py
thessem: Actually, upgrading this into a blocker. We can make this 3x quicker and save 100s of msecs.

```
$ time git clone git@github.com:10gen/mothra.git
Cloning into 'mothra'...
remote: Enumerating objects: 12393, done.
remote: Counting objects: 100% (1176/1176), done.
remote: Compressing objects: 100% (523/523), done.
remote: Total 12393 (delta 838), reused 947 (delta 647), pack-reused 11217
Receiving objects: 100% (12393/12393), 3.10 MiB | 2.89 MiB/s, done.
Resolving deltas: 100% (9220/9220), done.

________________________________________________________
Executed in    4.52 secs      fish           external
   usr time  384.12 millis    0.17 millis  383.96 millis
   sys time  197.60 millis    1.87 millis  195.73 millis
```

vs

```
$ time git clone --depth=1 git@github.com:10gen/mothra.git
Cloning into 'mothra'...
remote: Enumerating objects: 291, done.
remote: Counting objects: 100% (291/291), done.
remote: Compressing objects: 100% (277/277), done.
remote: Total 291 (delta 21), reused 140 (delta 7), pack-reused 0
Receiving objects: 100% (291/291), 634.19 KiB | 1.04 MiB/s, done.
Resolving deltas: 100% (21/21), done.

________________________________________________________
Executed in    3.93 secs      fish           external
   usr time  128.48 millis    0.17 millis  128.32 millis
   sys time  106.07 millis    1.71 millis  104.36 millis
```

These were repeatable results (M1 macbook, on a residential internet connection)

thessem (changes requested):
Thanks for changing your approach here, I am much happier with the clone technique. I still have a few comments but nothing that seems structural.

I want to make sure @lgfang and yourself have a plan on the SSH repository.

I'd like to possibly recommend cloning it out in a well known directory (similar to how DSI does it), and skipping if we're already up-to-date or whatever. DSI uses this logic, it clones on developer boxes but uses evergreen to clone repos (in the project YAML config), we'll side-step needing API keys in Evergreen at all, and give Genny the same functionality as DSI which is less surprising.

jimoleary (commented):
consider a factory method but close to done.

thessem:
> @thessem In terms of changing the cloning approach to use a well-defined directory I'd prefer to keep this using a TempDir as that simplifies things in terms of implementation and this ensures a similar environment between Evergreen and running locally. Mothra is pretty small at this point, so it doesn't seem like it's a necessity that we check if the repo is already up to date or not because cloning is relatively quick.

This wasn't about speed, it was about consistency with how DSI works, and avoiding the need to add an auth token to evergreen at all. It is my preference in CI to have evergreen manage all git checkouts. Especially given the SSH <-> GitHub long lived token security requirements.

I haven't had a chance to do another re-review yet (busy morning) but this is just a preference and not a blocker, especially if the SSH -> HTTPS work is being carried out post-merge.

thessem (approved):

src/lamplib/src/genny/cli.py
```diff
@@ -582,17 +582,17 @@ def self_test(ctx: click.Context):
 )
 @click.pass_context
 def lint_yaml(ctx: click.Context, lint_format: bool):
-    from genny.tasks import yaml_linter
+    from genny.tasks.yaml_linter import YamlLinter
 
-    yaml_linter.main(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"], lint_format=lint_format)
+    YamlLinter().lint(genny_repo_root=ctx.obj["GENNY_REPO_ROOT"], lint_format=lint_format)
```
jimoleary, line 587: Updating to Linter and Generator classes feels much better for this!

jimoleary (approved):
LGTM

IamXander:
This is so awesome!

PR #1197, opened 2024-04-22
DEVPROD-6718 - Create documentation page for TPCH and Streams workload

MAhmadShah:
JIRA Ticket: https://jira.mongodb.org/browse/DEVPROD-6718

Whats Changed:
We have added new documentation pages for the TPCH and Streams workloads. These workloads are run through test control files in DSI and call several Genny workloads within them. We will use these pages to link via the test control files.


thessem (approved):
Happy with this as-is, although potentially add a comment saying that the docs you're linking into are auto-generated and may break.

Additionally, is it possible to auto-generate these tables?

MAhmadShah:
> Happy with this as-is, although potentially add a comment saying that the docs you're linking into are auto-generated and may break.
> 
> Additionally, is it possible to auto-generate these tables? Note this review is a tick so unless this is easy don't worry.

Great ideas, thanks James. I'll take a look to see if we can auto-gen these pages along with the other Genny docs.

PR #1061, opened 2023-11-09
DEVPROD-1999: Run Actor tests on as many topologies as possible 

PyGeek03:
**Jira Ticket:** [DEVPROD-1999](https://jira.mongodb.org/browse/DEVPROD-1999)

**Whats Changed:**  
Run Actor tests on as many topologies as possible.

Since [genny_standalone.yml was removed](https://github.com/mongodb/genny/pull/418/files#diff-3d7334da0f966176da042540b0d5f1f24cd6a657bda76d8d6b8c31457a84b06a) and hence the tag `[standalone]` disabled, some actor test cases have stopped running. For some actors, this resulted in a drift between an actor's new modifications and its test code, for example `CollectorScanner` actor. Such test code has been fixed in this PR.

However, I discovered 3 Actors that might have correctness bugs: StreamStatsReporter (for all topologies), MultiCollectionQuery (for `[three_node_replset]` and `[sharded]`), and DbCheckActor (for `[sharded]`). I disabled their tests for now, but I think these should be addressed in another ticket to ensure test results' integrity.

**Patch testing results:** 
Not necessary because all changes are in test code.

PR #1198, opened 2024-04-23
BF-32909: Provide tenantId and processorId to SP commands.

sandeepgd:
**Jira Ticket:** [BF-32909](https://jira.mongodb.org/browse/BF-32909)

### Whats Changed

Provide tenantId, processorId and options.featureFlags to SP commands.

### DIFF
--- a/src/cast_core/src/StreamStatsReporter.cpp
+++ b/src/cast_core/src/StreamStatsReporter.cpp
@@ -64,9 +64,15 @@ struct StreamStatsReporter::PhaseConfig {
             BOOST_THROW_EXCEPTION(InvalidConfigurationException("StreamStatsReporter only allows a single repeat"));
         }
 
+        std::string tenantId = phaseContext["TenantId"].to<std::string>();
         std::string streamProcessorName = phaseContext["StreamProcessorName"].to<std::string>();
         bsoncxx::builder::stream::document builder;
-        statsCommand = builder << "streams_getStats" << "" << "name" << streamProcessorName << bsoncxx::builder::stream::finalize;
+        statsCommand = builder
+            << "streams_getStats" << ""
+            << "tenantId" << tenantId
+            << "name" << streamProcessorName
+            << "processorId" << streamProcessorName
+            << bsoncxx::builder::stream::finalize;
     }
 };
 
--- a/src/cast_core/src/StreamStatsReporter.cpp
+++ b/src/cast_core/src/StreamStatsReporter.cpp
@@ -66,12 +66,13 @@ struct StreamStatsReporter::PhaseConfig {
 
         std::string tenantId = phaseContext["TenantId"].to<std::string>();
         std::string streamProcessorName = phaseContext["StreamProcessorName"].to<std::string>();
+        std::string streamProcessorId = phaseContext["StreamProcessorId"].to<std::string>();
         bsoncxx::builder::stream::document builder;
         statsCommand = builder
             << "streams_getStats" << ""
             << "tenantId" << tenantId
             << "name" << streamProcessorName
-            << "processorId" << streamProcessorName
+            << "processorId" << streamProcessorId
             << bsoncxx::builder::stream::finalize;
     }
 };
--- a/src/cast_core/test/StreamStatsReporter_test.cpp
+++ b/src/cast_core/test/StreamStatsReporter_test.cpp
@@ -48,6 +48,7 @@ TEST_CASE_METHOD(MongoTestFixture, "StreamStatsReporter successfully connects to
           Phases:
           - Repeat: 1
             StreamProcessorName: sp
+            StreamProcessorId: spid
             ExpectedDocumentCount: 1
     )", __FILE__);
 
--- a/src/cast_core/test/StreamStatsReporter_test.cpp
+++ b/src/cast_core/test/StreamStatsReporter_test.cpp
@@ -58,7 +58,7 @@ TEST_CASE_METHOD(MongoTestFixture, "StreamStatsReporter successfully connects to
                 << "streams_startStreamProcessor" << ""
                 << "tenantId" << "test"
                 << "name" << "sp"
-                << "processorId" << "sp"
+                << "processorId" << "test_spid"
                 << "pipeline" << bsoncxx::builder::stream::open_array
                 << bsoncxx::builder::stream::open_document
                 << "$source" << bsoncxx::builder::stream::open_document << "connectionName" << "__testMemory" << bsoncxx::builder::stream::close_document
--- a/src/cast_core/test/StreamStatsReporter_test.cpp
+++ b/src/cast_core/test/StreamStatsReporter_test.cpp
@@ -75,6 +75,9 @@ TEST_CASE_METHOD(MongoTestFixture, "StreamStatsReporter successfully connects to
                 << "options" << bsoncxx::builder::stream::open_document << bsoncxx::builder::stream::close_document
                 << bsoncxx::builder::stream::close_document
                 << bsoncxx::builder::stream::close_array
+                << "options" << bsoncxx::builder::stream::open_document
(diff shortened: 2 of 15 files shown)


src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -64,9 +64,15 @@ struct StreamStatsReporter::PhaseConfig {
...
 
+        std::string tenantId = phaseContext["TenantId"].to<std::string>();
         std::string streamProcessorName = phaseContext["StreamProcessorName"].to<std::string>();
         bsoncxx::builder::stream::document builder;
-        statsCommand = builder << "streams_getStats" << "" << "name" << streamProcessorName << bsoncxx::builder::stream::finalize;
+        statsCommand = builder
+            << "streams_getStats" << ""
+            << "tenantId" << tenantId
+            << "name" << streamProcessorName
+            << "processorId" << streamProcessorName
```
thessem, line 74: `processorId` is getting set to the `streamProcessorName` here, but added to the YAML files as a separate parameter, that does not get read here?

src/cast_core/test/StreamStatsReporter_test.cpp
```diff
@@ -56,7 +56,9 @@ TEST_CASE_METHOD(MongoTestFixture, "StreamStatsReporter successfully connects to
         try {
             auto startCommand = bsoncxx::builder::stream::document{}
                 << "streams_startStreamProcessor" << ""
+                << "tenantId" << "test"
                 << "name" << "sp"
+                << "processorId" << "sp"
```
thessem, line 61: I'd like to see this tested with "test_processorId" instead of repeating "sp", since I think it's not catching a bug in the implementation.

thessem (changes requested):

thessem:
@jimoleary as @mongodb/performance is a code-owner I set you as a reviewer manually since you've historically reviewed my Genny code. Hope you don't mind.

src/workloads/docs/StreamStatsReporter.yml
```diff
@@ -29,7 +30,9 @@ Actors:
       OperationName: RunCommand
       OperationCommand:
         streams_startStreamProcessor: ""
+        tenantId: *TenantId
         name: *StreamProcessorName
+        processorId: *StreamProcessorName
```
jimoleary, line 35: As noted above `processorId` is never actually used in this code. Either remove it or change the cpp to get it.

src/cast_core/src/StreamStatsReporter.cpp
jimoleary: Either remove `processorId` from all the yaml files or add and use the following (which will default it to `streamProcessorName` , feel free to simplify if this isn't what you want):
```
        auto processorId = phaseContext["processorId"].maybe<std::string>()
                                      .value_or(streamProcessorName);
```

jimoleary (changes requested):
`processorId` isn't used as expected.

Do we need a patch for this?

src/cast_core/src/StreamStatsReporter.cpp
sandeepgd: Ah, thanks for pointing this out. I removed processorId from the yml files.

src/cast_core/test/StreamStatsReporter_test.cpp
sandeepgd: The `processorId` value here does not matter too much for us, I just need to specify some value here since we plan to make it a required field. But yeah, I can use test_processorId as well. Done.

sandeepgd (commented):
Sorry for the late response. Addressed all your comments.

sandeepgd:
Jim/James, can you also advice what I need to fix this [patch](https://parsley.mongodb.com/evergreen/genny_amazon2023_arm64_t_compile_and_dry_run_asan_patch_d2bfd3e27087ba12a1b793758b0c124f6f53441c_662709294e37eb0007c71d04_24_04_23_01_04_43/0/task?bookmarks=0,2214,2223,2260) build failure. Do I need to specify TenantId under Phases like done [here](https://github.com/mongodb/genny/blob/e4f3077ab80ef4e4874e21b8765311c700419fe0/src/workloads/docs/StreamStatsReporter.yml#L26)?

src/cast_core/src/StreamStatsReporter.cpp
```diff
@@ -64,9 +64,15 @@ struct StreamStatsReporter::PhaseConfig {
             BOOST_THROW_EXCEPTION(InvalidConfigurationException("StreamStatsReporter only allows a single repeat"));
         }
 
+        std::string tenantId = phaseContext["TenantId"].to<std::string>();
```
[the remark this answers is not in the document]
sandeepgd: Ah! Thanks for the code pointer, made the change.

src/cast_core/test/StreamStatsReporter_test.cpp
```diff
@@ -56,7 +56,9 @@ TEST_CASE_METHOD(MongoTestFixture, "StreamStatsReporter successfully connects to
         try {
             auto startCommand = bsoncxx::builder::stream::document{}
                 << "streams_startStreamProcessor" << ""
+                << "tenantId" << "test"
                 << "name" << "sp"
+                << "processorId" << "test_spid"
```
jimoleary, line 61: Shouldn't this be the same as `name` ?

jimoleary (approved):
LGTM mod the question about the value of `processorId` in the test.

src/cast_core/test/StreamStatsReporter_test.cpp
thessem: I'm fairly sure this test isn't running in CI because we don't have an integration test setup for it yet. Otherwise I am a little confused as to how this is passing.

thessem (commented):
I am happy with this, aside from wondering why the unit test seems to pass without it looking correct. If you can demo a passing run of this unit test + a patch run of this functionality working I'm happy to tick it.

src/cast_core/test/StreamStatsReporter_test.cpp
sandeepgd: processorId can be anything, our stream processor does not require it to be the same as the name. Let me get a successful patch build to resolve your concerns.

sandeepgd: Alright here is a successful patch build: 
https://spruce.mongodb.com/version/66352ff5c0fc690007de5172/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I made a change since the last revision. Now I specify processorId and also options.featureFlags fields which are required fields.

PyGeek03: Hi @sandeepgd, I tried to enable this test as part of #1061, but this test failed and I couldn't find a way to fix it, so I decided to [leave it disable](https://github.com/mongodb/genny/pull/1061/commits/b94c417c4140379338f20fca576e6a3d82f12f02). If possible, could you try to enable and fix this test? Thank you!

sandeepgd: @PyGeek03 I see! Would you be able to file a separate ticket for this with some context? I am not the expert on how our perf tests are set up. @harendra-c would be able to look into the ticket separately.

sandeepgd:
@thessem can you also approve the pr please since you requested changes?

IamXander:
@thessem is out @lgfang or @PyGeek03 could one of you confirm that @thessem 's concerns are addressed?

lgfang:
> @thessem is out @lgfang or @PyGeek03 could one of you confirm that @thessem 's concerns are addressed?

 Approving this as I've noticed that a patch has been executed as per James' request.

lgfang (approved):

PR #1199, opened 2024-04-23
replacing 1 node aff tags with 3 node

jawwadasghar:
**Jira Ticket:** [FOO-123](https://jira.mongodb.org/browse/FOO-123)

### DIFF
--- a/src/workloads/execution/BackgroundIndexConstruction.yml
+++ b/src/workloads/execution/BackgroundIndexConstruction.yml
@@ -228,7 +228,7 @@ AutoRun:
       $eq:
       - single-replica
       - single-replica-80-feature-flags
-      - single-replica-all-feature-flags
+      - replica-all-feature-flags
     branch_name:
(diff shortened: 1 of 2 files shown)


jimoleary (approved):
LGTM
I have to admint I spent too much time trying to figure out what aff was ... until I looked at the changed files.

jawwadasghar:
> LGTM I have to admint I spent too much time trying to figure out what aff was ... until I looked at the changed files.

It's a new acronym I'm tossing out there ;)

jordiolivares (approved):
LGTM. The intention behind the single node was to make the benchmarks cheaper to run. It doesn't add much value to toss in some secondary nodes since we're only interested in the primary here

jawwadasghar:
@jordiolivares Make sense! 
We are not supporting the 1 node anymore as we think the additional (slightly cheaper) variant isn't worth the additional complexity / overhead.

PR #1202, opened 2024-04-25
PERF-5094 Commit QE Range release criteria tests

marksg07:
See: https://github.com/10gen/mongo/pull/21731 For change in mongo repo.

**Jira Ticket:** [PERF-5094](https://jira.mongodb.org/browse/PERF-5094)

### Whats Changed

- Adds full support for Range V2 with trim factor + query type name update (rangePreview -> range)
- Adds six new workloads under `workloads/encrypted` which test different parameter combinations for QE Range, along with required data files housed in `workloads/contrib/qe_range_testing` to make them work
  - These workloads are all generated, and each consist of an insert phase and a mixed query/update phase
  - Each phase has 16 actors, each of which works on a fixed subset of the data.
- Adds a new task, `qe_range_balance_50_50_high_value`, which runs every 4 hours (in line with the other "high_value" tasks). The actual genny workload we run takes about 10 minutes total.
- Adds the python scripts used for data generation and workload creation along with rudimentary documentation in `workloads/contrib/qe_range_testing` 

### Patch Testing Results

https://spruce.mongodb.com/version/66296c32cf97390007df718b/tasks

### Workload Submission form

Filled out.


thessem (approved):
As codeowner of the `.cpp` file changes, LGTM.

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+  
+    EncryptionOptions:
+      KeyVaultDatabase: "keyvault"
+      KeyVaultCollection: "datakeys"
+      EncryptedCollections:
+      - genny_release_criteria.my_coll
+  
+
+Actors:
+
```
jimoleary, line 52: nit: whitespace?

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+    - *nop
+    - Repeat: 6250
+      Collection: *encrypted_coll
+      MetricsName: "load"
+      Operations:
+      - OperationName: insertOne
+        OperationMetricsName: inserts
+        OperationCommand:
+          Document:
+            tm_retail_tx: {^ConvertToInt: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_timestamps.txt", "sequential": true, "startFromLine": 0}}}}
```
jimoleary, line 70: Can this path be made relative? It may not be possible but I don't see how I would run this workload locally without having to artificially create the DSI directory layout which would be irritating.

Although the ```CryptSharedLibPath``` also makes this difficult.

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+
+  
+  
+  - Name: InsertActor_Thread0
+    Type: CrudActor
+    Threads: 1
+    Database: *encrypted_db
+    ClientName: EncryptedPool
+    Phases:
+    - *nop
```
jimoleary, line 61: Is there a particular reason why there is nothing in phase 0?

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+      KeyVaultCollection: "datakeys"
+      EncryptedCollections:
+      - genny_release_criteria.my_coll
+  
+
+Actors:
+
+  
+  
+  - Name: InsertActor_Thread0
```
jimoleary, line 55: Can we template / parameterize these insert actors the only changes are the startFromLine AFAICS.

See template definition [here](https://github.com/mongodb/genny/blob/f97023de49be506ec51b49f7caaecb85a1835d24/src/phases/query/CollScanSimplifiablePredicate.yml#L14-L31) and usage [here](https://github.com/mongodb/genny/blob/f97023de49be506ec51b49f7caaecb85a1835d24/src/phases/query/CollScanSimplifiablePredicate.yml#L459-L477)

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+      Operations:
+      - OperationName: insertOne
+        OperationMetricsName: inserts
+        OperationCommand:
+          Document:
+            tm_retail_tx: {^ConvertToInt: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_timestamps.txt", "sequential": true, "startFromLine": 0}}}}
+            age_hospitals: {^ConvertToInt32: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_ages.txt", "sequential": true, "startFromLine": 0}}}}
+            bnk_bal: {^ConvertToDecimal: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_balances.txt", "sequential": true, "startFromLine": 0}}}}
+
+            _id: {^Inc: {start: 0, multiplier: 500000}}
```
jimoleary, line 74: Do you actually care what type and value the _id is ? It doesn't seem to be used anywhere else?

src/workloads/encrypted/qe-range-age-100-0.yml
jimoleary, line 55: The multitude of insert actors seems to look as if it is just a way to load the data in parallel. Is there a functional reason for this or is it just to speed things up?

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,1036 @@
...
+        - Name: Update
+          Operations:
+          - OperationName: updateOne
+            OperationMetricsName: update
+            OperationCommand:
+              Filter: { age_hospitals: {^ConvertToInt32: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_ages_up.txt"}}}}}
+              Update: {$set: { age_hospitals: {^ConvertToInt32: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_ages.txt"} } } } } }
+          Transitions: *transitions
+
+  - Name: FSMActor_Thread5
```
jimoleary, line 600: These actors could also be templated as startFromLine seems to be the only variation.

jimoleary (changes requested):
I'm only really reviewing the cpp and yml files as I'm not convinced the python and temp code should be checked in here. How often are these files likely to change and could we just check in the generated files instead?

It also looks like this code will only run on an ubuntu spawn host , which is not particularly clear.


There is a huge amount of duplication in the workloads that could be removed through actor templates.

marksg07:
@jimoleary 

> I'm only really reviewing the cpp and yml files as I'm not convinced the python and temp code should be checked in here. How often are these files likely to change and could we just check in the generated files instead?

I was under the impression that the `workloads/contrib` directory was intended for stuff like workload/data generation code? We already have checked in `workloads/contrib/qe_test_gen`, which contains very similar contents (a python script that does workload/data gen).

> It also looks like this code will only run on an ubuntu spawn host , which is not particularly clear.

I think the only system-related dependency we have is on the path to libmongocrypt, and that's only a dependency for running genny locally, which is why I have the README section about changing that. Without that, you can still generate the workloads + data.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: The path already is relative; this is just how the directory structure is set up on Evergreen. Unfortunately I don't think there's a way to make a workload which has paths that work for both local and Evergreen, which is why I write to the separate `workloads/local` and `workloads/evergreen` paths in my experiment generation script. This is an existing problem though; see https://github.com/mongodb/genny/blob/1ce2f0de326b66764d5bdb66349178c386a5a168/src/workloads/docs/ChooseFromDataset.yml#L8-L9

src/workloads/contrib/qe_range_testing/qe_range_testing.py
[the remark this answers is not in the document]
marksg07: Yeah, this is a Python thing, it's needed to make a Python package called `qe_range_testing` in this directory which the Jinja2 `PackageLoader` can then look at to know where to load templates from.

src/workloads/contrib/qe_range_testing/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Then, using this directory as your current working directory, invoke Poetry to install the required dependencies into your local virtual environment, using the specific versions enumerated in `poetry.lock`.
+
+```sh
+poetry install
+```
+
+## General Usage
+
+Note: For local development, you will have to edit the `MONGO_CRYPT_PATH` constant in `experiment_generator.py` to point to `mongo_crypt_v1.so` on your local machine. The easiest way to get this file is to run the `crypt_create_lib` task in the `Amazon Linux 2 arm64 Crypt Compile` build variant and download the resulting shared library file. Otherwise, you'll have to compile the library yourself; instructions not attached.
```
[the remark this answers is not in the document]
marksg07: I hesitate to spend the time on automating this since this step is only necessary if someone wants to regenerate the workload files and run them locally -- the shared library is not necessary for testing with Evergreen. I can make that more clear in the README.

src/gennylib/src/encryption.cpp
```diff
@@ -403,7 +405,7 @@ void QueryableEncryptedField::appendEncryptInfo(sub_document subdoc) const {
                         throw InvalidConfigurationException(ss.str());
                     }
                 };
-                if(query.queryType == "rangePreview") {
+                if(query.queryType == "range") {
```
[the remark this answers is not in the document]
marksg07: We don't have any committed workloads testing rangePreview, so I don't think this should be a problem. This was for earlier testing of QE Range which was never committed to the repo.

src/workloads/contrib/qe_range_testing/experiment_generator.py
```diff
@@ -0,0 +1,203 @@
+from jinja2 import Environment, PackageLoader, select_autoescape
+from data_generation import generate_all_data
+
+env = Environment(
+    loader=PackageLoader("qe_range_testing"), autoescape=select_autoescape()
+)
+
+# Path to mongo crypt locally, point to your local mongo_crypt_v1.so object
+MONGO_CRYPT_PATH = "/home/ubuntu/mongo_crypt/lib/mongo_crypt_v1.so"
```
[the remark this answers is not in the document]
marksg07: If the user wants to run workloads locally, they will have to point this to the shared lib, it's not OS specific.

src/workloads/contrib/qe_range_testing/README.md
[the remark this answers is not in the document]
marksg07: I was thinking that it could be helpful at least for reproducibility purposes. I don't anticipate that we will sorely miss the ability to generate workloads that can be run locally, but given the chance that we might want that functionality in the future, and I'm checking the Evergreen workload generation code in anyways, it doesn't seem prudent to remove the local generation code.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: Nope, deleted.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: I've done some templatization here, let me know if that is what you're looking for.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: Nope, removed.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: It functions both as a way to speed up load, and as a test of parallel insertion with many threads, which is an important test for QE Range.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: Templated here as well.

src/workloads/encrypted/qe-range-age-100-0.yml
marksg07: Fixed some of the whitespace issues caused by generation.

src/workloads/encrypted/qe-range-age-100-0.yml
jimoleary, line 1: Thanks for the templates ... this make it much easier to follow and reduces the duplication.

src/workloads/encrypted/qe-range-age-100-0.yml
```diff
@@ -0,0 +1,318 @@
...
+    Phases: 
+    - Repeat: 6250
+      Collection: *encrypted_coll
+      MetricsName: "load"
+      Operations:
+      - OperationName: insertOne
+        OperationMetricsName: inserts
+        OperationCommand:
+          Document:
+            tm_retail_tx: {^ConvertToInt: {from: {^ChooseFromDataset:{"path": "./src/genny/src/workloads/contrib/qe_range_testing/data/rc_timestamps.txt", "sequential": true, "startFromLine": &StartLine {^Parameter: {Name: "StartLine", Default: -1}}}}}}
```
jimoleary, line 62: I've created DEVPROD-7205 to address the paths, it's irritating that you need to know the DSI directory structure.

src/workloads/encrypted/qe-range-age-100-0.yml
jimoleary: I've raised DEVPROD-7205 for this. It's a bit brittle and annoying.

jimoleary (approved):
LGTM on the tests.

Do we have any idea how noisy these tests are going to be?

marksg07:
> LGTM on the tests.
> 
> Do we have any idea how noisy these tests are going to be?

@jimoleary I'm hoping that the answer is "not very". The important metrics (query/read latency) are measured over a period of a few minutes, so they shouldn't be super spiky. And the relevant code shouldn't change all that often. 

Going ahead with the merge now, thank you for reviewing.

PR #1204, opened 2024-04-28
PERF-1885 unique index benchmark

wh5a:
**Jira Ticket:** [PERF-1885](https://jira.mongodb.org/browse/PERF-1885)

### Whats Changed

A new workload that tests performance of unique indexes, in particular concurrent insertion of duplicate keys. It also tests concurrent update and deletion of documents using the indexes, starting from 8 threads up to 128 threads.

### Patch Testing Results

https://spruce.mongodb.com/version/662d950c48f968000712eaf6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/execution/UniqueIndexes.yml
+++ b/src/workloads/execution/UniqueIndexes.yml
@@ -1,0 +1,214 @@
+SchemaVersion: 2018-07-01
+Owner: Storage Execution
+Description: |
+  This workload tests concurrent insert/update/delete performance with unique indexes,
+  a v2 unique index on 'a' and a v1 unique index on 'b'. The documents being inserted have
+  duplicate keys across threads, and the uniqueness of indexes allows only one insertion
+  to succeed. It first performs the operations with 8 threads, then 32, up to 128.
+
+Keywords:
+- insert
+- update
+- delete
+- unique indexes
+
+string_length: &string_length 10
+max_phases: &max_phases 9
+db: &db test
+coll: &coll test
+iterations: &iterations 12800
+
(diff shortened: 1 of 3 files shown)


thessem (approved):
Only leaving a review for `docs/using.md` as that's the file my team is marked as owning. Looks good to me.

src/workloads/execution/UniqueIndexes.yml
```diff
@@ -0,0 +1,214 @@
...
+      OnlyActiveInPhases:
+        Active: {^Parameter: {Name: "ActivePhases", Default: []}}
+        NopInPhasesUpTo: *max_phases
+        PhaseConfig:
+          Collection: *coll
+          Repeat: *iterations
+          ThrowOnFailure: false
+          Operations:
+          - &InsertOp
+            OperationName: insertMany
```
louiswilliams, line 43: In thinking about diagnosability and debugability, I worry that there are a few too many variables that might affect the latency of this phase, which would make it hard to interpret performance changes:
1. The latency of a successful insert
2. The latency of an unsuccessful insert due to an already-present key
3. The latency of an unsuccessful insert that fails due to a concurrent insertion of the same key

We can isolate these distinct behaviors by separating them into different phases. We can add an uncontended load phase to handle case 1, free of any duplicate key errors or concurrent writes. We can also isolate cases 2 & 3 by running an insert phase with a single thread.

Additionally, we could further isolate the behavior here by having one phase that only inserts duplicate keys.

PR #1205, opened 2024-04-29
DEVPROD-5390 Fix linting errors and enable yamllint for evergreen checks

lgfang:
**Jira Ticket:** [DEVPROD-5390](https://jira.mongodb.org/browse/DEVPROD-5390)

**Whats Changed:**

Revise `.yamllint` to match `DSI` and `PrivateWorkloads`, then fix linting errors and turn on yamllint in evergreen checks. Lastly, removing leading `0` of values in `maps_medical.yml` using a sed command.

This PR contains multiple commits, which are a mixture of `yamlfmt`, `prettier` and manual fixes.

**Patch testing results:**

None

**Related PRs:**

None


evergreen.yml
```diff
@@ -41,276 +41,273 @@
...
+          args:
+            - -v
+            - resmoke-test
+            - --create-new-actor-test-suite
 
 ##                ⚡️ Task Groups ⚡️
 
 task_groups:
+  - &a_compile_and_test
+
```
jimoleary, line 242: This new line is odd, any idea why it got added?

evergreen.yml
```diff
@@ -41,276 +41,273 @@
...
+      # For this task group, the implicit dependency is that t_*_test has to
+      # run after t_compile_and_dry_run.
+      - t_compile_and_dry_run
+      - t_python_test
+      # - t_lint_python
+      - t_lint_workloads
+      - t_check_generated_docs
+      - t_cmake_test
+
+  - !!merge <<: *a_compile_and_test
```
jimoleary, line 264: Should we should be using merge expansions. Do they work with out parser and / or are they standard?

evergreen.yml
```diff
@@ -320,199 +317,197 @@ functions:
...
-      - -v
-      - lint-yaml
-
+    - command: subprocess.exec
+      params:
+        binary: ./src/genny/run-genny
+        args:
+          - -v
+          - lint-yaml
+          - --format
```
jimoleary, line 461: Did you add this? I was hoping / assuming that all these were just whitespace linting changes.

src/lamplib/src/genny/tasks/yaml_linter.py
```diff
@@ -56,7 +56,12 @@ def lint(self, genny_repo_root: str, lint_format: bool):
 
         if lint_format:
             config_file_path = path.join(genny_repo_root, ".yamllint")
-            yamllint_argv = ["--strict", "--config-file", config_file_path] + all_yamls
+            yamllint_argv = (
```
jimoleary, line 59: why change this line from `all_yamls` ... it the same thing?

src/phases/execution/ClusteredCollection.yml
```diff
@@ -6,8 +6,14 @@ Description: |
   key such as monotonically increasing inserts, range queries and range deletions.
 
 GlobalDefaults:
-  RecordIdExtraPaddingParam: &RecordIdExtraPaddingParam {^Parameter: {Name: "RecordIdExtraPadding", Default: 0}}
-  PaddingGenerator: &PaddingGenerator {^FastRandomString: {length: *RecordIdExtraPaddingParam, alphabet: "a"}}
+  RecordIdExtraPaddingParam:
```
jimoleary, line 9: I'm assuming any changes after this are lint / whitespace only?

src/phases/query/RepeatedPathTraversal.yml
```diff
@@ -20,305 +20,258 @@ GlobalDefaults:
...
+          Repeat: 1
+          Database: *Database
+          Threads: 1
+          CollectionCount: 1
+          DocumentCount: *DocumentCount
+          BatchSize: 1000
+          Document: {
+            a: [{^Repeat: {count: 3, fromGenerator: {b: [&InnerDoc {c: 0, d: 0, e: 0, f: 0, g: 0, h: 0, i: 0, j: 0, k: 0, l: 0}]}}}],
+            # 5 levels deep
+            deep5: [{f1: [{f2: [{f3: [{f4: [{f5: [*InnerDoc]}, {f5: [*InnerDoc]}]}]}]}]}],
```
jimoleary, line 76: I find it a bit odd that it adds spaces for some `}}` combos but not for others. e.g [here](https://github.com/mongodb/genny/pull/1205/files?diff=split&w=1#diff-171b956f545f03dd1edd5321d1bcf1e9a081f51a4bc4298d2a6480f044da87a6R16) or [here](https://github.com/mongodb/genny/pull/1205/files?diff=split&w=1#diff-a9d48c8939e97aec7c441433a0b31da6cf17c0e3cdde684c181cd9214f5de6d0R45)

src/phases/replication/startup/StartupPhasesTemplate.yml
```diff
@@ -49,40 +49,40 @@ Document: &Doc
...
-  name: "index_int0_int1_int2"
-- key: {int3: 1}
-  name: "index_int3"
-- key: {int4: 1}
-  name: "index_int4"
-- key: {int5: 1, int6: 1}
-  name: "index_int5_int6"
+IndexesForCreateCmd: &IndexesForCreateCmd # 10 indexes
+  - key: {t: 1, w: 1}
+    name: "index_t_w" # 10 indexes
```
jimoleary, line 67: Did this comment move manually or automatically ?

jimoleary (commented):
We're going to need a full patch for this change.

**Do these changes automaticlly check and enforce lint format as part of the commit Q?** It looks like they do, but they must otherwise we'll end up with a mix of valid and invalid yml. We should also ensure that the commit Q gives a resonable error message if they do.

I didn't exhaustively review the yml files, there are too many of them (>200) and I assume the changes were automated ... but there is some oddness in there with whitespace handling and comments moving from anchor definitions to where it is used (which makes me think some the changes were manual).

thessem:
> I didn't exhaustively review the yml files, there are too many of them (>200) and I assume the changes were automated ... but there is some oddness in there with whitespace handling and comments moving from anchor definitions to where it is used (which makes me think some the changes were manual).

This may be due to one of the formatting passes being used, I can imagine this occurring because if a formatter is parsing an AST and then re-outputting it, it could lose where the comment was anchored.

Either that or there was manual changes...

thessem:
@lgfang could I ask that you rebase this branch to group together the manual changes, and have one commit for each automated formatting pass? That way this can be reviewed commit-by-commit.

src/lamplib/src/genny/tasks/yaml_linter.py
lgfang: `all_yamls` is a list of all yaml files, which is really long. This change is to display the top level directories only instead of the entire list. IMO, this improves the readability of the screen output.

evergreen.yml
lgfang: yes, I did it. Without this option, this command does not check the format of the yaml files.

src/phases/execution/ClusteredCollection.yml
lgfang: The [commits](https://github.com/mongodb/genny/pull/1205/commits) tab lists all the commits. Only the following commits are not lint/whitespace related. That said, I totally understand that the mixture of manual fixes and different formatters has made the PR hard to review. Please allow me sometime to re-org the changes to make it hopefully easier to review.
- [DEVRPOD-5390 update .yamllint](https://github.com/mongodb/genny/pull/1205/commits/5689626ff6475b39d872503e8e44a3babc063cb2)
- [turn on yaml linting for evergreen check](https://github.com/mongodb/genny/pull/1205/commits/5c5c8b0673b7b6d4532c88e8e15a1e45a5f99ec7)
- [Merge branch 'master' into lungang.fang/DEVPROD-5390_lint-genny-take2](https://github.com/mongodb/genny/pull/1205/commits/fa26a8cf84a4f8cf72336deb10185eeb0a043413)

src/phases/replication/startup/StartupPhasesTemplate.yml
lgfang: automatically

src/lamplib/src/genny/tasks/yaml_linter.py
jimoleary: yeah, that makes sense as long as ```yamllint.cli.run(yamllint_argv)``` matches the same files (and it would be surprising if it didn't).

Maybe move the code from line 36 to line 65,  ```all_yamls``` is only used in the log line now?

evergreen.yml
jimoleary: :+1:

src/phases/execution/ClusteredCollection.yml
jimoleary: Thanks ! that helps alot. I was a bit confused by the merge commit but then I was also on the other PR review.

jimoleary (commented):
So with the explanation of what was manual and what was automatic, I feel a bit better about the PR.  

But we still need a full patch and I'm still concerned about the usage of ```!!merge``` in evergreen.yml Although this is not a direct issue for genny, and may just be an issue for me.

thessem (changes requested):
To check this PR I ran a PyYAML (which is YAML 1.1) script to convert all YAML files to JSON files. I had to set it to enable multiple anchors and multiple documents. We have some messy YAMLS.

I ran the script twice, once on master and once on this PR. The following is a diff of the two.

``` diff
diff --git a/src/phases/tpch/normalized/Q1.0.json b/src/phases/tpch/normalized/Q1.0.json
index adfab50a4..c72da56c0 100644
--- a/src/phases/tpch/normalized/Q1.0.json
+++ b/src/phases/tpch/normalized/Q1.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 1 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 1 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query1Delta": {
         "^Parameter": {
             "Name": "Query1Delta",
             "Default": -90
diff --git a/src/phases/tpch/normalized/Q10.0.json b/src/phases/tpch/normalized/Q10.0.json
index 6f2a8a960..7fc13a7ec 100644
--- a/src/phases/tpch/normalized/Q10.0.json
+++ b/src/phases/tpch/normalized/Q10.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 10 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 10 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query10Date": {
         "^Parameter": {
             "Name": "Query10Date",
             "Default": "1993-10-01"
diff --git a/src/phases/tpch/normalized/Q11.0.json b/src/phases/tpch/normalized/Q11.0.json
index 3d400a063..da2b8c74e 100644
--- a/src/phases/tpch/normalized/Q11.0.json
+++ b/src/phases/tpch/normalized/Q11.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 11 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 11 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query11Nation": {
         "^Parameter": {
             "Name": "Query11Nation",
             "Default": "GERMANY"
diff --git a/src/phases/tpch/normalized/Q12.0.json b/src/phases/tpch/normalized/Q12.0.json
index 0452a7f1a..d2dd1cfe0 100644
--- a/src/phases/tpch/normalized/Q12.0.json
+++ b/src/phases/tpch/normalized/Q12.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 12 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 12 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query12ShipMode1": {
         "^Parameter": {
             "Name": "Query12ShipMode1",
             "Default": "MAIL"
diff --git a/src/phases/tpch/normalized/Q13.0.json b/src/phases/tpch/normalized/Q13.0.json
index 5ef77b577..bcf450cf7 100644
--- a/src/phases/tpch/normalized/Q13.0.json
+++ b/src/phases/tpch/normalized/Q13.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 13 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 13 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query13Regex": {
         "^Parameter": {
             "Name": "Query13Regex",
             "Default": "^.*special.*requests.*$"
diff --git a/src/phases/tpch/normalized/Q14.0.json b/src/phases/tpch/normalized/Q14.0.json
index b25a34038..d93819c7d 100644
--- a/src/phases/tpch/normalized/Q14.0.json
+++ b/src/phases/tpch/normalized/Q14.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 14 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 14 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query14Date": {
         "^Parameter": {
             "Name": "Query14Date",
             "Default": "1995-09-01"
diff --git a/src/phases/tpch/normalized/Q15.0.json b/src/phases/tpch/normalized/Q15.0.json
index 14fea0c89..09ea30a65 100644
--- a/src/phases/tpch/normalized/Q15.0.json
+++ b/src/phases/tpch/normalized/Q15.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 15 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 15 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query15Date": {
         "^Parameter": {
             "Name": "Query15Date",
             "Default": "1996-01-01"
diff --git a/src/phases/tpch/normalized/Q16.0.json b/src/phases/tpch/normalized/Q16.0.json
index 37c4ed0f0..b4e6716a7 100644
--- a/src/phases/tpch/normalized/Q16.0.json
+++ b/src/phases/tpch/normalized/Q16.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 16 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 16 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query16Brand": {
         "^Parameter": {
             "Name": "Query16Brand",
             "Default": "Brand#45"
diff --git a/src/phases/tpch/normalized/Q17.0.json b/src/phases/tpch/normalized/Q17.0.json
index a20ba3be2..77fca7f84 100644
--- a/src/phases/tpch/normalized/Q17.0.json
+++ b/src/phases/tpch/normalized/Q17.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 17 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 17 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query17Brand": {
         "^Parameter": {
             "Name": "Query17Brand",
             "Default": "Brand#23"
diff --git a/src/phases/tpch/normalized/Q18.0.json b/src/phases/tpch/normalized/Q18.0.json
index 242939f0d..f48721589 100644
--- a/src/phases/tpch/normalized/Q18.0.json
+++ b/src/phases/tpch/normalized/Q18.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 18 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 18 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query18Quantity": {
         "^Parameter": {
             "Name": "Query18Quantity",
             "Default": 300
diff --git a/src/phases/tpch/normalized/Q19.0.json b/src/phases/tpch/normalized/Q19.0.json
index d3331a188..a951664c0 100644
--- a/src/phases/tpch/normalized/Q19.0.json
+++ b/src/phases/tpch/normalized/Q19.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 19 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 19 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query19Brand1": {
         "^Parameter": {
             "Name": "Query19Brand1",
             "Default": "Brand#12"
diff --git a/src/phases/tpch/normalized/Q20.0.json b/src/phases/tpch/normalized/Q20.0.json
index be4d8af35..781f90872 100644
--- a/src/phases/tpch/normalized/Q20.0.json
+++ b/src/phases/tpch/normalized/Q20.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 20 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 20 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query20Nation": {
         "^Parameter": {
             "Name": "Query20Nation",
             "Default": "CANADA"
diff --git a/src/phases/tpch/normalized/Q21.0.json b/src/phases/tpch/normalized/Q21.0.json
index b4405eb9b..089bcf23c 100644
--- a/src/phases/tpch/normalized/Q21.0.json
+++ b/src/phases/tpch/normalized/Q21.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 21 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 21 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query21Nation": {
         "^Parameter": {
             "Name": "Query1Delta",
             "Default": "SAUDI ARABIA"
diff --git a/src/phases/tpch/normalized/Q22.0.json b/src/phases/tpch/normalized/Q22.0.json
index 840b883c8..1ed3cb075 100644
--- a/src/phases/tpch/normalized/Q22.0.json
+++ b/src/phases/tpch/normalized/Q22.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 22 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 22 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query22CountryCodes": {
         "^Parameter": {
             "Name": "Query22CountryCodes",
             "Default": [
diff --git a/src/phases/tpch/normalized/Q3.0.json b/src/phases/tpch/normalized/Q3.0.json
index 4706ff728..20481a052 100644
--- a/src/phases/tpch/normalized/Q3.0.json
+++ b/src/phases/tpch/normalized/Q3.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 3 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 3 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query3Segment": {
         "^Parameter": {
             "Name": "Query3Segment",
             "Default": "BUILDING"
diff --git a/src/phases/tpch/normalized/Q4.0.json b/src/phases/tpch/normalized/Q4.0.json
index 85e3c81a0..1f220480c 100644
--- a/src/phases/tpch/normalized/Q4.0.json
+++ b/src/phases/tpch/normalized/Q4.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 4 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 4 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query4Date": {
         "^Parameter": {
             "Name": "Query4Date",
             "Default": "1993-07-01"
diff --git a/src/phases/tpch/normalized/Q5.0.json b/src/phases/tpch/normalized/Q5.0.json
index 71b467e16..e8e4eac3f 100644
--- a/src/phases/tpch/normalized/Q5.0.json
+++ b/src/phases/tpch/normalized/Q5.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 5 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 5 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query5Region": {
         "^Parameter": {
             "Name": "Query5Region",
             "Default": "ASIA"
diff --git a/src/phases/tpch/normalized/Q6.0.json b/src/phases/tpch/normalized/Q6.0.json
index 3336fef7f..b1eb06f49 100644
--- a/src/phases/tpch/normalized/Q6.0.json
+++ b/src/phases/tpch/normalized/Q6.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 6 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 6 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query6Date": {
         "^Parameter": {
             "Name": "Query6Date",
             "Default": "1994-01-01"
diff --git a/src/phases/tpch/normalized/Q7.0.json b/src/phases/tpch/normalized/Q7.0.json
index a80eaeef9..72543de7b 100644
--- a/src/phases/tpch/normalized/Q7.0.json
+++ b/src/phases/tpch/normalized/Q7.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 7 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 7 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query7Nation1": {
         "^Parameter": {
             "Name": "Query7Nation1",
             "Default": "FRANCE"
diff --git a/src/phases/tpch/normalized/Q8.0.json b/src/phases/tpch/normalized/Q8.0.json
index 45b35b70f..29f876e6c 100644
--- a/src/phases/tpch/normalized/Q8.0.json
+++ b/src/phases/tpch/normalized/Q8.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 8 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 8 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query8Type": {
         "^Parameter": {
             "Name": "Query8Type",
             "Default": "ECONOMY ANODIZED STEEL"
diff --git a/src/phases/tpch/normalized/Q9.0.json b/src/phases/tpch/normalized/Q9.0.json
index 8366a84e0..41fe41fb2 100644
--- a/src/phases/tpch/normalized/Q9.0.json
+++ b/src/phases/tpch/normalized/Q9.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/product-query",
-    "Description": "Run TPC-H query 9 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema.\nUsing an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
+    "Description": "Run TPC-H query 9 (see http://tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.0.pdf) against the normalized schema. Using an 'executionStats' explain causes each command to run its execution plan until no\ndocuments remain, which ensures that the query executes in its entirety.\n",
     "batchSize": {
         "^Parameter": {
             "Name": "BatchSize",
             "Default": 101
         }
     },
     "query9Color": {
         "^Parameter": {
             "Name": "Query9Color",
             "Default": "^.*green.*$"
diff --git a/src/workloads/contrib/qe_test_gen/patchConfig.0.json b/src/workloads/contrib/qe_test_gen/patchConfig.0.json
index 2f516ddf2..97707e715 100644
--- a/src/workloads/contrib/qe_test_gen/patchConfig.0.json
+++ b/src/workloads/contrib/qe_test_gen/patchConfig.0.json
@@ -1,14 +1,14 @@
 {
     "workload_name": "qe_test_insert",
     "Owner": "Server Security",
-    "Description": "This workload evaluates the performance of Queryable Encryption against the established \u201cQueryable Encryption Performance Release Criteria.\u201d\n",
+    "Description": "This workload evaluates the performance of Queryable Encryption against the established \u201cQueryable Encryption Performance Release Criteria.\u201d \n",
     "patches": {
         "ReplaceMeWithPatchID": {
             "linux-1-node-replSet-fle.2022-11": [
                 "medical_workload_states_100_0",
                 "medical_workload_states_95_5",
                 "medical_workload_states_50_50",
                 "medical_workload_diagnosis_100_0",
                 "medical_workload_diagnosis_95_5",
                 "medical_workload_diagnosis_50_50",
                 "medical_workload_status_100_0",
diff --git a/src/workloads/docs/ExternalScriptActor.0.json b/src/workloads/docs/ExternalScriptActor.0.json
index cc83d151f..e99db99a9 100644
--- a/src/workloads/docs/ExternalScriptActor.0.json
+++ b/src/workloads/docs/ExternalScriptActor.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "This workload was created to test an external script runner as per PERF-3198.\nThe execution stats of the script will be collected with metrics name \"ExternalScript\"\nIf the script writes and only writes an integer to stdout as result, the result will be\ncollected to the specified metrics name (DefaultMetricsName as default)\n",
+    "Description": "This workload was created to test an external script runner as per PERF-3198.\nThe execution stats of the script will be collected with metrics name \"ExternalScript\"\nIf the script writes and only writes an integer to stdout as result, the result will be collected to the specified metrics name (DefaultMetricsName as default)\n",
     "Actors": [
         {
             "Name": "MongoshScriptRunner",
             "Type": "ExternalScriptRunner",
             "Threads": 5,
             "Phases": [
                 {
                     "Repeat": 5,
                     "Command": "mongosh",
                     "MetricsName": "MyMetrics",
diff --git a/src/workloads/docs/HelloWorld-MultiplePhases.0.json b/src/workloads/docs/HelloWorld-MultiplePhases.0.json
index 8051a6e51..a7ab4582f 100644
--- a/src/workloads/docs/HelloWorld-MultiplePhases.0.json
+++ b/src/workloads/docs/HelloWorld-MultiplePhases.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Performance Analysis",
-    "Description": "This workload illustrates how to think of Phases happening\nsimultaneously and how to configure Repeat and Duration\nwhen you have more than one Actor in a Workload.\n\nStart by quickly looking at the actual workload yaml below\n\nAlthough we write workloads by listing each Actor and then\nlisting each Phase for that Actor, Genny workloads actually\noperate in order of each Phase. We construct each Actor and\nthen start Phase 0. We wait until all Actors say that they\nare done with Phase 0 and then we move on to Phase 1, etc.\nActors using PhaseLoop automatically do this.\n\nIt may help to think of the Actor yamls as being written in\nparallel rather than in sequence\n\n    Actors:\n    - Type: HelloWorld             - Type: HelloWorld\n      Name: A                        Name: B\n      Threads: 10                    Threads: 2\n      Phases:                        Phases:\n\n      - Message: A.0                 - Message: B.0\n        Duration: 2 seconds            Repeat: 10\n\n      - Message: A.1                 - Message: B.1\n        Blocking: None                 Repeat: 1e3\n\nThe HelloWorld Actor echoes its \"Message\" parameter in a\nloop.\n\nThere will be 10 threads running HelloWorld with name \"A\"\nand 2 threads running with name \"B\".\n\n== Phase 0 ==\n\n          A                          B\n      Message: A.0              Message: B.0\n      Duration: 2 seconds       Repeat: 10\n\nThe first phase (Phase 0) starts and all 12 threads (10 from\nA and 2 from B) start echoing their respective messages. At\neach iteration, the PhaseLoop checks to see if it should\niterate again.\n\n    Actor A is configured to run its iteration in Phase 0\n    for 2 seconds, so each of its 10 threads will loop and\n    echo \"A.0\" for 2 seconds.\n\n    Actor B is configured to run its iteration in Phase 1\n    only 10 times. So it will echo \"B.0\" 10 times per thread.\n    Almost certainly this will take less than 2 seconds (the\n    amount of time Actor A requires to complete Phase 0). In\n    this case, Actor B is \"done\" with its actions in the\n    current Phase so its 2 threads will wait and do nothing\n    until the next Phase. Compare this with what Actor A does\n    in Phase 1 described next.\n\n== Phase 1 ==\n\nAfter Phase 0 comes Phase 1.\n\n        A                  B\n    Message: A.1      Message: B.1\n    Blocking: None    Repeat: 1e3\n\nNotice how Actor A does not specify either a Repeat nor a\nDuration. This means A is considered a \"background\" Actor\nfor this Phase. Since it doesn't specify Repeat or Duration\nit must specify Blocking: None as a precaution against\nfootgunning. It will run as long as it can while other\nActors are doing their work. This is useful if you want to\nhave some \"background\" operation such as doing a db.ping()\nor a collection.drop() etc.\n\nActor B's 2 threads will echo \"B.1\" 1e3 (1000) times each\nand Actor A's 10 threads will echo \"A.1\" for however long\nit takes B to do this.\n\n== Potential Gotchas ==\n\n1. You can specify both `Repeat` and `Duration` for a Phase.\n   The larger of the two wins. If you specify `Repeat: 1e40`\n   and `Duration: 1ms` it will repeat 1e40 times even if it\n   takes much longer than 1ms.\n\n2. It is undefined behavior if no Actors specify a Duration\n   or Repeat for a Phase. It does not currently cause an\n   infinite-loop, but we don't check for this situation\n   (which is almost certainly a mis-configuration) at setup\n   time.\n",
+    "Description": "\nThis workload illustrates how to think of Phases happening\nsimultaneously and how to configure Repeat and Duration\nwhen you have more than one Actor in a Workload.\n\nStart by quickly looking at the actual workload yaml below\n\nAlthough we write workloads by listing each Actor and then\nlisting each Phase for that Actor, Genny workloads actually\noperate in order of each Phase. We construct each Actor and\nthen start Phase 0. We wait until all Actors say that they\nare done with Phase 0 and then we move on to Phase 1, etc.\nActors using PhaseLoop automatically do this.\n\nIt may help to think of the Actor yamls as being written in\nparallel rather than in sequence\n\n    Actors:\n    - Type: HelloWorld             - Type: HelloWorld\n      Name: A                        Name: B\n      Threads: 10                    Threads: 2\n      Phases:                        Phases:\n\n      - Message: A.0                 - Message: B.0\n        Duration: 2 seconds            Repeat: 10\n\n      - Message: A.1                 - Message: B.1\n        Blocking: None                 Repeat: 1e3\n\nThe HelloWorld Actor echoes its \"Message\" parameter in a\nloop.\n\nThere will be 10 threads running HelloWorld with name \"A\"\nand 2 threads running with name \"B\".\n\n== Phase 0 ==\n\n          A                          B\n      Message: A.0              Message: B.0\n      Duration: 2 seconds       Repeat: 10\n\nThe first phase (Phase 0) starts and all 12 threads (10 from\nA and 2 from B) start echoing their respective messages. At\neach iteration, the PhaseLoop checks to see if it should\niterate again.\n\n    Actor A is configured to run its iteration in Phase 0\n    for 2 seconds, so each of its 10 threads will loop and\n    echo \"A.0\" for 2 seconds.\n\n    Actor B is configured to run its iteration in Phase 1\n    only 10 times. So it will echo \"B.0\" 10 times per thread.\n    Almost certainly this will take less than 2 seconds (the\n    amount of time Actor A requires to complete Phase 0). In\n    this case, Actor B is \"done\" with its actions in the\n    current Phase so its 2 threads will wait and do nothing\n    until the next Phase. Compare this with what Actor A does\n    in Phase 1 described next.\n\n== Phase 1 ==\n\nAfter Phase 0 comes Phase 1.\n\n        A                  B\n    Message: A.1      Message: B.1\n    Blocking: None    Repeat: 1e3\n\nNotice how Actor A does not specify either a Repeat nor a\nDuration. This means A is considered a \"background\" Actor\nfor this Phase. Since it doesn't specify Repeat or Duration\nit must specify Blocking: None as a precaution against\nfootgunning. It will run as long as it can while other\nActors are doing their work. This is useful if you want to\nhave some \"background\" operation such as doing a db.ping()\nor a collection.drop() etc.\n\nActor B's 2 threads will echo \"B.1\" 1e3 (1000) times each\nand Actor A's 10 threads will echo \"A.1\" for however long\nit takes B to do this.\n\n== Potential Gotchas ==\n\n1. You can specify both `Repeat` and `Duration` for a Phase.\n   The larger of the two wins. If you specify `Repeat: 1e40`\n   and `Duration: 1ms` it will repeat 1e40 times even if it\n   takes much longer than 1ms.\n\n2. It is undefined behavior if no Actors specify a Duration\n   or Repeat for a Phase. It does not currently cause an\n   infinite-loop, but we don't check for this situation\n   (which is almost certainly a mis-configuration) at setup\n   time.\n",
     "Actors": [
         {
             "Type": "HelloWorld",
             "Name": "A",
             "Threads": 10,
             "Phases": [
                 {
                     "Message": "A.0",
                     "Duration": "2 seconds"
                 },
diff --git a/src/workloads/execution/MixedMultiDeletesBatched.0.json b/src/workloads/execution/MixedMultiDeletesBatched.0.json
index 0a4e20507..210767947 100644
--- a/src/workloads/execution/MixedMultiDeletesBatched.0.json
+++ b/src/workloads/execution/MixedMultiDeletesBatched.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Storage Execution",
-    "Description": "Deletes a range of documents using the BATCHED_DELETE query exec stage both in isolation and while\nperforming writes on another collection. Tests deletes on documents of size ~1KB then deletes on\ndocuments of ~10MB.\n",
+    "Description": "Deletes a range of documents using the BATCHED_DELETE query exec stage both in isolation and while performing writes on another collection. Tests deletes on documents of size ~1KB then deletes on documents of ~10MB.\n",
     "Keywords": [
         "RunCommand",
         "Loader",
         "LoggingActor",
         "CrudActor",
         "insert",
         "delete",
         "batch",
         "deleteMany",
         "latency"
diff --git a/src/workloads/filesystem/Flushing.0.json b/src/workloads/filesystem/Flushing.0.json
index 027b3014f..b0d404308 100644
--- a/src/workloads/filesystem/Flushing.0.json
+++ b/src/workloads/filesystem/Flushing.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Product Performance",
-    "Description": "Run large insertMany's to stress checkpointing, journal flush and dirty memory with {w:majority}.\nThe workload runs a single CrudActor targeting:\n  - 'test' database, configurable via the Database parameter.\n  - 'Collection0' collection, configurable via the Collection parameter.\n  - 10 threads, configurable via the Threads parameter.\n  - 5000 iterations, configurable via the Iterations parameter.\n  - 1500 documents, configurable via the Documents parameter.\n  - {w:majority} insert options, configurable via the Options parameter.\nThe default documents have a very simple / flat structure as follows:\n  - _id: ObjectId\n  - x: a ~1KB fixed string\n  - id: 'User%07d' incrementing field.\nThere are no additional indexes (just _id).\nThe important ftdc metrics to look at are:\n  - ss opcounters insert\n  - ss average latency writes\n  - system memory dirty\n  - ss wt log log sync time duration\n  - ss wt transaction transaction checkpoint currently running\n",
+    "Description": "Run large insertMany's to stress checkpointing, journal flush and dirty memory with {w:majority}.\nThe workload runs a single CrudActor targeting:\n  - 'test' database, configurable via the Database parameter.\n  - 'Collection0' collection, configurable via the Collection parameter.\n  - 10 threads, configurable via the Threads parameter.\n  - 5000 iterations, configurable via the Iterations parameter.\n  - 1500 documents, configurable via the Documents parameter.\n  - {w:majority} insert options, configurable via the Options parameter.\nThe default documents have a very simple / flat structure as follows:\n  - _id: ObjectId\n  - x: a ~1KB fixed string \n  - id: 'User%07d' incrementing field.\nThere are no additional indexes (just _id).\nThe important ftdc metrics to look at are:\n  - ss opcounters insert\n  - ss average latency writes\n  - system memory dirty\n  - ss wt log log sync time duration\n  - ss wt transaction transaction checkpoint currently running\n",
     "Keywords": [
         "Flushing",
         "IO",
         "bulk insert",
         "stress",
         "checkpoint",
         "journal"
     ],
     "GlobalConstants": [
         "test",
diff --git a/src/workloads/query/DensifyMilliseconds.0.json b/src/workloads/query/DensifyMilliseconds.0.json
index caebaf394..2250dd4b6 100644
--- a/src/workloads/query/DensifyMilliseconds.0.json
+++ b/src/workloads/query/DensifyMilliseconds.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "This workload tests the performance of the $densify stage with millisecond step intervals.\nThe text of each Densify[Unit] workload is identical from from the Actors list to the end of the\nfile, with all workload-specific details defined as anchors under the GlobalDefaults key. Any\nchanges to the actors should make use of these anchors and be copy-pasted between the Densify[Unit]\nworkloads to keep them in sync.\n\nEach densify workload has a metric to cover possible codepaths through the densify stage, namely\ndifferent bounds types and whether or not to generate documents along partitions.\n\nSince $densify is a new aggregation stage, there isn't a known \"good\" runtime for aggregations\nthat include it. Rather than comparing against a baseline initially, workloads with different\noptions that generate similar numbers of documents will be compared to ensure that there isn't an\noption that relies on a specific codepath that slows down one kind of densification more than another.\n\nInitial investigation has shown that holding other options equal, numeric and millisecond step units\nhave similar latencies, and all other step units have larger latencies which are all similar to each other.\nThis is due to the need for date arithmetic to perform extra processing to add an interval of a\ncertain size to a date. Since dates are stored as milliseconds since the Unix epoch under the hood,\nthe millisecond step unit can skip most of this processing and just performs the same arithmetic\nas the numeric step unit.\n\nThe working assumption is that latency will scale linearly with the number of documents generated\nby the $densify stage: If N is the number of documents in the collection, and M is the number of\ndocuments generated by $densify, then the theoretical runtime should be O(N+M).\n\nThe factors that affect the number of documents generated include, but are not limited to:\n\n- The range that is being densified. This is pretty self-evident, the larger the range that $densify\n  is working over, the more documents will be generated. This includes large ranges specified\n  explicitly in the bounds option, and implicitly large ranges when densifying over collections\n  where the minimum and maximum documents are farther from each other in the\n  `bounds: \"full\"|\"partition\"` cases.\n\n- The step size. Smaller steps produce more documents when the range is held constant.\n\n- The cardinality of the fields specified in the `partitionByFields`, if set. One document\n  will be generated with every value within the range for each partition, so the higher the\n  cardinality, the more documents that will be generated. Specifying a `partitionByFields` will\n  never decrease the number of documents generated, holding all other options equal.\n",
+    "Description": "This workload tests the performance of the $densify stage with millisecond step intervals.\nThe text of each Densify[Unit] workload is identical from from the Actors list to the end of the\nfile, with all workload-specific details defined as anchors under the GlobalDefaults key. Any\nchanges to the actors should make use of these anchors and be copy-pasted between the Densify[Unit]\nworkloads to keep them in sync.\n\nEach densify workload has a metric to cover possible codepaths through the densify stage, namely\ndifferent bounds types and whether or not to generate documents along partitions.\n\n\nSince $densify is a new aggregation stage, there isn't a known \"good\" runtime for aggregations\nthat include it. Rather than comparing against a baseline initially, workloads with different\noptions that generate similar numbers of documents will be compared to ensure that there isn't an\noption that relies on a specific codepath that slows down one kind of densification more than another.\n\nInitial investigation has shown that holding other options equal, numeric and millisecond step units\nhave similar latencies, and all other step units have larger latencies which are all similar to each other.\nThis is due to the need for date arithmetic to perform extra processing to add an interval of a\ncertain size to a date. Since dates are stored as milliseconds since the Unix epoch under the hood,\nthe millisecond step unit can skip most of this processing and just performs the same arithmetic\nas the numeric step unit.\n\nThe working assumption is that latency will scale linearly with the number of documents generated\nby the $densify stage: If N is the number of documents in the collection, and M is the number of\ndocuments generated by $densify, then the theoretical runtime should be O(N+M).\n\nThe factors that affect the number of documents generated include, but are not limited to:\n\n- The range that is being densified. This is pretty self-evident, the larger the range that $densify\n  is working over, the more documents will be generated. This includes large ranges specified\n  explicitly in the bounds option, and implicitly large ranges when densifying over collections\n  where the minimum and maximum documents are farther from each other in the\n  `bounds: \"full\"|\"partition\"` cases.\n\n- The step size. Smaller steps produce more documents when the range is held constant.\n\n- The cardinality of the fields specified in the `partitionByFields`, if set. One document\n  will be generated with every value within the range for each partition, so the higher the\n  cardinality, the more documents that will be generated. Specifying a `partitionByFields` will\n  never decrease the number of documents generated, holding all other options equal.\n",
     "GlobalDefaults": {
         "dbname": "test",
         "MaxPhases": 20,
         "limit": 30000,
         "fieldName": "milliseconds",
         "unitName": "millisecond",
         "index": {
             "keys": {
                 "milliseconds": 1
             },
diff --git a/src/workloads/query/GroupStagesOnComputedFields.0.json b/src/workloads/query/GroupStagesOnComputedFields.0.json
index 70e798081..c0a260e20 100644
--- a/src/workloads/query/GroupStagesOnComputedFields.0.json
+++ b/src/workloads/query/GroupStagesOnComputedFields.0.json
@@ -215,21 +215,21 @@
                         "$sort": {
                             "time": 1
                         }
                     },
                     {
                         "$addFields": {
                             "t": {
                                 "$dateTrunc": {
                                     "date": "$time",
                                     "unit": "second",
-                                    "binSize": ""
+                                    "binSize": null
                                 }
                             }
                         }
                     },
                     {
                         "$group": {
                             "_id": "$t",
                             "sample": {
                                 "$first": "$$ROOT"
                             }
@@ -499,21 +499,21 @@
                             "$sort": {
                                 "time": 1
                             }
                         },
                         {
                             "$addFields": {
                                 "t": {
                                     "$dateTrunc": {
                                         "date": "$time",
                                         "unit": "second",
-                                        "binSize": ""
+                                        "binSize": null
                                     }
                                 }
                             }
                         },
                         {
                             "$group": {
                                 "_id": "$t",
                                 "sample": {
                                     "$first": "$$ROOT"
                                 }
diff --git a/src/workloads/query/TimeseriesCount.0.json b/src/workloads/query/TimeseriesCount.0.json
index 081272268..5f165ba9a 100644
--- a/src/workloads/query/TimeseriesCount.0.json
+++ b/src/workloads/query/TimeseriesCount.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "The queries in this workload exercise group stage that uses the $count accumulator and the $count\naggregation stage. On FCV greater than or equal to 7.1 $group using $count is optimized to remove\nthe $unpack stage.\n",
+    "Description": "The queries in this workload exercise group stage that uses the $count accumulator and the $count \naggregation stage. On FCV greater than or equal to 7.1 $group using $count is optimized to remove \nthe $unpack stage.\n",
     "Keywords": [
         "timeseries",
         "aggregate",
         "group"
     ],
     "GlobalDefaults": {
         "Database": "test",
         "Collection": "Collection0",
         "MaxPhases": 10,
         "repeat": 10
diff --git a/src/workloads/query/TimeseriesFixedBucketing.0.json b/src/workloads/query/TimeseriesFixedBucketing.0.json
index 962f4e286..e3cdb2e17 100644
--- a/src/workloads/query/TimeseriesFixedBucketing.0.json
+++ b/src/workloads/query/TimeseriesFixedBucketing.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Query Integration",
-    "Description": "This test exercises the behavior of the time series optimization when the collection has\nfixed buckets. This workload uses tsbs data that is imported in the dsi configuration. See\n'configurations/test_control/test_control.fixed_bucketing.yml' for all the details.\nThe data is set up with the fixed bucketing parameters set to 3600 and the timeField is \"time\"\nand the metaField is \"tags\". There are 20736000 documents in the collection.\n",
+    "Description": "This test exercises the behavior of the time series optimization when the collection has \nfixed buckets. This workload uses tsbs data that is imported in the dsi configuration. See \n'configurations/test_control/test_control.fixed_bucketing.yml' for all the details.\nThe data is set up with the fixed bucketing parameters set to 3600 and the timeField is \"time\"\nand the metaField is \"tags\". There are 20736000 documents in the collection.\n",
     "Keywords": [
         "timeseries",
         "aggregate",
         "group"
     ],
     "GlobalDefaults": {
         "Database": "test",
         "Collection": "coll",
         "maxPhases": 10,
         "Repeat": 5
diff --git a/src/workloads/query/TimeseriesTsbsExpressionQuery.0.json b/src/workloads/query/TimeseriesTsbsExpressionQuery.0.json
index 6ba811920..ebe1e8363 100644
--- a/src/workloads/query/TimeseriesTsbsExpressionQuery.0.json
+++ b/src/workloads/query/TimeseriesTsbsExpressionQuery.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "This workload runs different math expressions on time-series optimizations with the tsbs dataset.\nDuring development, This workload runs queries from the tsbs_query benchmark in Genny. It allows us\nto cross-check the results of the TSBS benchmarks, collect additional measurements, profile individual\nqueries, and extend some of the queries as needed. The data is preloaded by the dsi configuration. See\n'configurations/test_control/test_control.tsbs_query_genny.yml' for all  the details. There are 20736000\ndocuments in the collection. There are some differences between the queries in this workload and tsbs,\nbut we do not expect these differences to significantly affect runtime:\n1. TSBS will generate a random start time between the start/end time of the workload for every query type.\nWe hard code our time values, since genny does not currently support date arithmetic, nor storing a fixed\nvalue from randomized generation.\n2. TSBS randomizes which measurement fields to group by, we do not randomize these fields. Since all the fields\nare random integers between 0-100 we don't expect different fields to have significant runtime differences.\n",
+    "Description": "This workload runs different math expressions on time-series optimizations with the tsbs dataset. \nDuring development, This workload runs queries from the tsbs_query benchmark in Genny. It allows us \nto cross-check the results of the TSBS benchmarks, collect additional measurements, profile individual \nqueries, and extend some of the queries as needed. The data is preloaded by the dsi configuration. See\n'configurations/test_control/test_control.tsbs_query_genny.yml' for all  the details. There are 20736000\ndocuments in the collection. There are some differences between the queries in this workload and tsbs,\nbut we do not expect these differences to significantly affect runtime:\n1. TSBS will generate a random start time between the start/end time of the workload for every query type.\nWe hard code our time values, since genny does not currently support date arithmetic, nor storing a fixed\nvalue from randomized generation.\n2. TSBS randomizes which measurement fields to group by, we do not randomize these fields. Since all the fields\nare random integers between 0-100 we don't expect different fields to have significant runtime differences.\n",
     "Keywords": [
         "timeseries",
         "aggregate",
         "group",
         "sort"
     ],
     "Clients": {
         "Default": {
             "QueryOptions": {
                 "maxPoolSize": 4000
diff --git a/src/workloads/query/multiplanner/CompoundIndexes.0.json b/src/workloads/query/multiplanner/CompoundIndexes.0.json
index c4e25f365..c9d3a5764 100644
--- a/src/workloads/query/multiplanner/CompoundIndexes.0.json
+++ b/src/workloads/query/multiplanner/CompoundIndexes.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "The goal of this test is to exercise multiplanning. We create as many compound indexes as possible, each\ncomprising of tenantId and another field. We then run a query with an equality predicate on tenantId and\nother predicates a small number of fields, so we get all the plans that use relevant indexes.\n\nWe expect classic to have better latency and throughput than SBE on this workload,\nand we expect the combination of classic planner + SBE execution (PM-3591) to perform about\nas well as classic.\n",
+    "Description": "The goal of this test is to exercise multiplanning. We create as many compound indexes as possible, each\ncomprising of tenantId and another field. We then run a query with an equality predicate on tenantId and \nother predicates a small number of fields, so we get all the plans that use relevant indexes.\n\nWe expect classic to have better latency and throughput than SBE on this workload,\nand we expect the combination of classic planner + SBE execution (PM-3591) to perform about\nas well as classic.\n",
     "GlobalDefaults": {
         "dbname": "test",
         "coll": "Collection0",
         "docCount": "1e5",
         "resultCount": 101,
         "selectivity": {
             "^NumExpr": {
                 "withExpression": "resultCount / docCount",
                 "andValues": {
                     "resultCount": 101,
diff --git a/src/workloads/query/multiplanner/MultikeyIndexes.0.json b/src/workloads/query/multiplanner/MultikeyIndexes.0.json
index 0bd8cb7bd..de8c4a707 100644
--- a/src/workloads/query/multiplanner/MultikeyIndexes.0.json
+++ b/src/workloads/query/multiplanner/MultikeyIndexes.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "The goal of this test is to exercise multiplanning with multikey indexes. We create many indexes and\nrun a query that makes all of them eligible, so we get as many competing plans as possible. Because\nan IXSCAN of a multikey index has to deduplicate RIDs, a lot of space will be used. The classic\nmulti-planner will behave more optimally than the SBE multiplanner because it will cut off execution\nwhen the one good plan reaches the end.\n\nWe expect classic to have better latency and throughput than SBE on this workload,\nand we expect the combination of classic planner + SBE execution (PM-3591) to perform about\nas well as classic.\n",
+    "Description": "The goal of this test is to exercise multiplanning with multikey indexes. We create many indexes and\nrun a query that makes all of them eligible, so we get as many competing plans as possible. Because \nan IXSCAN of a multikey index has to deduplicate RIDs, a lot of space will be used. The classic \nmulti-planner will behave more optimally than the SBE multiplanner because it will cut off execution\nwhen the one good plan reaches the end.\n\nWe expect classic to have better latency and throughput than SBE on this workload,\nand we expect the combination of classic planner + SBE execution (PM-3591) to perform about\nas well as classic.\n",
     "GlobalDefaults": {
         "dbname": "test",
         "coll": "Collection0",
         "docCount": "1e5",
         "resultCount": 101,
         "selectivity": {
             "^NumExpr": {
                 "withExpression": "resultCount / docCount",
                 "andValues": {
                     "resultCount": 101,
diff --git a/src/workloads/scale/CursorStormMongos.0.json b/src/workloads/scale/CursorStormMongos.0.json
index 514535669..e0385ff20 100644
--- a/src/workloads/scale/CursorStormMongos.0.json
+++ b/src/workloads/scale/CursorStormMongos.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "@mongodb/query",
-    "Description": "Used to test performance of the router under memory pressure caused by accumulating\nmany heavy cursors. The workload is expected to fail due to host(s) being unreachable as a\nresult of mongos running out of memory.\n\nTo achieve this, many threads are spawned to run an unfiltered find on a collection.\nThe number and size of documents in that collection are tuned such, that the mongos is able\nto exhaust cursors on shards when pre-filling its buffers [<16MB per shard]. As a result,\nmemory pressure on the shards remains low, while it's kept large on the mongos.\n",
+    "Description": "Used to test performance of the router under memory pressure caused by accumulating\nmany heavy cursors. The workload is expected to fail due to host(s) being unreachable as a\nresult of mongos running out of memory.\n\nTo achieve this, many threads are spawned to run an unfiltered find on a collection.\nThe number and size of documents in that collection are tuned such, that the mongos is able\nto exhaust cursors on shards when pre-filling its buffers [<16MB per shard]. As a result, \nmemory pressure on the shards remains low, while it's kept large on the mongos.\n",
     "Keywords": [
         "scale",
         "memory stress",
         "cursor storm",
         "mongos",
         "fail",
         "oom",
         "out of memory"
     ],
     "GlobalDefaults": {
diff --git a/src/workloads/selftests/GennyOverhead.0.json b/src/workloads/selftests/GennyOverhead.0.json
index 507ee4e0b..b74ae15ac 100644
--- a/src/workloads/selftests/GennyOverhead.0.json
+++ b/src/workloads/selftests/GennyOverhead.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Performance Analysis",
-    "Description": "This workload measures the overhead of Genny itself. There are 2 different\nconfigurations that run consecutively for 5 phases each. The first\nconfiguration runs with 100 threads and the second configuration runs\nwith 1 thread.\n\nThe 5 phases of each configuration are as follows:\n\n1. Loop for 10 seconds using the \"Duration\" keyword\n2. Loop for 3M iterations using the \"Repeat\" keyword. Each iteration\n   takes around 5\u00b5s when run in a single thread.\n3. Loop for 10k iterations with a 1 per 1ms rate limit\n4. Loop for 10k iterations while sleeping for 1ms before each iteration\n5. Loop for 10k iterations while sleeping for 1ms after each iteration\n\nEach iteration is craftered to take around 10s. More configurations\ncan be added as needed, but unfortunately the \"5 phases\" building block\ncan't be reused en masse because YAML doesn't support merging lists.\n\nThe primary metric recorded in this workload is the average duration\nof each iteration.\n\nThis workload is intended to stress the Genny client itself, so should\nbe run with the smallest MongoDB setup.\n",
+    "Description": "\nThis workload measures the overhead of Genny itself. There are 2 different\nconfigurations that run consecutively for 5 phases each. The first\nconfiguration runs with 100 threads and the second configuration runs\nwith 1 thread.\n\nThe 5 phases of each configuration are as follows:\n\n1. Loop for 10 seconds using the \"Duration\" keyword\n2. Loop for 3M iterations using the \"Repeat\" keyword. Each iteration\n   takes around 5\u00b5s when run in a single thread.\n3. Loop for 10k iterations with a 1 per 1ms rate limit\n4. Loop for 10k iterations while sleeping for 1ms before each iteration\n5. Loop for 10k iterations while sleeping for 1ms after each iteration\n\nEach iteration is craftered to take around 10s. More configurations\ncan be added as needed, but unfortunately the \"5 phases\" building block\ncan't be reused en masse because YAML doesn't support merging lists.\n\nThe primary metric recorded in this workload is the average duration\nof each iteration.\n\nThis workload is intended to stress the Genny client itself, so should\nbe run with the smallest MongoDB setup.\n",
     "Actors": [
         {
             "Name": "GennyOverhead100T",
             "Type": "NopMetrics",
             "Threads": 100,
             "Phases": [
                 {
                     "MetricsName": "LoopOnlyFixedTime",
                     "Duration": "10 milliseconds"
                 },
diff --git a/src/workloads/sharding/ReshardCollectionWithIndexes.0.json b/src/workloads/sharding/ReshardCollectionWithIndexes.0.json
index 4585ab50b..c32a67026 100644
--- a/src/workloads/sharding/ReshardCollectionWithIndexes.0.json
+++ b/src/workloads/sharding/ReshardCollectionWithIndexes.0.json
@@ -1,14 +1,14 @@
 {
     "SchemaVersion": "2018-07-01",
     "Owner": "Cluster Scalability",
-    "Description": "This test measures the time for a sharded cluster to reshard a collection from one shard to two\nshards then to three shards. It was added August 2023 as part of PM-2322, to demonstrate planned\nresharding performance improvements. Note that the goal of this test is to show the performance\ngain on this setup rather than the performance difference on different kinds of data type.\n\nThe test expects the target cluster is created using ebs snapshot with 1 billion 1KB records and\nhas 3 shards, named shard-00, shard-01, shard-02. The collection has 10 indexes including\n_id index, single-key index and compound index. The whole dataset is on 1 shard at the beginning.\nReshardCollection should use same-key resharding and use shardDistribution parameter to reshard\ninto 2 shards then 3 shards. There will be random CRUD operations during resharding but should\nrun at a very low rate.\n\nThe workload consists of 3 phases:\n  1. Turning off balancer and make the sharded collection exists on only 1 shard.\n  2. Running read and write operations on the collection while it is being resharded to 2 shards.\n  3. Running read and write operations on the collection while it is being resharded to 3 shards.\n\nAll documents are generated through genny' data loader where the integer and length of short\nstring fields are randomly generated. The fields are designed like to form different combinations\nof indexes. The goal of having 10 indexes is to test resharding performance with indexes and the\nnumber 10 comes from design, which is arbitrary from testing perspective.\n\nThe inserted documents have the following form:\n\n    {\n      _id: integer(default _id index),\n      shardKey: integer(for the shard key),\n      counter: integer(used for counting updates),\n      num1: integer(random generated between [1, 1000000000]),\n      num2: integer(random generated between [1, 1000000000]),\n      str1: string(length <= 20),\n      str2: string(length <= 20),\n      padding: string(random string of bytes ...)\n    }\n\nThe indexes are:\n    [\n      {_id: 1},\n      {shardKey: 1},\n      {counter: 1},\n      {num1: 1},\n      {num2: 1},\n      {str1: 1},\n      {str2: 1},\n      {shardKey: 1, counter:1},\n      {str1: 1, num1: 1},\n      {num2:1, str2: 1}\n    ]\n",
+    "Description": "This test measures the time for a sharded cluster to reshard a collection from one shard to two \nshards then to three shards. It was added August 2023 as part of PM-2322, to demonstrate planned \nresharding performance improvements. Note that the goal of this test is to show the performance \ngain on this setup rather than the performance difference on different kinds of data type.\n\nThe test expects the target cluster is created using ebs snapshot with 1 billion 1KB records and \nhas 3 shards, named shard-00, shard-01, shard-02. The collection has 10 indexes including \n_id index, single-key index and compound index. The whole dataset is on 1 shard at the beginning.\nReshardCollection should use same-key resharding and use shardDistribution parameter to reshard \ninto 2 shards then 3 shards. There will be random CRUD operations during resharding but should \nrun at a very low rate.\n\nThe workload consists of 3 phases:\n  1. Turning off balancer and make the sharded collection exists on only 1 shard.\n  2. Running read and write operations on the collection while it is being resharded to 2 shards.\n  3. Running read and write operations on the collection while it is being resharded to 3 shards.\n\nAll documents are generated through genny' data loader where the integer and length of short \nstring fields are randomly generated. The fields are designed like to form different combinations \nof indexes. The goal of having 10 indexes is to test resharding performance with indexes and the \nnumber 10 comes from design, which is arbitrary from testing perspective.\n  \nThe inserted documents have the following form:\n\n    {\n      _id: integer(default _id index),\n      shardKey: integer(for the shard key),\n      counter: integer(used for counting updates),\n      num1: integer(random generated between [1, 1000000000]),\n      num2: integer(random generated between [1, 1000000000]),\n      str1: string(length <= 20),\n      str2: string(length <= 20),\n      padding: string(random string of bytes ...)\n    }\n\nThe indexes are:\n    [\n      {_id: 1},\n      {shardKey: 1},\n      {counter: 1},\n      {num1: 1},\n      {num2: 1},\n      {str1: 1},\n      {str2: 1},\n      {shardKey: 1, counter:1},\n      {str1: 1, num1: 1},\n      {num2:1, str2: 1}\n    ]\n",
     "Keywords": [
         "resharding",
         "indexes",
         "replication",
         "collection copy"
     ],
     "GlobalDefaults": {
         "Nop": {
             "Nop": true
         },

```

I have `src/workloads/contrib/qe_test_gen/maps_medical.yml` which had significant changes in the IDs, but I am assuming these are not relevant. Here's an example:

```
diff --git a/src/workloads/contrib/qe_test_gen/maps_medical.0.json b/src/workloads/contrib/qe_test_gen/maps_medical.0.json
index ffbb9ecfa..2036f81b8 100644
--- a/src/workloads/contrib/qe_test_gen/maps_medical.0.json
+++ b/src/workloads/contrib/qe_test_gen/maps_medical.0.json
@@ -5229,21 +5229,21 @@
                 "4473593503484549": 10,
                 "180087465900982": 10,
                 "30574338471269": 10,
                 "6011287264803104": 10,
                 "30556056625135": 10,
                 "4652337490417184": 10,
                 "6011410716819052": 10,
                 "3576392257439017": 10,
                 "4700259598102249": 10,
                 "3551414001466010": 10,
-                "60407444235": 10,
+                "6511544477": 10,
                 "4430543919718856065": 10,
                 "3555295550272883": 10,
                 "2424261691829743": 10,
                 "344730861119748": 10,
                 "30125659877707": 10,
                 "4209103143031337": 10,
                 "3598791918977469": 10,
                 "3592839727722349": 10,
                 "4510574044149520759": 10,
                 "213168367018452": 10,
```

I have four concerns here:
- `src/cast_core/test/CrudActorYamlTests.yml` has a duplicate `comment` anchor that should be getting linted away.
- `src/workloads/query/GroupStagesOnComputedFields.yaml` seems to have some invalid JSON in it, `{$addFields: {"t": {$dateTrunc: {date: "$time",unit: "second",binSize}}` doesn't look correct to me. This should be getting picked up by linting right?
- **This is a blocker** please add information for how to lint YAML to `.github/pull_request_template.md`, adding YAML to this repo is the most common thing people do, I think discovering our linter settings one-by-one is fairly unpleasant.
- The description newline removals were intended, right?

src/lamplib/src/genny/tasks/yaml_linter.py
lgfang: done

evergreen.yml
lgfang: I believe it is standard. My test code written using PyYAML had no with it. That said, I'm going to run a full patch as you requested in an earlier comment.

lgfang:
Thank you very much for reviewing this PR. Converting the files into JSON ones and do the comparison is genius. It appears that in the \`git diff\` output, the "original" copy refers to my version, while the "new" version refers to the master branch.

> I have src/workloads/contrib/qe\_test\_gen/maps\_medical.yml which had significant changes in the IDs, but I am assuming these are not relevant. &#x2026;
> 
> ```text
> -                "60407444235": 10,
> +                "6511544477": 10,
> ```

Such changes are intended. As we removed the leading zero(s), such values are no longer treated as octal one. Take the above value as an example, in the master branch, since the string \`060407444235\` is an octal string, its decimal value 6511544477 was dumped to the JSON file. After I removed the leading zero, the value remained unchanged.

```text
$ echo $(("060407444235"))
6511544477
```

> src/cast\_core/test/CrudActorYamlTests.yml has a duplicate comment anchor that should be getting linted away.

The directory `src/cast_core` is not checked by this PR, which covers the following (i.e. workloads, resmoke and evergreen):

-   src/workloads
-   src/phases
-   src/resmokeconfig
-   evergreen.yml

> src/workloads/query/GroupStagesOnComputedFields.yaml seems to have some invalid JSON in it, {$addFields: {"t": {$dateTrunc: {date: "$time",unit: "second",binSize}} doesn't look correct to me.

The code snippet you've provided is from the master branch. In my pull request, it's been "fixed" by yamlfmt to an object equivalent to the following:

```yaml
{$addFields: { "t": {$dateTrunc: { date: "$time", unit: "second", binSize: "" }}}},
```

> This is a blocker please add information for how to lint YAML to .github/pull\_request\_template.md, adding YAML to this repo is the most common thing people do, I think discovering our linter settings one-by-one is fairly unpleasant.

Done.

> The description newline removals were intended, right?

If you swap the "old" and "new" versions in the git diff command, you'll observe that contrary to removing newlines, I have in fact added some. This was a deliberate action to reduce the line length to shorter than 160 characters.

jimoleary (approved):
LGTM

I think I've gone off the  full patch as it will be large / expensive.  Also the longer this review goes on ... the more merges you will accrue.

Given that these changes are formatting ... we can hope that the changes are non-functional and fixing any issues afterwards.

It's probably worth asking the Build Barons to keep a close eye for oddness after this commit.

thessem (approved):
Happy that my concerns were addressed. Thanks

lgfang:
Re the commit [Fix and format some new yaml files](<https://github.com/mongodb/genny/pull/1205/commits/1547edbdcb2ab5b722a1604566e6775351db9b31>)

These new YAML files triggered yamllint errors. But neither `prettier` nor `yamlfmt` was able to format them due to a space missing after the colon in `{^ChooseFromDataset:{"path"`. I ran the following commands for each of them to fix it:

```bash
# confirm all errors are exactly the same
grep -o '[^ ]*:{[^ ]*' $each
# Insert a white space between : and {
sed -i -e 's/{^ChooseFromDataset:{"path"/{^ChooseFromDataset: {"path"/' $each
# format using npx
npx prettier $each --write
```

PR #1210, opened 2024-05-06
DEVPROD-7139 Create the default time-series index for timeseries sort compound

galon1:
**Jira Ticket:** [DEVPROD-7139](https://jira.mongodb.org/browse/DEVPROD-7139)

### Whats Changed
We fixed how the hint was passed into this workload to solve a regression BF. However, this workload runs on 6.0, and the index used in the hint is created by default in 7.0.0. So the aggregation now fails, since the index hint can't be applied to the query plan.

### DIFF
--- a/src/workloads/query/TimeSeriesSortCompound.yml
+++ b/src/workloads/query/TimeSeriesSortCompound.yml
@@ -31,6 +31,8 @@ Actors:
   - *Nop
   - *Nop
   - *Nop
+  - *Nop
+  - *Nop
   - Repeat: 1
     Database: test
     Operations:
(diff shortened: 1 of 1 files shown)


galon1 (commented):
Patch on [master](https://spruce.mongodb.com/version/6638ec28cca3830007708b09/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) and patch on [6.0](https://spruce.mongodb.com/version/6638ec719b684400079daa81/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). The task passes on 6.0 now, and shows no performance gains on master (which is expected, since the workload shouldn't change on master).

serhiitea:
How come hint application didn't fail on 6.0 when we were using `RunCommand` actor?

serhiitea (approved):
lgtm

galon1:
> DEVPROD-7139

When we were using `RunCommand` the workload was consistently failing from May 15 2023 (when the hint was added) to January 2 (when the commit to use `CrudActor` went in). The failure was happening on 6.0 only, but it is concerning it didn't cause a BF. You can see this on the sys-perf 6.0 [waterfall](https://spruce.mongodb.com/commits/sys-perf-6.0?skipOrderNumber=1884&taskNames=time_series_sort_compound&view=FAILED).

hilldani (approved):

PR #1218, opened 2024-05-10
PERF-5441 Adds the `time_series_lastpoint_high_cardinality` workload.

yun-soo:
**Jira Ticket:** [PERF-5441](https://jira.mongodb.org/browse/PERF-5441)

### Whats Changed

Adds the `time_series_lastpoint_high_cardinality` workload.

### Patch Testing Results

https://spruce.mongodb.com/version/663f01f365c3b80007df9990/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


src/phases/query/TimeSeriesLastpoint.yml
```diff
@@ -6,79 +6,57 @@ Description: |
...
-      OperationMetricsName: CreateIndex
-      OperationCommand:
-        createIndexes:
-          { ^Parameter: { Name: "Collection", Default: "Collection0" } }
-        indexes:
-          - {
-            key: { ^Parameter: { Name: "IndexPattern", Default: {} } },
-            name: { ^Parameter: { Name: "IndexName", Default: "" } },
-          }
+  Operation:
```
yun-soo, line 9: These changes are from #1219 and after it's merged, this will be gone.

src/workloads/query/TimeSeriesLastpoint.yml
```diff
@@ -13,6 +13,7 @@ Description: |
 db: &db test
 coll: &coll Collection0
 phasePath: &phasePath ../../phases/query/TimeSeriesLastpoint.yml
+MaxPhases: &MaxPhases 26
```
yun-soo, line 16: These changes are from #1219 and after it's merged, this will be gone.

alicedoherty:
Thanks @yun-soo! Would you be able to fill out the [New Workload Request](https://docs.google.com/forms/d/e/1FAIpQLSf1oh23Ddo1khjLZMqZ9DHbRdObnpeH20PSXTBuXVg-7mxxTA/viewform) form? Also just want to confirm that for this PR only the TimeSeriesLastpointHighCardinality workload file needs to be reviewed, and the changes to the TimeSeriesLastpoint workload/phase file will be dealt with in https://github.com/mongodb/genny/pull/1219?

Do you have an idea of how noisy this workload is? We'd generally recommend running x5 patches to get a sense of this. You can use the performance analyzer to automate this (see the third method [here](https://github.com/10gen/performance-tooling-docs/blob/main/data_analysis/performance_analyzer_docs.md#interface)).

yun-soo:
Thanks a lot for the review!

> Would you be able to fill out the [New Workload Request](https://docs.google.com/forms/d/e/1FAIpQLSf1oh23Ddo1khjLZMqZ9DHbRdObnpeH20PSXTBuXVg-7mxxTA/viewform) form?

Filed the request. I noticed that build variant list is not consistent with the up-to-date one.

> Also just want to confirm that for this PR only the TimeSeriesLastpointHighCardinality workload file needs to be reviewed, and the changes to the TimeSeriesLastpoint workload/phase file will be dealt with in #1219?

Yes, only TimeSeriesLastpointHighCardinality needs to be reviewed and they will in #1219

> Do you have an idea of how noisy this workload is? We'd generally recommend running x5 patches to get a sense of this. You can use the performance analyzer to automate this (see the third method [here](https://github.com/10gen/performance-tooling-docs/blob/main/data_analysis/performance_analyzer_docs.md#interface)).

Just scheduled one [here](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=4bb7a283-995e-4414-89f7-d909cbdc6782). Will update this as soon as I will get results

src/workloads/query/TimeSeriesLastpointHighCardinality.yml
```diff
@@ -0,0 +1,337 @@
+SchemaVersion: 2018-07-01
+Owner: "@mongodb/query-integration"
+Description: |
```
alicedoherty, line 3: In the form you mention that latencies for MetaAscTimeAsc.RunLastPointQuery*, MetaAscTimeDesc.RunLastPointQuery*, MetaDescTimeDesc.RunLastPointQuery*, MetaDescTimeAsc.RunLastPointQuery* are the key metrics we care about here. Can you add a line in the description saying this.

(Also in the form you call out AverageLatency. I'm pretty sure for Genny workloads we only really track Latency50thPercentile and Latency95thPercentile)

src/workloads/query/TimeSeriesLastpointHighCardinality.yml
```diff
@@ -0,0 +1,337 @@
...
+        Active: [3, 5, 8, 10, 13, 15, 18, 20]
+        MaxPhases: *MaxPhases
+
+AutoRun:
+  - When:
+      branch_name:
+        $gte: v7.0
+      mongodb_setup:
+        $eq:
+          - standalone
```
alicedoherty, line 332: We've gotten rid of our regular standalone variants. Having this line will just mean it will only be running on the ARM and Intel 3-Node ReplSet 2023-11 variants.

alicedoherty (commented):
Do you have any idea why the Latency95thPercentile for MetaDescTimeDesc.RunLastPointQueryWithSmallPredicate is so noisy? Since this seems like one of the key metrics you care about this would make triaging changepoints pretty hard for this measurement.

Also as an FYI in the [perf analyzer](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=4bb7a283-995e-4414-89f7-d909cbdc6782&selected_tab=data-table&test_filter=MetaAscTimeDesc%5C.RunLastPointQuery%5BA-Za-z%5D%2A&test_filter=MetaAscTimeAsc%5C.RunLastPointQuery%5BA-Za-z%5D%2A&test_filter=MetaDescTimeDesc%5C.RunLastPointQuery%5BA-Za-z%5D%2A&test_filter=MetaDescTimeAsc%5C.RunLastPointQuery%5BA-Za-z%5D%2A&percent_filter=0%7C%7C100&z_filter=0%7C%7C10) if you select to show the "CoV (%)" it will show the calculated coefficient of variation (the main measure of noise our team uses) for each measurement in your patches to save you some time when doing a noise analysis.

src/workloads/query/TimeSeriesLastpointHighCardinality.yml
yun-soo: Apologies but I don't understand what you're suggesting here. Could you clarify on it so that I know what should be changed? I could not exactly correlate these names with the updated build variants.

These lines are exactly same as in the companion workload `TimeSeriesLastpoint.yml`. And I was able to see that this workload is enabled on perf-3-node-replSet-intel.intel.aws.2023-11 and perf-3-node-replSet.arm.aws.2023-11 and perf-3-node-replSet-all-feature-flags.arm.2023-11. I think these would be enough.

alicedoherty: Sorry the way I worded that was confusing 😅 We've removed our regular standalone variants ([SERVER-89499](https://jira.mongodb.org/browse/SERVER-89499)). Having `$eq: standalone, standalone-XXX` won't cause any issues, it just means it won't match an existing variant, and therefore won't run on any standalone variants. I just wanted to call this out so you weren't surprised that the test isn't running on standalone variants despite having this line in your code.

For clarity it's probably best to remove the `standalone` lines here.

(JFYI WRT the correlation between these lines and our variants - our variants are defined in [perf.yml](https://github.com/mongodb/mongo/blob/master/etc/evergreen_yml_components/variants/perf/perf.yml) and [perf_branching.yml](https://github.com/mongodb/mongo/blob/master/etc/evergreen_yml_components/variants/perf/perf_branching.yml). Each variant has a `mongodb_setup: XXX` line. In your Genny file if you write `AutoRun: When: mongodb_setup: $eq: replica` this means it will run on any variant that has `mongodb_setup: replica`)

alicedoherty (commented):
> Do you happen to know what could cause such noisiness in other workloads?

Looking at the existing [time_series_lastpoint](https://spruce.mongodb.com/task/sys_perf_perf_3_node_replSet.arm.aws.2023_11_time_series_lastpoint_3f0a2b4d625c37d142dc46bd47e2eddd80463023_24_05_20_22_50_07/trend-charts?execution=0&sortBy=STATUS&sortDir=ASC) test it seems much of the underlying test design/queries are noisy. If you're getting useful signal from that existing test I think it would be okay to merge this test with similar levels, however, I definitely think it's worth looking into the noise of Latency95thPercentile for MetaDescTimeDesc.RunLastPointQueryWithSmallPredicate as a CV of 73% (!) and 12% seems like an indication that something is not testing what you'd expect. Since the really bad noise is limited to this specific phase and (potentially) ARM it might be a good idea to do some profiling, t2 analysis, etc. on that specific part of the test. There are some general resources documented [here](https://github.com/10gen/performance-tooling-docs/blob/main/workloads/workload_noise/identify_sources_noise.md), but feel free to ask in #ask-devprod-performance if you need more guidance.

> What's the threshold which the team decides a metric is too noisy on?

We've struggled to define a specific threshold. Ideally we would like it <1% but at the moment with our current testing infrastructure this seems a little unrealistic. Personally, for tests like this I'd be concerned about metrics with CV >5%.

If this is a time-sensitive feature you want to test - it could also be an option to tell the build barons to ignore the really noisy metric (since it's not the whole task) and open a follow-on ticket to investigate this further.

src/workloads/query/TimeSeriesLastpointHighCardinality.yml
yun-soo: Ah~ makes sense now!! Thanks for detailed explanation. Really appreciate it! Removed `standalone(-*)` for clarity just like you suggested.

yun-soo (commented):
> I definitely think it's worth looking into the noise of Latency95thPercentile for MetaDescTimeDesc.RunLastPointQueryWithSmallPredicate as a CV of 73% (!) and 12% seems like an indication that something is not testing what you'd expect. Since the really bad noise is limited to this specific phase and (potentially) ARM it might be a good idea to do some profiling, t2 analysis, etc. on that specific part of the test (looking at one of the really good vs bad runs). There are some general resources documented [here](https://github.com/10gen/performance-tooling-docs/blob/main/workloads/workload_noise/identify_sources_noise.md), but feel free to ask in #ask-devprod-performance if you need more guidance.

Thanks for detailed guidance and really appreciate it! **I'll definitely look into those two metrics.** I wanted to measure improved performance for the higher cardinality workload thanks to a new optimization because `TimeSeriesLastpoint.yml` had a bug (so a noise) and also didn't give my team a good signal about the new optimization but don't want to get too much noise from this new workload either.

> If this is a time-sensitive feature you want to test - it could also be an option to tell the build barons to ignore the really noisy metric (since it's not the whole task) and open a follow-on ticket to investigate this further.

Actually, that's not the case. I confirmed that the feature actually improves the performance in other workloads and this workload is for future performance regression testing, which is not urgent. I wanted to give a closure to the feature. I'm in the middle of something for the other project and so will get back to you with my analysis and any changes necessary to this workload.

src/workloads/query/TimeSeriesLastpoint.yml
```diff
@@ -53,369 +54,330 @@ LastpointQueryMetaAscTimeAsc: &LastpointQueryMetaAscTimeAsc
...
+            Parameters:
+              Collection: *coll
+              IndexName: *metaDescTimeDesc
 
   - LoadConfig:
       Path: *phasePath
       Key: QuiesceActor
       Parameters:
-        MaxPhases: 23
+        MaxPhases: *MaxPhases
```
jawwadasghar, line 380: nit: It might be clearer if the active phases for the quiesce actor are defined here rather than inheriting the defaults. 
That way all the phase info is in this file, and it lines up with what is done for TimeSeriesLastpointHighCardinality.yml

src/workloads/query/TimeSeriesLastpointHighCardinality.yml
```diff
@@ -0,0 +1,338 @@
...
+            Parameters:
+              Collection: *coll
+              IndexName: *metaAscTimeDesc
+
+  # QuiesceActor will run in phase 18.
+  ##################################################################################################
+
+  ##################################################################################################
+  # Following 4 actors including the hidden QuiesceActor in phase 20 run a lastpoint query with a
+  # sort and group on meta subfield ascending and time descending.
```
jawwadasghar, line 278: Should this say meta subfield _descending_?

src/workloads/query/TimeSeriesLastpoint.yml
```diff
@@ -53,369 +54,330 @@ LastpointQueryMetaAscTimeAsc: &LastpointQueryMetaAscTimeAsc
...
 
 AutoRun:
```
jawwadasghar, line 382: We are deprecating support for standalone variants. 
Can you remove the AutoRun conditions for the standalone setups? 
This would also then match when the HighCardinality workload gets scheduled. 

Also should the branch conditional be "$gte: v7.0" like it is for the HighCardinality one?

src/workloads/query/TimeSeriesLastpoint.yml
yun-soo: This is done in #1219.

src/workloads/query/TimeSeriesLastpoint.yml
yun-soo: > Can you remove the AutoRun conditions for the standalone setups?

This is done in #1219.

> should the branch conditional be "$gte: v7.0" like it is for the HighCardinality one?

This is answered in #1219.

yun-soo (commented):
@jawwadasghar, addressed your comments

jawwadasghar (approved):
LGTM, thanks!

PR #1220, opened 2024-05-13
BF-33222 set `binSize` to null instead of "" (empty string)

lgfang:
**Jira Ticket:** [BF-33222](https://jira.mongodb.org/browse/BF-33222)

**Whats Changed:**

Set a "binSize" value to null. Originally it lacked a value. During YAML linting and formatting, we set its value to an empty string. Now, change it to `null` as the corresponding test failed with the value being an empty string.

**Patch testing results:**

[patch](https://spruce.mongodb.com/version/664154a8cd210900071d3a14/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

**Related PRs:**

### DIFF
--- a/src/workloads/query/GroupStagesOnComputedFields.yml
+++ b/src/workloads/query/GroupStagesOnComputedFields.yml
@@ -148,7 +148,7 @@ GroupStageAfterSortAndAddFields: &GroupStageAfterSortAndAddFields
               {
                 "t":
                   {
-                    $dateTrunc: { date: "$time", unit: "second", binSize: "" },
+                    $dateTrunc: { date: "$time", unit: "second", binSize: null },
                   },
               },
           },


jimoleary (approved):
LGTM

Ugh looks like fixing the follwing went from an implicit null to an explicit "":
```yaml
      pipeline: [{$sort: {"time": 1}},
                 {$addFields: {
                    "t": {$dateTrunc: {
                    date: "$time",
                    unit: "second",
                    binSize
                  }}
```

We might want to look through the changes to see if there are any more but I'm not sure how we would do that automatically :-(

thessem:
@jimoleary if it makes you feel less anxious I did pick up the inconsistency in https://github.com/mongodb/genny/pull/1205#pullrequestreview-2032908245, it just seems like the fix was the wrong way of going about it. It was only the `binSize` and some of the octal->non-octal changes that worry me, so this PR removes half my worries.

PR #1224, opened 2024-05-30
PERF-5373 Fix NoSuchField.yml so that it hits 'max works' as intended

dstorch:
**Jira Ticket:** [PERF-5373](https://jira.mongodb.org/browse/PERF-5373)

### Whats Changed

I confirmed that as originally implemented, the multi-planning trial period would end after the plan using index `{x1: 1}` hit EOF. By changing this predicate from `{x: {$lt: *selectivity}}` to `{x1: { $lte: 1}}`, I've made it so that all of the indexed plans need to perform a large index scan. After this change, I've confirmed that none of the plans will hit EOF, and the trial period will instead terminate by hitting the "max works" condition.

In addition, I noticed that we were unnecessarily repeating the `collMod` command to hide indexes 1000 times, so I fixed this to just hide indexes once.

### Patch Testing Results

I've confirmed that this lints locally by running `./run-genny -v lint-yaml --format`.

Kicked off a patch build: https://spruce.mongodb.com/version/6658beef88b0fc0007615e39/tasks. As of this writing, I'm letting it auto-generate tasks first and will need to come back and schedule the generated Genny task for this workload later.

### DIFF
--- a/docs/generated/workloads.md
+++ b/docs/generated/workloads.md
@@ -4396,13 +4396,19 @@ as well as classic.
 
 
 ### Description
-The goal of this test is to exercise multiplanning. We create as many indexes as possible, and run a
-query that makes all of them eligible, so we get as many competing plans as possible. Here, we add
-an additional predicate: {no_such_field: "none"} to guarantee that we hit getTrialPeriodMaxWorks().
+The goal of this test is to exercise the "max works" case of multi-planning. The test is similar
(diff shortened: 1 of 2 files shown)


dpercy:
> In addition, I noticed that we were unnecessarily repeating the collMod command to hide indexes 1000 times, so I fixed this to just hide indexes once.

Nice catch!

mbor-dev (approved):

dpercy (approved):

PR #1228, opened 2024-06-05
PERF-5355 Clean up query stats genny auto-run

shaargtz:
**Jira Ticket:** [PERF-5355](https://jira.mongodb.org/browse/PERF-5355)

### Whats Changed

Since the query stats backport is pretty much wrapped up, this PR will cleanup the modifications made to auto run configurations in different workloads to go back to the original states pre-project. The QueryStats workload will be left as >=v7.0 since we'll be [getting rid](https://jira.mongodb.org/browse/PERF-5456) of the `*-query-stats` configurations in DSI. 

### Patch Testing Results

[7.0](https://spruce.mongodb.com/version/665f6fcdd17bd600079ff5da/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[master](https://spruce.mongodb.com/version/6660a8085455de0007c0cbe6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

The patches are only running the QueryStats and QueryStatsQueryShapes workloads, but the query stats variants are gone.


awagenmaker (commented):
One high-level question: do we know if `atlas-like-replica.2023-09` is still in use anywhere? I'm thinking that we should remove all references to it that were added in [PERF-5152](https://jira.mongodb.org/browse/PERF-5152). If the variant is still in use, then we have some extra unnecessary auto-runs leftover from that backport change. If the variant is not still in use, these are stale references and we might as well remove them. 

Otherwise these changes look good!

shaargtz:
> do we know if atlas-like-replica.2023-09 is still in use anywhere?
nice catch @awagenmaker, it's only our workloads that are still using it, i'll go ahead and remove it.

@mongodb/performance i found there's only a compile for it in v8.0, and the setup file is still in DSI, i'm guessing that after this patch lands we could get rid of the configuration.

awagenmaker (approved):
Cool, thanks for checking on that @shaargtz ! LGTM

blagoev (approved):

ghartnett (approved):

dalyd (approved):
Adding an LGTM to see if it satisfies the Codeowners requirement. It should be satisfied by @blagoev 's review.

blagoev:
I was added to the group after my review, so perhaps that's the reason it didn't go through initially.

PR #1230, opened 2024-06-10
PERF-5566 Fix TPCH Q15 scale 10 phases

parker-felix:
**Jira Ticket:** [PERF-5566](https://jira.mongodb.org/browse/PERF-5566)

### Whats Changed

Before this change, the views that the queries are supposed to run against weren't being created so the queries returned immediately without doing any actual work.

### Patch Testing Results

[sys-perf patch](https://spruce.mongodb.com/version/66678bd32e682f0007eefe09/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/tpch/denormalized/10/Q15.yml
+++ b/src/workloads/tpch/denormalized/10/Q15.yml
@@ -9,10 +9,25 @@ Clients:
       socketTimeoutMS: -1
 
 Actors:
+  # This query creates a view before running any workloads, and then destroys the view afterwards.
+  - Name: TPCHDenormalizedQuery15SetupAndCleanup
+    Type: RunCommand
+    Database: &db tpch
+    Phases:
+      - LoadConfig:
+          Path: ../../../../phases/tpch/denormalized/Q15.yml
+          Key: TPCHDenormalizedQuery15CreateView
+      - &Nop { Nop: true }
+      - LoadConfig:
+          Path: ../../../../phases/tpch/denormalized/Q15.yml
+          Key: TPCHDenormalizedQuery15DropView
+
   - Name: TPCHDenormalizedQuery15Explain
     Type: RunCommand
     Database: tpch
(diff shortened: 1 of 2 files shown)


alicedoherty (changes requested):
Thanks @parker-felix for submitting this fix. Does this only affect TPCH Q15? From a quick look at the others, it looks like none of them create a view before the query is run.

alyacb (approved):
My bad- I was convinced this was done for scale 10 as well. Thank you for fixing this! LGTM.

src/workloads/tpch/denormalized/10/Q15.yml
```diff
@@ -9,10 +9,25 @@ Clients:
...
+          Path: ../../../../phases/tpch/denormalized/Q15.yml
+          Key: TPCHDenormalizedQuery15CreateView
+      - &Nop { Nop: true }
+      - LoadConfig:
+          Path: ../../../../phases/tpch/denormalized/Q15.yml
+          Key: TPCHDenormalizedQuery15DropView
+
   - Name: TPCHDenormalizedQuery15Explain
     Type: RunCommand
     Database: tpch
```
alicedoherty, line 27: nit: you've created a db anchor above (line 15), can we use that here

alyacb:
> From a quick look at the others, it looks like none of them create a view before the query is run.

@alicedoherty see https://github.com/mongodb/genny/blob/master/src/workloads/tpch/normalized/1/validate.yml (and denormalized equivalent)- this is a workload that verifies the correctness of individual queries. Q15 is the only one that creates a view. I also vaguely recall from the TPCH specification it was the only one that needed a view, but at any rate, it was the only one we created in the TPCH benchmark project.

alicedoherty (approved):
Thanks for the clarification Alya, LGTM in that case

PR #1231, opened 2024-06-12
DEVPROD-8018: Give Genny the ability to output JSON rollups

zamj:
**Jira Ticket:** [DEVPROD-8018](https://jira.mongodb.org/browse/DEVPROD-8018)

### Whats Changed

This adds a new flag to Genny that allows for generating JSON rollups from the FTDC files output from workloads. It does so by iterating over the output data folder and calling out to Curator using a CLI. 

I gave it a default of "Off" to preserve the current default behavior. When testing locally, I did not see a significant amount of time added to the runtime of Genny when doing these rollups. 

### Patch Testing Results

[Sys-perf](https://spruce.mongodb.com/version/666a071466c328000796b15c/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[Sys-perf with flag](https://spruce.mongodb.com/version/666a071466c328000796b15c/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=mixed)

### DIFF
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -133,6 +133,7 @@ def translate(
     )
     subprocess.run(args, check=True)
 
+
 # Calculate rollups from all the FTDC outputs that are not from canaries.
 def calculate_rollups(output_dir: str, workspace_root: str, genny_repo_root: str) -> None:
     curator = _find_curator(workspace_root, genny_repo_root)
@@ -150,10 +151,9 @@ def calculate_rollups(output_dir: str, workspace_root: str, genny_repo_root: str
                         "--inputFile",
                         ftdc_file_name,
                         "--outputFile",
-                        rollup_file_name
+                        rollup_file_name,
                     ]
                     subprocess.run(args, stderr=subprocess.STDOUT, text=True)
-
 
 
 @contextmanager
--- a/src/lamplib/src/genny/curator.py
+++ b/src/lamplib/src/genny/curator.py
@@ -145,6 +145,11 @@ def calculate_rollups(output_dir: str, workspace_root: str, genny_repo_root: str
                 if file.endswith(".ftdc") and "canary" not in file:
                     ftdc_file_name = os.path.join(root, file)
                     rollup_file_name = ftdc_file_name.replace(".ftdc", ".json")
+                    SLOG.info(
+                        "Creating perf rollup from FTDC file.",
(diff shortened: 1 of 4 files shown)


zamj:
Right now, DSI will fail with the new flag. This is because DSI is trying to treat the new JSON rollups as FTDC files. 

The default case will continue to work as it does today and I'll update DSI to be able to handle the new JSON files appropriately.

src/lamplib/src/genny/curator.py
```diff
@@ -134,6 +134,28 @@ def translate(
...
+def calculate_rollups(output_dir: str, workspace_root: str, genny_repo_root: str) -> None:
+    curator = _find_curator(workspace_root, genny_repo_root)
+    if not curator:
+        raise OSError("Could not find Curator.")
+    for root, dirs, files in os.walk(output_dir):
+        if "internal" not in root:
+            for file in files:
+                if file.endswith(".ftdc") and "canary" not in file:
+                    ftdc_file_name = os.path.join(root, file)
+                    rollup_file_name = ftdc_file_name.replace(".ftdc", ".json")
```
thessem, line 147: Would you mind adding some logging around this that includes the file being processed and the file that's being output?

src/lamplib/src/genny/curator.py
```diff
@@ -244,7 +266,7 @@ class CuratorDownloader(Downloader):
     # These build IDs are from the Curator Evergreen task.
     # https://evergreen.mongodb.com/waterfall/curator
 
-    CURATOR_VERSION = "a1293258b30e04ea503e6b91b6b1628789e90093"
+    CURATOR_VERSION = "fa0dfec710dfbe2fb47c15d5800e978717f23614"
```
thessem, line 269: Is this an incidental version bump or required for this?

thessem (approved):
Some minor comments but otherwise happy with this, I'll leave a tick on it because I don't think I need to re-review logging lines.

src/lamplib/src/genny/curator.py
zamj: This is needed - I added the Curator command that Genny will be using in this version.

src/lamplib/src/genny/curator.py
zamj: Good point, added

ghartnett (approved):

PR #1232, opened 2024-06-14
PERF-4866: remove standalone variants for filter_with_complex_logical_expression and array_traversal

miguelangelnieto:
Remove standalone variants for filter_with_complex_logical_expression and array_traversal tests.

### DIFF
--- a/src/workloads/query/ArrayTraversal.yml
+++ b/src/workloads/query/ArrayTraversal.yml
@@ -227,14 +227,6 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone
-          - standalone-80-feature-flags
-          - standalone-all-feature-flags
-          - standalone-classic-query-engine
-          - standalone-heuristic-bonsai
-          - standalone-query-stats
-          - standalone-sampling-bonsai
-          - standalone-sbe
           - replica
       branch_name:
         $neq:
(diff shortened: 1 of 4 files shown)


alicedoherty (approved):
LGTM but will there be a mongo PR to schedule `array_traversal_high_value` on a 3-Node ReplSet?

alicedoherty:
Wait I know we deleted the regular standalone variant, but I think we may have kept some of the query standalone variants around. Can we double check the other standalone variants you've deleted still aren't run?

src/workloads/query/ArrayTraversal.yml
```diff
@@ -227,14 +227,6 @@ AutoRun:
...
       mongodb_setup:
         $eq:
-          - standalone
-          - standalone-80-feature-flags
-          - standalone-all-feature-flags
-          - standalone-classic-query-engine
-          - standalone-heuristic-bonsai
-          - standalone-query-stats
-          - standalone-sampling-bonsai
-          - standalone-sbe
```
alicedoherty, line 237: `standalone-sbe` and `standalone-classic-query-engine` are still in use, we might not want to delete those (or check with query)

alicedoherty (changes requested):

alicedoherty (approved):
LGTM thanks Miguel

PR #1234, opened 2024-06-18
DEVPROD-8215 Remove ssh cloning of Mothra repo for documentation generation.

ahartschen:
**Jira Ticket:** [DEVPROD-8215](https://jira.mongodb.org/browse/DEVPROD-8215)

### Whats Changed

SSH cloning is being deprecated. This change relies on evergreen modules and the user manually checking out the mothra repo for local development.

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -452,9 +452,9 @@ Try using `python test_result_summary.py --help` for more options.
 
 2.  Run the self-tests:
 
-    To run `self-test` it is required that the Mothra repository exists in your local storage at src/mothra.
+    To run `self-test` it is required that the [Mothra repository](https://github.com/10gen/mothra/) exists at the root of the repository.
     ```bash
-    cd src && git clone git@github.com:10gen/mothra.git && cd ..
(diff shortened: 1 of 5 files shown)


docs/using.md
```diff
@@ -452,10 +452,16 @@ Try using `python test_result_summary.py --help` for more options.
 
 2.  Run the self-tests:
 
+    To run `self-test` it is required that the Mothra repository exists in your local storage at src/mothra.
     ```bash
-	./run-genny lint-yaml  # Lint all YAML files
-	./run-genny cmake-test  # Run C++ Unit test - only necessary if editing core C++ code
-	./run-genny resmoke-test  # Run Actor integration tests - only necessary if adding/editing Actors
+    cd src && git clone git@github.com:10gen/mothra.git && cd ..
```
johndaniels, line 457: Does it maybe make sense to put mothra at the top level rather than in src so it doesn't get confused with the Genny source code?

johndaniels (approved):

PR #1236, opened 2024-06-20
DEVPROD-7308: Run multiplanning Genny workloads on single-node all-feature-flags variants

wqian94:
**Jira Ticket:** [DEVPROD-7308](https://jira.mongodb.org/browse/DEVPROD-7308)

### Whats Changed

Add single replica all feature flags variants to multiplanner Genny workloads.

### Patch Testing Results

[Patch](https://spruce.mongodb.com/version/66a2b4eab87fd00007bc4d87).

### DIFF
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -395,11 +395,10 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - single-replica
-          - single-replica-all-feature-flags
-          - single-replica-sbe
-          - single-replica-classic-query-engine
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - replica
           - replica-all-feature-flags
+          - replica-query-engine-sbe
+          - replica-query-engine-classic
+          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -399,6 +399,6 @@ AutoRun:
           - replica-all-feature-flags
           - replica-query-engine-sbe
           - replica-query-engine-classic
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - standalone-80-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -395,10 +395,11 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone-sbe
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-all-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-classic-query-engine
           - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
+++ b/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
@@ -155,11 +155,11 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags
-          - standalone-all-feature-flags
-          - standalone-classic-query-engine
-          - standalone-sbe
           - replica
           - replica-all-feature-flags
       branch_name:
--- a/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
+++ b/src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
@@ -155,12 +155,10 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - single-replica
-          - single-replica-all-feature-flags
-          - single-replica-sbe
-          - single-replica-classic-query-engine
-          - standalone-80-feature-flags
           - replica
           - replica-all-feature-flags
+          - replica-query-engine-sbe
+          - replica-query-engine-classic
+          - standalone-80-feature-flags
       branch_name:
         $gte: v8.0
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -413,10 +413,11 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone-sbe
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-all-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-classic-query-engine
           - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -417,6 +417,6 @@ AutoRun:
           - replica-all-feature-flags
           - replica-query-engine-sbe
           - replica-query-engine-classic
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - standalone-80-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -413,11 +413,10 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - single-replica
-          - single-replica-all-feature-flags
-          - single-replica-sbe
-          - single-replica-classic-query-engine
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - replica
           - replica-all-feature-flags
+          - replica-query-engine-sbe
+          - replica-query-engine-classic
+          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/CompoundIndexes.yml
+++ b/src/workloads/query/multiplanner/CompoundIndexes.yml
@@ -405,10 +405,11 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone-sbe
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-all-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-classic-query-engine
           - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/CompoundIndexes.yml
+++ b/src/workloads/query/multiplanner/CompoundIndexes.yml
@@ -405,11 +405,10 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - single-replica
-          - single-replica-all-feature-flags
-          - single-replica-sbe
-          - single-replica-classic-query-engine
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - replica
           - replica-all-feature-flags
+          - replica-query-engine-sbe
+          - replica-query-engine-classic
+          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/CompoundIndexes.yml
+++ b/src/workloads/query/multiplanner/CompoundIndexes.yml
@@ -409,6 +409,6 @@ AutoRun:
           - replica-all-feature-flags
           - replica-query-engine-sbe
           - replica-query-engine-classic
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
+          - standalone-80-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ManyIndexSeeks.yml
+++ b/src/workloads/query/multiplanner/ManyIndexSeeks.yml
@@ -405,10 +405,11 @@ AutoRun:
   - When:
       mongodb_setup:
         $eq:
-          - standalone-sbe
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-all-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-classic-query-engine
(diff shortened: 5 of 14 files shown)


dalyd (approved):

dstorch (commented):
Thanks for picking this up William! Two things:

- My first question is the same as the one @dpercy posed on the ticket -- are the "graviton lite" variants just the normal single node replica set variants that we run these days?
- Should we also run these tests on a single node replica set in a release configuration? Right now we run on SBE, classic, and all feature flags, but we never run against the server in its regular configuration. This seems potentially valuable to help catch multi-planning regressions that can affect customers.

wqian94:
- According to [DSI configs](https://github.com/10gen/dsi/blob/master/configurations/infrastructure_provisioning/infrastructure_provisioning.graviton-single-lite.2022-11.yml), `graviton-single-lite` is `amz2-arm64` on `c6g.large` -- `c6g` being the Graviton CPU variant of `c6`s. `single-replica` runs on whatever the default setup is, which I'm guessing is not `c6g`s because the linked file above overrides the default values.
- I can add single-node repl set release. I think that's just `single-replica`.

wqian94:
Oh, it looks like the [default](https://github.com/10gen/dsi/blob/master/configurations/infrastructure_provisioning/defaults.2022-11.yml) for 2022-11 setups is `c6g`. There are two other "default" files for [2023-09](https://github.com/10gen/dsi/blob/master/configurations/infrastructure_provisioning/defaults.2023-09.yml) and [undated](https://github.com/10gen/dsi/blob/master/configurations/infrastructure_provisioning/defaults.original.yml).

dstorch:
Gotcha, thanks for looking into the business about the host types! Sounds like these compute-optimized instances with Graviton CPUs are just our instance type of choice for 1-node replsets on ARM.

 > I can add single-node repl set release. I think that's just single-replica.

That seems like a good idea to me. That way we are getting coverage in the release configuration as well as some special configurations with various flags (all-feature-flags, SBE, classic). Does that sound good to you as well @dalyd?

dstorch:
Ping @wqian94! Is this PR still active?

dalyd:
> Gotcha, thanks for looking into the business about the host types! Sounds like these compute-optimized instances with Graviton CPUs are just our instance type of choice for 1-node replsets on ARM.
> 
> > I can add single-node repl set release. I think that's just single-replica.
> 
> That seems like a good idea to me. That way we are getting coverage in the release configuration as well as some special configurations with various flags (all-feature-flags, SBE, classic). Does that sound good to you as well @dalyd?

I'm not a huge fan of the single-replica configs just because they are special butterflies, not always representative of real world behavior. That said, I don't think that shows up on read only workloads. So long as we're consistent with how we are running other query workloads, this sounds fine to me.

dalyd (approved):

wqian94:
@dstorch thoughts? I don't think it'd be that difficult to make the change, but I don't know how this will affect the trendlines that are currently in place.

wqian94:
@dstorch @alicedoherty how about I add the `single-replica` variants right now and not delete the `standalone` variants yet?

dstorch:
I think we should either be running these tests on standalone configurations or single replica set configurations, but not both. If the standalone configs are going away, then I would say they should run only on the single replica set configurations. That said, do all the single replica set configurations we are specifying in `AutoRun` actually exist? I didn't see any pre-existing instances in the code base of `single-replica` or `single-replica-all-feature-flags`? Could you list the build variants against which the multiplanning workloads will run after this change?

dstorch (changes requested):
Marking as "request changes" so you can re-request review when I should look again.

wqian94:
image

These are the variants that are run without the singletons that have been removed. [Patch](https://spruce.mongodb.com/version/66a2b4eab87fd00007bc4d87).

image

These are the variants that are run with the singletons. [Patch](https://spruce.mongodb.com/version/66a2b3b5d8e68d0007f59b35).

wqian94:
It appears that PERF-5527 (Commit 2a5810f4aff0e43c9705ad213ed156117004782a #1240) introduced some replica conditions that have created merge conflicts. Will need to rebase and force-push.

src/workloads/query/multiplanner/MultiPlanningReadsALotOfData.yml
```diff
@@ -155,11 +155,11 @@ AutoRun:
...
-          - standalone
+          - single-replica
+          - single-replica-all-feature-flags
+          - single-replica-sbe
+          - single-replica-classic-query-engine
           - standalone-80-feature-flags
-          - standalone-all-feature-flags
-          - standalone-classic-query-engine
-          - standalone-sbe
           - replica
```
alicedoherty, line 163: There's some inconsistencies in the variants we're running across these workloads. Some have `replica` and `replica-all-feature-flags`, others just have `replica-all-feature-flags`. Assuming that's unintentional can we make them all consistent?

alicedoherty (commented):
Thanks @wqian94 for those changes, just a small nit/question.

dstorch:
I've become a little confused here. The original goal of this ticket was to make it so that the genny multi-planning workloads run in these four configurations:
* SBE
* Classic
* All feature flags
* Release

(It's possible that we could drop SBE and get coverage through all feature flags, but let's leave that issue aside.) Given the nature of these benchmarks, whether they run on standalone or single-node replsets is immaterial. We can just take the advice of the perf team on this. If the single-node replica set variants aren't available yet, then we should use the standalone variants that are available. Also, I don't think we need all four variants to be covered on both ARM and Intel, but we can take the perf folks' advice here as well.

Looking at the screenshots you posted, it looks like neither version of the patch covers all four configurations -- the release configuration is missing. I actually don't know where to look to understand what `AutoRun` values are available under `mongodb_setup` and what build variants they map to. @alicedoherty perhaps you could give us some pointers on this?

alicedoherty:
> I actually don't know where to look to understand what AutoRun values are available under mongodb_setup and what build variants they map to. @alicedoherty perhaps you could give us some pointers on this?

All of the build variants running on sys-perf are defined in either [variants.yml](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml) or [variants_branching.yml](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants_branching.yml) (variants_branching includes all the variants we want on our release branches so is a smaller prioritised subset of all our variants). Any of the variants relevant here should be in variants.yml.

The values under `AutoRun` directly map to what `mongodb_setup` is defined in the YAML of a build variant, e.g., searching for "single-replica" in variants.yml tells me this condition will match the [`1-Node ReplSet With Test Commands ARM AWS 2023-11`](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L436-L441), [`Long-Running 1-Node ReplSet ARM AWS 2023-11`](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1283-L1288), and [`M60-like Large Volume 1-Node ReplSet ARM AWS 2023-11`](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1327-L1332). (Note it should be an exact match, e.g., `single-replica` won't match a variant that uses `mongodb_setup: single-replica-sbe`) Now this is where I started to get confused because looking at @wqian94 screenshot shows that none of these variants were used despite having `single-replica` under `AutoRun`. This is because none of those variants have the [`schedule_variant_auto_tasks`](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L270) task running on them, which is the "task" that goes through the Genny workloads and schedules any workloads that have a matching condition in their `AutoRun`. **_I think_ this is a mistake on our team's behalf and I will follow-up on this.**

To close the loop on mapping Genny's `AutoRun` to the actual variant configs, that value beside `mongodb_setup` then maps to a config file in DSI, in this example case [mongodb_setup.single-replica.yml](https://github.com/10gen/dsi/blob/master/configurations/mongodb_setup/mongodb_setup.single-replica.yml)

> Given the nature of these benchmarks, whether they run on standalone or single-node replsets is immaterial. We can just take the advice of the perf team on this.

My opinion here would be to use `single-replica-X` over `standalone-X`. That gives us the following variants:

**SBE** (`single-replica-sbe`)
- [SBE 1-Node ReplSet Intel AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L232C19-L232C55)
- [SBE 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L416C19-L416C53)

**Classic** (`single-replica-classic-query-engine`)
- [Classic Query Engine Intel 1-Node ReplSet Intel AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L212C19-L212C78)
- [Classic Query Engine 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L396C19-L396C70)

**All feature flags** (`single-replica-all-feature-flags`)
- [graviton lite, all FF 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L252C19-L252C71)
- [Long-Running all feature flags 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1305C19-L1305C80)
- [M60-like Large Volume all feature flags 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1348C19-L1348C89)

**Release**
- I can't find any variants here that match `single-replica-80-feature-flags` or `standalone-80-feature-flags` (I assume this standalone one existed at some stage but was deleted by perf). If this is important we might have to create this variant for you.

Is there a need for the variants that match `single-replica` (the ones listed in the example above)? Or do the variants above cover all the cases you need?

> I don't think we need all four variants to be covered on both ARM and Intel, but we can take the perf folks' advice here as well

Running on ARM and Intel probably isn't necessary in this case, but since the `AutoRun` condition will match both the ARM and Intel version of the variant I think it's much easier to just leave these tests running on both.

Let me know if this actually answered your questions @dstorch or if things can be further clarified/more detail is needed!

wqian94:
@alicedoherty how long do you think it would take to make sure that `schedule_variant_auto_tasks` works on the `single-replica` variants?

Also, just to be sure, which of the `single-replica` variants are "release" variants? Presumably `single-replica-sbe` and `single-replica-classic-query-engine`?

alicedoherty:
@wqian94 The change shouldn't take long - I can have a PR up by the end of today if my assumptions are correct. _But_ do you actually need all `single-replica` variants ([1-Node ReplSet With Test Commands ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L436-L441), [Long-Running 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1283-L1288), and [M60-like Large Volume 1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/master/etc/system_perf_yml_components/variants.yml#L1327-L1332))? There are [no other Genny tasks "AutoRunning" on those variants](https://github.com/search?q=repo%3Amongodb%2Fgenny+single-replica&type=code&p=1).

> Also, just to be sure, which of the single-replica variants are "release" variants? Presumably single-replica-sbe and single-replica-classic-query-engine?

I wasn't sure exactly what you meant by release variants. In my previously comment I was making the assumption that these were variants with the 8.0 feature flags enabled (`single-replica-80-feature-flags` or `standalone-80-feature-flags`). But like I mentioned above I couldn't find any variants that matched that.

wqian94:
@dstorch correct me if I'm wrong. The ticket's original intention is to ensure that this test runs on both all feature flags configurations and on the release configuration. The latter is basically whatever is expected to become the release candidate, I think? The goal is to make sure that we're gathering data from not just so feature flags variants.

alicedoherty:
Hold on, I think what we want here is the plain [1-Node ReplSet ARM AWS 2023-11](https://github.com/10gen/mongo/blob/v8.0/etc/system_perf_yml_components/variants.yml#L226C17-L226C47) variant. That should be the exact same as the SBE and classic query engine variants I've listed above, just w/o SBE/classic query engine enabled. _But_, for some reason that variant isn't running on master (plain sys-perf), just sys-perf-8.0. Maybe we can take this offline and come back here with a more solid proposal - otherwise I think we're going to go in circles 😅

wqian94:
As discussed last week, awaiting the conclusion of `sys-perf` configs to the DSI repo before:
- Adding new Query-specific variants.
- Updating the variants list in these tests with `replica-*` variants.

alicedoherty (commented):
Thanks @wqian94 for adding those new 3-Node ReplSet SBE/Classic Engine variants! These changes in general LGTM but obviously hold off merging until https://github.com/10gen/dsi/pull/1868 is in (and run a patch to sanity check the AutoRun conditions get picked up).

I know I mentioned potentially keeping this running on standalone for a period to build up history on the replica variants - will `standalone-80-feature-flags` give you enough for that?

wqian94:
[DSI PR](https://github.com/10gen/dsi/pull/1868) has been merged!

[Here](https://spruce.mongodb.com/version/66ba46ba2019f20007a404ac) is an updated patch, with [comparison analysis](https://performance-analyzer.server-tig.prod.corp.mongodb.com/perf-analyzer-viz/?comparison_id=20ba865d-2206-4a9b-8fff-875a66a7685a&selected_tab=data-table&test_filter=.%2A%5C.find&measurement_filter=OperationThroughput&percent_filter=0%7C%7C100&z_filter=0%7C%7C10).

wqian94:
> Thanks @wqian94 for adding those new 3-Node ReplSet SBE/Classic Engine variants! These changes in general LGTM but obviously hold off merging until [10gen/dsi#1868](https://github.com/10gen/dsi/pull/1868) is in (and run a patch to sanity check the AutoRun conditions get picked up).
> 
> I know I mentioned potentially keeping this running on standalone for a period to build up history on the replica variants - will `standalone-80-feature-flags` give you enough for that?

I'm okay with having the blip in the data. We can just try to connect it together later.

Can we keep the `standalone-80-feature-flags` variant until DEVPROD-9461?

alicedoherty (approved):
LGTM - thanks @wqian94 for adding those variants. I'm happy for this to be merged once you've run a patch with just `schedule_variant_auto_tasks_task` scheduled and validated all these multiplanner tests are scheduled.

Also repeating here for completeness that these won't be run on `sys-perf-8.0` until [BACKPORT-22859](https://jira.mongodb.org/browse/BACKPORT-22859) is completed. (You don't have to do anything when that merges, your new variant should automatically start running then.)

src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -395,10 +395,10 @@ AutoRun:
...
         $eq:
-          - standalone-sbe
-          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-all-feature-flags # At time of writing this will enable PM-3591.
-          - standalone-classic-query-engine
+          - replica
           - replica-all-feature-flags
+          - replica-query-engine-sbe
+          - replica-query-engine-classic
+          - standalone-80-feature-flags # At time of writing this will enable PM-3591.
```
dstorch, line 402: I think you should drop the "At time of writing" comment. [SPM-3591](https://jira.mongodb.org/browse/SPM-3591) is fully enabled in master and 8.0, so the comment has grown stale. Same for the changes in all the other workload files.

dstorch (changes requested):
This makes sense! Just so I'm clear on the effect of this change, could you copy-paste (or screenshot) the list of build variants on which the multi-planner benchmarks will run now?

Also, I think there are a few more multi-planner benchmark files where you should make the same change:

```
Simple.yml
Subplanning.yml
UseClusteredIndex.yml
VariedSelectivity.yml
```

wqian94:
Variants:
image

Tasks:
image

dstorch (approved):
LGTM, thanks for closing this out!

PR #1237, opened 2024-06-25
DEVPROD-8018: Add documentation on new JSON files

zamj:
**Jira Ticket:** [DEVPROD-8018](https://jira.mongodb.org/browse/DEVPROD-8018)

### Whats Changed

Documentation on the new JSON files and the new Genny flag.

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -380,6 +380,8 @@ Workloads can be run with the following Genny command:
 
 If your workload requires a MongoDB connection (most do), then you can pass it in with `-u`. See [Connecting to the Server](#orgd6b0450) for more details.
 
+If you'd like to have your results summarized locally in a JSON format, you can pass in the `-r` flag.
+
 
 <a id="orgec88ad4"></a>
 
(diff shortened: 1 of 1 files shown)


jimoleary (approved):
Thanks for the extra docs!

Question: Should we also add _**-r**_ to most of the ['./run-genny workload' invocations](https://github.com/search?q=repo%3Amongodb%2Fgenny+%22.%2Frun-genny+workload%22+path%3Adocs%2Fusing.md&type=code)?

zamj:
> Thanks for the extra docs!
> 
> Question: Should we also add _**-r**_ to most of the ['./run-genny workload' invocations](https://github.com/search?q=repo%3Amongodb%2Fgenny+%22.%2Frun-genny+workload%22+path%3Adocs%2Fusing.md&type=code)?

We can, I just wasn't sure if this was a behavior people would always want. If you think so, I can add it there. Curious to hear @thessem 's thoughts on this as well

thessem (approved):
I don't have a strong opinion on the -r thing, weak opinion is once we're using it as our main way of getting data it should be the default/updated everywhere.

jimoleary:
Feel free to merge this as is. I'm happy with it as it stands, we can revisit it in the future.

PR #1238, opened 2024-06-26
PERF-5628 Reduce runtime of TimeseriesEnum test

borosaurus:
**Jira Ticket:** [PERF-5628](https://jira.mongodb.org/browse/PERF-5628)

### Whats Changed

Changing the test to run on a smaller data set. This will help avoid timeouts (some of the queries currently take 6+ minutes). At the same time, this will not make the queries run so quickly that we don't get a good signal on the query's performance.

### Patch Testing Results

Patch build running here:
https://spruce.mongodb.com/version/667c46abea070d00079381c4/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/src/workloads/query/TimeseriesEnum.yml
+++ b/src/workloads/query/TimeseriesEnum.yml
@@ -11,9 +11,9 @@ Keywords:
 GlobalDefaults:
   Database: &database test
   TSCollection: &tsCollection Collection0
-  DocumentCount: &documentCount 1e7
+  DocumentCount: &documentCount 1e6
   EnumCollection: &enumCollection Collection1
-  repeat: &repeat 20
+  repeat: &repeat 5
   MetaCount: &metaCount 10
   MaxPhases: &maxPhases 15
 


src/workloads/query/TimeseriesEnum.yml
```diff
@@ -11,9 +11,9 @@ Keywords:
 GlobalDefaults:
   Database: &database test
   TSCollection: &tsCollection Collection0
-  DocumentCount: &documentCount 1e7
+  DocumentCount: &documentCount 1e6
```
borosaurus, line 14: The timeout was caused by each individual query running for too long, so I reduced the size of the data set. The queries still take many seconds to execute, which is more than enough to get good signal. I also reduced the repeat count, since the task currently takes >2 hours, which seems unnecessarily long.

fotiniAlvanaki (approved):
LGTM

blagoev (approved):

PR #1240, opened 2024-07-02
PERF-5527 add replica condition where standalone only is defined

blagoev:
Standalone variants are being removed. 
For genny tasks that have a standalone conditions and would otherwise not run, we add in a replica condition.

### DIFF
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -399,5 +399,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -399,7 +399,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
-          - replica
           - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/BlockingSort.yml
+++ b/src/workloads/query/multiplanner/BlockingSort.yml
@@ -400,5 +400,6 @@ AutoRun:
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
           - replica
+          - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -417,7 +417,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
-          - replica
           - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -418,5 +418,6 @@ AutoRun:
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
           - replica
+          - replica-all-feature-flags
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/ClusteredCollection.yml
+++ b/src/workloads/query/multiplanner/ClusteredCollection.yml
@@ -417,5 +417,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/CompoundIndexes.yml
+++ b/src/workloads/query/multiplanner/CompoundIndexes.yml
@@ -409,5 +409,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
       branch_name:
         $gte: v7.3
--- a/src/workloads/query/multiplanner/CompoundIndexes.yml
+++ b/src/workloads/query/multiplanner/CompoundIndexes.yml
@@ -410,5 +410,6 @@ AutoRun:
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
           - replica
+          - replica-all-feature-flags
       branch_name:
         $gte: v7.3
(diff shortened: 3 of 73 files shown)


jawwadasghar:
Thanks Lubo!
I think to maintain similar test coverage, where we see "standalone-all-feature-flags" we should add "replica-all-feature-flags". 

Also, can you create a separate ticket to remove the standalone condition all-together once we ship 8.0?

blagoev:
@jawwadasghar Will fix

blagoev:
added a follow up ticket https://jira.mongodb.org/browse/PERF-5643

blagoev:
@jawwadasghar I have added `replica-all-feature-flags` to all workloads which had `standalone-all-feature-flags`. Could you re-review when you have the time?

src/workloads/query/multiplanner/VariedSelectivity.yml
```diff
@@ -254,5 +254,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 257: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/UseClusteredIndex.yml
```diff
@@ -414,5 +414,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 417: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/Subplanning.yml
```diff
@@ -356,5 +356,7 @@ AutoRun:
           - standalone-sbe
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 359: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/Simple.yml
```diff
@@ -408,5 +408,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 411: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/NonBlockingVsBlocking.yml
```diff
@@ -222,5 +222,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 225: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/MultikeyIndexes.yml
```diff
@@ -412,5 +412,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 415: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/ManyIndexSeeks.yml
```diff
@@ -409,5 +409,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 412: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/CompoundIndexes.yml
```diff
@@ -409,5 +409,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 412: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

src/workloads/query/multiplanner/ClusteredCollection.yml
```diff
@@ -417,5 +417,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 420: Same case - I think we could remove the non-feature-flag replica case here as there is no standard standalone case running

jawwadasghar (changes requested):
I don't think we need to add a replica case when no standard standalone case is called. 
The standalone-sbe / standalone-classic-query-engine cases will need to get replaced with new replica set equivalents that we don't have yet 

Also can you kick off a patch test that runs all the genny tests for a variant that has mongodb_setup equal to the replica / replica-all-feature-flags values we are adding?

blagoev:
@jawwadasghar I will remove the `replica` as suggested in the review comments. Just to clarify this PR did not add `replica` on the places where standard standalone case is NOT present. 
I am not sure why `replica` was added there but will remove it as you suggested.
I will run a patch.

src/workloads/query/multiplanner/BlockingSort.yml
```diff
@@ -399,5 +399,7 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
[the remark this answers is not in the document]
blagoev: fixed

src/workloads/query/multiplanner/ClusteredCollection.yml
blagoev: fixed

src/workloads/query/multiplanner/CompoundIndexes.yml
blagoev: fixed

blagoev:
Patch here https://spruce.mongodb.com/version/66a0b651427ede00076a53ba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
I hope I have scheduled all genny tests the way you suggested.

src/workloads/query/multiplanner/NoSuchField.yml
```diff
@@ -408,5 +408,6 @@ AutoRun:
           - standalone-80-feature-flags # At time of writing this will enable PM-3591.
           - standalone-all-feature-flags # At time of writing this will enable PM-3591.
           - standalone-classic-query-engine
+          - replica
```
jawwadasghar, line 411: I think this one should be replica-all-feature-flag

jawwadasghar (approved):
I think there is a typo in NoSuchField.yml, but other than that looks good!
Approving, but would be good to fix that before you push. 

Most of the tasks in the patch look to have succeeded. 
I think the majority_reads10_k_threads failure is covered by BF-31431, but would be good to double check that. 
I don't think you touched those files anyway

blagoev:
@jawwadasghar Thanks, will fix the NoSuchField.yml before merging

PR #1242, opened 2024-07-03
[PERF-5596] Create EventCollection.yml workload

cannonmongo:
[PERF-5596](https://jira.mongodb.org/browse/PERF-5596)


src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,71 @@
...
+Actors:
+  - Name: CreateEventCollection
+    Type: RunCommand
+    Phases:
+      - Repeat: 1
+        Database: *db
+        Operation:
+          OperationName: RunCommand
+          OperationCommand:
+            drop: *Collection
```
gregorynoma, line 19: Do other workloads start with a drop like this? I think we probably don't want this -- it should always be running on a system where the collection doesn't yet exist, as far as I know

src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,71 @@
+SchemaVersion: 2018-07-01
+Owner: Storage Execution
+Description: |
+  Inserts many documents into an event collection.
```
gregorynoma, line 4:
```suggestion
  Inserts documents into an event collection.
```

src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,71 @@
...
+            ]
+          point:
+            &point {
+              type: Point,
+              coordinates:
+                [
+                  { ^RandomInt: { min: -180, max: 180 } },
+                  { ^RandomInt: { min: -90, max: 90 } },
+                ],
+            }
```
gregorynoma, line 61: Since the types in the data don't really matter much as there are no indexes, for simplicity maybe we should stick with just numbers and strings

src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,71 @@
...
+    Phases:
+      - *Nop
+      - *Nop
+      - Repeat: 1
+        Database: *db
+        Threads: 1
+        CollectionCount: 1
+        DocumentCount: 1000000
+        BatchSize: 1000
+        Document:
```
gregorynoma, line 42: How many bytes does each document end up being?

src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,71 @@
...
+  - Name: Insert
+    Type: Loader
+    Database: *db
+    Threads: 1
+    Phases:
+      - *Nop
+      - *Nop
+      - Repeat: 1
+        Database: *db
+        Threads: 1
```
gregorynoma, line 38: Does this mean all of the writes are happening on a single thread? We'll want the workload to test when there are multiple threads writing as well, since that's where any insert serialization we implement will be noticeable. In fact, maybe we want multiple loader phases each with a different level of concurrency? Maybe 1, 4, 8, 16?

src/workloads/execution/EventCollection.yml
gregorynoma, line 1: How long does this workload take to run?

cannonmongo: `Done with load phase. All 1000000 documents loaded into Collection0 in 8.87381 seconds at 1.1269e+05 docs/s`

src/workloads/execution/EventCollection.yml
```diff
@@ -0,0 +1,107 @@
...
+        Active: [1]
+        NopInPhasesUpTo: *MaxPhases
+        PhaseConfig:
+          Repeat: 1
+          Database: *db
+          Threads: 1
+          CollectionCount: 1
+          DocumentCount: 1000000
+          BatchSize: 1000
+          Document:
```
cannonmongo, line 43: Each document, with field names + values, should be 156 byes (assuming there is no overhead)

src/workloads/execution/EventCollection.yml
cannonmongo: Each document, with field names + values, should be 156 byes (assuming there is no overhead)

src/workloads/execution/EventCollection.yml
binhvomongodb: It sounds like we'll want a much bigger load so we can see the behavior once truncates are involved

src/workloads/execution/EventCollectionWrites.yml
```diff
@@ -0,0 +1,172 @@
...
+Keywords:
+  - event
+  - RunCommand
+  - Loader
+
+GlobalDefaults:
+  Collection: &Collection Collection0
+  Database: &db test
+  MaxPhases: &MaxPhases 8
+  DocumentCount: &DocumentCount 3000000
```
cannonmongo, line 15: This many documents takes just over 1 minute for the single threaded actor

PR #1244, opened 2024-07-16
PERF-5554 Add a multi-threaded read-only Genny workload

samanca:
Adds a new multi-threaded, read-only workload to collect stable performance metrics for `findOne` operations.

### DIFF
--- a/src/workloads/basic/ReadOnlyMultiThreaded.yml
+++ b/src/workloads/basic/ReadOnlyMultiThreaded.yml
@@ -1,0 +1,91 @@
+SchemaVersion: 2018-07-01
+Owner: Product Performance
+Description: |
+  Runs a multi-threaded, read-only workload comprising of `findOne` commands that all target the 
+  same document. The workload inserts a single document and then spawns multiple worker threads to 
+  retrieve it.
+
+GlobalDefaults:
+  Nop: &Nop {Nop: true}
+  Database: &Database test
+  Collection: &Collection test
+  PaddingLength: &PaddingLength 64 # bytes
+  WarmUpPeriod: &WarmUpPeriod 30 seconds
+  RunPeriod: &RunPeriod 5 minutes
+  Threads: &Threads 64
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 500
+
+Actors:
+- Name: CRUDOps
+  Type: CrudActor
+  Threads: 1
+  Phases:
+  - Repeat: 1
+    Database: *Database
+    Collection: *Collection
+    Operations:
+    - OperationName: drop
+  - Repeat: 1 # only 1 document is needed for this workload
+    Database: *Database
+    Collection: *Collection
+    Operations:
+    - OperationName: insertOne
+      OperationCommand:
+        Document:
+          _id: 1
+          padding: {^FastRandomString: {length: *PaddingLength}}
+  - *Nop
+  - *Nop
+- Name: RunCommand
+  Type: RunCommand
+  Threads: *Threads
+  Phases:
+  - *Nop
+  - *Nop
+  - Duration: *WarmUpPeriod
+    Database: *Database
+    Collection: *Collection
+    Operations:
+    - OperationMetricsName: hello
+      OperationName: RunCommand
+      OperationIsQuiet: true
+      OperationCommand:
+        hello: 1
+  - *Nop
+- Name: MTCrudOps
+  Type: CrudActor
+  Threads: *Threads
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - Duration: *RunPeriod
+    Database: *Database
+    Collection: *Collection
+    Operations:
+    - OperationName: findOne
+      OperationCommand:
+        Filter: {_id: 1}
+
+# Guard against timeout for no output.
+- Name: LoggingActor
+  Type: LoggingActor
+  Threads: 1
+  Phases:
+  - *Nop
+  - *Nop
+  - *Nop
+  - LogEvery: 2 minutes
+    Blocking: None
+
+AutoRun:
+- When:
+    mongodb_setup:
+      $eq:
+      - standalone
+      - replica
+      - shard-single


samanca:
Evergreen: https://spruce.mongodb.com/version/6696c72bedf09f0007cb4780

I just don't know if making the UI show metrics collected by Genny should be done as part of this ticket.

jimoleary:
> I just don't know if making the UI show metrics collected by Genny should be done as part of this ticket.

If your metrics are missing then that is a bug in genny / DSI or cedar. But AFAICS your metrics are there,  they are just not visible from the default context.

If you remove the 'default sys-perf metrics' context here:

context 2024-07-17 09-06-26

Then your metrics are visible:

no context 2024-07-17 09-07-12

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
+SchemaVersion: 2018-07-01
+Owner: Product Performance
```
jimoleary, line 2: Is product perf  the long term owner of this workload?

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
...
+Description: |
+  Runs a multi-threaded, read-only workload comprising of `findOne` commands that all target the
```
jimoleary, line 4: What are the main metrics to look at for this workload, can you document them here?

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
+SchemaVersion: 2018-07-01
+Owner: Product Performance
+Description: |
+  Runs a multi-threaded, read-only workload comprising of `findOne` commands that all target the 
+  same document. The workload inserts a single document and then spawns multiple worker threads to 
+  retrieve it.
+
```
jimoleary, line 7: Can you add a list of Keywords? 

For example:

```
Keywords:
  - findOne
  - MultiThreaded
  - stable
  - autoReverter

```

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
...
+          padding: {^FastRandomString: {length: *PaddingLength}}
+  - *Nop
+  - *Nop
+- Name: RunCommand
+  Type: RunCommand
+  Threads: *Threads
+  Phases:
+  - *Nop
+  - *Nop
+  - Duration: *WarmUpPeriod
```
jimoleary, line 49: What is this warming, the connection pool?

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
...
+      OperationCommand:
+        Document:
+          _id: 1
+          padding: {^FastRandomString: {length: *PaddingLength}}
+  - *Nop
+  - *Nop
+- Name: RunCommand
+  Type: RunCommand
+  Threads: *Threads
+  Phases:
```
jimoleary, line 46: You could also consider using [OnlyActiveInPhases](https://github.com/mongodb/genny/blob/c4ce0d6dbe195629a3348eba9573f25381890683/docs/using.md#orgf9c328f) but I'm not convinced it would save you anything here.

src/workloads/basic/ReadOnlyMultiThreaded.yml
jimoleary: Could you add some of the rationale for this test from the other tickets?

```
We will benefit from very stable benchmarks, where the primary objective is to identify regressions introduced by individual commits, and not necessarily measuring the end-to-end database performance. Such benchmarks are the building blocks for adding performance tests to the auto-reverter.
```

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,91 @@
...
+  Collection: &Collection test
+  PaddingLength: &PaddingLength 64 # bytes
+  WarmUpPeriod: &WarmUpPeriod 30 seconds
+  RunPeriod: &RunPeriod 5 minutes
+  Threads: &Threads 64
+
+Clients:
+  Default:
+    QueryOptions:
+      maxPoolSize: 500
```
jimoleary, line 20: Do you actually need a custom client here? The number of threads is less than the default [maxPoolSize](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/connection-pools/#:~:text=to%20the%20minPoolSize.-,maxPoolSize,-The%20maximum%20number) of 100.

jimoleary (changes requested):
Some suggestions about documentation etc 

How stable / noisy are the results here? What is the Coefficient of Variation for 5 or 6 runs?

It looks like you were trying to run this task on the [2024 variant](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/variants_branching.yml#L378) but right now this variant has no genny workloads configured.

If you wanted to patch test this you would need to add either:
 1. A specific [task](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/tasks.yml#L1385-L1390) 
 2. A new [test_control](https://github.com/10gen/dsi/blob/master/configurations/test_control/test_control.tsbs_query_genny.yml) 
 3. The [auto tasks](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/variants_branching.yml#L463-L464) to the 2024 variant.

Long term option 1 would probably be the preference.

samanca:
> https://spruce.mongodb.com/version/6696c72bedf09f0007cb4780

Thank you, I didn't notice that it's filtered :)

src/workloads/basic/ReadOnlyMultiThreaded.yml
samanca: I have this set to 200 and 100, and Genny failed on some runs due to connection limit issues, so opted to set it to a value that I am sure is never exhausted by the benchmark. LMK if that's acceptable or if you want us to root cause the issue as part of this PR.

src/workloads/basic/ReadOnlyMultiThreaded.yml
samanca: Yes, added a comment explaining this.

src/workloads/basic/ReadOnlyMultiThreaded.yml
samanca: I believe so, since the long term objective is to use this as a stable benchmark and potentially add it to auto-reverter.

samanca:
> Some suggestions about documentation etc
> 
> How stable / noisy are the results here? What is the Coefficient of Variation for 5 or 6 runs?
> 
> It looks like you were trying to run this task on the [2024 variant](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/variants_branching.yml#L378) but right now this variant has no genny workloads configured.
> 
> If you wanted to patch test this you would need to add either:
> 
> 1. A specific [task](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/tasks.yml#L1385-L1390)
> 2. A new [test_control](https://github.com/10gen/dsi/blob/master/configurations/test_control/test_control.tsbs_query_genny.yml)
> 3. The [auto tasks](https://github.com/mongodb/mongo/blob/a8f7db25763df9647df2a14d63ccad5014eb2031/etc/system_perf_yml_components/variants_branching.yml#L463-L464) to the 2024 variant.
> 
> Long term option 1 would probably be the preference.

Also regarding the noise levels, I have done some research for this specific workload that is available here: https://docs.google.com/document/d/1OKirHDh1gz2TxfILzUQUOtmYw8CgiQSca1A1aT6im2s

src/workloads/basic/ReadOnlyMultiThreaded.yml
jimoleary: Sounds good, thanks!

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,99 @@
+SchemaVersion: 2018-07-01
+Owner: Product Performance
+Description: |
+  Runs a multi-threaded, read-only workload comprising of `findOne` commands that all target the
```
jimoleary, line 4: TYVM!

src/workloads/basic/ReadOnlyMultiThreaded.yml
jimoleary: That is fine ... but peculiar. Can you raise a PERF ticket to note the issues you found. I'm not sure how much effort it will get as genny is going to be mothballed but I'd like to track the issues you found.

I'm confused as to why a test with 64 threads required many more connections .. unless this test caused a mini-connection storm which would be interesting in itself.

I was going to suggest that [minPoolSize](https://www.mongodb.com/docs/languages/cpp/cpp-driver/current/connection-pools/#:~:text=to%20the%20pool.-,minPoolSize,-Sets%20a%20target) might avoid the need for a warmup but on reading the docs that doesn't seem to be the case.

src/workloads/basic/ReadOnlyMultiThreaded.yml
```diff
@@ -0,0 +1,99 @@
...
+      - *Nop
+      - LogEvery: 2 minutes
+        Blocking: None
+
+AutoRun:
+  - When:
+      mongodb_setup:
+        $eq:
+          - standalone
+          - replica
```
jimoleary, line 98: > 
> IIUC, all of the suggestions require changes to other repos, correct? If so, I'll send another PR to include this workload in `tasks.yml`.

Actually it looks like someone smuggled the `auto_task` tasks into the variant so it maybe as simple as adding `replica.2024-05` to the list.

jimoleary (approved):
LGTM mod testing the `replica.2024-05`.

> Also regarding the noise levels, I have done some research for this specific workload that is available here: https://docs.google.com/document/d/1OKirHDh1gz2TxfILzUQUOtmYw8CgiQSca1A1aT6im2s

**Using a multi-threaded Genny workload running against a vanilla MongoDB server, we can achieve noise levels below 1% on dedicated EC2 VMs.**

Nice!

We are currently in the process of moving from genny to locust ... this workload would be perfect as a comparison of 2 workload generators.

I'm going to create some PERF tickets to have a look at porting this workload and scheduling / comparing performance and noise profiles.

jimoleary:
I've created [PERF-5673](https://jira.mongodb.org/browse/PERF-5673) to look at writing a locust version of this workload (once it's merged). I think it's a perfect workload to investigate differences between genny / locust.

src/workloads/basic/ReadOnlyMultiThreaded.yml
samanca: Filed [PERF-5681](https://jira.mongodb.org/browse/PERF-5681).

samanca:
Confirmed that the workload now runs on `replica.2024-05`: https://spruce.mongodb.com/version/66992ee4688ab00007fb4f6d

jimoleary (approved):
LGTM still ... also great that we only need one commit!

PR #1245, opened 2024-07-22
PERF-5507 Fix AggregateExpressions.yml query definition and data generation

ktbooker:
**Jira Ticket:** [PERF-5507](https://jira.mongodb.org/browse/PERF-5507)

### Whats Changed

Fixed generation of array data to populate arrays of size 10, previously was only making arrays of size 1. 

Additionally fixed a query using an undefined field.


parker-felix (commented):
Per offline discussion, we are going to modify data generation so that each document has the same arrays, to match the original design of the workload.

parker-felix (approved):
LGTM

parker-felix:
Verified that updated data generation looks good offline so still LGTM.

leozzx (approved):
lgtm

miguelangelnieto (approved):

PR #1249, opened 2024-07-31
PERF-5711 remove quiesce so we can rolling restart

hilldani:
**Jira Ticket:** PERF-5711

### Whats Changed
Replace quiesce with noop which appears to not impact the results in this test thankfully. This lets us then use this genny test to test rolling restarts. Instead of making genny quiesce handle a rolling restart we should keep this in mind for Locust

### Patch Testing Results
[patch](https://spruce.mongodb.com/version/66a9417b7f65a40007ac9e15/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/workloads/scale/MixedWorkloadsGenny.yml
+++ b/src/workloads/scale/MixedWorkloadsGenny.yml
@@ -148,117 +148,6 @@ Actors:
       - *nop
       - *nop
 
-  # Update Actors
-  - ActorFromTemplate:
-      TemplateName: UpdateTemplate
-      TemplateParameters:
-        Name: Update_1
-        Threads: 1
-        OnlyActiveInPhase: 2
-
-  - ActorFromTemplate:
-      TemplateName: UpdateTemplate
-      TemplateParameters:
-        Name: Update_16
-        Threads: 16
-        OnlyActiveInPhase: 4
-
-  - ActorFromTemplate:
-      TemplateName: UpdateTemplate
-      TemplateParameters:
-        Name: Update_32
-        Threads: 32
-        OnlyActiveInPhase: 6
-
-  - ActorFromTemplate:
-      TemplateName: UpdateTemplate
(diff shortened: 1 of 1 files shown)


lgfang (commented):
After the code change, the whole `QuiesceBetweenLevels` does nothing. Can we just remove this actor?

hilldani:
> After the code change, the whole `QuiesceBetweenLevels` does nothing. Can we just remove this actor?

good point. removed and retested in the patch

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -128,61 +128,44 @@ Actors:
         Indexes:
           - keys: {id: 1}
           - keys: {a: 1}
-      - Phase: 1..10
+      - Phase: 1..6
```
jawwadasghar, line 131: The max phase is 5 no?

src/workloads/scale/MixedWorkloadsGenny.yml
```diff
@@ -102,7 +102,7 @@ ActorTemplates:
       Phases:
         OnlyActiveInPhases:
           Active: [{^Parameter: {Name: "OnlyActiveInPhase", Default: 1}}]
-          NopInPhasesUpTo: 10
+          NopInPhasesUpTo: 6
```
jawwadasghar, line 105: If max phase is indeed 5, would change this to reflect that

jawwadasghar:
Does it make sense to run a multipatch to see if removing the quiesce increases CV?

jawwadasghar (changes requested):
Would suggest a multipatch in case removing quiesce adds noise. 
Might be fine even if it does, but would be good to know the extent if any. 
Also think the max phase is 5 rather than 6.

jawwadasghar (approved):
LGTM

PR #1258, opened 2024-09-09
PERF-5868 Edit comment in TenMDocCollection_IntId about {_id: {$eq: <val>}}

militsasotirova:
**Jira Ticket:** [PERF-5868](https://jira.mongodb.org/browse/PERF-5868)

### Whats Changed

Needed to change a comment, and maybe a test name, now that we use IDHACK for a query of the shape `{_id: {$eq: <value>}}`. And maybe we should add another query that doesn't use IDHACK to replace the one we lost here.

### DIFF
--- a/src/workloads/query/TenMDocCollection_IntId.yml
+++ b/src/workloads/query/TenMDocCollection_IntId.yml
@@ -111,9 +111,9 @@ Actors:
   - ActorFromTemplate:
       TemplateName: CrudTemplate
       TemplateParameters:
-        Name: IDFieldPointProject128Threads
+        Name: IDHackEqPointProject128Threads
         Find:
-          # Using $eq turns this from an IDHACK find into a scan find.
+          # Queries of the shape {_id: {$eq: <value>}} are eligible for IDHACK.
           Filter: {_id: {$eq: {^RandomInt: {min: 0, max: 9999999}}}}
           Options:
             Projection: {_id: 1}


src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -111,9 +111,9 @@ Actors:
   - ActorFromTemplate:
       TemplateName: CrudTemplate
       TemplateParameters:
-        Name: IDFieldPointProject128Threads
+        Name: IDHackEqPointProject128Threads
```
militsasotirova, line 114: Does this rename cause the history of the test to be lost/messed up?

src/workloads/query/TenMDocCollection_IntId.yml
```diff
@@ -111,9 +111,9 @@ Actors:
...
         Find:
-          # Using $eq turns this from an IDHACK find into a scan find.
+          # Queries of the shape {_id: {$eq: <value>}} are eligible for IDHACK.
           Filter: {_id: {$eq: {^RandomInt: {min: 0, max: 9999999}}}}
```
militsasotirova, line 117: Since this query now uses IDHACK, we no longer have a query on _id that doesn't use IDHACK in this workload. Should I add one? One idea would just be to have a query with an additional predicate, something like
`{_id: {$eq: <random int>, $lt: <random int>}}`

militsasotirova:
I also [asked](https://mongodb.slack.com/archives/CKABWR2CT/p1725895746211229) in #query-execution to see if anyone there has opinions on these changes since they own express/IDHACK.

src/workloads/query/TenMDocCollection_IntId.yml
miguelangelnieto: If you rename it the history will be lost because this will be a "new" metric. But taking into account that the query optimisation is now completely different (IDHACK) and therefore the performance is not comparable, I do not think the history really matters in this case.

src/workloads/query/TenMDocCollection_IntId.yml
miguelangelnieto: I think this is something the owner of the workload should answer (@mongodb/query).

cswanson310 (approved):
Updates LGTM!

militsasotirova:
> Updates LGTM!

Thanks for the review! So you don't think I should add another query to the workload right?

src/workloads/query/TenMDocCollection_IntId.yml
militsasotirova: As discussed offline with @cswanson310 I won't be adding another query here. We will be benchmarking the optimizer via https://jira.mongodb.org/browse/SPM-2658, which should include range predicate index scans. @miguelangelnieto do you have any other comments about this PR?

miguelangelnieto (approved):
LGTM

PR #1259, opened 2024-09-12
DEVPROD-10486 Move to local task generation

SimonEismann:
**Jira Ticket:** [DEVPROD-10486](https://jira.mongodb.org/browse/DEVPROD-10486)

### What's Changed

This PR replaces the task generation in evergreen with a locally pre-generated definition and a CI task that validates that the auto-generated task definitions are up to date.

The main driver here is that a lot of the analysis tooling gets more complicated with auto-generated tests. Additionally, this means that users can schedule Genny tasks directly through the evergreen UI like normal tasks. 

The downside is that users need to re-run the generation script if they are adding a new workload or if they are adding a new variant on which genny tasks are supposed to run (both of these should get rarer in the future since we are slowly moving away from Genny).

Companion PRs:
https://github.com/10gen/dsi/pull/1984
https://github.com/10gen/mongo/pull/26993

### Patch Testing Results
* Example of the Genny CI task detecting differences in the generated files:
https://spruce.mongodb.com/task/genny_validate_genny_tasks_validate_genny_tasks_patch_764d6015fe8100a4b0a58174ebb4ba778a248c02_66e2b399e435ee00074aee2e_24_09_12_09_25_52/logs?execution=0
* Patch that is running genny tasks from both the genny repo itself and the private workloads repo:
https://spruce.mongodb.com/version/66e2f21b9f86e70007412741
* To validate that the same tasks as before get scheduled, select all tasks in [this patch with the local taskgen](https://spruce.mongodb.com/patch/66e2f21b9f86e70007412741/configure/tasks) and in [this baseline patch](https://spruce.mongodb.com/patch/66e2e51e03471a0007c4f1d1/configure/tasks) where generate_all_variant_autotasks ran.

### DIFF
--- a/src/lamplib/src/genny/tasks/auto_tasks_local.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks_local.py
@@ -69,7 +69,7 @@ def main(workspace_root: str, running_in_evergreen: bool) -> None:
         SLOG.info("Cloning the PrivateWorkloads repo to look for tests")
         GitRepo.clone_from(PRIVATE_WORKLOADS_REPO_URL, PRIVATE_WORKLOADS_TMP_PATH)
 
-    for branch_name in PROJECT_FILES.keys():
+    for branch_name in PROJECT_FILES:
         SLOG.info("Creating the configuration", branch_name=branch_name)
         builds = get_builds(branch_name)
         lister = WorkloadLister(
--- a/src/lamplib/src/genny/tasks/auto_tasks_local.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks_local.py
@@ -58,10 +58,8 @@ def get_builds(branch_name: str):
 def fix_auto_workload_path(command: CommandDefinition):
     if command._vars[WORKLOAD_PATH_KEY].startswith("src"):
         command._vars[WORKLOAD_PATH_KEY] = "src/genny/" + command._vars[WORKLOAD_PATH_KEY]
-    if command._vars[WORKLOAD_PATH_KEY].startswith("tmp-private-workloads"):
-        command._vars[WORKLOAD_PATH_KEY] = command._vars[WORKLOAD_PATH_KEY].replace(
-            "tmp-private-workloads", "src/PrivateWorkloads"
-        )
+    if command._vars[WORKLOAD_PATH_KEY].startswith("PrivateWorkloads"):
+        command._vars[WORKLOAD_PATH_KEY] = "src/" + command._vars[WORKLOAD_PATH_KEY]
 
 
 def main(workspace_root: str, running_in_evergreen: bool) -> None:
--- a/src/lamplib/src/genny/tasks/auto_tasks_local.py
+++ b/src/lamplib/src/genny/tasks/auto_tasks_local.py
@@ -1,0 +1,105 @@
+import os
+import shutil
+
+import structlog
+from git import Repo as GitRepo
+from shrub.command import CommandDefinition
+
+from genny.tasks.auto_tasks import (
+    ConfigWriter,
+    Repo,
+    WorkloadLister,
+    YamlReader,
+)
+from genny.tasks.auto_tasks_all import create_configuration, get_all_builds
+
+SLOG = structlog.get_logger("genny.tasks.auto_tasks.local")
+DSI_TMP_PATH = "./dsi"
+DSI_REPO_URL = "https://github.com/10gen/dsi.git"
+PRIVATE_WORKLOADS_TMP_PATH = "./PrivateWorkloads"
+PRIVATE_WORKLOADS_REPO_URL = "https://github.com/10gen/PrivateWorkloads.git"
+WORKLOAD_FILE_PATTERN = os.path.join("**", "src", "workloads", "**", "*.yml")
+FIRST_EXECUTION = 0
+WORKLOAD_PATH_KEY = "auto_workload_path"
+DSI_TASK_NAME = "f_run_dsi_workload"
+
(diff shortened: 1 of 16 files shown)


thessem (commented):
I think John is technicaly part of @mongodb/devprod-performance-infrastructure so I will delegate my team's review of this to him

src/lamplib/src/genny/tasks/auto_tasks_local.py
```diff
@@ -0,0 +1,103 @@
...
+    for branch_name in PROJECT_FILES.keys():
+        SLOG.info("Creating the configuration", branch_name=branch_name)
+        builds = get_builds(branch_name)
+        lister = WorkloadLister(
+            workspace_root=workspace_root, workload_file_pattern=WORKLOAD_FILE_PATTERN
+        )
+        repo = Repo(lister=lister, reader=YamlReader(), workspace_root=workspace_root)
+        config = create_configuration(repo, builds, False, [])
+
+        SLOG.info("Removing global exec_timeout_secs as it is already in the main project")
```
johndaniels, line 81: Is it worth worth seeing if we should just remove this entirely? exec_timeout should be configured globally in sys-perf so is this setting ever used even in regular task generation?

I don't love patching the config to remove it after the fact. 

We also set exec_timeout using timeout.update when we generate the tasks, which seems like it should override the global timeout anyway?

src/lamplib/src/genny/tasks/auto_tasks_local.py
```diff
@@ -0,0 +1,103 @@
...
+        config._exec_timeout_secs = None
+
+        SLOG.info("Fixing the auto_workload_path to match the DSI folder structure")
+        for task in config._tasks:
+            for command in task._commands._cmd_seq:
+                if command._function_name == DSI_TASK_NAME:
+                    fix_auto_workload_path(command)
```
johndaniels, line 88: I feel like there's a a number of these hacks scattered around because all the tooling is expecting a directory structure of the form:
```
src
  genny/src/workloads/**
  PrivateWorkloads/src/workloads/**
```

but you're checking out things as private child directories of the genny repo.

Since you're making temporary folders anyway, you could probably get this all working using a couple of simlinks and avoid some headache when running in evergreen.

Specifically, you could make a temporary directory and symlink any relevant folders to get a similar directory structure, and cut down on the fixing up.

src/lamplib/src/genny/tasks/auto_tasks_local.py
```diff
@@ -0,0 +1,103 @@
...
+PRIVATE_WORKLOADS_TMP_PATH = "./PrivateWorkloads"
+PRIVATE_WORKLOADS_REPO_URL = "https://github.com/10gen/PrivateWorkloads.git"
+WORKLOAD_FILE_PATTERN = os.path.join("**", "src", "workloads", "**", "*.yml")
+FIRST_EXECUTION = 0
+WORKLOAD_PATH_KEY = "auto_workload_path"
+DSI_TASK_NAME = "f_run_dsi_workload"
+
+PROJECT_FILES = {
+    "master": [
+        os.path.join(DSI_TMP_PATH, "evergreen", "system_perf", "master", "variants.yml"),
```
johndaniels, line 28: Won't these paths break when run in evergreen, since it will want dsi at `src/dsi`?

johndaniels (commented):
I'm curious if setting up a better directory structure is simpler here, but otherwise this seems straightforward.

src/lamplib/src/genny/tasks/auto_tasks_local.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def main(workspace_root: str, running_in_evergreen: bool) -> None:
+    if not running_in_evergreen:
+        SLOG.info("Cloning the DSI repo to look for variants")
+        GitRepo.clone_from(DSI_REPO_URL, DSI_TMP_PATH)
+        SLOG.info("Cloning the PrivateWorkloads repo to look for tests")
+        GitRepo.clone_from(PRIVATE_WORKLOADS_REPO_URL, PRIVATE_WORKLOADS_TMP_PATH)
+
+    for branch_name in PROJECT_FILES.keys():
```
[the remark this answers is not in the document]
SimonEismann: TIL

src/lamplib/src/genny/tasks/auto_tasks_local.py
SimonEismann: I'm also not a great fan of this workaround. But I think de-risking the PR by making sure the existing behavior that will be live on the 8.0 branch and on old mongo commits for a while is a good idea. Especially, since this is a band-aid solution to bridge us over until Genny is removed and not "forever code".

What do you think?

src/lamplib/src/genny/tasks/auto_tasks_local.py
SimonEismann: I've hit a few bumps with this/tried a few things:
* Cloning the private workloads repo next to the genny repo does not work, as this would modify something outside the repo, which is a bad practice.
* Symlinking the folders to `src/*/workloads` inside the genny repo is not very clean, as there is already a src/genny folder in the genny repo.
* Symlinking the folders to a temporary folder does not work as some of the code call git commands, meaning we need to be inside the correct git folder. 
* Symlinking the folders to some other folder in the genny folder results in a prefix we need to remove again.

I don't think anything here is insurmountable, but the effort/payoff ratio for something that will hopefully only live for ~6 months does not seem worth it. Let me know if you disagree with this :)

johndaniels (approved):

PR #1261, opened 2024-09-19
 DEVPROD-10486 Remove requirement for auto-tasks

SimonEismann:
**Jira Ticket:** [DEVPROD-10486](https://jira.mongodb.org/browse/DEVPROD-10486)

### What's Changed

This PR removes the requirement for the schedule_variant_auto_tasks when running local taskgen.

### DIFF
--- a/evergreen/system_perf/master/genny_tasks.yml
+++ b/evergreen/system_perf/master/genny_tasks.yml
@@ -459,6 +459,12 @@ buildvariants:
   - name: write_one_replica_set
   - name: llt_mixed
   - name: llt_mixed_small
+- name: perf-3-node-replSet.arm.aws.2024-05
+  tasks:
+  - name: read_only_multi_threaded
+- name: perf-3-node-replSet.intel.aws.2024-05
+  tasks:
+  - name: read_only_multi_threaded
 - name: perf-standalone-classic-query-engine.intel.aws.2023-11
   tasks:
   - name: multi_planning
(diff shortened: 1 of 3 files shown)


evergreen/system_perf/master/genny_tasks.yml
```diff
@@ -998,6 +1008,14 @@ buildvariants:
...
   tasks:
   - name: transport_layer_connect_timing
+- name: perf-1-node-replSet-longRunning.arm.aws.2023-11
+  tasks:
+  - name: device_monitoring
+  - name: non_search_hybrid_scoring
+- name: perf-1-node-replSet-m60-like-large-volume.arm.aws.2023-11
+  tasks:
+  - name: device_monitoring
+  - name: non_search_hybrid_scoring
```
SimonEismann, line 1018: I am not sure if we want these to get scheduled here to be honest. We might want to update the corresponding auto-run sections.

johndaniels (approved):

johndaniels:
This doesn't break the BF I fixed here: https://jira.mongodb.org/browse/BF-31453

This seems fine based on the tasks that its adding. It's neither anything really unusual nor is it like a huge pile of tasks.

johndaniels:
I will point out that using mongodb_setup for AutoRun is fairly broken, since people add new mongodb_setups all the time and so part of the reason for this change being so small is that very few variants are specifically "single_replica".

Generally we should probably just avoid using this for anything and configure the runnable genny tasks in DSI.

PR #1262, opened 2024-09-23
DEVPROD-11215 Create benchmarks for TS bucket-level optimizations

mattBoros:
Extends `TimeseriesTsbsOptimizations.yml` with benchmarks that will be affected by no longer doing the bucket-level rewrites in SERVER-93874.
Created `TimeseriesExtendedRange.yml` since there's future work for extended range optimizations, and also since SERVER-93874 will remove a block-processing optimization on $max. We can keep all benchmarks specific to extended range in `TimeseriesExtendedRange.yml`.


lgfang (approved):

src/workloads/query/TimeseriesTsbsOptimizations.yml
```diff
@@ -104,6 +104,60 @@ Actors:
...
+                  [
+                    {
+                      $group:
+                        {
+                          _id: "$tags.hostname",
+                          min_time: { $min: "$time" },
+                        },
+                    },
+                  ]
+
```
mattBoros, line 160: The other query we talked about `{$group: {_id: time, a: {$min/max: meta}}}` wasn't affected, since it can't use bucket-level rewrites anyway.

mattBoros:
I wrote the two benchmark files, all other files are automatically generated by genny

src/workloads/query/TimeseriesExtendedRange.yml
```diff
@@ -0,0 +1,163 @@
...
+          Database: *database
+          Collection: *collection
+          Operations:
+            - OperationMetricsName: TsBlockExtendedRangeMinTime
+              OperationName: aggregate
+              OperationCommand:
+                Pipeline:
+                  [
+                    {$project: {time: 1, meta: 1}},
+                    {$group: {_id: "$meta", gb: {$min: "$time"}}}
```
galon1, line 129: Both extended range and not extended range won't be applicable for $group rewrites with $min on the timeField because it's a rounded down value right? So we shouldn't be losing performance here between extended range and not. I'm not saying this isn't relevant here but it should be around the same for extended range and normal time-series collections

galon1 (commented):
I just had one main question but most of the changes look good

docs/generated/workloads.md
```diff
@@ -3864,6 +3864,26 @@ The queries in this workload exercise group stage that uses an enum like field f
 timeseries, aggregate, group 
 
 
+## [TimeseriesExtendedRange](https://www.github.com/mongodb/genny/blob/master/src/workloads/query/TimeseriesExtendedRange.yml)
+### Owner 
+Query Execution 
+
+
+### Support Channel
+[#query-execution](https://mongodb.enterprise.slack.com/archives/CKABWR2CT)
```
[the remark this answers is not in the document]
mattBoros: Oh yea, QI will probably be more involved in this benchmark in the future. I'll change it

src/workloads/query/TimeseriesExtendedRange.yml
mattBoros: Yea! That's true. I was thinking it's probably best to cover both $min/$max instead of just $max, while were writing these benchmarks. Like if in the future we run into a similar $min bug, we'll already have this benchmark and it's perf history.

docs/generated/workloads.md
```diff
@@ -3864,6 +3864,26 @@ The queries in this workload exercise group stage that uses an enum like field f
 timeseries, aggregate, group 
 
 
+## [TimeseriesExtendedRange](https://www.github.com/mongodb/genny/blob/master/src/workloads/query/TimeseriesExtendedRange.yml)
+### Owner 
+Query Integration 
+
+
+### Support Channel
+[#query-integration](https://mongodb.enterprise.slack.com/archives/C04PDS7GAFM)
```
parker-felix, line 3873: I would say that we shouldn't be linking to slack channels on a public readme but looks like we are already doing that elsewhere, and removing all of those links is definitely outside of scope for this ticket.

src/workloads/query/TimeseriesExtendedRange.yml
```diff
@@ -0,0 +1,163 @@
...
+
+Keywords:
+  - timeseries
+  - aggregate
+
+GlobalDefaults:
+  Database: &database test
+  Collection: &collection Collection0
+  DocumentCount: &documentCount 1e7
+  Repeat: &repeat 200
```
brad-devlugt, line 16: It looks like this benchmark only repeats the queries 200 times on a really small dataset. I think this is likely to lead to really noisy results that won't be representative of a real customer use case. 
- Have we done patch testing to validate that we are getting good signal here?
- If there was a regression in this workload, would it be high priority to fix?

brad-devlugt (changes requested):

evergreen/system_perf/8.0/genny_tasks.yml
```diff
@@ -339,6 +339,12 @@ buildvariants:
   - name: write_one_replica_set
   - name: llt_mixed
   - name: llt_mixed_small
+- name: perf-3-node-replSet.arm.aws.2024-05
+  tasks:
+  - name: read_only_multi_threaded
+- name: perf-3-node-replSet.intel.aws.2024-05
+  tasks:
+  - name: read_only_multi_threaded
```
brad-devlugt, line 347: It looks like we're adding another workload here - do we need to schedule this one?

docs/generated/workloads.md
mattBoros: Yea, this was auto-generated. I'm guessing the slack is private so the link just goes nowhere for public viewers

src/workloads/query/TimeseriesExtendedRange.yml
mattBoros: I could increase the number of queries run. Also, I didn't realize 10^7 is considered a small dataset. What is a realistic dataset size?

> Have we done patch testing to validate that we are getting good signal here?

Yea, the signal was not very noisy and also reflected the regression we are expecting to see after fixing https://github.com/10gen/mongo/pull/26614

> If there was a regression in this workload, would it be high priority to fix?

Yes, this is a benchmark we specifically optimize for, since it is a common pattern.

brad-devlugt: Thanks Matt. How noisy was it? What does 10^7 documents map to in terms of GiB?

mattBoros: I realized the benchmarks actually changed a bit since the start of this PR, so I'll remeasure the noise and get back to you.

The collection size is interesting. The initial size was small, so I added two more random double fields. With these fields, I think it's still affected by the fact that WiredTiger compresses data, and also time-series collections bucket similar data together. The size of the TS collection is 0.17 GB (using `db.collection.totalSize()`). The size of the equivalent non-TS collection is 0.52 GB.

Since both TS and non-TS collections are compressed, the uncompressed collection size (the data processed by the query engine) is probably in the GB

mattBoros: I ran three patches with the latest benchmark, it looks not too noisy:

```
BlockProcessingExtendedRangeMinTime
ARM (AverageLatency, ns)
236,650,000
235,400,000
236,030,000
range = 1,250,000
range / avg measurement = 0.53%

Intel (AverageLatency, ns)
129,160,000
129,800,000
127,810,000
range = 1,990,000
range / avg measurement = 1.54%

BlockProcessingExtendedRangeMaxTime
ARM (AverageLatency, ns)
61,750,000
61,200,000
60,680,000
range = 1,070,000
range / avg measurement = 1.75%

Intel (AverageLatency, ns)
36,260,000
36,600,000
35,770,000
range = 830,000
range / avg measurement = 2.29%

```

evergreen/system_perf/8.0/genny_tasks.yml
mattBoros: Had to pull and then rerun `auto-generate-tasks`, thanks for pointing this out

galon1 (approved):
LGTM! sorry for the delay on this!

src/workloads/query/TimeseriesExtendedRange.yml
brad-devlugt: Thanks @mattBoros, 0.17GiB seems pretty small - is that how large these collections would be in practice? If we optimize for this smaller dataset, is there any risk we'd miss problems that occur on a larger collection?

mattBoros: I could double the size, but I tried doing 10x more documents and insertion was taking a long time.

I think the uncompressed data size matters more than the compressed size for this benchmark, since the query system would be operating on the uncompressed data, which is probably at least a GB. I am curious if other benchmarks insert lots of data, like is it typical to have long insertion times? Then I'd be happy to increase it more

brad-devlugt (approved):
Matt & I synced on slack, it sounds like adding more queries or more data just increases the runtime without reducing noise (which is already low), so this LGTM

PR #1269, opened 2024-10-09
DEVPROD-11940 Add a flag to suppress metrics from the RunCommand actor

SimonEismann:
**Jira Ticket:** [DEVPROD-11940](https://jira.mongodb.org/browse/DEVPROD-11940)

### What's Changed

This PR adds a `ReportMetrics ` flag to the RunCommand actor that is true by default. Additionally, I went through all uses of the RunCommand actor and labeled anything that was clearly not a measurement as `ReportMetrics: false`. This is usually indicated by Repeat: 1, and the command creating indexes/dropping index/creating collections or setting database parameters.

### Patch Testing Results
https://spruce.mongodb.com/version/670690abed45040007c733dc/

### DIFF
--- a/src/cast_core/src/RunCommand.cpp
+++ b/src/cast_core/src/RunCommand.cpp
@@ -123,12 +123,14 @@ public:
                       const std::string& databaseName,
                       mongocxx::database database,
                       DocumentGenerator commandExpr,
-                      OpConfig opts)
+                      OpConfig opts,
+                      bool reportMetrics)
         : _databaseName{databaseName},
           _database{std::move(database)},
           _commandExpr{std::move(commandExpr)},
           _options{std::move(opts)},
           _awaitStepdown{opts.awaitStepdown},
+          _reportMetrics{reportMetrics},
           // Record metrics for the operation or the phase depending on
           // whether metricsName is set for the operation.
           //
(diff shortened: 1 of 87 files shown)


PyGeek03:
Did you mean `ReportMetrics` instead of `RecordMetrics`? There's no mention of `RecordMetrics` in the PR, only `ReportMetrics`.

SimonEismann:
Good catch, I've updated the PR accordingly.

thessem (approved):
I am happy with this so long as @mongodb/performance are happy with not having this metric available any longer for debugging regressions.

alicedoherty:
(More just thinking out loud here) but we have a lot of parallel work going on to clean up reported metrics ([PERF-5867](https://jira.mongodb.org/browse/PERF-5867), [scope for adding tags to time-series](https://docs.google.com/document/d/1cR3AwnERS_1wHgXX1UueT0V98zpGbFWdUogNzOxX13I/edit), and this. I assume we're going this change now because it'll get in much quicker than that DEVPROD project for adding tags?

SimonEismann:
>(More just thinking out loud here) but we have a lot of parallel work going on to clean up reported metrics ([PERF-5867](https://jira.mongodb.org/browse/PERF-5867), [scope for adding tags to time-series](https://docs.google.com/document/d/1cR3AwnERS_1wHgXX1UueT0V98zpGbFWdUogNzOxX13I/edit), and this. I assume we're going this change now because it'll get in much quicker than that DEVPROD project for adding tags? 
You are right. The best solution for this would be to wait for tags to live and then tag the metrics as setup or something. But that would need to wait until tags are available and require more complex genny plumbing. Given that we are looking to deprecate Genny soonish, a band-aid fix like this seems preferable to investing in a more complex solution.

alicedoherty (approved):
LGTM, but maybe give a heads up in the server-perf channel about this in case anyone has conflicting opinions

PR #1280, opened 2024-11-07
PERF-5985 Remove the workload running primary reads with stepDown

samanca:
**Jira Ticket:** [PERF-5985](https://jira.mongodb.org/browse/PERF-5985)

### Whats Changed
Removed `ServiceArchitectureWorkloads`. The workload aims to measure the effects of primary reads during an election: it runs primary reads and issues an election after 60 seconds. However, there are no metrics to evaluate how elections impact performance.


lgfang (approved):

PyGeek03:
Nice. With this removal, we could remove [this weird hack](https://github.com/10gen/dsi/blob/a8715df4d24ece00b2fb4b8cbab13ee3edec330e/src/dsi/libanalysis/log_analysis.py#L99-L106) in DSI analysis that's only for `ServiceArchitectureWorkloads`.

PR #1282, opened 2024-11-27
PERF-6114 Add $lookup tests to 1-shard variants

nzolnierzmdb:


evergreen/system_perf/master/genny_tasks.yml
```diff
@@ -770,6 +770,54 @@ buildvariants:
 - name: perf-3-node-replSet-limited-10000-query-stats.arm.aws.2023-11
   tasks:
   - name: non_search_hybrid_scoring
+- name: perf-1-node-replSet-default-sharding.arm.aws.2023-11
```
HanaPearlman, line 773: So we're the first ones to want to use any of these variants for perf tests? Interesting

Relatedly, how did you pick this list? I can't tell from the names here exactly what they represent, but they're all basically 1-shard variants?

src/workloads/query/Lookup.yml
```diff
@@ -326,3 +326,8 @@ AutoRun:
         $eq:
           - shard
           - shard-all-feature-flags
+          - 1-shard-dedicated.2024-05
```
HanaPearlman, line 329: How do the changes in the workload files relate to the task changes above? I can't tell how these are supposed to correlate. Does this list mean: schedule this task whenever we're running one of these configs? And if so, then what do the changes in `genny_tasks.yml` mean?

HanaPearlman (commented):
Some newbie questions!

IamXander:
@brad-devlugt @johndaniels @thessem should this be in locust at this point? I am not totally sure but just want to make sure you were aware of this.

nzolnierzmdb:
> @brad-devlugt @johndaniels @thessem should this be in locust at this point? I am not totally sure but just want to make sure you were aware of this.

@hilldani mentioned the same thing when I reached out. Given that this PR does not change the workload (only the configuration), I'd prefer to keep the port to locust as a separate task. We've talked about scheduling a project on query to port existing Genny tests but didn't end up scheduling this Q.

evergreen/system_perf/master/genny_tasks.yml
nzolnierzmdb: We're the first ones to add Genny tasks to these variants, they're currently defined in dsi https://github.com/10gen/dsi/blob/3a032e9a89abcafc9a8400b3e2b0d4cee182b97e/evergreen/system_perf/master/master_variants.yml#L1493

Unfortunately the names here are not immediately obvious that they're related to invisible sharding (/all clusters sharded) but you can see the comment in the link above for why that's the case.

src/workloads/query/Lookup.yml
nzolnierzmdb: I think this is related to some history with how Genny tasks were historically scheduled, where the auto* task may run on variants where some of the individual workloads were not compatible. But yes at a high level I believe this must match the `mongodb_setup` for each of the Genny tasks where this workload is included. [Here is a patch](https://spruce.mongodb.com/task/genny_validate_genny_tasks_validate_genny_tasks_patch_46408e2d9fabae647be29746fe3e04560e03a286_674df8801a9b7d0007e5ecec_24_12_02_18_12_17/logs?execution=0) where they don't align and the error `To resolve this issue, please run ./run-genny auto-tasks-local and commit the new files in the evergreen folder.`


What's nice is that you can add the conditions here and then run `./run-genny auto-tasks-local` to automatically populate the `genny_tasks.yml`. Hopefully that makes sense but I'm certainly not the expert on this!

nzolnierzmdb:
[Patch](https://spruce.mongodb.com/version/674f16ffe2e0380007b26948/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

HanaPearlman (approved):

blagoev (approved):
lgtm

brad-devlugt:
> > @brad-devlugt @johndaniels @thessem should this be in locust at this point? I am not totally sure but just want to make sure you were aware of this.
> 
> @hilldani mentioned the same thing when I reached out. Given that this PR does not change the workload (only the configuration), I'd prefer to keep the port to locust as a separate task. We've talked about scheduling a project on query to port existing Genny tests but didn't end up scheduling this Q.



> > @brad-devlugt @johndaniels @thessem should this be in locust at this point? I am not totally sure but just want to make sure you were aware of this.
> 
> @hilldani mentioned the same thing when I reached out. Given that this PR does not change the workload (only the configuration), I'd prefer to keep the port to locust as a separate task. We've talked about scheduling a project on query to port existing Genny tests but didn't end up scheduling this Q.

I agree, porting other workloads should come after DEVPROD-253 is complete

johndaniels (approved):

nzolnierzmdb:
macOS dry run originally failed, and in the latest attempt it failed on a `git clone` step also on macOS. I've been grepping around slack and it _seems_ like restarting should resolve, otherwise I'm working on getting Genny installed locally to see if it repros.

PR #1284, opened 2024-12-16
DEVPROD-13512: Disable task generation in Genny.

johndaniels:
We are switching over to using manually maintained files based in the DSI repo to run all Genny tasks. This removes any task generation functionality so that it doesn't cause confusion


docs/using.md
```diff
@@ -538,13 +538,16 @@ Examples with these and other Actors can be found in [./src/workloads/docs](../s
...
 
-AutoRun searches in `[workspace]/src/*/src/workloads` and assumes that all workload repos, including Genny itself, are checked out in `[workspace]/src`.
+### Configuring AutoRun
 
-Based on these, Genny generates task defintion files that are used by Evergreen. See [Auto-generated Genny Tasks](evergreen/system_perf/README.md) for more information.
+!!! warning AutoRun in Genny is deprecated and no longer used
+
+    AutoRun is no longer available in Genny.
+    This documentation is kept for historical purposes only.
+
```
SimonEismann, line 549: Let's use [caution blocks](https://github.com/orgs/community/discussions/16925) here that should be harder to miss.

SimonEismann (changes requested):
As discussed at the onsite during the perf collab session, this is a good idea because the genny tasks are no longer expected to change much and we've learned the painful lesson that having evergreen configurations spread across multiple repos is hard to work with.

I would like to see the AutoRun blocks commented out in all workloads, with a corresponding comment explaining that it's deprecated. If possible, checking if the test defintion has an AutoRun block and failing immediately would provide a safe guard for any newly written tests. If the latter makes sense depends on how much work it would be.

docs/using.md
```diff
@@ -538,13 +538,16 @@ Examples with these and other Actors can be found in [./src/workloads/docs](../s
 
 ### What is AutoRun?
```
PyGeek03, line 539: I think it would be nice to modify the header to include the deprecation as well:

```suggestion
### What was AutoRun?
```

docs/using.md
```diff
@@ -538,13 +538,16 @@ Examples with these and other Actors can be found in [./src/workloads/docs](../s
...
 
-AutoRun is a utility to allow workload authors to determine scheduling of their workloads without having to commit to a separate repo. The utility is specifically designed for users who are using DSI through Evergreen. To use AutoRun, make sure your project has integrated DSI with Evergreen as explained [here](https://github.com/10gen/dsi/wiki/DSI-In-Evergreen).
+AutoRun was a utility to allow workload authors to determine scheduling of their workloads without having to commit to a separate repo. Autorun is now deprecated, and any modifications to what Genny tasks should run should happen in the Evergreen Configuration found in the [DSI Repo](https://github.com/10gen/dsi/tree/master/evergreen/system_perf).
```
PyGeek03, line 541: Nit: Some wording changes to reduce duplication:

```suggestion
AutoRun was a utility to allow workload authors to determine scheduling of their workloads without having to commit to a separate repo. Autorun is now deprecated, and any modifications regarding which Genny tasks to run should happen in the Evergreen Configuration found in the [DSI Repo](https://github.com/10gen/dsi/tree/master/evergreen/system_perf).
```

evergreen/system_perf/README.md
```diff
@@ -1,8 +1,18 @@
...
+
+
 This folder contains the task defintions and buildvariant assignments for the
-Genny workloads with `AutoRun` sections. These files are automatically
-generated and should never be updated manually.
+Genny workloads with `AutoRun` sections. They are kept for backwards 
+compatibility purposes only for older versions of the mongo repository, 
+and should not be modified.
+
+Any modifications to what Genny tasks should run should happen in the Evergreen
```
PyGeek03, line 14: Same nit as above:

```suggestion
Any modifications regarding which Genny tasks to run should happen in the Evergreen
```

PyGeek03:
> I would like to see the AutoRun blocks commented out in all workloads, with a corresponding comment explaining that it's deprecated. If possible, checking if the test defintion has an AutoRun block and failing immediately would provide a safe guard for any newly written tests. If the latter makes sense depends on how much work it would be.

Initially I thought this is a good thing to do in this PR as well. However, as mentioned in this PR [here](https://github.com/mongodb/genny/pull/1284/files#diff-481bd291277476289a4b35f80c52c0df7f2a8d7bf31d212410bc800a015396f9R10-R12), they are kept for backward compatibility with older older versions of the mongo repository. Do you think backward compatibility would be an obstacle for removing AutoRun blocks?

PyGeek03 (commented):
So happy to see Genny autotask going away, thank you team!

docs/using.md
johndaniels: @SimonEismann This is the way to do admonitions in the Pine devprod docs, so it will be more visible there.

johndaniels:
I actually have a separate change on my machine to delete AutoRun blocks that I can apply whenever. I didn't want to muddy this PR with it.

docs/using.md
SimonEismann: Pine also supports notes now:
https://docs.devprod.prod.corp.mongodb.com/performance/profiling_workflow/#heap-profiling-heaptrack
image

But I can see just doing the thing that looks good in Pine as well.

johndaniels: 😱

johndaniels:
> I would like to see the AutoRun blocks commented out in all workloads, with a corresponding comment explaining that it's deprecated. If possible, checking if the test defintion has an AutoRun block and failing immediately would provide a safe guard for any newly written tests. If the latter makes sense depends on how much work it would be.

@SimonEismann per the above comment I have a followup PR that replaces all the autorun blocks with `HasTasks` (since we use the presence of AutoRun to determine if there is likely a linkeable task). I wanted to keep the big change separate from the more focussed changes

@PyGeek03 I think this change is already backwards incompatible with any PR that does task generation.

This seems fine... it would only apply when using the latest version of Genny from an old version of Mongo and the workaround would be to mannually add the genny task info you want into DSI. I can add the workaround info to https://docs.google.com/document/d/1YzzrmKgIcpLG-60IdejYD4pxJzUQR-dy_nrrBgnLbBA/edit?tab=t.0 if that makes sense to do.

SimonEismann (approved):

PR #1285, opened 2024-12-18
DEVPROD-13594 Adding an SLOG statement to the genny_runner to collect Python version info

cjeffries:
### Whats Changed
As part of #1277 - upgrading the requests package drops support for Python 3.7. Lets check that Genny doesn't use 3.7 before accepting the upgrade. This should output a easy to discover log we can collate to see what versions of Python are being used by Genny.

Also removed some unused imports

### DIFF
--- a/src/lamplib/src/genny/tasks/genny_runner.py
+++ b/src/lamplib/src/genny/tasks/genny_runner.py
@@ -37,7 +37,7 @@ def main_genny_runner(
         # about what version of Python Genny is running
         # This will help inform whether upgrading packages that deprecate
         # use of Python versions that Genny supports will cause problems
-        SLOG.info(f"Genny is running the Python version: {sys.version}")
+        SLOG.info(f"Python version being used by Genny", python_version=f'"{sys.version}"')
 
         path = os.path.join(genny_repo_root, "dist", "bin", "genny_core")
         if not os.path.exists(path):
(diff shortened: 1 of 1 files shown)


src/lamplib/src/genny/tasks/genny_runner.py
```diff
@@ -34,6 +33,12 @@ def main_genny_runner(
         workspace_root=workspace_root,
         genny_repo_root=genny_repo_root,
     ):
+        # Inserting an Info log to gather some information 
+        # about what version of Python Genny is running
+        # This will help inform whether upgrading packages that deprecate
+        # use of Python versions that Genny supports will cause problems
+        SLOG.info(f"Genny is running the Python version: {sys.version}")
```
PyGeek03, line 40: How about we make this log line more structured? (I added the quotes around the output of `sys.version` to (hopefully) make it easier to parse, as the output is a string with multiple spaces)

```suggestion
        SLOG.info(f"Python version being used by Genny", python_version=f'"{sys.version}"')
```

PyGeek03:
Btw this is a log line, not a print statement, so it might be a good idea to modify the PR title.

src/lamplib/src/genny/tasks/genny_runner.py
cjeffries: Sounds great, thank you!

PyGeek03 (approved):

PR #1290, opened 2025-01-06
Updating documentation and checks to use Python 3.8 as the earliest version allowed

cjeffries:
### Whats Changed
As discussed in #1277, the requests library drops support for Python 3.7. This version of Python is also EOL (for more than 12 months) and Genny does not explicitly rely on it, so this PR makes sure the docs are also up to date in that regard.


edwardgauld (approved):
LGTM

PR #1293, opened 2025-01-13
SERVER-99058 Modify existing Resharding Performance Tests to include all CRUD ops.

janetio:
**Jira Ticket:** [SERVER-99058](https://jira.mongodb.org/browse/ERVER-99058)

### Whats Changed

Added `insertOne` and `deleteOne` write operations to resharding workloads.

### Patch Testing Results
https://spruce.mongodb.com/version/6781430bfd13090007dd54b6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


src/workloads/sharding/ReshardCollection.yml
```diff
@@ -29,9 +29,8 @@ GlobalDefaults:
   # fields in the document.
   ApproxDocumentSize: &ApproxDocumentSize 10000 # = 10kB
   ApproxDocumentSize80Pct: &ApproxDocumentSize80Pct 8000 # = 8kB
-  # TODO: Increase DocumentCount to 4000000 so there will be ~20GB of data inserted on each shard
-  # and increase NumReshardedChunks to 400 so there continue to be ~50MB of data per chunk.
-  DocumentCount: &DocumentCount 100000 # for an approximate total of 1GB
+  # TODO: Increase NumReshardedChunks to 400 so there continue to be ~50MB of data per chunk.
+  DocumentCount: &DocumentCount 4000000 # for an approximate total of 20GB
```
janetio, line 33: Had to increase since was failing due to insufficient shard key cardinality

cheahuychou: Hmm, I don't know if we want to change the collection size (especially by this much). I think the cause of this is that you are not specifying the new shard key in the document being inserted here so all newly inserted documents have `newKey: null`, making the new shard key have low cardinality and high frequency because of the "null" value. The insert here should specify `oldKey` and `newKey` just like the `LoadInitialData` inserts. That is, the docs in the collection should have the same schema. The same applies to the other workloads. 

```
oldKey: {^RandomInt: {min: *ShardKeyValueMin, max: *ShardKeyValueMax}}
newKey: {^RandomInt: {min: *ShardKeyValueMin, max: *ShardKeyValueMax}}
```

cheahuychou (changes requested):

src/workloads/sharding/ReshardCollection.yml
```diff
@@ -29,9 +29,8 @@ GlobalDefaults:
   # fields in the document.
   ApproxDocumentSize: &ApproxDocumentSize 10000 # = 10kB
   ApproxDocumentSize80Pct: &ApproxDocumentSize80Pct 8000 # = 8kB
-  # TODO: Increase DocumentCount to 4000000 so there will be ~20GB of data inserted on each shard
-  # and increase NumReshardedChunks to 400 so there continue to be ~50MB of data per chunk.
-  DocumentCount: &DocumentCount 100000 # for an approximate total of 1GB
+  # TODO: Increase NumReshardedChunks to 400 so there continue to be ~50MB of data per chunk.
+  DocumentCount: &DocumentCount 100000 # for an approximate total of 20GB
```
cheahuychou, line 33: Is this comment accurate? Nothing has changed, correct? So can we keep the same comment?

src/workloads/sharding/ReshardCollection.yml
```diff
@@ -44,6 +43,16 @@ GlobalDefaults:
       OperationCommand:
         Filter: {_id: {^RandomInt: {min: 1, max: *DocumentCount}}}
         Update: {$inc: {counter: 1}}
+    - OperationName: insertOne
+      OperationCommand:
+        Document:
+          _id: {^RandomInt: {min: *DocumentCount, max: 9000000}}
```
cheahuychou, line 49: I think we should define a constant under `GlobalDefaults` for the min and max, e.g. name it `DocumentIdMinForInserts` and `DocumentIdMaxForInserts`? and add a comment to explain why that is the range to use, e.g. by pointing out that the `MonotonicSingleLoader` used by the `LoadInitialData` phase generates documents with monotonically increasing `_id`, starting from `_id: 0`. Same for the other workloads.

[question] Why is it 900'000 instead of some round number like 1'000'000?

evergreen/system_perf/8.0/genny_tasks.yml
```diff
@@ -571,6 +571,353 @@ buildvariants:
   - name: write_one_replica_set
   - name: llt_mixed
   - name: llt_mixed_small
+- name: mongotune-perf-3-shard.arm.aws.2023-11
```
cheahuychou, line 574: Where did the genny_tasks.yml changes come from? Why are they needed?

cheahuychou (changes requested):

src/workloads/sharding/ReshardCollection.yml
janetio: Oh yes, I just didn't change the comment back yet because I wanted to see if changing it back passed the test.

PR #1295, opened 2025-01-28
DEVPROD-13512: Explicitly deprecate updating Genny.

johndaniels:
It's now the end of January. We should go ahead and deprecate new workloads in Genny. This change just adds warnings to make it official. When we're ready we can just fully archive the repository.


cjeffries (approved):
👍 👍 👍 LGTM

SimonEismann (approved):

PR #1300, opened 2025-03-20
DEVPROD-15706 Include a dependabot.yml file that should disable all updates

cjeffries:
**Jira Ticket:** [DEVPROD-15706](https://jira.mongodb.org/browse/DEVPROD-15706)

### Whats Changed

Disabling Dependabot YML


cjeffries:
If this file doesn't work, we can also try specifying: `ignore-dependencies: *`. I haven't really been able to find a tool that assesses what dependabot will do with a given config file which has been very annoying

thessem (approved):

PR #1303, opened 2025-03-31
DEVPROD-16216: Fixing linting and mothra service teams files

cjeffries:
Fixing test suite and checks so #1301 can be merged


cjeffries:
Other than the change of the file name in the mothra team check, no code has been changed, it has only been formatted. This PR doesn't require approval from the Performance team in that vein. It is only here to unblock #1301

thessem (approved):

PR #1304, opened 2025-04-15
PERF-6583 Modify time_series_lastpoint to use different data set

erinzhu001:
**Jira Ticket:** [FOO-123](https://jira.mongodb.org/browse/FOO-123)

### DIFF
--- a/src/workloads/query/TimeSeriesLastpoint.yml
+++ b/src/workloads/query/TimeSeriesLastpoint.yml
@@ -10,8 +10,8 @@ Description: |
     3. any of the above pipelines with a preceding match predicate on a meta field.
 
 # Parameters reused in multiple Actors.
-db: &db test
-coll: &coll Collection0
+db: &db cpu
+coll: &coll point_data
 phasePath: &phasePath ../../phases/query/TimeSeriesLastpoint.yml
 MaxPhases: &MaxPhases 26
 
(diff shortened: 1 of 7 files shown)


docs/generated/workloads.md
```diff
@@ -4126,9 +4126,10 @@ currently supported lastpoint aggregate pipelines that are tested here include:
   3. any of the above pipelines with a preceding match predicate on a meta field.
 
 
-[Task Page](https://evergreen.mongodb.com/task_history/sys-perf/time_series_lastpoint)
```
erinzhu001, line 4129: This file is autogenerated -- I think this was removed because I removed the definitions from genny_tasks.yml.

thessem (approved):
My team shouldn't be code-owners on these files, but as the Readme.md says "Adding new workloads in Genny is deprecated." so we probably wont re-do the codeowners on this repository. I am approving to get out of the way.

hilldani:
FYI @alicedoherty who is currently moving genny workloads.

hilldani:
@erinzhu001 are there plans to move this to locust? Ideally we would in place depreciate genny and new changes or workloads would go to locust

erinzhu001:
@hilldani I filed a [ticket to move them here](https://jira.mongodb.org/browse/PERF-6672) to be done as part of a timeseries tech debt project.

alicedoherty (approved):

src/workloads/query/TimeSeriesLastpoint.yml
```diff
@@ -95,59 +104,42 @@ LastpointQueryMetaAscTimeAscRepeat100: &LastpointQueryMetaAscTimeAscRepeat100
       Repeat: 100
       Collection: *coll
```
galon1, line 105: [optional] to reduce the runtime of this workload we could consider dropping this to 50 and not 100 since we have way more documents now.

src/workloads/query/TimeSeriesLastpoint.yml
```diff
@@ -95,59 +104,42 @@ LastpointQueryMetaAscTimeAscRepeat100: &LastpointQueryMetaAscTimeAscRepeat100
...
-          Database: *db
-          Collection: *coll
-          Operation:
-            OperationName: drop
+  # The data is already loaded into the collection from the dsi configuration.
+  # QuiesceActor will run for phase 0.
 
-  - Name: CreateTimeSeriesCollection
+  - Name: DropExistingIndexForMetaTime
     Type: RunCommand
```
galon1, line 115: I think you can remove this actor because the test in phase 4 RunLastpointQueryWithMetaSubfieldAscendingTimeDescendingNoIndex that runs with no index wouldn't match this index. We also don't expect users to drop the default meta_1_time_1 index for timeseries collections so it might be better to remove this actor. In the old version of the test this index would also have existed and not been dropped because it's the default timeseries index

galon1 (commented):
I just left two comments. I don't feel too strongly about either of them but wanted to hear your thoughts before approving

src/workloads/query/TimeSeriesLastpoint.yml
erinzhu001: Removed! Rerunning the test on a few variants to make sure it works properly. 
[master](https://spruce.mongodb.com/version/680260bbd12832000723b603/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[8.0](https://spruce.mongodb.com/version/680260ec66560300071e52b3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[6.0](https://spruce.mongodb.com/version/6802614b66560300071e5413/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

galon1 (approved):
Thanks Erin for these changes!

PR #1313, opened 2025-07-18
DEVPROD-19700 Fallback to mongodbtoolchain v5 if v4 is missing

thessem:
The `mongodbtoolchain` for AL2 seems to be missing v4, but has v5. We should fallback to v5 if v4 is missing


cjeffries (approved):
LGTM!

PR #1326, opened 2026-07-29
PERF-9005 Rename InsertData actor to Setup in MatchWithLargeExpression

henrinikku:
Renames the `InsertData` Loader actor in `MatchWithLargeExpression.yml` to `Setup`.


alicedoherty (approved):
LGTM but please validate the patch passes and the loader metrics are no longer recorded as expected before merging

